PR #1792, opened 2023-02-18
ci: increase mobile integration test speed and reliability

zackpollard:
Introduces a bunch of caching as well as a complete overhaul of the way the tests wait for elements to be available to click/interact with.

### DIFF
--- a/mobile/integration_test/test_utils/general_helper.dart
+++ b/mobile/integration_test/test_utils/general_helper.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:easy_localization/easy_localization.dart';
 import 'package:flutter_test/flutter_test.dart';
 import 'package:hive/hive.dart';
@@ -43,7 +45,6 @@ class ImmichTestHelper {
     // Load main Widget
     await tester.pumpWidget(app.getMainWidget(db));
     // Post run tasks
-    await tester.pumpAndSettle();
     await EasyLocalization.ensureInitialized();
   }
 }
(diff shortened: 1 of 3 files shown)


mobile/integration_test/test_utils/general_helper.dart
```diff
@@ -62,3 +63,21 @@ void immichWidgetTest(
...
+    Finder finder, {
+      Duration timeout = const Duration(seconds: 120),
+    }) async {
+  bool timerDone = false;
+  final timer = Timer(timeout, () => throw TimeoutException("Pump until has timed out"));
+  while (timerDone != true) {
+    await tester.pump();
+
+    final found = tester.any(finder);
+    if (found) {
```
matthinc, line 78: I had to think about this for a while.
What about

```dart
final found = false;
//...
while (!found) {
//...
found = tester.any(finder);
//...
}

```

zackpollard: I took this code from somewhere else, I am not sure refactoring it in this way really makes it any more or less readable. Would rather leave it alone given it works perfectly :smile:

matthinc: My concern is just the naming here. Because `timerDone` is the opposite of what this variable is actually doing :D

zackpollard: Yea true, reading it again I think we can just drop the variable. Will quickly test now

matthinc (approved):

matthinc (approved):

PR #1804, opened 2023-02-20
feat(mobile): improve date formatting

michelheusschen:
- Removed hardcoded en_US locales
- Removed all dates from translation files, it shouldn't be necessary to let users translate dates and it can lead to issues (see #1771)
- Removed unused file `mobile/lib/modules/home/ui/asset_grid/monthly_title_text.dart`

Some dates are not being localized, I think this is because Intl has no locale set or DateFormat being used outside of a BuildContext. I have currently no fix for this, but added a TODO comment.


alextran1502 (approved):

alextran1502 (approved):

PeterBasista:
What does this change mean? Will US date formats be the default for each country from now on? Because if so I'm not sure if this is a change for the better or a change backwards. For example, the date format that was currently in immich is not very clear for my country and by default we use dd.m.yyyy.

samip5:
I also think this might be a change backwards as we use dd.mm.yyyy format...

alextran1502:
@michelheusschen I think I will revert this PR as others have valid points.

michelheusschen:
I'm a bit confused why you think this is a step backwards. All the changed dates are replaced with a built-in version that contains the date format and translation for specific locales. Only the dates on the timeline are still an issue, but I've clearly marked it as such.

Below some examples with before (left) and after (right) screenshots for the Finnish locale. If I'm missing the point you guys are trying to make, please correct me.

[album_viewer_page.dart](https://github.com/immich-app/immich/pull/1804/files#diff-003a12ab07eceb0c11d0d6c02e1ac042a802dcb231831ce5125732996f9d344b)
album_finnish_before album_finnish_after


[current_backup_asset_info_box.dart](https://github.com/immich-app/immich/pull/1804/files#diff-44eb45e2c4e0890d918d2b65facf0556b4349811d0515574ecb555c44fafc5dc)
backup_finnish_before backup_finnish_after

[exif_bottom_sheet.dart](https://github.com/immich-app/immich/pull/1804/files#diff-8b87843ef2c1c52551461a95d5ef7bb606d04010d4ffda7178054c9bb28b060c)
mobile_exif_finnish_before mobile_exif_finnish_after

samip5:
Huh, okay. I withdraw my comment. That seems better.

P.S Seems so wierd to see Finnish in a GH issue.

PeterBasista:
Thank you for the clarification. Yes this step is definitely a step for the better. I thought this was a fixed definition of the format. This solution is then very good, as there is no need to define formats for each language separately

PR #1810, opened 2023-02-21
fix(mobiile): Fixed load translations in background sync

martyfuhry:
Fixes a regression which accidentally removed the `loadTranslations()` function in #1758.


martyfuhry:
This is ready for now so that we don't regress the Android notifications. I'm working on fixing this for iOS back in #1790 so we can remove the platform guard.

alextran1502 (approved):

PR #1839, opened 2023-02-22
feat(mobile): Background app refresh status

martyfuhry:
Adds a message that App Refresh is currently disabled and a settings button to navigate to app settings.

Unfortunately, the Go To Settings takes you to  Settings > Immich instead of Settings > General > Background App Refresh.

- There is not a safe way to navigate to the latter. Maybe the message should say that's where they need to go?
- If Background App Refresh is on globally, but off for Immich, the system still reports that it's "available". This will be obvious for folks with 0 processes queued, though.
- App Refresh tasks can now be scheduled even when connected power is required; they will check to see if they have power and if so, run, otherwise immediately complete.
- Fixes an issue where turning off and on background app refresh would clear all Immich tasks until a full app restart. Now you just need to reopen the Immich app to bring it back to foreground to reschedule the background tasks.

image

### DIFF
--- a/mobile/lib/modules/backup/providers/ios_background_settings.provider.dart
+++ b/mobile/lib/modules/backup/providers/ios_background_settings.provider.dart
@@ -15,9 +15,9 @@ class IOSBackgroundSettings {
   });
 }
 
-class IOSBackgroundSettingsState extends StateNotifier<IOSBackgroundSettings?> {
+class IOSBackgroundSettingsNotifier extends StateNotifier<IOSBackgroundSettings?> {
   final BackgroundService _service;
-  IOSBackgroundSettingsState(this._service) : super(null);
+  IOSBackgroundSettingsNotifier(this._service) : super(null);
 
   IOSBackgroundSettings? get settings => state;
 
@@ -51,7 +51,7 @@ class IOSBackgroundSettingsState extends StateNotifier<IOSBackgroundSettings?> {
 
 }
 
-final iOSBackgroundSettingsProvider = StateProvider<IOSBackgroundSettingsState>(
-  (ref) => IOSBackgroundSettingsState(ref.watch(backgroundServiceProvider)),
+final iOSBackgroundSettingsProvider = StateNotifierProvider<IOSBackgroundSettingsNotifier, IOSBackgroundSettings?>(
+  (ref) => IOSBackgroundSettingsNotifier(ref.watch(backgroundServiceProvider)),
 );
 
--- a/mobile/lib/modules/backup/providers/ios_background_settings.provider.dart
+++ b/mobile/lib/modules/backup/providers/ios_background_settings.provider.dart
@@ -52,6 +52,6 @@ class IOSBackgroundSettingsState extends StateNotifier<IOSBackgroundSettings?> {
 }
 
 final iOSBackgroundSettingsProvider = StateProvider<IOSBackgroundSettingsState>(
-  (ref) => IOSBackgroundSettingsState(ref.read(backgroundServiceProvider)),
+  (ref) => IOSBackgroundSettingsState(ref.watch(backgroundServiceProvider)),
 );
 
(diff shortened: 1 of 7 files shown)


mobile/lib/modules/backup/providers/ios_background_settings.provider.dart
```diff
@@ -0,0 +1,57 @@
...
+    );
+
+    state = settings;
+    return settings;
+  }
+
+}
+
+final iOSBackgroundSettingsProvider = StateProvider<IOSBackgroundSettingsState>(
+  (ref) => IOSBackgroundSettingsState(ref.read(backgroundServiceProvider)),
```
alextran1502, line 55: I think we should use `ref.watch` here? is there any reason you are choosing to use `ref.read`?

martyfuhry: Not really sure. I guess watch will update this provider when it changes? This only barely depends on the `backgroundServiceProvider` at all; it just had all the method channel communication functions in it, so I piggybacked there.

I'm still not super familiar with RiverPod and there's likely something I'm not doing correct here. But I made the change and it still seems happily working here.

alextran1502: Riverpod suggests using `ref.watch` as much as possible I believe. Only using `ref.read` in button pressing context.

Here is some good info
https://riverpod.dev/docs/concepts/reading#using-ref-to-interact-with-providers

martyfuhry: Thanks! I gave it a read and it's becoming clearer to me. 

In any event, I changed it to `ref.watch` and looked through the code to find other `ref.read` statements. In this case, the other `ref.read` statements are justified as they are just one-off refreshes (app resume and page load).

mobile/lib/modules/backup/providers/ios_background_settings.provider.dart
```diff
@@ -0,0 +1,57 @@
...
+      timeOfLastProcessing: lastProcessingTime,
+    );
+
+    state = settings;
+    return settings;
+  }
+
+}
+
+final iOSBackgroundSettingsProvider = StateProvider<IOSBackgroundSettingsState>(
```
alextran1502, line 54: If you are using `StateNotifier` I believe you have to use `StateNotifierProvider` here instead of `StateProvider`. Is using `StateProvider` do what you are expecting?

mobile/lib/modules/backup/views/backup_controller_page.dart
```diff
@@ -48,7 +48,7 @@ class BackupControllerPage extends HookConsumerWidget {
         }
 
         if (Platform.isIOS) {
-          ref.read(iOSBackgroundSettingsProvider).refresh();
+          ref.read(iOSBackgroundSettingsProvider.notifier).refresh();
```
alextran1502, line 51: You probably want to use `ref.watch` here as well?

martyfuhry: I thought you weren't supposed to put `ref.watch` in a `useEffect` block. I just wanted a place to refresh the data when we load this page.

You said:
> Riverpod state that you should not user ref.watch in any async/initialization/button press, user ref.read instead
image

And I thought `useEffect` was an initialization? But I may be misunderstanding, because there is a `watch` right above it in this block. Happy to change it to a `watch`.

martyfuhry: Either way, I changed this to `watch`. Thanks!

mobile/lib/modules/backup/providers/ios_background_settings.provider.dart
martyfuhry: Forgot to say I resolved this.

alextran1502 (approved):

PR #1840, opened 2023-02-22
Manually Remove Backup Assets

TCVinNYC:
Allow users to have the option to remove assets already backed up in their Immich server.

(I'm back for this again haha)


zackpollard:
Hey,

Thanks for the PR, we will need to discuss internally if this is the way we want to go about this, however it's definitely a feature we do want, but we've just not really discussed it up to now. There's also a large amount of work going on in the app right now to add offline support and such which takes priority over most other things, so this still might not be a good time to merge something like this, although that work should be finished in a couple weeks I believe. For now, could you include a bit more of an explanation of how this is intended to work, as well as some screenshots from both platforms iOS and Android, or even better, a video showing the process.

Cheers!

TCVinNYC:
Gotcha, no worries! I'll get a little demo video + explanation this Friday/Saturday.

Might give me more time to rethink doing the automated deletion feature instead.

PR #1844, opened 2023-02-23
fix(ci): Mobile build should not run on fork PRs

samip5:
It doesn't have the necessary secrets exposed for it succeed in PR context.

### DIFF
--- a/.github/workflows/build-mobile.yml
+++ b/.github/workflows/build-mobile.yml
@@ -7,6 +7,7 @@ on:
       ref:
         required: false
         type: string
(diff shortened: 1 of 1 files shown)


bo0tzz (changes requested):
It should still run on internal PRs, only on a fork does it not have the necessary secrets, so then it should probably just exit early. Here's how we check for a fork PR in the docker workflow: https://github.com/immich-app/immich/blob/main/.github/workflows/docker.yml#L64

bo0tzz (approved):

PR #1853, opened 2023-02-24
feat(server): improve API specification

michelheusschen:
Changed the `Authenticated` decorator to include the `key` URL query string. Also added `cookie` as an authentication method and included both Bearer and cookie auth in the decorator, so `@Authenticated` doesn't need additional decorators.

Also modified endpoints that return files to have the correct return type. The `{ type: 'string', format: 'binary' }` format used comes from the OpenAPI 3 [spec for file uploads](https://spec.openapis.org/oas/v3.0.3#considerations-for-file-uploads).

As a result of these changes had to make minor adjustments to API calls from the web.

### DIFF
--- a/mobile/openapi/README.md
+++ b/mobile/openapi/README.md
@@ -39,6 +39,16 @@ Please follow the [installation procedure](#installation--usage) and then run th
 ```dart
 import 'package:openapi/api.dart';
 
+// TODO Configure HTTP Bearer authorization: bearer
+// Case 1. Use String Token
+//defaultApiClient.getAuthentication<HttpBearerAuth>('bearer').setAccessToken('YOUR_ACCESS_TOKEN');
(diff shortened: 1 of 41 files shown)


jrasm91 (approved):
I haven't tested it yet, but all the changes should be strictly related to open API correct?

I like auto setting the appropriate decorators automatically in the `@Authenticated` implementation, including the optional key query param. Nice work!

michelheusschen:
> I haven't tested it yet, but all the changes should be strictly related to open API correct?
> 
> I like auto setting the appropriate decorators automatically in the `@Authenticated` implementation, including the optional key query param. Nice work!

Correct, server-side there are only OpenAPI related changes. That has the side effect of changing the generated client APIs and had to change some minor things for the web API.

PR #1858, opened 2023-02-24
fix(web): sharing of access token in server API

michelheusschen:
Fixes a potential bug/security risk, because the web server would retain the access token from the previous request. Theoretically it would be possible for data of another user to be leaked under pretty specific circumstances, but because no API data was being passed from the server to the client, I think there was no risk of data being leaked and haven't been able to produce such a case. I can't say this with absolute certainty however.

The issue originates from creating a global instance of `ImmichAPI` where the access token is replaced for every request. This shouldn't happen and SvelteKit also [makes note](https://kit.svelte.dev/docs/load#shared-state) of this. A new instance of `ImmichAPI` is now created in `hooks.server.ts` for every request and stored in `event.locals` that is available in every load function server-side.

### DIFF
--- a/web/src/api/api.ts
+++ b/web/src/api/api.ts
@@ -1,11 +1,10 @@
-import { browser } from '$app/environment';
-import { env } from '$env/dynamic/public';
 import {
 	AlbumApi,
 	APIKeyApi,
 	AssetApi,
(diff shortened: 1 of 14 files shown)


jrasm91 (approved):
:+1:

PR #1865, opened 2023-02-25
feat(mobile): Adds onboarding for permissions

martyfuhry:
- Adds guards for gallery viewer permissions
- When permissions are required, takes you to an onboarding page which asks for the permissions
- When permissions are denied, shows a message on the onboarding page to go to settings

image

image

image

image


zoodyy:
I've installed a release version of this on an older Android phone it cannot get past "Grant Permission" screen. Pressing the button does not do anything. Using a debug build I get the following log error on the console `permissions_handler(32376): No permissions found in manifest for: []9` My test device has Android 7.1.1 on it :sweat_smile:

martyfuhry:
> I've installed a release version of this on an older Android phone it cannot get past "Grant Permission" screen. Pressing the button does not do anything. Using a debug build I get the following log error on the console `permissions_handler(32376): No permissions found in manifest for: []9` My test device has Android 7.1.1 on it sweat_smile

https://github.com/Baseflow/flutter-permission-handler/issues/944 

Looks like a known issue. There's a simple workaround, so I'll fix it with that quickly.

martyfuhry:
@zoodyy I can't reproduce it, but I saw that this is a new bug in `10.2.0` of the `permission_handler` plugin. I've downgraded to `10.1.0`, as people in that above linked thread suggest that it worked in `10.1.0`. Please try it again with the new version and see if that works.

If not, I can implement the workaround provided in that thread. But it would be better to let them fix the issue and just upgrade when it's fixed.

zoodyy:
@martyfuhry thanks for checking! and tried with the downgraded version to no avail. I guess we need the workaround :-/

martyfuhry:
> @martyfuhry thanks for checking! and tried with the downgraded version to no avail. I guess we need the workaround :-/

Thanks for helping me test! I've added the workaround and some comments about the general workflow. It diverges based on OS, but this is very customizable and extensible. I think this will do exactly what we are looking for, but this needs some testing on:
- [ ] iOS 14+ (limited access)
- [ ] iOS (not limited access)
- [ ] Android 32 and below
- [X] Android 33

martyfuhry:
> I guess we need the workaround

Hey could you try it again and see if this is working for you?

zoodyy:
> > I guess we need the workaround
> 
> Hey could you try it again and see if this is working for you?

Just did! It works now on my old test device! :-)

martyfuhry:
> > > I guess we need the workaround
> > 
> > 
> > Hey could you try it again and see if this is working for you?
> 
> Just did! It works now on my old test device! :-)

Thanks so much! I'm satisfied with the testing suite, then. We'll let @alextran1502 have the last say.

fyfrey (approved):

alextran1502:
Hi Marty, 

Thank you for the PR. I can confirm that the permission-granting process works on iOS.

There are two things I noticed
1. After all permissions are granted, if I sign out and sign back in again, I think it would be a normal flow not to show the permission page again, even with the message confirming that all permissions have been granted. This helps with less clicking on the user, thus improving user experience.
2. The album selection page album card is designed to scroll horizontally. If we want to change the design, the album card look and feel will also need to be redesigned. I suggest we revert to the original design, and we can open a different PR to focus on redesigning that page for better UI/UX.

Please let me know what do you think about those points.

jrasm91:
I have to say, I really dislike the horizonal scrolling and think the vertical cards would be an improvement.

alextran1502:
> I have to say, I really dislike the horizonal scrolling and think the vertical cards would be an improvement.

I don't disagree with you. However, my point was about the card's design for vertical scrolling. Right now on my phone, it shows as one big card, taking up an entire screen's width, which I think it doesn't look right

image

jrasm91:
Ah, I see. Marty's screenshot had two columns, which I thought looked fine for now.

martyfuhry:
>  1. After all permissions are granted, if I sign out and sign back in again, I think it would be a normal flow not to show the permission page again, even with the message confirming that all permissions have been granted. This helps with less clicking on the user, thus improving user experience.

Are you saying that after signing out and signing back in, you are taken to the permission page again? This is certainly undesired behavior. I'll get this fixed.

> 2. The album selection page album card is designed to scroll horizontally. If we want to change the design, the album card look and feel will also need to be redesigned. I suggest we revert to the original design, and we can open a different PR to focus on redesigning that page for better UI/UX.

I agree. I will cherry pick out the commit into a new PR so we can remain focused on onboarding. The grid view needs more responsive design and more serious consideration than what I've given here.

Thank you!

PR #1871, opened 2023-02-25
fix(web): don't log unauthorized errors

michelheusschen:
Fixes minor regression from #1858 where the API gets called when no access token is present. This led to some errors being logged, but it didn't impact any functionality.

This PR goes back to the old behavior of only calling the API when an access token is present and logs the error for a failed API call.

### DIFF
--- a/web/src/hooks.server.ts
+++ b/web/src/hooks.server.ts
@@ -16,8 +16,10 @@ export const handle = (async ({ event, resolve }) => {
 			const { data: user } = await api.userApi.getMyUserInfo();
 			event.locals.user = user;
 		} catch (err) {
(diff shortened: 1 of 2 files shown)


jrasm91 (approved):

michelheusschen:
Please hold off on merging until the Docker build error is fixed

jrasm91:
Just a note - `AxiosError` is used in the `handleError` implementation as well, but only as a type via `(error as AxiosError)?.response` instead of a value/class (via `instanceof`). You could probably do the same here to fix the build problem.

michelheusschen:
Build error is now fixed

PR #1881, opened 2023-02-26
fix(web): layout nesting

### LINKED ISSUE #1827: [BUG] Admin page to photos page routing issue
### The bug

Opening the photos page by clicking the immich logo from admin page, will result to a view as shown in the screenshot below.

image

Ubuntu 22.04

### Version of Immich Server

v1,48

v1.48

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
version: "3.8"

services:
  immich-server:
    container_name: immich_server
    volumes:
      - /mnt/sda/srv/immich/upload:/usr/src/app/upload
    env_file:
      - .env
    environment:
      - NODE_ENV=production
      - TZ=Asia/Singapore
      - IMMICH_MACHINE_LEARNING_URL=false
    depends_on:
      - redis
      - database
    restart: always

  immich-microservices:
    volumes:
      - /mnt/sda/srv/immich/upload:/usr/src/app/upload
    env_file:
      - .env
    environment:
      - NODE_ENV=production
      - IMMICH_MACHINE_LEARNING_URL=false
    depends_on:
      - redis
      - database
    restart: always


  immich-web:
    image: altran1502/immich-web:release
    container_name: immich_web
    entrypoint: ["/bin/sh", "./entrypoint.sh"]
    env_file:
      - .env
    restart: always

  redis:
    container_name: immich_redis
    image: redis:7
    restart: always

  database:
    image: postgres:15
    env_file:
      - .env
    environment:
    #  POSTGRES_PASSWORD: ${DB_PASSWORD}
    #  POSTGRES_USER: ${DB_USERNAME}
    #  POSTGRES_DB: ${DB_DATABASE_NAME}
    volumes:
      - /srv/immich/pgdata:/var/lib/postgresql/data
    restart: always


  immich-proxy:
    container_name: immich_proxy
    image: altran1502/immich-proxy:release

    logging:
      driver: none
    depends_on:
      - immich-server
    restart: always
```

### Your .env content

```Shell
DB_HOSTNAME=immich_postgres
DB_USERNAME=
DB_PASSWORD=
DB_DATABASE_NAME=immich
REDIS_HOSTNAME=immich_redis
UPLOAD_LOCATION=/srv/immich/upload
JWT_SECRET=
ENABLE_MAPBOX=false
```

### Reproduction steps

```bash
1. Open admin page.
2. Click the immich logo to proceed to photos page.
3. The photos page view will show up below the admin page view. The url(/photos) is correct but the view is weird.
4. Refreshing the page will correct the view.
...
```

### Additional information

_No response_

michelheusschen:
- Fixes wrong layout nesting as described below. Fixes #1827
- Drag and drop overlay no longer flickers when dragging a file across the logo or text
- Removed drag and drop overlay from `/admin` and `/auth` pages
- Removed version announcement notification, because it will never show without the `appVersion` item already present in localStorage. Can easily be restored after fixing some issues.

Layouts in Svelte can be nested. We were using layouts as if they would overwrite eachother, but that isn't the case. You can still apply a layout to certain pages by grouping them in a directory with parentheses, like I've done by creating the `(user)` group.

I'm actually pretty surprised this hasn't led to problems earlier on. When switching from a `/admin` page to another, the pages would *sometimes* become stacked on top of eachother. I've tried to reproduce it across 5 instances and only for one instance this would happen. I can't use this instance to validate if the problem is fixed, but theoretically it should.


jrasm91 (commented):
Why do we need to move everything and use this new nested layout? It seems like everything would be quite a bit simpler the way is was before. Basically why do we need to use (user)/layout.ts to fix the issue instead of just preventing upload behavior on certain pages?

michelheusschen:
We are already using a layout for the `/admin` paths. This layout gets nested inside the root layout by SvelteKit, causing the stacked page issue. I've considered exporting something like `{ meta: { allowFileUpload: true } }` from every `+page.server.ts` and then conditionally enabling file uploads. This would work for file uploads, but all pages that are moved to the `(user)` folder in this PR use the general layout which is now duplicated in every page.

Another option is to create a component for this and wrap every page in that, but then you're basically imitating what layouts do. Besides grouping all pages like I've done you can also create a subpath, but either way you have to move all those pages to a subdirectory. A subpath would probably break a lot more things, so that's why I decided to go for the grouping method.

jrasm91:
Maybe because it is hard to visualize the file moves on mobile, but does this basically lead to three top level folders? Admin, user, and auth? If that's the case, it actually would make quite a lot of sense. I assume the parens around user just tell svelte to not include that in the rendered url, so you keep /photos like we already had. I'm actually ok with this as it matches our layout and business logic more accurately without having to change our urls themselves.

michelheusschen:
Correct this leads to three top level folders: `(user)`, `admin` and `auth`. The parenthesis indeed group routes together without affecting the URL.

jrasm91 (approved):

PR #1887, opened 2023-02-27
feat(web): re-add version announcement

michelheusschen:
- Version announcement now actually shows up if `appVersion` wasn't previously present in localStorage
- Use server version to compare against, instead of the last `appVersion` value
- Announcement only shows for admin users
- Small redesign: text better readable and dark mode support

image


web/src/lib/components/shared-components/version-announcement-box.svelte
```diff
@@ -0,0 +1,82 @@
...
+		try {
+			if (localStorage.getItem('appVersion') === serverVersionName) {
+				// Updated version has already been acknowledged.
+				return;
+			}
+
+			githubVersion = await getGithubVersion();
+			if (githubVersion !== serverVersionName) {
+				showModal = true;
+			}
```
bo0tzz, line 34: I think this logic is not quite correct. If the server is on, say, 1.49.0, and that version has been acknowledged and stored locally, this check will return before ever comparing to the githubVersion.

michelheusschen: Good catch! It would be correct for only showing the notification once after the server gets outdated. But I've fixed it and the notification will now show for every update.

bo0tzz (approved):

jrasm91 (approved):

PR #1894, opened 2023-02-28
Add timezone to exif entity

AndreAle94:
Hello,

This pull request introduces a new field in EXIF entity to persist the timezone. For images, this information is extracted by exif-vendored library using [different heuristics](https://github.com/photostructure/exiftool-vendored.js#dates). For videos, the timezone is derived based on the gps location (heuristic 2, if location is available).

The front-end part has been modified to format the timestamp with the correct timezone (if returned by the server). Here is an example with a picture I took in Los Angeles (UTC-7) while the timezone of the web client is UTC+2
Before:
1894_before
After:
1894_after

This pull request implements the first point described in this discussion (#1589). I would like to work also on the second point in order to organize media in the correct bucket in case there is an offset from UTC (in a different PR).


server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -323,6 +337,14 @@ export class MetadataExtractionProcessor {
...
       }
 
+      if (newExif.longitude && newExif.latitude) {
+        try {
+          newExif.timeZone = tz_lookup(newExif.latitude, newExif.longitude);
+        } catch (err) {
+          console.log('Error while calculating timezone from gps coordinates', err);
+        }
+      }
+
```
jrasm91, line 347: This code is unnecessary I believe. Exiftool-vendored actually does this exact process to determine the original timezone information, so we should not need to do it a second time ourselves.

see: https://github.com/photostructure/exiftool-vendored.js#dates

The extraction information is already available here (added to `tags`)

image

jrasm91: Oh, I'm just re-reading your original comment... and we do indeed need this for videos still. My bad :smile: I keep forgetting that we haven't migrated to use this tool for videos as well.

jrasm91: Can you update the console.log to be `this.logger.error` instead?

jrasm91 (approved):
The change as is, is a display change only if I'm understanding correctly. Preserving the original timezone where/when the photo was taken, and then displaying _that_ timezone in the client. Seems like a welcome enhancement.

You mention a "part 2", where the asset will be grouped by the appropriate date in the main timeline, taking the timezone into account. Do you have a plan on how you will achieve that? Currently the assets are sorted by absolute date using standard UTC (0 offset) and the queries (at the api and database level) use that column to for all the timeline related calculations. It is unclear to me how that timeline can (simply) be updated to support different assets taken in different timezones around the same time.

AndreAle94:
Exactly, this change will only impact the UI of the details panel: if a timezone is found, the timestamp will be shown taking it into account. Example: I'm based in Italy (UTC+1 now) and I'm checking the pictures taken in US (UTC-7), I'll see the original time when I took the picture, not the converted one in my local time.

Regarding the second point, I decided to wait to work on it and gather your opinion before doing any change. Since all the timestamps are persisted in UTC, the offset information is lost. My proposal is to add an additional `timestamp` field in Asset entity (not `timestamptz` to avoid conversions on the client side) and persist the `dateTimeOriginal` timestamp already converted in the original timezone. This way, the timestamp will be static regardless the timezone of the client. Then, we could modify the current logic that is grouping the media in buckets to use this new field, instead of `fileCreatedAt`.

Example to clarify the steps: let's consider an image with the following metadata:
- dateTimeOriginal = "2022-09-03 20:44:56-07:00"
- timeZone = "America/Los_Angeles"

After this change, the asset entity will contain the following timestamps:
- fileCreatedAt = "2022-09-04 03:44:56+00:00" -> date is converted in UTC, that's why it's placed in the wrong bucket
- fileCreatedLocal = "2022-09-03 20:44:56" -> no timezone info, the timestamp is already converted in UTC-7

The logic that groups the media in buckets will be updated to use `fileCreatedLocal` instead of `fileCreatedAt`.

The same idea is used on similar solutions. What do you think?

jrasm91:
That would fix the problem of showing the images in wrong buckets, but will it also not lead to assets potentially showing up in the wrong order?

AndreAle94:
The only case it will show the assets in the wrong order will be if they are taken in different timezones in a short amount of time or if one asset doesn't have timezone information. Sorting them by UTC time may help keeping the correct order, we could use the new timestamp only for identifing the correct bucket.

jrasm91:
That may be an option we'll have to see how complicated it is and how well it works. In google photos it looks like they actually have two different buckets for the same day if there are pictures from different timezones in that one day. Probably to help split out photos taken on the same day in different parts of the world, it's easier to view them separately rather than have pictures from two different events inter-mixed.

AndreAle94:
Not sure it's the best approach to have multiple buckets for the same day, it could create confusion from my point of view.

I just tried to upload a couple of picture on Google Photos with different timezone to see what is the behavior there:
- image1 (Death Valley) = "2022:08:27 19:30:15.206-07:00" -> "20220828_023015206 UTC"
- image2 (Horseshoe Band) = "2022:08:27 19:58:25.088-06:00" -> "20220828_015825088 UTC"

In the timeline, they are shown in the same bucket (08/27) without any distinction and they appear in the following order:
- image1 (Death Valley) -> as most recent
- image2 (Horseshoe Band) -> as oldest

immagine

So, it seems they are grouped by local time and then sorted by UTC timestamp, regardless the specific timezone

jrasm91:
Can you help resolve the conflicts and we'll get this merged in.

AndreAle94:
Conflicts resolved!

PR #1916, opened 2023-03-01
Feat/ml image optimisations

ollywelch:
Recreating this PR, as I've realised where I made a mistake. When creating the requirements.txt file I forgot to add in the --pre flag to the pip install torch, so it wasn't downloading the correct version.
I've gone back to the old format of just installing the dependencies using separate commands, I feel like a requirements file doesn't actually make things any easier here!

@alextran1502 I did manage to reproduce the error you saw by building and running on my laptop (x86_64 arch), and have confirmed that after reverting the requirements.txt changes the build works


machine-learning/Dockerfile
```diff
@@ -1,19 +1,25 @@
...
-    PYTHONDONTWRITEBYTECODE=1 \
+ENV PYTHONDONTWRITEBYTECODE=1 \
     PYTHONUNBUFFERED=1 \
     PIP_NO_CACHE_DIR=true
 
-WORKDIR /usr/src/app
+RUN python -m venv /opt/venv && \
+    /opt/venv/bin/pip install --pre torch  -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html && \
+    /opt/venv/bin/pip install transformers tqdm numpy scikit-learn scipy nltk sentencepiece flask Pillow gunicorn && \
+    /opt/venv/bin/pip install --no-deps sentence-transformers
```
jrasm91, line 10: Since this is a builder image, is there any advantage to combining these commands instead of having them separated?

jrasm91 (approved):

machine-learning/Dockerfile
ollywelch: You're right, there's probably no need for this. Will update it to be separate commands :)

PR #1918, opened 2023-03-02
Fix timezone mismatch in server tests

brighteyed:
My local timezone is `UTC+3`. I cloned repository, switched to v1.50.1 tag and run server tests. Two tests failed:
1. All `checkValidTimestamp` tests pass UTC time as input values but implementation of `checkValidTimestamp` uses local time in its internal checks. As a result test `check for year before 1583` fails. So I changed input values from UTC to local time

2. Test `should handle a file upload` passes input asset with UTC time but upload directory is named after local time. So I modified `fileModifiedAt`, `fileCreatedAt` and `updatedAt` values in test asset from UTC to local time

All tests are green now

**UPDATE:** Instead of changing input values from UTC to local time I set timezone in jest config

### DIFF
--- a/server/libs/common/src/utils/time-utils.spec.ts
+++ b/server/libs/common/src/utils/time-utils.spec.ts
@@ -3,6 +3,12 @@
 import { timeUtils } from './time-utils';
 
 describe('Time Utilities', () => {
+  describe('timezone', () => {
+    it('should always be UTC', () => {
+      expect(new Date().getTimezoneOffset()).toBe(0);
+    });
+  });
+
   describe('checkValidTimestamp', () => {
     it('check for year 0000', () => {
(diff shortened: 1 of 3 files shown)


jrasm91:
We do want to test behavior with UTC dates, so I don't think the solution should be removing them from the timestamps.

I think there are two things we should do here:

1. Change check valid timestamp method to use getUTCFullYear() for the comparisons

2. Set the timezone for the tests to UTC in jest, so tests should not break in the future. 

I believe number two should be possible by using a global setup test file to specify an process env for timezone, probably setting it to UTC.

brighteyed:
Thanks for clear explanation! I tried to revert my changes and modify command to `TZ=utc jest`. And that works!

Should I close this PR now or can I modify package.json to change all `jest` commands?

jrasm91 (approved):
LGTM!

PR #1937, opened 2023-03-04
feat(server): improve get albums performance

michelheusschen:
Improved performance for `/api/album` endpoint by optimizing the used query and no longer selecting all assets of all albums. Removed `AlbumRepository.getListByAssetId(ownerId, assetId)` because it has become redundant. Also extended the E2E tests to include shared links and tests for the different query params that the endpoint uses.

Below are some query times from benchmarks ran directly against the DB. I only measured the API time for the 100.000 assets one and before it took 2132ms on average, now the API responds in 40ms.

| Asset count | Old query  | New query |
| ------------- | ------------- |------------- |
| 5 x 5000 = 25.000 | 127ms  | 4.9ms |
| 10 x 5000 = 50.000 | 501ms | 7.8ms  |
| 20 x 5000 = 100.000 | 994ms | 14.9ms  |


michelheusschen:
Will close this for now as maintainability is an issue. This does need improvement though and will get back to this later once I've reworked some things.

PR #1957, opened 2023-03-06
feat(web): improve search bar + add to search page

michelheusschen:
Improved search bar by adding a search icon and a button to quickly clear the search bar. To make follow up searches easier, the search bar is added to the search page. Also made some small styling changes, mostly for better responsive design and zooming in/out. Below some screenshots, but probably best to try it out yourself.

Search in navigation bar
image

Header for `/search` page
image

### DIFF
--- a/web/src/lib/components/shared-components/control-app-bar.svelte
+++ b/web/src/lib/components/shared-components/control-app-bar.svelte
@@ -56,7 +56,9 @@
 			<slot name="leading" />
 		</div>
 
-		<div class="flex place-items-center gap-1 mr-4 ">
+		<slot />
+
+		<div class="flex place-items-center gap-1 mr-4">
 			<slot name="trailing" />
 		</div>
 	</div>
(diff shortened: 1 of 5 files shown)


bo0tzz:
Very nice :D

If you do two searches in a row, the back button ends up toggling between them rather than returning to the page you originally came from.

michelheusschen:
> Very nice :D
> 
> If you do two searches in a row, the back button ends up toggling between them rather than returning to the page you originally came from.

Yup, that should be an easy fix though

bo0tzz:
Thanks :) Now previous searches aren't shown when using the back button, instead it returns directly to the gallery. That's not a big issue to me, just thought I'd mention it.

michelheusschen:
> Thanks :) Now previous searches aren't shown when using the back button, instead it returns directly to the gallery. That's not a big issue to me, just thought I'd mention it.

Not showing previous searches is intentional because I think that's the best approach in this situation. The main method to navigate away from the search page is the back button and having to go through all your previous searches would be pretty annoying. If we were to add the header and sidebar to the search page or add another way of navigating, then going back should cycle through the previous searches.

bo0tzz (approved):
<3

jrasm91 (approved):

PR #1962, opened 2023-03-07
feat(server): start migration to prisma

michelheusschen:
This pull request starts the migration from TypeORM to Prisma. Prisma offers improved productivity, code maintainability and tooling over TypeORM. While this migration has benefits, I've stumbled upon some potential drawbacks that we need to consider.

1. tsvector data type for searching is unsupported, queries that need this column might have to use $queryRaw (exif table)
2. Prisma can't exclude fields by default (e.g. user password). Can be solved by mapping response to a new type with excluded fields, but there's a bit more risk involved.

To kickstart this migration, I've generated a Prisma scheme using introspection with some changes to follow Prisma's naming convention. Many-to-many relations were explicitly defined and I've kept it that way, because using the implicit variant has some caveats..

The repository for API keys now uses Prisma to see it in action. Migrations will still be done by TypeORM, until we've mostly or completely transitioned to Prisma. This does mean we need to maintain two different schema definitions until the migration is complete.


uhthomas:
There has been no activity for 4 months. I'm going to close this.

PR #1981, opened 2023-03-11
feat(web): Make scaling of albums overview more responsive

matthinc:
Problem:

If an album in the overview does not fit in a row it gets pushed to the next one leaving a gap.
This, for example, happens on my 13" MBP and looks following:

Screenshot_2023-03-11_at_16_42_55

Solution:

Use tailwinds grid system together with media breakpoints and a fixed aspect ratio of the preview image and scale thumbnails to always fill the whole row.

Same screen size after the change:

Screenshot 2023-03-11 at 19 35 31

### DIFF
--- a/web/src/routes/(user)/albums/+page.svelte
+++ b/web/src/routes/(user)/albums/+page.svelte
@@ -70,9 +70,7 @@
 			</div>
 
 			<!-- Album Card -->
-			<div
-				class="grid sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-3 xl:grid-cols-4 2xl:grid-cols-5 3xl:grid-cols-6 4xl:grid-cols-7 5xl:grid-cols-8 gap-8"
-			>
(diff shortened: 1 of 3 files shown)


web/src/routes/(user)/albums/+page.svelte
```diff
@@ -70,7 +70,9 @@
 			</div>
 
 			<!-- Album Card -->
-			<div class="flex flex-wrap gap-8">
+			<div
+				class="grid sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-3 xl:grid-cols-4 2xl:grid-cols-5 3xl:grid-cols-6 4xl:grid-cols-7 5xl:grid-cols-8 gap-8"
```
michelheusschen, line 74: Maybe better to use something like `grid gap-8 grid-cols-[repeat(auto-fill,minmax(15rem,1fr))]` instead of defining the amount of columns for every breakpoint and even having to add breakpoints

alextran1502 (approved):

PR #1983, opened 2023-03-12
feat(web): Better scaling of albums and favorites

matthinc:
Rework albums and favorites to match different screen sizes better without scaling thumbnails too large.

Screenshot 2023-03-12 at 13 30 53

Screenshot 2023-03-12 at 13 31 03

### DIFF
--- a/web/src/lib/components/album-page/album-viewer.svelte
+++ b/web/src/lib/components/album-page/album-viewer.svelte
@@ -479,7 +479,7 @@
 		</ControlAppBar>
 	{/if}
 
-	<section class="flex flex-col my-[160px] px-6 sm:px-12 md:px-24 lg:px-40">
+	<section class="flex flex-col my-[160px] px-6 xl:px-40">
 		<input
 			on:keydown={(e) => {
 				if (e.key == 'Enter') {
@@ -522,7 +522,12 @@
 		{/if}
 
 		{#if album.assetCount > 0}
-			<GalleryViewer assets={album.assets} {sharedLink} bind:selectedAssets={multiSelectAsset} />
+			<GalleryViewer
+				assets={album.assets}
+				{sharedLink}
+				bind:selectedAssets={multiSelectAsset}
+				dynamicResponsiveGrid={true}
+			/>
 		{:else}
 			<!-- Album is empty - Show asset selectection buttons -->
 			<section id="empty-album" class=" mt-[200px] flex place-content-center place-items-center">
--- a/web/src/lib/components/shared-components/gallery-viewer/gallery-viewer.svelte
+++ b/web/src/lib/components/shared-components/gallery-viewer/gallery-viewer.svelte
@@ -9,6 +9,7 @@
 	export let assets: AssetResponseDto[];
 	export let sharedLink: SharedLinkResponseDto | undefined = undefined;
 	export let selectedAssets: Set<AssetResponseDto> = new Set();
+	export let dynamicResponsiveGrid: boolean = false;
 
 	let isShowAssetViewer = false;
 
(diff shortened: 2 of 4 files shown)


web/src/lib/components/album-page/album-viewer.svelte
```diff
@@ -479,7 +479,7 @@
 		</ControlAppBar>
 	{/if}
 
-	<section class="flex flex-col my-[160px] px-6 sm:px-12 md:px-24 lg:px-40">
+	<section class="flex flex-col my-[160px] px-6 xl:px-40">
```
matthinc, line 482: Avoid too much "jumping" when scaling the browser window

web/src/lib/components/shared-components/immich-thumbnail.svelte
```diff
@@ -149,8 +154,6 @@
 
 <IntersectionObserver once={false} let:intersecting>
 	<div
-		style:width={`${thumbnailSize}px`}
-		style:height={`${thumbnailSize}px`}
```
matthinc, line 153: Is there a reason we had that? AFAIK this information was already provided by the CSS class?

web/src/lib/components/shared-components/gallery-viewer/gallery-viewer.svelte
```diff
@@ -32,6 +33,12 @@
 		}
 	}
 
+	const gridLayout = `${
+		dynamicResponsiveGrid
+			? 'grid grid-cols-3 lg:grid-cols-[repeat(auto-fill,minmax(15rem,1fr))]'
```
michelheusschen, line 38: Not necessarily wrong, but why a minimum of 3 columns?

web/src/lib/components/shared-components/immich-thumbnail.svelte
michelheusschen: This probably doesn't work. Class names shouldn't contain dynamic values, use style instead

web/src/lib/components/shared-components/immich-thumbnail.svelte
```diff
@@ -250,11 +253,9 @@
 		{#if intersecting}
 			<img
 				id={asset.id}
-				style:width={`${thumbnailSize}px`}
-				style:height={`${thumbnailSize}px`}
 				src={`/api/asset/thumbnail/${asset.id}?format=${format}&key=${publicSharedKey}`}
 				alt={asset.id}
-				class={`object-cover ${getSize()} transition-all z-0 ${getThumbnailBorderStyle()}`}
+				class={`object-cover ${getSize()} h-full v-full transition-all z-0 ${getThumbnailBorderStyle()}`}
```
michelheusschen, line 258: `v-full` doesn't exist right?

michelheusschen (changes requested):
Overall the `ImmichThumbnail` component isn't designed that well. Changing something will likely have unintended side effects in other components or pages that use it. Not saying you should refactor the whole component, but everyting should at least keep working.

Some issues I've seen are:
- Live photos and videos have wrong dimensions when starting playback by hovering over them
- The /search and /explore pages have incorrect height/width for thumbnails

web/src/lib/components/shared-components/immich-thumbnail.svelte
matthinc:
```js
		if (thumbnailSize) {
			return `w-[${thumbnailSize}px] h-[${thumbnailSize}px]`;
		}
```

This was already in `getSize`

web/src/lib/components/shared-components/gallery-viewer/gallery-viewer.svelte
matthinc: Just IMO looks better on mobile screens.

Issue #1801, opened 2023-02-20
[Feature]: (mobile) Persist album sorting preference


matthinc:
It'd be nice if the album sorting preference (most recent, title) would be persistent between app launches.
Should be rather simple using Hive.

![image](https://user-images.githubusercontent.com/6898797/220044380-b6c3da69-1cd5-4b65-83ed-6443c8e8322f.png)

zoodyy:
> @matthinc  I would like to try this one.

Excellent. Use the settings service to persist it

PR #1997, opened 2023-03-14
feat(mobile): persist album sort order

truedem0n:
Problem:
see https://github.com/immich-app/immich/issues/1801

Solution:
persist sorting order using hive


etnoy (commented):
Looks good to me

alextran1502 (approved):

PR #2015, opened 2023-03-18
feat(server): Fallback to text search if machine-learning is disabled

bo0tzz:
I've also reordered the switch-case so that text search is the default (This shouldn't usually make a difference since we always have either the CLIP or TEXT strategy, but I think it makes the code a little clearer anyways).

### DIFF
--- a/server/libs/domain/src/search/search.service.ts
+++ b/server/libs/domain/src/search/search.service.ts
@@ -99,21 +99,18 @@ export class SearchService {
     this.assertEnabled();
 
     const query = dto.q || dto.query || '*';
-    const strategy = dto.clip ? SearchStrategy.CLIP : SearchStrategy.TEXT;
+    const strategy = dto.clip && MACHINE_LEARNING_ENABLED ? SearchStrategy.CLIP : SearchStrategy.TEXT;
     const filters = { userId: authUser.id, ...dto };
 
     let assets: SearchResult<AssetEntity>;
     switch (strategy) {
-      case SearchStrategy.TEXT:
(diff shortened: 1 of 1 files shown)


server/libs/domain/src/search/search.service.ts
```diff
@@ -99,21 +99,18 @@ export class SearchService {
     this.assertEnabled();
 
     const query = dto.q || dto.query || '*';
-    const strategy = dto.clip ? SearchStrategy.CLIP : SearchStrategy.TEXT;
+    const strategy = dto.clip && MACHINE_LEARNING_ENABLED ? SearchStrategy.CLIP : SearchStrategy.TEXT;
```
jrasm91, line 102: Does this need parens for the ternary to apply to both the first two conditions?

jrasm91 (commented):
Looks good

jrasm91 (approved):

server/libs/domain/src/search/search.service.ts
bo0tzz: I added parens at first because I preferred it for readability and the formatter removed them, so I assumed they're not needed for the conditional either.

bo0tzz: Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#table, `&&` takes precedence over ternaries.

jrasm91: Ok perfect.

PR #2019, opened 2023-03-19
fix(nginx): x-forwarded-* headers

michelheusschen:
Fixes issues surrounding X-Forwarded-* headers for both server and web:
1. Invalid login attempts are logged on the server with the wrong IP address
2. Cookies are never set with the `Secure` flag
3. URL.protocol is always `http` and URL.host doesn't include the port
4. `app.set('trust proxy')` on the server doesn't have a value so X-Forwarded-* headers are ignored

These issues have been resolved. Below some more details and choices that have been made for the various headers. I've also added a guide to the documentation regarding reverse proxies.

**Host / X-Forwarded-Host**
Changed to `$http_host` which also includes the port while `$host` strips the port. Removing the port leads to issues when using a different port than `80` or `443`.

**X-Forwarded-Proto**
Nginx only listens on HTTP so the value is always `http`. Often, another reverse proxy handles TLS termination and sets the 'X-Forwarded-Proto' header, allowing us to reuse that value while using '$scheme' as a fallback.

**X-Forwarded-For**
This is the most tricky header, because getting it wrong means the user can supply any IP. The header contains a comma-separated list of IP addresses, with the first IP in the list being the original client IP, followed by any intermediary proxies.

One common method for extracting the correct client IP address from the X-Forwarded-For header, is to supply the proxy depth and then extract the IP address accordingly. However, this requires manual configuration by every user and is therefore more error-prone.

The second method works by getting the first public IP address when checking the list from right to left. This method is incompatible with SvelteKit, but we don't need the client's IP there. I've chosen this method only for ease of configuration.

### DIFF
--- a/docs/docs/guides/reverse-proxy.md
+++ b/docs/docs/guides/reverse-proxy.md
@@ -1,0 +1,26 @@
+---
+sidebar_position: 2
+---
+
+# Reverse Proxy
+
+When deploying Immich it is important to understand that a reverse proxy is required in front of the server and web container. The reverse proxy acts as an intermediary between the user and container, forwarding requests to the correct container based on the URL path.
+
+## Default Reverse Proxy
+
+Immich provides a default nginx reverse proxy preconfigured to perform the correct routing and set the necessary headers for the server and web container to use. These headers are crucial to redirect to the correct URL and determine the client's IP address.
+
+## Using a Different Reverse Proxy
+
+While the reverse proxy provided by Immich works well for basic deployments, some users may want to use a different reverse proxy. Fortunately, Immich is flexible enough to accommodate different reverse proxies. Users can either:
+
+1. Add another reverse proxy on top of Immich's reverse proxy
+2. Completely replace the default reverse proxy
+
+## Adding a Custom Reverse Proxy
+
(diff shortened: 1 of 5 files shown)


jrasm91 (approved):

docs/docs/guides/reverse-proxy.md
bo0tzz, line 1: Should this live in the `administration` category instead?

docs/docs/guides/reverse-proxy.md
```diff
@@ -0,0 +1,26 @@
...
+1. Add another reverse proxy on top of Immich's reverse proxy
+2. Completely replace the default reverse proxy
+
+## Adding a Custom Reverse Proxy
+
+Users can deploy a custom reverse proxy that forwards requests to Immich's reverse proxy. This way, the new reverse proxy can handle TLS termination, load balancing, or other advanced features, while still delegating routing decisions to Immich's reverse proxy. All reverse proxies between Immich and the user must forward all headers and set the `Host`, `X-Forwarded-Host`, `X-Forwarded-Proto` and `X-Forwarded-For` headers to their appropriate values. By following these practices, you ensure that all custom reverse proxies are fully compatible with Immich.
+
+## Replacing the Default Reverse Proxy
+
+Replacing Immich's default reverse proxy is an advanced deployment and support may be limited. When replacing Immich's default proxy it is important to ensure that requests to `/api/*` are routed to the server container and all other requests to the web container. Additionally, the previously mentioned headers should be configured accordingly.
```
bo0tzz, line 26: Would it be helpful to link our default nginx config as a reference here?

docs/docs/guides/reverse-proxy.md
michelheusschen: I was initially hesitating between the administration and guides categories. I thought administration only included changing settings via the web interface, but I now also see server commands. Will move it to the administration category.

docs/docs/guides/reverse-proxy.md
michelheusschen: Definitely, I'll add it

bo0tzz (approved):

PR #2022, opened 2023-03-20
fix(server): Return the original path for gif playback

samip5:
Usually browser is able to play them directly.

### DIFF
--- a/server/apps/immich/src/api-v1/asset/asset.service.ts
+++ b/server/apps/immich/src/api-v1/asset/asset.service.ts
@@ -279,19 +279,7 @@ export class AssetService {
         /**
          * Serve file viewer on the web
          */
-        if (query.isWeb && asset.mimeType == 'image/gif') {
-          res.set({
-            'Content-Type': asset.mimeType,
-          });
-
-          if (await processETag(asset.originalPath, res, headers)) {
-            return;
-          }
-          await fs.access(asset.originalPath, constants.R_OK | constants.W_OK);
-          fileReadStream = createReadStream(asset.originalPath);
-
-          return new StreamableFile(fileReadStream);
-        } else if (query.isWeb) {
+        if (query.isWeb) {
           res.set({
             'Content-Type': 'image/jpeg',
           });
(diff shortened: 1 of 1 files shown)


server/apps/immich/src/api-v1/asset/asset.service.ts
```diff
@@ -279,17 +279,32 @@ export class AssetService {
...
+          });
+
+          if (await processETag(asset.originalPath, res, headers)) {
+            return;
+          }
+          await fs.access(asset.originalPath, constants.R_OK | constants.W_OK);
+          fileReadStream = createReadStream(asset.originalPath);
+
+          return new StreamableFile(fileReadStream);
+        } else if (query.isWeb) {
```
jrasm91, line 294: Man, all of these situations are basically the same:
- Some conditions
- processETag
- fs.access check
- create read stream
- return streamable file

We should probably refactor all the if/else statements to just set a `filepath` variable, and then do the above steps _once_ with the path (or throw an error if the path is empty).  You don't need to do that, I'm just ranting btw.

It looks like your "return the original file" code is exactly the same as another condition further down. Can you update that condition instead of adding a new one? (`if((!query.isThumb && allowOriginalFile) || (query.isWeb && asset.mimeType === 'image/gif'))`

samip5: Somehow I had missed that block further down... Added it there now. :)

server/apps/immich/src/api-v1/asset/asset.service.ts
```diff
@@ -299,7 +302,7 @@ export class AssetService {
         /**
          * Serve thumbnail image for both web and mobile app
          */
-        if (!query.isThumb && allowOriginalFile) {
+        if ((!query.isThumb && allowOriginalFile) || (query.isWeb && asset.mimeType === 'image/gif')) {
```
jrasm91, line 305: I don't think you'll reach this statement, right? I think you need `asset.mimeType !== 'image/gif'` on line 282.

jrasm91 (approved):

PR #2036, opened 2023-03-21
refactor(mobile): migrate all Hive boxes to Isar database

fyfrey:
This PR removes all Hive usage across the mobile app and replaces it with the custom `Store` class.
The `Store` is now also type-safe.

Migrations for existing Hive boxes are performed once on app start. After successful migration, the Hive boxes are removed.

We can delete the (generated) Hive classes and migration once enough releases have passed (so that users had a chance to update).

### DIFF
--- a/mobile/lib/modules/home/ui/profile_drawer/profile_drawer_header.dart
+++ b/mobile/lib/modules/home/ui/profile_drawer/profile_drawer_header.dart
@@ -1,14 +1,13 @@
 import 'dart:math';
 
 import 'package:flutter/material.dart';
-import 'package:flutter_hooks/flutter_hooks.dart';
-import 'package:hive_flutter/hive_flutter.dart';
+import 'package:flutter_hooks/flutter_hooks.dart' hide Store;
 import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:image_picker/image_picker.dart';
-import 'package:immich_mobile/constants/hive_box.dart';
 import 'package:immich_mobile/modules/home/providers/upload_profile_image.provider.dart';
 import 'package:immich_mobile/modules/login/models/authentication_state.model.dart';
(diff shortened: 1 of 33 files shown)


mobile/lib/modules/home/ui/profile_drawer/profile_drawer_header.dart
```diff
@@ -19,7 +18,7 @@ class ProfileDrawerHeader extends HookConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    String endpoint = Hive.box(userInfoBox).get(serverEndpointKey);
+    final String? endpoint = Store.get(StoreKey.serverEndpoint);
```
martyfuhry, line 21: `final String endpoint` since `Store.get` will throw an exception if the value is null. Maybe you wanted `tryGet`.

mobile/lib/shared/models/store.dart
```diff
@@ -72,76 +84,113 @@ class StoreValue {
...
   final int id;
   final Type type;
-  final dynamic Function(Isar, int)? fromDb;
-  final Future<int> Function(Isar, dynamic)? toDb;
-  final Function(dynamic)? fromJson;
+  final T? Function<T>(Isar, int)? fromDb;
+  final Future<int> Function<T>(Isar, T)? toDb;
+}
+
+T? _getUser<T>(Isar db, int i) {
```
martyfuhry, line 186: Top level private functions are a bit odd. What are these for again?

martyfuhry (commented):
Looks great! I haven't had the chance to run it here yet, but I will soon. Thanks!

mobile/lib/modules/home/ui/profile_drawer/profile_drawer_header.dart
fyfrey: removed the `?`

PR #2043, opened 2023-03-22
feat(mobile): Use new search API and GridView for Places / Locations

martyfuhry:
## Places and Things
- Uses a `GridView` with 140 max extent, so wider displays will show more elements per row at a max size of 140 device pixels.
- Same layout for Things, but I don't have any Things to show... I'm not sure if "Things" are relevant for the new search, so maybe they should be suppressed?
- I didn't include the "broken image" looking element for when there are no places or things. I don't have any UI element to show that there are no places or things currently; it's just empty with the headers visible. I'd be happy to put something in for the empty element case!
- Still shows the loading throbbers while waiting to load these in.

image

## Search Results
- Uses CLIP with the new `SearchApi`.
- Didn't change anything else; they seem to show up in a "timeline" like view.

image

### DIFF
--- a/mobile/assets/i18n/en-US.json
+++ b/mobile/assets/i18n/en-US.json
@@ -246,5 +246,7 @@
   "permission_onboarding_log_out": "Log out",
   "login_form_next_button": "Next",
   "album_thumbnail_shared_by": "Shared by {}",
-  "album_thumbnail_owned": "Owned"
+  "album_thumbnail_owned": "Owned",
+  "curated_object_page_title": "Things",
+  "curated_location_page_title": "Places"
 }
--- a/mobile/assets/i18n/en-US.json
+++ b/mobile/assets/i18n/en-US.json
@@ -248,5 +248,6 @@
   "album_thumbnail_shared_by": "Shared by {}",
   "album_thumbnail_owned": "Owned",
   "curated_object_page_title": "Things",
-  "curated_location_page_title": "Places"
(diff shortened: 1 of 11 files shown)


jrasm91:
Can you add a screenshot?

martyfuhry:
> Can you add a screenshot?

I sure can! I didn't want to add a screenshot of the Explore Locations section for obvious reasons, but I can blur stuff out.

alextran1502:
@martyfuhry Thank you for the PR. The search integration is perfect, I love how minimal of a change we need to do.

I would like to keep the layout of the Places and Things to be horizontally scrolling but with some small tweak

1. Only show the maximum of 4 results for each category. Then on the row where we put Things and Places, you can have a button to "View all" which then navigate to a different page to show all the entry with the grid layout that you have. This would keep the UI/UX similar to GPhotos's flow.
2. The reason that it is showing in a horizontal fashion is that we will incorporate more grouping categories, like the Explore pages on the web. Here is how GPhotos does the grouping
image

martyfuhry:
I've made the adjustments:
- I thought it was better to show (at most) 10 in a horizontally scrolling list here. I can change that, of course.  For example, I could show all of them, or don't let you scroll at all. Up to you.
- Tapping View all shows the grid view page.
- When the curated content is empty, the grid just shows a broken image looking thing, just like in the search row.

**Search Page**
image

**View More Grid Page**
image

**Empty Objects / Locations**
image

PR #2048, opened 2023-03-22
feat(server): improve and refactor get all albums

michelheusschen:
Continuation of #1937 to improve the performance of the `/api/album` endpoint. Used default TypeORM functions instead of the query builder where possible. Although I originally planned to refactor the entire album repo, I decided against it because it would make tracking changes more difficult.

- Improved listing albums performance. Listing 20 albums with 5000 assets each (100.000 assets total) now takes less than 50ms (before >2s)
- Fix bug where `/api/album?shared=false` would include albums with shared links
- Better validation for query paramters
- Extended album E2E tests
- Refactored to using domain/infra
- Refactored updating thumbnails


jrasm91 (approved):
The code all looks good. Great job! This is a lot of core functionality that's being touched, so we should make sure to test it as much as possible. I'll try to spin it up later today.

PR #2058, opened 2023-03-23
fix(server): invalid video duration format

michelheusschen:
Formatting of video duration doesn't account for decimal numbers, so the duration ends up like `01:02:03.456.000000` instead of the intended `01:02:03.000000`. I've used Luxon instead for formatting and added support for millisecond precision. Now, durations will be formatted like `01:02:03.456`.


bo0tzz (approved):
Very good

jrasm91 (approved):

PR #2059, opened 2023-03-23
 chore: bump openapi version to v1.51.2

michelheusschen:
Can't this be automated?


bo0tzz:
I thought it was automated 🤔. It definitely should be.

zackpollard:
It was missed in the original release automation, it needs to be added

PR #2072, opened 2023-03-24
feat(server): Add support for client-side hashing

etnoy:
This is the first major PR related to client side hashing. Here, we focus on getting the server-side plumbing done. 

* New server endpoint for checking for duplicates via checksum
* No changes to how mobile uploads work, future PRs will be able to improve dupe handling with checksums
* Removed the duplication checking when uploading via web, dedupes will be handled server side only in this case
* Updating checksum column to be non-nullable, this was determined to be safe: https://discordapp.com/channels/979116623879368755/1063891307149414410/1088819797518712882
* CLI uploads will require the hash-enabled CLI version: https://github.com/immich-app/CLI/tree/hash


bo0tzz:
> Mobile apps don't do hashing yet

Is it worth doing that in a separate PR, to keep review sizes down?

server/libs/infra/src/db/entities/asset.entity.ts
```diff
@@ -71,7 +71,7 @@ export class AssetEntity {
 
   @Column({ type: 'bytea', nullable: true, select: false })
   @Index({ where: `'checksum' IS NOT NULL` }) // avoid null index
```
michelheusschen, line 73: Removing null doesn't make the column nullable, you have change that in the `@Column` decorator and for this specific instance the `@Index` also. But is this change backwards compatible?

server/libs/domain/src/media/media.service.ts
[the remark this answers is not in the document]
bo0tzz: I don't think we need to change the paths in this PR at all, and instead leave that to #2047

server/libs/infra/src/db/migrations/1679572446996-DeviceIdOptional.ts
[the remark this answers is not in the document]
etnoy: I could need some help in getting the migrations right. The goal is to make deviceId and deviceAssetId optional (ie remove the NOT NULL requirement)

server/libs/infra/src/db/entities/asset.entity.ts
etnoy: Thanks, can you look at the new version please?

server/libs/domain/src/media/media.service.ts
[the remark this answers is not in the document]
etnoy: Never mind, I was confused about the diff view.

These changes are needed because deviceId is no longer available here. I belive the impact will be minimal, only new items will get the new path.

server/libs/infra/src/db/entities/asset.entity.ts
michelheusschen: Yes that's good, you can also remove the `@Index` because checksum can never be null

server/libs/infra/src/db/migrations/1679572446996-DeviceIdOptional.ts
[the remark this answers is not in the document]
etnoy: Done

server/libs/infra/src/db/entities/asset.entity.ts
michelheusschen: Sorry, I meant to say remove the where part, so you just have `@Index()`. There is still a composite index generated for ownerId and checksum, but it's probably safer to add the index back.

server/libs/infra/src/db/migrations/1679572446996-DeviceIdOptional.ts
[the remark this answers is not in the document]
michelheusschen: Make sure to reflect making these columns nullable in DTOs and other places where it's used

Also, always check the generated migrations, sometimes it creates weird migrations. Now it drops `UQ_userid_checksum` only to recreate it right after, not a big issue now since it's just a constraint but sometimes it does that to columns.

server/libs/domain/src/media/media.service.ts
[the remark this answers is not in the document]
bo0tzz: The PR I mentioned also removes deviceId. At the moment this diff is necessary for this branch to build of course, but once that PR merges it should replace this change IMO.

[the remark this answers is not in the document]
etnoy: I discussed with @jrasm91 and this PR will be a band-aid fix to get things going, the real change will be in #2047. 

We could also skip changing media service here, but then thumbs will be in `thumbs/[Object object]` and originals will be in `originals/undefined` :)

server/libs/infra/src/db/entities/asset.entity.ts
etnoy: After a discussion in discord I think we keep the composite index only

server/libs/infra/src/db/migrations/1679662469076-RemoveChecksumIndex.ts
```diff
@@ -0,0 +1,21 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class removeChecksumIndex1679662469076 implements MigrationInterface {
+  name = 'removeChecksumIndex1679662469076';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX "public"."IDX_64c507300988dd1764f9a6530c"`);
+    await queryRunner.query(`ALTER TABLE "assets" DROP CONSTRAINT "UQ_userid_checksum"`);
```
michelheusschen, line 8: I'm guessing you manually removed the `ADD CONSTRAINT "UQ_userid_checksum"`? You should also remove this query, otherwise you're only removing the constraint here and not adding it back later.

server/apps/immich/src/api-v1/asset/asset-repository.ts
```diff
@@ -285,25 +285,22 @@ export class AssetRepository implements IAssetRepository {
       },
       select: ['deviceAssetId'],
     });
-    const res: string[] = [];
+    const res: (string | null)[] = [];
```
michelheusschen, line 288: Probably better to filter out null values if there's no reason to keep them included

server/apps/immich/src/api-v1/asset/dto/check-existence-of-assets.dto.ts
```diff
@@ -0,0 +1,19 @@
+import { Type } from 'class-transformer';
+import { IsArray, IsNotEmpty, IsString, ValidateNested } from 'class-validator';
+
+export class CheckExistenceOfAssetByChecksumDto {
+  @IsString({ each: true })
+  @IsNotEmpty({ each: true })
+  id!: string;
+
+  @IsString({ each: true })
+  @IsNotEmpty({ each: true })
```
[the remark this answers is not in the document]
etnoy: Thanks

server/apps/immich/src/api-v1/asset/asset.service.ts
```diff
@@ -451,6 +459,41 @@ export class AssetService {
...
 
+  async checkExistenceOfAssetsByChecksum(
+    authUser: AuthUserDto,
+    checkExistenceOfAssetsDto: CheckExistenceOfAssetsByChecksumDto,
+  ): Promise<CheckExistenceOfAssetsResponseDto> {
+    const queriedChecksums: Buffer[] = checkExistenceOfAssetsDto.assets.map((asset) =>
+      Buffer.from(asset.checksum, 'hex'),
+    );
+
+    const existingAssets: Promise<AssetEntity[]> = this._assetRepository.getAssetsByChecksums(
```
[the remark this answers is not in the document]
etnoy: I'm new to ts, but wouldn't that block for every array element? I thought this is faster since it just blocks once, allowing all elements to finish asynchronously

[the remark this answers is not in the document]
michelheusschen: No it wouldn't block for every element. The repository will make one database call and return all elements. You can see the return type for getAssetsByChecksums is `Promise<AssetEntity[]>`. You can just await the Promise and the type will become `AssetEntity[]`.

If the return type would be `Promise<AssetEntity>[]` then you have to await each individual element.

server/libs/infra/src/db/migrations/1679662469076-RemoveChecksumIndex.ts
etnoy: We probably need to regenerate the migrations for this PR as they became such a mess. I might need some help on that when we approach merge tiem

michelheusschen: I've updated the migrations

web/src/lib/utils/file-uploader.ts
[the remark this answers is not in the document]
etnoy: File-uploader has now been updated to use this api

server/libs/infra/src/migrations/1679662469076-RemoveChecksumIndex.ts
michelheusschen, line 1: Please merge all migration files into one

server/apps/immich/src/api-v1/asset/asset.service.ts
```diff
@@ -130,7 +136,8 @@ export class AssetService {
 
       // handle duplicates with a success response
       if (error instanceof QueryFailedError && (error as any).constraint === 'UQ_userid_checksum') {
-        const duplicate = await this.getAssetByChecksum(authUser.id, file.checksum);
+        const checksums = [file.checksum, livePhotoFile?.checksum].filter((checksum) => !!checksum) as Buffer[];
```
michelheusschen, line 139:
```suggestion
        const checksums = [file.checksum, livePhotoFile?.checksum].filter((checksum): checksum is Buffer => !!checksum);
```

server/apps/immich/src/api-v1/asset/asset.controller.ts
```diff
@@ -344,6 +346,19 @@ export class AssetController {
     return await this.assetService.checkExistingAssets(authUser, checkExistingAssetsDto);
   }
 
+  /**
+   * Checks if assets exist by checksums
+   */
+  @Authenticated()
+  @Post('/bulk-upload-check')
+  @HttpCode(207)
```
michelheusschen, line 354: Why response code `207`?

server/apps/immich/src/api-v1/asset/asset-repository.ts
```diff
@@ -291,49 +292,48 @@ export class AssetRepository implements IAssetRepository {
...
         isVisible: true,
+        deviceAssetId: Not(IsNull()),
       },
-      select: ['deviceAssetId'],
     });
-    const res: string[] = [];
-    rows.forEach((v) => res.push(v.deviceAssetId));
 
-    return res;
+    return items.map((asset) => asset.deviceAssetId) as string[];
```
michelheusschen, line 305:
```suggestion
    return items.map((asset) => asset.deviceAssetId);
```

server/apps/immich/src/api-v1/asset/asset-repository.ts
```diff
@@ -291,49 +292,48 @@ export class AssetRepository implements IAssetRepository {
...
    */
   async getAllByDeviceId(ownerId: string, deviceId: string): Promise<string[]> {
-    const rows = await this.assetRepository.find({
+    const items = await this.assetRepository.find({
+      select: { deviceAssetId: true },
       where: {
         ownerId,
         deviceId,
         isVisible: true,
+        deviceAssetId: Not(IsNull()),
```
michelheusschen, line 301: The `deviceAssetId` check seems unnecessary, because it's not nullable

michelheusschen (dismissed):
The server implementation seems fine, but for web there are quite some things that would break upload in one way or another.

server/apps/immich/src/api-v1/asset/asset-repository.ts
etnoy: Thanks, removing it

web/src/lib/utils/file-uploader.ts
```diff
@@ -97,26 +105,27 @@ async function fileUploader(
...
+			status === 200 &&
+			data.results[0].reason === AssetBulkUploadCheckResultReasonEnum.Duplicate
+		) {
 			if (albumId) {
-				await addAssetsToAlbum(albumId, [data.id], sharedKey);
+				await addAssetsToAlbum(albumId, [data.results[0].id], sharedKey);
 			}
 
-			return data.id;
+			return asset.name;
```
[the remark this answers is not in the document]
etnoy: I discussed the web uploader with Alex and we decided to skip hashing altogether in the web uploader

server/libs/infra/src/entities/asset.entity.ts
```diff
@@ -27,7 +27,7 @@ export class AssetEntity {
   @PrimaryGeneratedColumn('uuid')
   id!: string;
 
-  @Column()
+  @Column({ type: 'varchar' })
```
jrasm91, line 30: I don't think this change is necessary, it should be the default type.

server/libs/infra/src/entities/asset.entity.ts
```diff
@@ -36,7 +36,7 @@ export class AssetEntity {
   @Column()
   ownerId!: string;
 
-  @Column()
+  @Column({ type: 'varchar' })
```
jrasm91, line 39: Same here

server/libs/infra/src/migrations/1679662469076-RemoveChecksumIndex.ts
jrasm91, line 1: There two migration files should be merged, and `remove` should be capitalized (`RemoveChecksum...`)

web/src/lib/utils/file-uploader.ts
```diff
@@ -102,23 +102,6 @@ async function fileUploader(
...
-		);
-
-		if (status === 200 && data.isExist && data.id) {
-			if (albumId) {
-				await addAssetsToAlbum(albumId, [data.id], sharedKey);
-			}
-
-			return data.id;
-		}
-
```
jrasm91, line 121: The web no longer does deduping, which I think is fine, since anything besides a hash is less than ideal and hashing in the UI is impractical. However, you've also removed some `addAssetsToAlbum` code. Did you uploading assets directly to an album? I'm assuming this needs to be added back.

server/apps/immich/src/api-v1/asset/asset-repository.ts
```diff
@@ -174,7 +175,7 @@ export class AssetRepository implements IAssetRepository {
   async getDetectedObjectsByUserId(userId: string): Promise<CuratedObjectsResponseDto[]> {
     return await this.assetRepository.query(
       `
-        SELECT DISTINCT ON (unnest(si.objects)) a.id, unnest(si.objects) as "object", a."resizePath", a."deviceAssetId", a."deviceId"
+        SELECT DISTINCT ON (unnest(si.objects)) a.id, unnest(si.objects) as "object", a."resizePath", a."checksum", a."deviceId"
```
jrasm91, line 178: I don't think this change is necessary, since we still have device asset id everywhere.

server/apps/immich/src/api-v1/asset/asset-repository.ts
```diff
@@ -188,7 +189,7 @@ export class AssetRepository implements IAssetRepository {
   async getLocationsByUserId(userId: string): Promise<CuratedLocationsResponseDto[]> {
     return await this.assetRepository.query(
       `
-        SELECT DISTINCT ON (e.city) a.id, e.city, a."resizePath", a."deviceAssetId", a."deviceId"
+        SELECT DISTINCT ON (e.city) a.id, e.city, a."resizePath", a."checksum", a."deviceId"
```
jrasm91, line 192: Same.

jrasm91 (commented):
I think the scope of this PR has changed a lot over time. It'd be good to update the description to match what's actually being changed here. From what I can tell it is:
- New server endpoint for checking for duplicates via checksum
- No changes to how mobile uploads work
- Removing unique check from web
- Updating checksum column to be non-nullable

If that is all that is changing, we should remove any unrelated changes, like changes to get the objects and locations for user.

server/apps/immich/src/api-v1/asset/asset.controller.ts
jrasm91: I think the original idea was to communicate that the response would containe multiple response codes, one for each item. However, we aren't returning multiple _http response_ codes, just our own custom enums, so I think a 200 would probably 
be more appropriate.

etnoy: Fixed

server/apps/immich/src/api-v1/asset/asset-repository.ts
etnoy: Device asset id isn't null, but it will often be empty. Does this change the equation?

web/src/lib/utils/file-uploader.ts
michelheusschen: Just below this, the file gets uploaded and `addAssetsToAlbum` is called.  So I assume uploading to albums will still work

etnoy: I think Michel is right

server/apps/immich/src/api-v1/asset/asset.service.ts
```diff
@@ -128,7 +134,8 @@ export class AssetService {
 
       // handle duplicates with a success response
       if (error instanceof QueryFailedError && (error as any).constraint === 'UQ_userid_checksum') {
-        const duplicate = await this.getAssetByChecksum(authUser.id, file.checksum);
+        const checksums = [file.checksum, livePhotoFile?.checksum].filter((checksum): checksum is Buffer => !!checksum);
+        const [duplicate] = await this._assetRepository.getAssetsByChecksums(authUser.id, checksums);
```
michelheusschen, line 138: When uploading a live photo, the duplicate could be either `file` or `livePhotoFile` because the return order is undefined, no?

server/apps/immich/src/api-v1/asset/asset.service.ts
```diff
@@ -463,7 +470,40 @@ export class AssetService {
...
+        if (duplicate) {
+          return {
+            id,
+            assetId: duplicate,
+            action: AssetUploadAction.REJECT,
+            reason: AssetRejectReason.DUPLICATE,
+          };
+        }
+
+        // TODO mime-check
```
michelheusschen, line 499: Is mime check planned for this PR?

server/apps/immich/src/api-v1/asset/asset.service.ts
jrasm91: There is an existing bug here, in the scenario where the live photo is a duplicate and the still is not uploaded yet. This code used to look up the still asset by checksum and return it, but since it didn't exist it would throw an EntityNotFound and that asset would fail to backup leading to some sync problem on the mobile app. It's unclear to me what should happen in that situation. I think returning the id to either part is fine. Really, I think, they should be two separate upload requests, which would also solve the issue. That requires mobile app changes and does not seem to be a priority for the mobile devs.

web/src/lib/utils/file-uploader.ts
jrasm91: Ah ok. This was just another code branch where it would skip the upload and still add to album. No applicable anymore :+1:.

server/apps/immich/src/api-v1/asset/asset.service.ts
etnoy: The TODO note is there because that's the perfect place to put a pre-upload mime check in the future

server/apps/immich/src/api-v1/asset/asset.service.ts
etnoy: Thanks, Jason. I think we can resolve this in this PR and track the issue somewhere else

jrasm91: Sounds good. I think this does semi-resolve an existing bug too btw.

jrasm91 (approved):

michelheusschen (approved):

server/apps/immich/src/api-v1/asset/dto/asset-check.dto.ts
```diff
@@ -0,0 +1,19 @@
+import { Type } from 'class-transformer';
+import { IsArray, IsNotEmpty, IsString, ValidateNested } from 'class-validator';
+
+export class AssetBulkUploadCheckItem {
+  @IsString()
+  @IsNotEmpty()
+  id!: string;
```
alextran1502, line 7: What is the would we fill in for the ID for the incoming asset?

etnoy: The client is free to choose any id that makes sense for it

jrasm91: Instead of assuming the return/response order is going to be the same, the client sends an ID to keep track of individual items. The ID is returned back along with the result (accepted/rejected).

alextran1502: Ah got it. Thank you for clarifying!

PR #2083, opened 2023-03-25
fix(machine-learning): Better memory management

samip5:
This PR does the following:

- Updates editorconfig to follow PEP8 for Python files.
- Cached models will expire after an hour to release RAM which will then be re-cached when needed.
- Use Flask's ThreadPoolExecutor.

### DIFF
--- a/machine-learning/src/main.py
+++ b/machine-learning/src/main.py
@@ -3,6 +3,7 @@ from flask import Flask, request
 from transformers import pipeline
 from sentence_transformers import SentenceTransformer, util
 from PIL import Image
+from multiprocessing import Process, Queue
 
 is_dev = os.getenv('NODE_ENV') == 'development'
 server_port = os.getenv('MACHINE_LEARNING_PORT', 3003)
@@ -14,6 +15,8 @@ clip_image_model = os.getenv('MACHINE_LEARNING_CLIP_IMAGE_MODEL', 'clip-ViT-B-32
 clip_text_model = os.getenv('MACHINE_LEARNING_CLIP_TEXT_MODEL', 'clip-ViT-B-32')
 
 _model_cache = {}
+
+
 def _get_model(model, task=None):
   global _model_cache
   key = '|'.join([model, str(task)])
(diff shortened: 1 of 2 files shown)


machine-learning/src/main.py
```diff
@@ -14,59 +17,87 @@
...
+        # set a timeout of 1 hour for the model object
+        _model_cache[key + '|timestamp'] = datetime.now() + timedelta(hours=1)
     else:
-      _model_cache[key] = SentenceTransformer(model)
-  return _model_cache[key]
+        # check if the model object has expired
+        timestamp = _model_cache.get(key + '|timestamp')
+        if timestamp and datetime.now() > timestamp:
+            del _model_cache[key]
+            del _model_cache[key + '|timestamp']
```
jrasm91, line 38: When I was testing earlier, I deleted the model from the cache, but the memory usage did not go down. Did you observe something different?

samip5: No I didn't see really chance but I also did not let it run for over an hour. I also think that the whole code needs to be profiled to figure out which portion is the largest memory hog (I suspect the model cache that wasn't being shared previously between workers). I'm going to try to do that tomorrow.

I was originally trying to use multiprocessing but couldn't quite get it to work with the clip models.

jrasm91: Why don't you change it to one minute and test? 😁

I deleted it immediately and it never went down... So idk lol.

thomasverelst: You might need to call Python's garbage collector to release the memory, otherwise it might take a while until it is released. 

```
import gc
gc.collect()
```

jrasm91: I instantiated the model, then deleted the object, then called gc collect and memory usage remain unchanged.

jrasm91: I gave up at that point. I assume there is something going on in the underlying libraries with how they manage loading models. They might have their own cache, etc. Would probably require more research to figure it out, but I moved on to other things 😛.

jrasm91: I gave up at that point. I assume there is something going on in the underlying libraries with how they manage loading models. They might have their own cache, etc. Would probably require more research to figure it out, but I moved on to other things 😛.

samip5:
Yeah, I don't know how to improve it as those that I tried did not really make a difference..

PR #2084, opened 2023-03-25
feat(server): add transcode presets

brighteyed:
I have a lot of 4K videos in my library. They require a decent network bandwidth (for example, 153MB for 17 seconds video) to be playable without interruptions.

I could use `transcode all` option but:

- it transcodes 1080p videos in my library also so they loose quality
- more storage space is used

I replaced `transcode all` switch with `<select>` with the following transcode presets:

- "Only videos not in the desired format" (default, the same as `transcodeAll=false` in the current version of Immich)
- "All videos" (the same as `transcodeAll=true` in the current version of Immich)
- "Videos higher than 1080p or not in the desired format" (the same as default but also transcodes videos with higher resolution than 1080p)

select-opened
select-closed-dark


server/libs/infra/src/db/entities/system-config.entity.ts
```diff
@@ -40,7 +40,7 @@ export interface SystemConfig {
     targetVideoCodec: string;
     targetAudioCodec: string;
     targetScaling: string;
-    transcodeAll: boolean;
+    transcode: string;
```
bo0tzz, line 43: It would be better to use an enum type instead of a string for this setting

bo0tzz (commented):
I like this idea, thank you!

server/libs/infra/src/db/entities/system-config.entity.ts
brighteyed: Done!

server/apps/microservices/src/processors/video-transcode.processor.ts
```diff
@@ -91,6 +92,13 @@ export class VideoTranscodeProcessor {
     if (longestVideoStream.codec_name !== config.ffmpeg.targetVideoCodec) {
       return this.runFFMPEGPipeLine(asset, savedEncodedPath);
     }
+
+    if (config.ffmpeg.transcode == TranscodePreset.OPTIMAL) {
+      const videoHeightThreshold = 1080;
+      if (!longestVideoStream.height || longestVideoStream.height > videoHeightThreshold) {
+        return this.runFFMPEGPipeLine(asset, savedEncodedPath);
+      }
+    }
```
michelheusschen, line 101: Thanks for the PR. A welcoming change, as I'm facing this issue myself. I do have a few notes/questions though:
1. 1080 is a pretty sensible value, it would be nice to make this configurable but I think it's fine for now
2. What if someone uploads a vertical video? Shouldn't it check if both the height or width exceed the threshold?
3. If the height is unavailable it will always run the transcoding pipeline. Is that intended behavior?

Also a general comment: please use strict equality checks (triple equals) whenever possible

web/src/lib/components/admin-page/settings/setting-transcode.svelte
michelheusschen, line 1: There's already a `SettingSelect` component, where you probably copied this from. It is preferable if you extend that component rather than duplicating it with minor changes.

server/apps/microservices/src/processors/video-transcode.processor.ts
brighteyed: > * 1080 is a pretty sensible value, it would be nice to make this configurable but I think it's fine for now
> * What if someone uploads a vertical video? Shouldn't it check if both the height or width exceed the threshold?

I decided to adapt my logic from Google's Storage Saver plan where they do the following: `Videos higher than 1080p will be resized to high-definition 1080p`. I suppose that's enough to start with

> If the height is unavailable it will always run the transcoding pipeline. Is that intended behavior?

I can hardly imagine a video with height unavailable. Not sure but decided to transcode video in this case

> Also a general comment: please use strict equality checks (triple equals) whenever possible

Thanks, I will fix it

web/src/lib/components/admin-page/settings/setting-transcode.svelte
brighteyed: > There's already a `SettingSelect` component, where you probably copied this from. It is preferable if you extend that component rather than duplicating it with minor changes.

Agree. I have almost zero experience in Typescrypt/Svelte so decided to copy file with minor changes. But I will try to extend an existing component

server/apps/microservices/src/processors/video-transcode.processor.ts
```diff
@@ -71,7 +72,7 @@ export class VideoTranscodeProcessor {
   async runVideoEncode(asset: AssetEntity, savedEncodedPath: string): Promise<void> {
     const config = await this.systemConfigService.getConfig();
 
-    if (config.ffmpeg.transcodeAll) {
+    if (config.ffmpeg.transcode == TranscodePreset.ALL) {
```
jrasm91, line 75: Would you mind changing the flow of the method to be more of something like

let transcode: boolean

if(all) 
  transcode = true

Else if ...
  run Ffprobe pipeline
  transcode = ...

Else
  transcode = false
  ...

And finally all conditions run this code:

If transcode
  Run pipeline

I think keeping the method linear with less branching and return statements makes it more readable. It might be even better to extract that logic into a method `needsTranscoding(asset): boolean`

jrasm91 (approved):
Looks pretty good! Thanks for the enhancement.

server/apps/microservices/src/processors/video-transcode.processor.ts
brighteyed: Sure. I will refactor this method

jrasm91: Awesome, thanks.

brighteyed: Done. Please review

web/src/lib/components/admin-page/settings/setting-transcode.svelte
brighteyed: Done. Now I reuse `SettingSelect` for transcoding presets. Please review

server/apps/microservices/src/processors/video-transcode.processor.ts
brighteyed: Replaced loose equality checks with scrict ones while refactoring `runVideoEncode` method

server/apps/microservices/src/processors/video-transcode.processor.ts
```diff
@@ -72,10 +73,45 @@ export class VideoTranscodeProcessor {
...
     }
+  }
+
+  async needsTranscoding(asset: AssetEntity, ffmpegConfig: SystemConfigFFmpegDto): Promise<boolean> {
+    switch (ffmpegConfig.transcode) {
+      case TranscodePreset.ALL:
+        return true;
+
+      case TranscodePreset.REQUIRED:
+        {
```
bo0tzz, line 89: I think these extra brackets around the case body should be unnecessary. Did you run the formatter?

server/apps/microservices/src/processors/video-transcode.processor.ts
```diff
@@ -72,10 +73,45 @@ export class VideoTranscodeProcessor {
...
+          }
+        }
+        break;
+
+      case TranscodePreset.OPTIMAL:
+        {
+          const videoStream = await this.getVideoStream(asset);
+          if (videoStream.codec_name !== ffmpegConfig.targetVideoCodec) {
+            return true;
+          }
```
bo0tzz, line 102: Can you deduplicate this code so it only exists once? You might need to move to if/else for that instead of using a switch, I think.

server/apps/microservices/src/processors/video-transcode.processor.ts
```diff
@@ -72,10 +73,45 @@ export class VideoTranscodeProcessor {
...
+
+          const videoHeightThreshold = 1080;
+          if (!videoStream.height || videoStream.height > videoHeightThreshold) {
+            return true;
+          }
```
bo0tzz, line 107: This can just be `return !videoStream.height || videoStream.height > videoHeightThreshold`

web/src/lib/components/admin-page/settings/setting-select.svelte
```diff
@@ -27,13 +28,13 @@
...
-		name="presets"
-		id="preset-select"
+		{name}
+		id="{name}-select"
 		bind:value
 		on:change={handleChange}
 	>
 		{#each options as option}
-			<option value={option}>{option}</option>
+			<option value={option.value}>{option.text}</option>
```
bo0tzz, line 37: Why is it necessary to change this to have a separate value and text, I believe those two values are the same in your usage?

server/apps/microservices/src/processors/video-transcode.processor.ts
brighteyed: Oh, sure. You're right

server/apps/microservices/src/processors/video-transcode.processor.ts
michelheusschen: They are necessary, because `videoStream` is used twice and the switch statement doesn't create [lexical scopes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#lexical_scoping)

web/src/lib/components/admin-page/settings/setting-select.svelte
brighteyed: This separation allows to have a short `value` (used as a key to identify selected option in code and stored in the database) and long `text` for the user (that maybe eventually should be localized). For example, in my case one of transcoding presets has `value="optimal"` and `text="Videos higher than 1080p or not in the desired format"`

michelheusschen: They aren't the same for transcoding at least, value will be the enum and text is the description the user sees. For codec selection the value and text are indeed the same.

One more thing and this was already wrong before but: the `for` attribute of label should link to the `id` of select, now those two values are different. Basically to fix this, also set `<label for="{name}-select"`

server/apps/microservices/src/processors/video-transcode.processor.ts
brighteyed: > Did you run the formatter?

Yes. In my implementation these braces are necessary to limit the scope of `videoStream` variable. Without them there is an error `Cannot redeclare block-scoped variable 'videoStream'`

server/apps/microservices/src/processors/video-transcode.processor.ts
brighteyed: Previous version of this method (before my refactoring) didn't have that duplication. But introducing 3 presets (all/required/optimal) worsened the readability of the method. So @jrasm91 suggested to explicitly split code into 3 branches that  resulted in this duplication. Maybe I misunderstood his intention but code after refactoring seems to me better. We now have 3 distinct branches where it is possible to tune transcoding.

Is it fine for now?

server/apps/microservices/src/processors/video-transcode.processor.ts
brighteyed: Done!

web/src/lib/components/admin-page/settings/setting-select.svelte
bo0tzz: I was only seeing the `targetVideoCodec` section, my bad!

server/libs/infra/src/db/migrations/1679751316282-UpdateTranscodeOption.ts
```diff
@@ -0,0 +1,27 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class UpdateTranscodeOption1679751316282 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
```
jrasm91, line 5: Should we delete the old key, or how would you handle the case it was set to false?

jrasm91 (approved):
That transcode section looks better, thanks for that. Had question about the migration otherwise looks good to me.

server/apps/microservices/src/processors/video-transcode.processor.ts
bo0tzz: I had a short attempt at coming up with a tidier approach, but it's a little less straightforward than I thought. All good!

bo0tzz (approved):

web/src/lib/components/admin-page/settings/setting-select.svelte
brighteyed: > Basically to fix this, also set `<label for="{name}-select"`

Done!

server/libs/infra/src/db/migrations/1679751316282-UpdateTranscodeOption.ts
brighteyed: I noticed that immich doesn't store the default value of `transcodeAll` (equals to `false`) into database. So `system_config` table could only store `transcodeAll` with `true` value

jrasm91: Actually that does sounds correct, I forgot 😁. Perfect!

PR #2093, opened 2023-03-27
feat(server): apply storage migration after exif completes

jrasm91:
Apply storage migration after exif process saves `fileCreatedAt`.


bo0tzz (approved):
`handleTemplateMigration` has a TODO about removing some livePhoto stuff, is that applicable yet? And if so, can that function be refactored just delegate to the `single` one?

alextran1502:
I tried to upload LivePhotos, and the motion part doesn't get moved correctly to the `library` folder

## Actual

image

## Expected

image

alextran1502 (changes requested):
Error in moving LivePhotos

jrasm91:
@alextran1502 - did a bit more testing with live photos and it seems to be working now. Feel free to retest.

alextran1502 (approved):

PR #2116, opened 2023-03-29
chore(server): use ioredis

jrasm91:
Prerequisite for #473.


EnochPrime:
Does this change impact whether or not redis 7+ is supported?

jrasm91:
I haven't done any testing with redis 7, this is just switching client libraries. It would probably work though? Maybe you can test and open a PR if everything looks good.

PR #2125, opened 2023-03-30
feat(web): pause and resume jobs

michelheusschen:
Adds pause and resume commands to the administration page for jobs. Also refactored the components a bit and added a colored status indicator to quickly see the status of each job. There is also an indicator for failed jobs, see screenshots below.

Screenshot 2023-03-30 at 11-24-53 Job Status
Screenshot 2023-03-30 at 11-23-32 Job Status
Screenshot 2023-03-30 at 11-24-26 Job Status

### DIFF
--- a/web/src/app.css
+++ b/web/src/app.css
@@ -111,6 +111,6 @@ input:focus-visible {
 	}
 
 	.job-play-button {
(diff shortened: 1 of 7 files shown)


jrasm91 (approved):

etnoy:
Haven't tested the code yet but this is a great feature for pausing jobs when large cli uploads are running

etnoy:
I have now tested this feature and it is amazing. By pausing the machine learning jobs my 40k queue of thumbnail generation was done within an hour or two which is an order of magnitude faster than otherwise.

etnoy (approved):

PR #2137, opened 2023-03-31
feat(server): apply ValidationPipe on controllers

michelheusschen:
Sets `ValidationPipe` on the controller level instead of passing it to `@Query()`, `@Param()`, `@Body()`, etc every time. It will also do conversions to primitive types. In the following example, `isAll` will automatically be converted to a boolean: `@Query('isAll') isAll: boolean`.

I can't think of a situation where you don't want to have validation and setting it as default seems sensible. This also reduces the likelihood of forgetting a `ValidationPipe`, as has happened for `ChangePasswordDto` which will be validated as a result of this change.

### DIFF
--- a/server/apps/immich/src/controllers/album.controller.ts
+++ b/server/apps/immich/src/controllers/album.controller.ts
@@ -1,6 +1,6 @@
 import { AlbumService, AuthUserDto } from '@app/domain';
 import { GetAlbumsDto } from '@app/domain/album/dto/get-albums.dto';
-import { Controller, Get, Query, ValidationPipe } from '@nestjs/common';
+import { Controller, Get, Query, UsePipes, ValidationPipe } from '@nestjs/common';
 import { ApiTags } from '@nestjs/swagger';
 import { GetAuthUser } from '../decorators/auth-user.decorator';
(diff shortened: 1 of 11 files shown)


jrasm91:
I'm fine with this, but you can actually apply it globally, not just on individual controllers;

https://docs.nestjs.com/pipes

At the bottom of this link there are examples of applying it to the app module or in main.ts as middleware.

michelheusschen:
Yeah I saw, but I decided not to. It hasn't been applied to all controllers, only those that are refactored. Also, there is apparently no easy way to disable it. I can't think of a reason right now why you'd want to disable the global `ValidationPipe`, but if you ever need to for a single method you have to refactor everything.

jrasm91:
Ok, we can slowly opt in and maybe switch to that at some point.

jrasm91:
You mention isAll as a boolean. Did you test this? In general class transformer converts all strings to boolran true even 'false'

michelheusschen:
Yes, I tested it and it worked as expected. If isAll is empty, `invalid`, `false` or `1` it converts to boolean `false`. Only the string `true` yields true

jrasm91 (approved):

PR #2141, opened 2023-03-31
feat(server): redis sentinel support

### LINKED ISSUE #473: [Feature]: Support for redis sentinel

samip5:
Fixes #473

### DIFF
--- a/server/libs/infra/src/infra.config.ts
+++ b/server/libs/infra/src/infra.config.ts
@@ -3,17 +3,35 @@ import { BullModuleOptions } from '@nestjs/bull';
 import { RedisOptions } from 'ioredis';
 import { ConfigurationOptions } from 'typesense/lib/Typesense/Configuration';
 
-export const redisConfig: RedisOptions = {
-  host: process.env.REDIS_HOSTNAME || 'immich_redis',
-  port: parseInt(process.env.REDIS_PORT || '6379'),
-  db: parseInt(process.env.REDIS_DBINDEX || '0'),
-  password: process.env.REDIS_PASSWORD || undefined,
-  path: process.env.REDIS_SOCKET || undefined,
-};
+export function redisConfig(): RedisOptions {
+  const redisUrl = process.env.REDIS_URL;
+  if (redisUrl != undefined && redisUrl.startsWith('ioredis://')) {
+    let customOptions = {};
+    try {
+      const decodedString = Buffer.from(redisUrl.slice(10), "base64").toString();
+      customOptions = JSON.parse(decodedString);
+    } catch (error) {
+      throw new Error(`Failed to decode redis options: ${error}`);
+    }
+    return customOptions;
+  }else {
+    {
+      return {
+        host: process.env.REDIS_HOSTNAME || 'immich_redis',
+        port: parseInt(process.env.REDIS_PORT || '6379'),
+        db: parseInt(process.env.REDIS_DBINDEX || '0'),
+        username: process.env.REDIS_USERNAME || undefined,
+        password: process.env.REDIS_PASSWORD || undefined,
+        path: process.env.REDIS_SOCKET || undefined,
+      };
+    }
+  }
+}
+
 
 export const bullConfig: BullModuleOptions = {
   prefix: 'immich_bull',
-  redis: redisConfig,
+  redis: redisConfig(),
   defaultJobOptions: {
     attempts: 3,
     removeOnComplete: true,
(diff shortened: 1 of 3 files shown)


samip5:
I have verified that it does properly work with Sentinel now. :)

server/libs/infra/src/infra.config.ts
```diff
@@ -3,17 +3,34 @@ import { BullModuleOptions } from '@nestjs/bull';
...
-  password: process.env.REDIS_PASSWORD || undefined,
-  path: process.env.REDIS_SOCKET || undefined,
-};
+export function redisConfig(): RedisOptions {
+  const redisUrl = process.env.REDIS_URL;
+  if (redisUrl != undefined && redisUrl.startsWith('ioredis://')) {
+    let customOptions = {};
+    try {
+      const decodedString = Buffer.from(redisUrl.slice(10), 'base64').toString();
+      customOptions = JSON.parse(decodedString);
```
jrasm91, line 12: Just return `JSON.parse(decodedString);` here

server/libs/infra/src/infra.config.ts
```diff
@@ -3,17 +3,34 @@ import { BullModuleOptions } from '@nestjs/bull';
...
+    } catch (error) {
+      throw new Error(`Failed to decode redis options: ${error}`);
+    }
+    return customOptions;
+  } else {
+    {
```
jrasm91, line 18: This set of brackets seems to be extraneous.

server/libs/infra/src/infra.config.ts
```diff
@@ -3,17 +3,34 @@ import { BullModuleOptions } from '@nestjs/bull';
...
+  if (redisUrl != undefined && redisUrl.startsWith('ioredis://')) {
+    let customOptions = {};
+    try {
+      const decodedString = Buffer.from(redisUrl.slice(10), 'base64').toString();
+      customOptions = JSON.parse(decodedString);
+    } catch (error) {
+      throw new Error(`Failed to decode redis options: ${error}`);
+    }
+    return customOptions;
+  } else {
```
jrasm91, line 17: If you return from all paths in the above condition you don't need the else condition explicitly anymore.

server/libs/infra/src/infra.config.ts
```diff
@@ -3,17 +3,34 @@ import { BullModuleOptions } from '@nestjs/bull';
...
-export const redisConfig: RedisOptions = {
-  host: process.env.REDIS_HOSTNAME || 'immich_redis',
-  port: parseInt(process.env.REDIS_PORT || '6379'),
-  db: parseInt(process.env.REDIS_DBINDEX || '0'),
-  password: process.env.REDIS_PASSWORD || undefined,
-  path: process.env.REDIS_SOCKET || undefined,
-};
+export function redisConfig(): RedisOptions {
+  const redisUrl = process.env.REDIS_URL;
+  if (redisUrl != undefined && redisUrl.startsWith('ioredis://')) {
```
jrasm91, line 8: It's probably better to do a falsy check instead of explicitly checking for `undefined`, so `redisUrl && ...`

server/libs/infra/src/infra.config.ts
```diff
@@ -3,17 +3,34 @@ import { BullModuleOptions } from '@nestjs/bull';
...
 import { ConfigurationOptions } from 'typesense/lib/Typesense/Configuration';
 
-export const redisConfig: RedisOptions = {
-  host: process.env.REDIS_HOSTNAME || 'immich_redis',
-  port: parseInt(process.env.REDIS_PORT || '6379'),
-  db: parseInt(process.env.REDIS_DBINDEX || '0'),
-  password: process.env.REDIS_PASSWORD || undefined,
-  path: process.env.REDIS_SOCKET || undefined,
-};
+export function redisConfig(): RedisOptions {
```
jrasm91, line 6: I'm fine with a function that returns the config, but can you still just export the config here? So like
```typescript
function parseRedisConfig(): RedisOptions {
...
export const redisConfig: RedisOptions = parseRedisConfig();
```

jrasm91 (approved):

PR #2144, opened 2023-04-01
feat(server/web): jobs clear button + queue status

michelheusschen:
Server:
- return job status after executing command
- include queue status in job response

Web:
- add clear button to job tile
- add delayed jobs count
- adjusted styling of badge
- adjusted button icon size
- created component for job tile button
- update status with data from server after sending command

light_paused_all_tags_v2
dark_paused_v2


jrasm91 (approved):

michelheusschen:
@alextran1502 moved the status bar to the left of the buttons, you can just revert if you don't like this

image

PR #2146, opened 2023-04-01
feat(server): Support TypeSense High-Availibility configuration

samip5:
I'm not sure how this should be done without duplicating the last 4 lines, for apikey etc.


jrasm91 (commented):
You can create an object at the top of the function that holds the common keys and then use that in both return objects via `...common`. Or, do like we do in the database config file and only return the node configuration and merge that into the rest of the (common) config. Ping me in discord if that doesn't make sense.

michelheusschen:
I also wanted to ask this in your redis PR, but it was already merged: why is the config value base64 encoded?

samip5:
> I also wanted to ask this in your redis PR, but it was already merged: why is the config value base64 encoded?

Simpler to parse mainly and in redis the whole object is given to ioredis.

michelheusschen:
It only seems like an extra conversion step, or am I missing something? Why not use something like this:

```ts
// Set this in .env or however you want to
TYPESENSE_NODES=[{"host":"typesense-1"},{"host":"typesense-2"}]

// Parse config in infra.config.ts
const nodes = JSON.parse(process.env.TYPESENSE_NODES);
```

samip5:
> It only seems like an extra conversion step, or am I missing something? Why not use something like this:
> 
> ```ts
> // Set this in .env or however you want to
> TYPESENSE_NODES=[{"host":"typesense-1"},{"host":"typesense-2"}]
> 
> // Parse config in infra.config.ts
> const nodes = JSON.parse(process.env.TYPESENSE_NODES);
> ```

I don't think ENV vars like RAW JSON... or at least haven't seen it.


Let's ask the other way, why not base64 encoded?

michelheusschen:
It's perfectly fine to keep the TypeSense configuration as a single variable, that was just an example. I have seen JSON being used in env variables and from a quick test it works just fine. I'm not completely against using base64, but sticking with JSON only seems more user-friendly because you don't have the extra conversion step and it simplifies troubleshooting any configuration issues.

Regardless, it's important to document the new environment variable and if base64 encoding is chosen, a note should be provided for users who may not be familiar with it.

jrasm91 (approved):

PR #2149, opened 2023-04-02
feat(server): improve validation in controllers

michelheusschen:
Add `whitelist: true` to `ValidationPipe` of all refactored controllers to only keep properties of incoming DTOs that have decorators. All other properties are stripped. The current behavior could become an issue when a dto is passed as an object and used without further validation, for example when passing a DTO directly to the update function of a repository.

Nested objects are only validated when they are an instance of a class. This means that `SystemConfigDto` didn't have proper validation and nested values were empty after applying `whitelist: true`. This has been fixed in this PR.

```ts
class Dto {
    @IsNotEmpty()
    @IsString()
    validated: string;
    
    unvalidated: string;
}

// Update endpoint in controller
update(@Body() dto: Dto)

// Request body to update endpoint
{ validated: 'value', unvalidated: 'value', other: 10 }


// Current value of dto:
{ validated: 'value', unvalidated: 'value', other: 10 }

// Value of dto after adding `whitelist: true`
{ validated: 'value' }
```

### DIFF
--- a/server/apps/immich/src/controllers/album.controller.ts
+++ b/server/apps/immich/src/controllers/album.controller.ts
@@ -1,14 +1,15 @@
 import { AlbumService, AuthUserDto } from '@app/domain';
 import { GetAlbumsDto } from '@app/domain/album/dto/get-albums.dto';
-import { Controller, Get, Query, UsePipes, ValidationPipe } from '@nestjs/common';
+import { Controller, Get, Query } from '@nestjs/common';
 import { ApiTags } from '@nestjs/swagger';
 import { GetAuthUser } from '../decorators/auth-user.decorator';
 import { Authenticated } from '../decorators/authenticated.decorator';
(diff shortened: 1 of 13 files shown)


jrasm91:
Might be worth making a @Validate() decorator instead of repeating the configuration in each controller. That could either alias the use pipe like you have it now. Or, use Set metadata and add a global pipe that conditionally applies the validation pipe if the metadata is present.

michelheusschen:
Good idea, I tried making a custom global pipe that conditionally disables validation, but it doesn't seem like pipes have access to the metadata set by decorators. So instead, I created a decorator that applies `ValidationPipe` with the configuration like you suggested.

jrasm91 (approved):
Nice. Yeah, I guess it would have to be an interceptor that updated the request, but this seems to work well enough.

PR #2159, opened 2023-04-03
chores(docs): fix bulk upload example commands to work with paths with spaces

MMauro94:
The examples commands didn't work when the result of the subcommand `$(pwd)` contained spaces. This CR adds the proper escaping for them to work. This change also fixes the behavior of the alias allowing the correct current directory to be passed when it is placed inside the `.bashrc` file.

### DIFF
--- a/docs/docs/features/bulk-upload.md
+++ b/docs/docs/features/bulk-upload.md
@@ -60,15 +60,15 @@ Be aware that as this runs inside a container, you need to mount the folder from
 
 ```bash title="Upload current directory"
 cd /DIRECTORY/WITH/IMAGES
-docker run -it --rm -v $(pwd):/import ghcr.io/immich-app/immich-cli:latest upload --key HFEJ38DNSDUEG --server http://192.168.1.216:2283/api
+docker run -it --rm -v "$(pwd):/import" ghcr.io/immich-app/immich-cli:latest upload --key HFEJ38DNSDUEG --server http://192.168.1.216:2283/api
 ```
 
 ```bash title="Upload target directory"
-docker run -it --rm -v /DIRECTORY/WITH/IMAGES:/import ghcr.io/immich-app/immich-cli:latest upload --key HFEJ38DNSDUEG --server http://192.168.1.216:2283/api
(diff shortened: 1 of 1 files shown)


docs/docs/features/bulk-upload.md
```diff
@@ -60,15 +60,15 @@ Be aware that as this runs inside a container, you need to mount the folder from
...
 ```
 
 ```bash title="Upload target directory"
-docker run -it --rm -v /DIRECTORY/WITH/IMAGES:/import ghcr.io/immich-app/immich-cli:latest upload --key HFEJ38DNSDUEG --server http://192.168.1.216:2283/api
+docker run -it --rm -v "/DIRECTORY/WITH/IMAGES:/import" ghcr.io/immich-app/immich-cli:latest upload --key HFEJ38DNSDUEG --server http://192.168.1.216:2283/api
 ```
 
 ```bash title="Create an alias"
-alias immich="docker run -it --rm -v $(pwd):/import ghcr.io/immich-app/immich-cli:latest"
+alias immich='docker run -it --rm -v "$(pwd):/import" ghcr.io/immich-app/immich-cli:latest'
```
jrasm91, line 71: I don't think one is right, since single quotes means no bash expansion and you will have the  literal $(pwd) string.

jrasm91: Or does this just defer bash expansion until you actually run the command?

MMauro94: Yes. From alias man page:

> The value of an alias can be set to an expression including spaces, command options and/or variables, the expression must be quoted, either with 'Single quotes' that will be evaluated dynamically each time the Alias is used, or "Double quotes" which will be evaluated once when the alias is created.

Test:
```
bash-5.1# alias echotest='echo "$(pwd)"'
bash-5.1# cd /home/                                     bash-5.1# echotest
/home                                                   bash-5.1# cd /media/
bash-5.1# echotest                                      
/media                                                  bash-5.1#
```

jrasm91: Oh, perfect. TIL!

jrasm91 (approved):

PR #2166, opened 2023-04-04
refactor(server): more consistent param validation

michelheusschen:
- Use a DTO for validating UUID params
- Add validation to params that were missing it
- Set format to uuid for openapi definition


jrasm91 (approved):
Only downside to this is that the connection between the param name (in the url and the dto) are now farther away. Dto key needs to match the string and it was easier to follow when they lived in the same file, one line apart.

michelheusschen:
I tried making a `@UUIDParam('id')` decorator, but that didn't work out. The DTO is destructured everywhere, so seeing the link between url and param should still be easy

PR #2167, opened 2023-04-04
refactor(server): reverse geocoding

jrasm91:
- Move reverse geocoding call to an interface
- Move reverse geocoding config to `infra.config.ts` 
- Start extraction queue after reverse geocoder is initialized
- Use promises instead of callbacks
- Remove unused reverse geocoding job


server/libs/infra/src/repositories/geocoding.repository.ts
```diff
@@ -0,0 +1,44 @@
...
+    this.logger.verbose(`Raw: ${JSON.stringify(address, null, 2)}`);
+
+    const { countryCode, name: city, admin1Code, admin2Code } = address[0] as GeoData;
+    const country = getName(countryCode, 'en');
+    const stateParts = [(admin2Code as AdminCode)?.name, (admin1Code as AdminCode)?.name].filter((name) => !!name);
+    const state = stateParts.length > 0 ? stateParts.join(', ') : '';
+    this.logger.debug(`Normalized: ${JSON.stringify({ country, state, city })}`);
+
+    return { country, state, city };
+  }
```
michelheusschen, line 43: State is always empty for me, probably because `admin1Code` and `admin2Code` are a code and there's no lookup happening like there is for country. Might also be better to try and minimize the amount of typecasts, which could make type errors hard to spot especially in the future with package upgrades. 

Also the state column is nullable in the database so why not use null instead of an empty string?

jrasm91: I mainly refactored the logic without trying to change how it functioned. It returned empty string before, etc. Happy to make more changes though, especially if there are bugs. 

Whoever implemented it originally had the typecasts. The typing from the library uses `string` for the admin codes, but the response I get back has objects. Do you have some files/gps data that returns strings instead?

michelheusschen: After recreating my dev instance the state is correctly populated. Before, the lookup function would only return strings for admin codes but I can no longer reproduce it and they are objects. Weird, because I tried it multiple times before. I guess it's solved then and only the typecasts could use some love but no major issues.

jrasm91: Looks like it's a problem in the typing for the library. The way it works is they match against the list of cities. The city match contains stuff like country code, state code, etc. They use those codes to create a key that should index into admin1 or admin2 csv file. If they get a match, they return the admin1 or admin2 record. Otherwise they return the original value.

I'm not sure in what situation you would have a record in the cities file that referenced countries/states that did not exist in the admin1 or admin2 files.

PR #2171, opened 2023-04-05
feat(all): transcoding improvements

zackpollard:
- Changes ffmpeg settings so audio codec is a dropdown menu
- Checks the audio codec of uploaded files and converts them if they don't match the target codec
- Checks the container of uploaded files and converts them if they don't match the target container
- Change to single concurrency for video transcoding queue
- Set niceness to 10 for ffmpeg process to correctly prioritise CPU time

### DIFF
--- a/server/libs/domain/src/media/media.service.ts
+++ b/server/libs/domain/src/media/media.service.ts
@@ -184,6 +184,9 @@ export class MediaService {
     const isLargerThanTargetResolution = Math.min(videoStream.height, videoStream.width) > targetResolution;
 
     switch (ffmpegConfig.transcode) {
+      case TranscodePreset.DISABLED:
+        return false;
+
       case TranscodePreset.ALL:
         return true;
 
@@ -199,8 +202,6 @@ export class MediaService {
   }
 
   private getFfmpegOptions(stream: VideoStreamInfo, ffmpeg: SystemConfigFFmpegDto) {
-    // TODO: If video or audio are already the correct format, don't re-encode, copy the stream
-
     const options = [
       `-crf ${ffmpeg.crf}`,
       `-preset ${ffmpeg.preset}`,
--- a/server/libs/domain/src/media/media.service.ts
+++ b/server/libs/domain/src/media/media.service.ts
@@ -143,9 +143,11 @@ export class MediaService {
       }
 
       const options = this.getFfmpegOptions(mainVideoStream, config);
+
+      this.logger.log(`Start encoding video ${asset.id} ${options}`);
       await this.mediaRepository.transcode(input, output, options);
 
-      this.logger.log(`Converting Success ${asset.id}`);
+      this.logger.log(`Encoding success ${asset.id}`);
 
       await this.assetRepository.save({ id: asset.id, encodedVideoPath: output });
     } catch (error: any) {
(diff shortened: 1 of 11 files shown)


jrasm91 (approved):

alextran1502:
Can you fix the web test linter?

image

server/libs/domain/src/media/media.service.ts
```diff
@@ -151,32 +153,45 @@ export class MediaService {
...
-      .sort((stream1, stream2) => stream2.frameCount - stream1.frameCount)[0];
+  private getMainVideoStream(streams: VideoStreamInfo[]): VideoStreamInfo | null {
+    return streams.sort((stream1, stream2) => stream2.frameCount - stream1.frameCount)[0];
   }
 
-  private isTranscodeRequired(stream: VideoStreamInfo, ffmpegConfig: SystemConfigFFmpegDto): boolean {
-    if (!stream.height || !stream.width) {
+  private getMainAudioStream(streams: AudioStreamInfo[]): AudioStreamInfo | null {
+    return streams[0];
+  }
```
michelheusschen, line 162: Uploading a file with multiple audio or video streams is probably not that common, but I noticed we select the video stream with the highest frame count and for audio the first stream. Is there any reason behind this? The selected streams are also not passed on to ffmpeg which uses it's own [stream selection](https://ffmpeg.org/ffmpeg.html#Automatic-stream-selection) algorithm.

server/libs/domain/src/media/media.service.ts
```diff
@@ -127,20 +127,22 @@ export class MediaService {
...
       this.storageRepository.mkdirSync(outputFolder);
 
-      const { streams } = await this.mediaRepository.probe(input);
-      const stream = await this.getLongestStream(streams);
-      if (!stream) {
+      const { videoStreams, audioStreams, format } = await this.mediaRepository.probe(input);
+      const mainVideoStream = await this.getMainVideoStream(videoStreams);
+      const mainAudioStream = await this.getMainAudioStream(audioStreams);
+      const containerExtension = format.formatName;
+      if (!mainVideoStream || !mainAudioStream || !containerExtension) {
```
michelheusschen, line 134: This causes audio or video-only files to never be transcoded. Is that intended behavior? Also unneeded awaits for selecting the streams

zackpollard: Not sure why my IDE wasn't complaining about the awaits, will remove those. We don't support audio only files in Immich, however video files without audio I guess could be an edge case we could consider.

server/libs/domain/src/media/media.service.ts
zackpollard: Improving the actual commands sent to ffmpeg is further down my list, we can look at passing the stream identifier directly into ffmpeg at a later date. The reason for just selecting the first audio stream is simply that there isn't really any metadata available to easily identify the "best" stream. I thought about doing it based on bitrate of the stream, but that is actually only metadata added by the original encoder rather than being calculated by ffprobe at the time, so it can be wrong.

michelheusschen: Was just curious about the different approach between them. FFmpeg uses the stream with most audio channels, but since this is such an edge case it can definitely be addressed later on.

PR #2173, opened 2023-04-05
chore(server): update typeorm

michelheusschen:
Updates TypeORM from version `0.3.11` to `0.3.12`, which includes a fix for redundant unique constraints. As a result, a migration file is necessary


zackpollard:
Did TypeORM generate these migrations itself? Can you confirm there are other constraints that still fulfill this criteria as I do believe we still want this to be unique, or is this handled directly by the column definition now?

michelheusschen:
TypeORM generated these, but I manually removed some index recreations. `assetId` is the primary key of both the `exif` and `smart_info` tables, there was only an additional unique constraint generated by TypeORM which is now fixed and removed by these migrations.

zackpollard (approved):

jrasm91 (approved):

PR #2177, opened 2023-04-05
fix(web): remove search bar focus after searching

michelheusschen:
Removes focus from the search bar after performing a search by using the `focusout` event, instead of relying on a click outside the search bar


jrasm91 (approved):

bo0tzz (approved):
Very quick, thank you!

PR #2185, opened 2023-04-06
Ensure email is readable in bright mode

nebulade:
Due to the disabled style, the text and background color end up being the same.


michelheusschen:
Well spotted! However, the `disabled` class doesn't exist and the 'not allowed' cursor is already set by the `immich-form-input` class. So, similar to the other inputs, simply using `class="immich-form-input"` would be sufficient.

Another option would be to enable the field, since changing the email address via the API is possible.

nebulade:
Feel free to close this then and just push a fix directly. This was just a quick and dirty css fix for me, as I have no real knowledge about the app stylesheet.

michelheusschen:
#2189

PR #2189, opened 2023-04-06
feat(web): allow admin to update email of users

michelheusschen:
Allows the admin to update the email address of users via the management page. Don't know why it's disabled currently, but there is no API limitation and users can already update their own email address via the web interface.

### DIFF
--- a/web/src/lib/components/forms/edit-user-form.svelte
+++ b/web/src/lib/components/forms/edit-user-form.svelte
@@ -14,21 +14,12 @@
 
 	const dispatch = createEventDispatcher();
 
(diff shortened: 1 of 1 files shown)


jrasm91 (approved):
Originally, email was not supposed to be able to be updated. We changed that a month or two ago and added in the user profile page as editable. I assume this section was just never updated to support that change. Enabling it seems fine.

re: OAuth - no issues that I can see

- Email is used to auto-link an existing immich user with an OAuth user, the first time they sign in
- Already linked accounts use oauthId after that
- You can link and unlink accounts with different email address via the OAuth settings in the profile page for these situations.

PR #2195, opened 2023-04-06
Update Norwegian locale to ICU standard

Allram:
Don't know if this is the only place this should be edited. But the Norwegian locale is not working on the Android app (still in English), And it seems like it should be 'nb-NO' and not 'no-NO'.

https://www.localeplanet.com/icu/nb-NO/index.html


alextran1502:
Did it show the language correctly on your device? This is what I see on the translation platform so I decided to go with the generic code 

image

PR #2196, opened 2023-04-06
chore: run web (only) with remote backend

jrasm91:


web/vite.config.js
```diff
@@ -9,6 +9,19 @@ const config = {
 			'@api': path.resolve('./src/api')
 		}
 	},
+	server: {
+		// connect to a remote backend during web-only development
```
bo0tzz, line 13: Should we config this so it only applies when NODE_ENV=development?

jrasm91: vite server is only ever used in development (`npm run dev`)

PR #2198, opened 2023-04-06
feat: sony raw support

jrasm91:


server/libs/infra/src/repositories/media.repository.ts
```diff
@@ -22,11 +28,31 @@ export class MediaRepository implements IMediaRepository {
...
+            .resize(options.size, options.size, { fit: 'outside', withoutEnlargement: true })
+            .jpeg()
+            .rotate()
+            .toFile(output);
+        } else {
+          try {
+            await stat(output);
+            await unlink(output);
+          } catch {}
+          const command = `darktable-cli '${input}' '${output}' --apply-custom-presets false --width ${options.size} --height ${options.size} --core --library :memory:`;
```
brighteyed, line 42: I've noticed that `--width/--height` options of `darktable-cli` limit the _maximum_ width/height of the image. But we need to guarantee that thumbnail width/height is as small as possible but _not less_ than `options.size`

server/libs/domain/src/media/media.service.ts
```diff
@@ -93,7 +107,7 @@ export class MediaService {
       return;
     }
 
-    const webpPath = asset.resizePath.replace('jpeg', 'webp');
+    const webpPath = asset.resizePath.replace('jpeg', 'webp').replace('jpg', 'webp');
```
brighteyed, line 110: If we changed thumbnail extension from `jpeg` to `jpg` in `handleGenerateJpegThumbnail` then it seem we don't need to replace `jpeg` here

jrasm91: Maybe. Existing database records will have `jpeg` and I wasn't 100% sure if it was possible for this job to run on some of those paths or not.

server/libs/infra/src/repositories/media.repository.ts
```diff
@@ -22,11 +28,31 @@ export class MediaRepository implements IMediaRepository {
...
+        if (!options.raw) {
+          await sharp(input, { failOnError: false })
+            .resize(options.size, options.size, { fit: 'outside', withoutEnlargement: true })
+            .jpeg()
+            .rotate()
+            .toFile(output);
+        } else {
+          try {
+            await stat(output);
+            await unlink(output);
```
uhthomas, line 40: What's the purpose of this? Why is it okay for it to fail?

server/libs/domain/src/media/media.service.ts
```diff
@@ -48,21 +48,35 @@ export class MediaService {
...
       const thumbnailDimension = 1440;
       if (asset.type == AssetType.IMAGE) {
-        try {
-          await this.mediaRepository.resize(asset.originalPath, jpegThumbnailPath, {
+        await this.mediaRepository
+          .resize(asset.originalPath, jpegThumbnailPath, {
             size: thumbnailDimension,
             format: 'jpeg',
+          })
+          .catch((error) => {
```
uhthomas, line 60: This repository seems to use await rather than callbacks, is there any reason to deviate from this?

server/libs/infra/src/repositories/media.repository.ts
jrasm91: The dark table cli don't ever overwrite a file that exists and trying to stat a file that doesn't exist throws an error.

server/libs/infra/src/repositories/media.repository.ts
```diff
@@ -22,11 +28,31 @@ export class MediaRepository implements IMediaRepository {
         return;
 
       case 'jpeg':
-        await sharp(input, { failOnError: false })
-          .resize(options.size, options.size, { fit: 'outside', withoutEnlargement: true })
-          .jpeg()
-          .rotate()
-          .toFile(output);
+        if (!options.raw) {
```
uhthomas, line 31: It may be possible to remove a level of indentation by returning early. Right now it looks like:

```ts
if (raw) {

} else {

}
```

It could look like:

```
if (raw) {
    return
}

// not raw code
```

Might be nice.

uhthomas, line 31: It might be nice to put these into their own functions. Should make for easier unit testing and may be easier to read.

```ts
class {
  resize()
  resizeRaw()
}
```

server/libs/infra/src/repositories/media.repository.ts
uhthomas: I think I see, thanks. In which case, something like this might be more suitable?

```ts
try {
  await unlink(output);
} catch(e) {
  if (e && e.code !== 'ENOENT') throw e;
}
```

What do you think?

jrasm91: Possibly, yeah

server/libs/domain/src/media/media.service.ts
jrasm91: Mostly that this is a chain of fallbacks and i didn't want nested try/catch statements.

server/libs/infra/src/repositories/media.repository.ts
jrasm91: Yeah, that's a good catch. I usually prefer code like that.

jrasm91: Yes, agree. I was trying to see if the approach was even worthwhile before going through the extra effort of splitting it out, mocking it, etc., but I think it's going to work so that's a good suggestion.

server/libs/infra/src/repositories/media.repository.ts
```diff
@@ -31,6 +37,35 @@ export class MediaRepository implements IMediaRepository {
     }
   }
 
+  async resizeRaw(input: string, output: string, options: ResizeRawOptions): Promise<void> {
+    try {
```
uhthomas, line 41: Linking https://github.com/immich-app/immich/pull/2198#discussion_r1164776892 again as it's marked as out-dated.

uhthomas:
Could it be possible for `sharp` to support raw images natively? https://github.com/lovell/sharp/issues/3612

If not, would the suggested package `libraw.js` be worth considering? https://github.com/justinkambic/libraw.js

jrasm91:
Interesting. Worth a shot. It sounds like installing vips magik could solve some of the issues with raw formats.

jrasm91:
Closing in favor of #2668

PR #2203, opened 2023-04-07
chore(server): remove unneeded debug logging

samip5:


bo0tzz:
No need to comment out code, just remove it ;)

jrasm91 (commented):
Can you actually change this to verbose instead?

jrasm91:
Inevitably there will be someone who can't blah blah blah some video file. It'll be nice for them to turn on verbose logging and see the details.

samip5:
For it to be more useful, I think the log should include assetID, but it's not passed though to the isTranscodeRequired function. :(

jrasm91 (approved):

PR #2212, opened 2023-04-08
chore(server): redis error handling

samip5:
Should prevent errors like:

```
missing 'error' handler on this Redis client
missing 'error' handler on this Redis client
```

### DIFF
--- a/server/libs/infra/src/infra.config.ts
+++ b/server/libs/infra/src/infra.config.ts
@@ -26,7 +26,11 @@ function parseRedisConfig(): RedisOptions {
 
 export const redisConfig: RedisOptions = parseRedisConfig();
 
-export const bullConfig: BullModuleOptions = {
+interface CustomBullModuleOptions extends BullModuleOptions {
+  onError?: (error: any) => void;
+}
+
+export const bullConfig: CustomBullModuleOptions = {
   prefix: 'immich_bull',
   redis: redisConfig,
   defaultJobOptions: {
@@ -34,9 +38,12 @@ export const bullConfig: BullModuleOptions = {
     removeOnComplete: true,
     removeOnFail: false,
   },
+  onError: (error) => {
+    console.log(error);
+  },
 };
 
-export const bullQueues: BullModuleOptions[] = Object.values(QueueName).map((name) => ({ name }));
+export const bullQueues: CustomBullModuleOptions[] = Object.values(QueueName).map((name) => ({ name }));
 
 function parseTypeSenseConfig(): ConfigurationOptions {
   const typesenseURL = process.env.TYPESENSE_URL;
--- a/server/libs/infra/src/infra.config.ts
+++ b/server/libs/infra/src/infra.config.ts
@@ -1,8 +1,11 @@
 import { QueueName } from '@app/domain';
+import { Logger } from '@nestjs/common';
 import { BullModuleOptions } from '@nestjs/bull';
 import { RedisOptions } from 'ioredis';
 import { InitOptions } from 'local-reverse-geocoder';
 import { ConfigurationOptions } from 'typesense/lib/Typesense/Configuration';
+
+const logger = new Logger('infra.config');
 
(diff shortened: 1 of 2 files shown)


server/libs/infra/src/infra.config.ts
```diff
@@ -26,17 +26,24 @@ function parseRedisConfig(): RedisOptions {
...
+export const bullConfig: CustomBullModuleOptions = {
   prefix: 'immich_bull',
   redis: redisConfig,
   defaultJobOptions: {
     attempts: 3,
     removeOnComplete: true,
     removeOnFail: false,
   },
+  onError: (error) => {
+    console.log(error);
```
bo0tzz, line 42: Can this use the nestjs logger instead of console.log?

jrasm91: Should be able to instantiate a new logger in this life and then use it here.

samip5: That's a good question as it's not inside a class so one cannot simply do this.logger, but should be possible.

bo0tzz (approved):

server/libs/infra/src/infra.config.ts
```diff
@@ -26,17 +29,24 @@ function parseRedisConfig(): RedisOptions {
...
     removeOnComplete: true,
     removeOnFail: false,
   },
+  onError: (error) => {
+    logger.error(error);
+  },
 };
 
-export const bullQueues: BullModuleOptions[] = Object.values(QueueName).map((name) => ({ name }));
+export const bullQueues: CustomBullModuleOptions[] = Object.values(QueueName).map((name) => ({ name }));
```
jrasm91, line 49: Is this change necessary?

samip5: I believe so as the BullModuleOptions doesn't expose the onError, but I could be mistaken too.

jrasm91: This is only the config for the queues for not main config.

samip5: Let me change it back as the main config should be enough.

server/libs/infra/src/infra.config.ts
```diff
@@ -26,17 +29,24 @@ function parseRedisConfig(): RedisOptions {
 
 export const redisConfig: RedisOptions = parseRedisConfig();
 
-export const bullConfig: BullModuleOptions = {
+interface CustomBullModuleOptions extends BullModuleOptions {
+  onError?: (error: any) => void;
```
jrasm91, line 33: If there is no type here, how did you figure it this gets called anywhere?

samip5: It seems the guidance on stackoverflow wasn't as great. :)

server/libs/infra/src/infra.config.ts
```diff
@@ -1,17 +1,20 @@
...
+
 function parseRedisConfig(): RedisOptions {
   const redisUrl = process.env.REDIS_URL;
   if (redisUrl && redisUrl.startsWith('ioredis://')) {
     try {
       const decodedString = Buffer.from(redisUrl.slice(10), 'base64').toString();
       return JSON.parse(decodedString);
     } catch (error) {
-      throw new Error(`Failed to decode redis options: ${error}`);
+      logger.error(`Failed to decode redis options: ${error}`);
```
jrasm91, line 17: This one should probably throw an error to prevent the app from starting if the config is invalid...

server/libs/infra/src/infra.config.ts
```diff
@@ -54,7 +57,7 @@ function parseTypeSenseConfig(): ConfigurationOptions {
         ...common,
       };
     } catch (error) {
-      throw new Error(`Failed to decode typesense options: ${error}`);
+      logger.error(`Failed to decode typesense options: ${error}`);
```
jrasm91, line 60: Same here

jrasm91 (approved):

PR #2251, opened 2023-04-14
fix(mobile): fix asset removal edge cases

fyfrey:
On iOS, assets can be in multiple local/device albums at once. This can lead to weird edge cases in the `SyncService`. One case is `_handleAssetRemoval` where the lists of candidates and existing assets may not contain duplicates to work correctly. This PR removes duplicates from these lists.


alextran1502 (approved):
> On iOS, assets can be in multiple local/device albums at once

Can you elaborate on this? What is the behavior on Android?

fyfrey:
> > On iOS, assets can be in multiple local/device albums at once
> 
> Can you elaborate on this? What is the behavior on Android?

Albums on iOS work like albums in Immich. The same asset can be in multiple albums at once.
On Android, this is not the case. Albums are more like folders and each asset can only be in one album.

PR #2255, opened 2023-04-15
feat(web): small responsivness improvements regarding mobile use

faupau03:
Small improvements on share tab, user management and jobs panel.
Prevent from overflowing, easier to interact on mobile
share tab
user management
jobs panel

### DIFF
--- a/web/src/lib/components/admin-page/jobs/job-tile.svelte
+++ b/web/src/lib/components/admin-page/jobs/job-tile.svelte
@@ -25,7 +25,7 @@
 </script>
 
 <div
-	class="flex xs:flex-row flex-col bg-gray-100 dark:bg-immich-dark-gray rounded-3xl overflow-hidden"
+	class="flex sm:flex-row flex-col bg-gray-100 dark:bg-immich-dark-gray rounded-3xl overflow-hidden"
 >
 	<div class="flex flex-col w-full">
 		{#if queueStatus.isPaused}
@@ -80,7 +80,7 @@
 			</div>
 		</div>
 	</div>
-	<div class="flex xs:flex-col flex-row xs:w-32 w-full overflow-hidden">
+	<div class="flex sm:flex-col flex-row sm:w-32 w-full overflow-hidden">
 		{#if !isIdle}
 			{#if waitingCount > 0}
 				<JobTileButton
--- a/web/src/lib/components/admin-page/jobs/job-tile.svelte
+++ b/web/src/lib/components/admin-page/jobs/job-tile.svelte
@@ -24,7 +24,7 @@
 	const dispatch = createEventDispatcher<{ command: JobCommandDto }>();
 </script>
 
-<div class="flex bg-gray-100 dark:bg-immich-dark-gray rounded-3xl overflow-hidden">
+<div class="xs:flex grid bg-gray-100 dark:bg-immich-dark-gray rounded-3xl overflow-hidden">
 	<div class="flex flex-col w-full">
 		{#if queueStatus.isPaused}
 			<JobTileStatus color="warning">Paused</JobTileStatus>
@@ -78,7 +78,7 @@
 			</div>
 		</div>
 	</div>
-	<div class="flex flex-col w-32 overflow-hidden">
+	<div class="flex xs:flex-col flex-row xs:w-32 w-full overflow-hidden">
 		{#if !isIdle}
 			{#if waitingCount > 0}
 				<JobTileButton
--- a/web/src/lib/components/admin-page/jobs/job-tile.svelte
+++ b/web/src/lib/components/admin-page/jobs/job-tile.svelte
@@ -24,7 +24,9 @@
 	const dispatch = createEventDispatcher<{ command: JobCommandDto }>();
 </script>
 
-<div class="xs:flex grid bg-gray-100 dark:bg-immich-dark-gray rounded-3xl overflow-hidden">
+<div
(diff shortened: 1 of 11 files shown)


jrasm91:
Can you add screenshots to the PR description?

web/src/routes/admin/+layout.svelte
```diff
@@ -86,7 +86,7 @@
 			</div>
 
 			<section id="setting-content" class="flex place-content-center mx-2">
-				<section class="w-[800px] pt-5 pb-28">
+				<section class="w-full pt-5 pb-28">
```
michelheusschen, line 89: While fullwidth is good for mobile, you have to be careful with larger screens. Right now (on a 2560x1440 screen) all admin pages look stretched out with lots of empty space between elements. You can have a look at [using a container](https://tailwindcss.com/docs/container) or just define a [max width](https://tailwindcss.com/docs/max-width) either for the layout or individual pages.

web/src/routes/(user)/sharing/+page.svelte
```diff
@@ -38,23 +38,23 @@
 		<LinkButton on:click={createSharedAlbum}>
 			<div class="flex place-items-center gap-1 text-sm">
 				<PlusBoxOutline size="18" />
-				Create shared album
+				<span class="max-xs:hidden">Create shared album</span>
```
michelheusschen, line 41: This removes an accessible name for screens smaller than breakpoint `xs`

web/src/lib/components/shared-components/side-bar/side-bar-button.svelte
```diff
@@ -22,7 +22,8 @@
     {isSelected
 		? 'bg-immich-primary/10 dark:bg-immich-dark-primary/10 text-immich-primary dark:text-[#adcbfa] hover:bg-immich-primary/25'
 		: ''}
-		{isCollapsed ? 'pl-5' : 'px-5'}
+		max-xs:pl-5
+		{isCollapsed ? 'pl-5' : 'xs:px-5'}
   "
```
michelheusschen, line 27: The JS based breakpoint `isCollapsed` and the CSS `xs` breakpoint are starting to get mixed here. I think it would be better to stick with either one. Using multiple breakpoint systems can lead to confusion and make debugging errors harder.

web/tailwind.config.cjs
```diff
@@ -19,6 +19,14 @@ module.exports = {
...
 				'immich-title': ['Snowburst One', 'cursive']
 			}
+		},
+		screens: {
+			xs: '430px',
+			sm: '640px',
+			md: '768px',
+			lg: '1024px',
+			xl: '1280px',
+			'2xl': '1536px'
```
michelheusschen, line 29: You've defined an extra `xs` breakpoint besides the default ones. I think you're trying to target mobile devices specifically with such a small breakpoint, but I often find the `sm` breakpoint to be better suited. For screens in the 430-640px range everything already starts to look crammed together, so the `xs` breakpoint might not be needed at all.

web/src/routes/admin/+layout.svelte
faupau03: fixed in 7b77a64205feefae7e1c40a6f3b5b3a6a39fa150

web/src/routes/(user)/sharing/+page.svelte
```diff
@@ -38,23 +38,23 @@
...
-				Shared links
+				<span class="max-xs:hidden">Shared links</span>
 			</div>
 		</LinkButton>
 	</div>
 
 	<section>
 		<!-- Share Album List -->
-		<div class="w-full flex flex-col place-items-center">
+		<div class="w-auto flex flex-col place-items-center">
```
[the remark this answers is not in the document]
faupau03: fixed in 7b77a64205feefae7e1c40a6f3b5b3a6a39fa150

web/src/routes/(user)/sharing/+page.svelte
faupau03: Fixed in 6baf73e0468a4ab5dffb89b639223c788e6f9d58

web/src/lib/components/admin-page/jobs/job-tile.svelte
```diff
@@ -24,7 +24,7 @@
 	const dispatch = createEventDispatcher<{ command: JobCommandDto }>();
 </script>
 
-<div class="flex bg-gray-100 dark:bg-immich-dark-gray rounded-3xl overflow-hidden">
+<div class="xs:flex grid bg-gray-100 dark:bg-immich-dark-gray rounded-3xl overflow-hidden">
```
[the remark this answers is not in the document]
faupau03: fixed in 5b91ed58d3c1c78158c9887e3cc93bc55b95314b

web/tailwind.config.cjs
faupau03: I think you're right. Changed in 41467184ee51e8bd04af67c649599384d2e465c4

web/src/lib/components/shared-components/side-bar/side-bar-button.svelte
faupau03: Changed in 607c55e4c6d6ea63a4e118f76f64b1d8f1231bad

michelheusschen (approved):
Good to see you went with the CSS only breakpoints. Not entirely sure about the user management table yet, maybe a card design would work better for both desktop and mobile. Also try to limit the amount of custom sizes in brackets and use `rem` instead of `px` so the layout properly scales when zooming.

But it's fine for now and that can be done in a future PR. Thank you for your efforts!

PR #2273, opened 2023-04-18
chore(web): update svelte kit to 1.15.7 (fixes CVE-2023-29008)

faupau03:


bo0tzz:
This is already covered by #2242, that just needs to go through. Thanks for looking out!

PR #2276, opened 2023-04-18
fix(web): accountinfopanel not closing on button press

faupau03:
When pressing on the profile button the accountinfopanel opens. When clicking outside of the panel it should close. But thats not the case if you press on the profile button again. Thats extremly annoying on mobile. This request fixes that.
accountinfopanel

### DIFF
--- a/web/src/lib/components/shared-components/navigation-bar/account-info-panel.svelte
+++ b/web/src/lib/components/shared-components/navigation-bar/account-info-panel.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-	import { clickOutside } from '$lib/utils/click-outside';
 	import { UserResponseDto } from '@api';
 	import { createEventDispatcher } from 'svelte';
 	import { page } from '$app/stores';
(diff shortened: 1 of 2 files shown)


michelheusschen:
Yep clicking the button indeed closes the infobox for a split second, only to reappear right after. That shouldn't happen, but I find your logic a bit hard to follow. What do you think of this solution?

All the events that are attached to the `div` should be attached to the `button`. See the example below, where I removed the `keydown` event, added a missing `blur` event (opposite of focus) and the `onclick` event now toggles the info panel. 


```svelte
<button
	class="flex place-items-center place-content-center rounded-full bg-immich-primary hover:bg-immich-primary/80 h-12 w-12 text-gray-100 dark:text-immich-dark-bg dark:bg-immich-dark-primary"
	on:mouseover={() => (shouldShowAccountInfo = true)}
	on:focus={() => (shouldShowAccountInfo = true)}
	on:blur={() => (shouldShowAccountInfo = false)}
	on:mouseleave={() => (shouldShowAccountInfo = false)}
	on:click={() => (shouldShowAccountInfoPanel = !shouldShowAccountInfoPanel)}
>
```

The infopanel can then be moved inside the div and the `clickOutside` event is removed from `AccountInfoPanel` and added to the parent div. Clicking outside will close the info panel and combined with the button events should lead to the desired behavior.

```svelte
<div use:clickOutside on:outclick={() => (shouldShowAccountInfoPanel = false)}>
  <button ...>

  {#if shouldShowAccountInfo && !shouldShowAccountInfoPanel}
    <div ...>
  {/if}

  {#if shouldShowAccountInfoPanel}
    <AccountInfoPanel {user} on:logout={logOut} />
  {/if}
</div>
```

michelheusschen (approved):

PR #2278, opened 2023-04-18
fix(web): noscript message

michelheusschen:
#2274 added a noscript message, but the image won't show in production builds because only assets in the `static` folder are copied to the production build. The noscript element was added to `app.html`, but since SvelteKit does SSR the message can be added to the root layout and use existing components.


alextran1502 (approved):

michelheusschen:
Woops misclick, didn't mean to close it

PR #2300, opened 2023-04-20
chore(web): flip album icon

jrasm91:
Before:

image
image

After:

image
image


michelheusschen (approved):
Looks all good, `flippedLogo` might be a better name instead of `flipped` but that's just a minor thing

alextran1502 (approved):

PR #2311, opened 2023-04-21
feat(web): show asset count in sharing tab and album viewer

faupau03:
Resolves #2301 partially. Mobile still needs to be done.
sharing-tab-shared
link-album-viewer
album-viewer


michelheusschen (approved):
Looks good after the code style issue is resolved

Issue #238, opened 2022-06-20
[Feature]: Global Map


matthinc:
### Feature detail

It would be nice to have a global map that shows the positions of all backed-up photos.

To limit the number of markers, [Marker Clustering](https://github.com/Leaflet/Leaflet.markercluster) could be used.

### Platform

Web

PR #2355, opened 2023-04-29
feat(web): Global map showing all assets with geo information

matthinc:
Screenshot 2023-04-29 at 10 27 51

TODO

- [x] Frontend map implementation
- [x] API
- [x] Dark Mode
- [x] Correct initial map centering
- [x] Asset grid showing all assets of a cluster _(Currently implemented using AssetViewer. Separate PR if needed)_
- ~_Optional:_ Heatmap UI~ (Will Do later)
- [x] Tests
- [x] Refactor / Linter

Implements #238


matthinc:
Since #2292 is abandoned.

klejejs:
@matthinc This is amazing, something that IMO Immich has been missing for me for a long time!
Just a personal opinion from my side - I really like how GPhotos has implemented the map by showing heat map instead of numbers on the map. Is that possible with Leaflet? Do you like this better than heat map?

web/src/lib/components/shared-components/side-bar/side-bar.svelte
```diff
@@ -108,6 +109,13 @@
 			isSelected={$page.route.id === '/(user)/explore'}
 		/>
 	</a>
+	<a data-sveltekit-preload-data="hover" href={AppRoute.MAP} draggable="false">
+		<SideBarButton
+			title="Map"
+			logo={Map}
+			flippedLogo={true}
```
jrasm91, line 116: Does this icon need to be flipped?

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
```diff
@@ -0,0 +1,71 @@
...
+
+		for (let asset of assets) {
+			if (asset.exifInfo) {
+				const lat = asset.exifInfo.latitude;
+				const lon = asset.exifInfo.longitude;
+
+				if (lat && lon) {
+					const icon = new Icon({
+						iconUrl: `/api/asset/thumbnail/${asset.id}?format=WEBP`,
+						iconRetinaUrl: `/api/asset/thumbnail/${asset.id}?format=WEBP`,
```
jrasm91, line 42: I think there is a method in `@api` for these urls.

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
```diff
@@ -0,0 +1,71 @@
...
+		cluster = new L.MarkerClusterGroup({
+			showCoverageOnHover: false,
+			zoomToBoundsOnClick: true,
+			spiderfyOnMaxZoom: true,
+			maxClusterRadius: 30,
+			spiderLegPolylineOptions: { opacity: 0 }
+		});
+
+		for (let asset of assets) {
+			if (asset.exifInfo) {
```
jrasm91, line 35: In the negative cases can we just continue and unindent everything below the if?

jrasm91 (commented):
Looks good! We're loading the whole asset list at the moment. Should we at least try only loading assets with valid gps data?

web/src/lib/components/shared-components/side-bar/side-bar.svelte
matthinc: I just copy-pasted this :P
Yeah, I'll remove that

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
matthinc: I'm aware of that method. I'm still in a very early stage in development and my goal is to get a first prototype pretty early. This will go through a refactoring later.

matthinc:
> @matthinc This is amazing, something that IMO Immich has been missing for me for a long time! Just a personal opinion from my side - I really like how GPhotos has implemented the map by showing heat map instead of numbers on the map. Is that possible with Leaflet? Do you like this better than heat map?

I _personally_ like clustered markers a bit better because they give you a bit more information IMO. But I also see that the heatmap might be superior in some situations. I think the heatmap would be super easy to implement since I know that there is a leaflet plugin for exactly that purpose. I'd suggest that I implement both options and add a setting.

matthinc:
> Looks good! We're loading the whole asset list at the moment. Should we at least try only loading assets with valid gps data?

I plan to copy this over from #2292. Parsing the whole `/api/asset` response is indeed overkill.

matthinc:
Screenshot 2023-05-01 at 21 48 31

Dark mode

jrasm91 (approved):
Can you explain the template changes you made for open API?

matthinc:
> Can you explain the template changes you made for open API?

The generator did this automatically - it downloaded a new version at some point.

web/src/lib/components/shared-components/leaflet/tile-layer.svelte
```diff
@@ -1,16 +1,26 @@
...
+	export let options: TileLayerFilterOptions | undefined = undefined;
 	let tileLayer: TileLayer;
 
 	const map = getMapContext();
 
 	onMount(() => {
-		tileLayer = new TileLayer(urlTemplate, options).addTo(map);
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+		// @ts-ignore
+		tileLayer = new L.tileLayer.colorFilter(urlTemplate, options).addTo(map);
```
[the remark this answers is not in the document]
matthinc: I use [this](https://github.com/xtk93x/Leaflet.TileLayer.ColorFilter) for the dark mode. Unfortunately the library does not support ts and is not in the `@types` library. I added a custom type for the parameters of this component. I thought it was ok for this single function call.

jrasm91:
Ah ok, that makes more sense

web/src/app.css
michelheusschen, line 1: These styles don't need to be in the global stylesheet. Also be careful with selectors like `.leaflet-marker-icon` because those will affect all instances of the map.

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
```diff
@@ -0,0 +1,90 @@
...
+
+	const dispatch = createEventDispatcher<{ view: { assets: string[] } }>();
+
+	export let markers: MapMarkerResponseDto[];
+
+	const map = getMapContext();
+
+	let cluster: L.MarkerClusterGroup;
+
+	setClusterContext(() => cluster);
```
michelheusschen, line 51: The cluster is added to the context, but never used

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
```diff
@@ -0,0 +1,90 @@
...
+	});
+
+	onDestroy(() => {
+		if (cluster) cluster.remove();
+	});
+</script>
+
+{#if cluster}
+	<slot />
+{/if}
```
michelheusschen, line 90: What is supposed to be passed to this slot?

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
```diff
@@ -0,0 +1,90 @@
...
+
+		for (let marker of markers) {
+			const leafletMarker = new AssetMarker(marker);
+
+			leafletMarker.on('click', () => {
+				dispatch('view', { assets: [marker.id] });
+			});
+
+			cluster.addLayer(leafletMarker);
+		}
```
michelheusschen, line 78: How does this perform with a large amount of assets? Will this block the main thread for a long time?

web/src/routes/(user)/map/+page.svelte
```diff
@@ -0,0 +1,90 @@
...
+		isViewingAssetStoreState,
+		viewingAssetStoreState
+	} from '$lib/stores/asset-interaction.store';
+	import { colorTheme } from '$lib/stores/preferences.store';
+
+	export let data: PageData;
+
+	let initialMapCenter: [number, number] = [48, 11];
+
+	if (data.mapMarkers.length) {
```
michelheusschen, line 18: This should be a reactive statement, since `data` could change in some circumstances

web/src/lib/components/shared-components/leaflet/tile-layer.svelte
michelheusschen, line 1: The benefits of the `leaflet.tilelayer.colorfilter` package aren't immediately obvious to me and I would suggest undoing the changes related to it and just apply the filter with regular css. Something like:

```svelte
<style>
	:global(.leaflet-layer) {
		filter: brightness(101%) contrast(101%) saturate(79%);
	}

	:global(.dark .leaflet-layer) {
		filter: invert(100%) brightness(127%) saturate(0%);
	}
</style>
```

web/src/routes/(user)/map/+page.server.ts
```diff
@@ -0,0 +1,23 @@
...
+import { redirect } from '@sveltejs/kit';
+import type { PageServerLoad } from './$types';
+
+export const load = (async ({ locals: { api, user } }) => {
+	if (!user) {
+		throw redirect(302, AppRoute.AUTH_LOGIN);
+	}
+
+	try {
+		const { data: mapMarkers } = await api.assetApi.getMapMarkers();
```
michelheusschen, line 11: Might be worth to only load the markers client-side to improve the initial page load time

web/src/routes/(user)/map/+page.svelte
```diff
@@ -0,0 +1,90 @@
...
+			assetInteractionStore.setViewingAssetId(viewingAssets[--viewingAssetCursor]);
+		}
+	}
+</script>
+
+<UserPageLayout user={data.user} title={data.meta.title}>
+	<div slot="buttons" />
+
+	<div class="h-[90%] w-full">
+		<Map latlng={initialMapCenter} zoom={7}>
```
michelheusschen, line 49: If I recall correctly, directly importing Leaflet (or components that import it) will lead to errors in production builds. Can you please validate that everything still works in production? I've previously [lazy loaded](https://github.com/immich-app/immich/blob/6acfb55dccf51d7a86a5f664cc9e40974cbc38dd/web/src/lib/components/asset-viewer/detail-panel.svelte#L213) all Leaflet components

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
matthinc: See below

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
matthinc: It's reasonably quick. I worked with the library before and even hundreds of thousands of markers are not an issue.

web/src/routes/(user)/map/+page.server.ts
matthinc: I experimented with that. My main reason was the following: The initial centering of the map depends on the loaded data. Loading client-side results in a change of the center point very quickly after opening the map tab.
This doesn't look very nice and also results in a lot of additional tile loads.

I found a little delay but complete map from the beginning a bit cleaner looking.

web/src/lib/components/shared-components/leaflet/tile-layer.svelte
matthinc: I have honestly no idea why this works - but it's brilliant!

web/src/routes/(user)/map/+page.svelte
matthinc: Strange, but thanks for the hint. I just copied your approach and it seems to work well.

web/src/lib/components/shared-components/leaflet/tile-layer.svelte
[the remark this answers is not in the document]
matthinc: Obsolte since the library was removed.

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
matthinc: I quickly checked on my prod instance. Absolutely no problem with 20k assets.

EPP100:
Glad to see that someone is continuing this feature :)

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
michelheusschen: Thanks for testing. After 10k assets I'm starting to see *some* slowdown, where loading takes ~500ms and seems to increase linearly. That's pretty good and definitely usable, but we should keep an eye on performance especially for users with a large library.

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
michelheusschen: Sure no problem

matthinc:
> Glad to see that someone is continuing this feature :)

Thanks for your PR. I took a lot of inspiration for the implementation from it!

web/src/routes/(user)/map/+page.server.ts
matthinc: @michelheusschen After sleeping a night, I think I'll try some hybrid implementation, you made ideed a valid point here. Maybe I extend the API endpoint so I can query the first, lets say 1000 assets server side and the rest client side.

web/src/routes/(user)/map/+page.svelte
michelheusschen: It happens because Leaflet accesses `window` which isn't available for server side rendered responses. Lazy loading worked previously because a user interaction would trigger the component load, so the server never attempted to render it. But now Leaflet is loaded on a page, which means the server will attempt to load this anyway and an error is logged (the page still works though).

To get rid of the logged error, I can think of two solutions:

1. Wrap the await import in a check to see if it's running in a browser.
```svelte
{#if browser}
	{#await import()}{/await}
{/if}
```

2. Remove the await import and disable SSR for the map page
```ts
// +page.server.ts
export const ssr = false;
```

matthinc:
@michelheusschen I'd like to tackle your performance suggestions in an upcoming PR. This would involve some changes to database code which would better be reviewed separately

web/src/routes/(user)/map/+page.svelte
michelheusschen: This still happens after the await fix. Loading the page doesn't error out anymore in production, but I'm still seeing an error being logged (without impacting usability) in the web container

michelheusschen (approved):
Performance can definitely be addressed in a future PR. Thanks for your work on this! There are some minor issues I noticed while testing, but nothing game breaking:
- The cluster marker doesn't seem to expand to individual assets when zooming in on a cluster with 2 assets for example (see preview environment)
- Cluster markers don't stand out very well in light mode

matthinc:
> Performance can definitely be addressed in a future PR. Thanks for your work on this! There are some minor issues I noticed while testing, but nothing game breaking:
> 
>     * The cluster marker doesn't seem to expand to individual assets when zooming in on a cluster with 2 assets for example (see preview environment)
> 
>     * Cluster markers don't stand out very well in light mode

Thanks a lot for the review! The preview is a bit of a special case here because lots of assets share geolocation.
The distance at which assets are clustered can be selected in the options of `leaflet.markercluster`, maybe we find a better setting one day.
As for the design - I'm sure Alex has tons of wishes so I'm waiting for his review :P

web/src/lib/components/shared-components/leaflet/tile-layer.svelte
jrasm91: Can we uninstall the plugin then?

web/src/routes/(user)/map/+page.svelte
matthinc: Weird. I don't see it in my logs locally.

web/src/lib/components/shared-components/leaflet/tile-layer.svelte
matthinc: It's not part of the PR any more (already removed from package.json)

jrasm91: Ah yes. I got it confused with the cluster plugin, which is still being used.

server/apps/immich/src/api-v1/asset/asset.controller.ts
```diff
@@ -260,6 +260,18 @@ export class AssetController {
     return await this.assetService.getAssetByTimeBucket(authUser, getAssetByTimeBucketDto);
   }
 
+  /**
+   * Get all assets that have GPS information embedded
+   */
+  @Authenticated()
+  @Get('/mapMarker')
```
alextran1502, line 267: I think this should be `map-marker`. It was my lack of understanding of writing REST API endpoint name to use camelCase :)

server/openapi-generator/templates/mobile/serialization/native/native_class.mustache
```diff
@@ -141,7 +141,7 @@ class {{{classname}}} {
         {{{name}}}: json[r'{{{baseName}}}'] is List
           ? (json[r'{{{baseName}}}'] as List).map((e) =>
               {{#items.complexType}}
-              {{items.complexType}}.listFromJson(json[r'{{{baseName}}}']){{#required}}{{^isNullable}}!{{/isNullable}}{{#uniqueItems}}.toSet(){{/uniqueItems}}{{/required}}{{^required}}{{#defaultValue}} ?? {{{.}}}{{/defaultValue}}{{/required}}
+              {{items.complexType}}.listFromJson(json[r'{{{baseName}}}']){{#uniqueItems}}.toSet(){{/uniqueItems}}
```
alextran1502, line 144: Are these changes auto generated?

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
```diff
@@ -0,0 +1,119 @@
+<script lang="ts" context="module">
+	import { createContext } from '$lib/utils/context';
+	import L from 'leaflet';
```
alextran1502, line 3: Can we use `import {MarkerClusterGroup} from 'leaflet'` instead of importing the whole library?

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
```diff
@@ -0,0 +1,119 @@
...
+
+	const { get: getContext, set: setClusterContext } = createContext<() => L.MarkerClusterGroup>();
+
+	export const getClusterContext = () => {
+		return getContext()();
+	};
+</script>
+
+<script lang="ts">
+	import 'leaflet.markercluster';
```
alextran1502, line 13: this is a very interesting way to import a functionality of a library. Is this a recommended way from the library? can we use something like `import {markerclsuter} from 'leaflet'`?

server/openapi-generator/templates/mobile/serialization/native/native_class.mustache
matthinc: Yes. The generator downloaded a new version at some point.

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -296,7 +296,7 @@
 
 <section
 	id="immich-asset-viewer"
-	class="fixed h-screen w-screen left-0 top-0 overflow-y-hidden bg-black z-[999] grid grid-rows-[64px_1fr] grid-cols-4"
```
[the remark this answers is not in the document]
matthinc: Yes, I tested it in all occurrences of this component. Problem was that leaflet uses z-index 0...1000 which caused display bugs.

web/src/lib/components/shared-components/leaflet/asset-marker-cluster.svelte
matthinc: I personally don't think this works because of the separation of TS types and actual implementation here. But I'll check tomorrow.

matthinc: Nope, does not work :/

PR #2382, opened 2023-05-05
feat(web): Global map performance optimizations

matthinc:
A collection of little improvements to the map's performance:

- Use array instead of map in `/asset/map-marker` response: Less descriptive but much smaller
```
{ "id": ..., "lat": ..., "lon": ...} ---> [id, lat, lon]
```
- Round lat/lon to 5 decimals (we don't need centimeter accuracy)
- Preload: Queries all markers without assetId during SSR and query assetId in a second request client-side
`/asset/map-marker?preload=1`
- Little refactorings
- Minor bug fixes


matthinc:
Please ignore the commit history. This was branched off #2355 before the merge.

server/apps/immich/src/api-v1/asset/response-dto/map-marker-response.dto.ts
```diff
@@ -0,0 +1,20 @@
...
+  number,
+  // longitude
+  number,
+];
+
+export function mapAssetMapMarker(asset: AssetEntity, preload: boolean): MapMarkerResponseDto {
+  const lat = asset.exifInfo?.latitude || 0;
+  const lon = asset.exifInfo?.longitude || 0;
+
+  const assetId = preload ? '' : asset.id;
```
jrasm91, line 17: This sends an empty string for each asset? They doesn't seem ideal.

web/src/api/api.ts
```diff
@@ -86,4 +86,35 @@ export class ImmichApi {
...
+	// assetId
+	string,
+	// latitude
+	number,
+	// longitude
+	number
+];
+
+export class MapMarkerResponseDtoDecorator {
+	private dto: MapMarkerResponseDto;
```
jrasm91, line 99: This line isn't necessary if you just add private in the constructor.

web/src/api/api.ts
```diff
@@ -86,4 +86,35 @@ export class ImmichApi {
...
+
+	constructor(dto: MapMarkerResponseDto) {
+		this.dto = dto;
+	}
+
+	static map(response: object[]): MapMarkerResponseDtoDecorator[] {
```
jrasm91, line 105: If you're going to have a class that represents this and take the time to convert from the array format into an instance for each record, you should just have named properties on this like private string assetId, etc. and they should be set in the constructor. Definitely should not be using unknown casting anywhere here.

web/src/routes/(user)/map/+page.svelte
```diff
@@ -8,17 +8,25 @@
...
+	let mapMarkers: MapMarkerResponseDtoDecorator[] = [];
 
 	let initialMapCenter: [number, number] = [48, 11];
 
-	$: {
-		if (data.mapMarkers.length) {
-			let firstMarker = data.mapMarkers[0];
-			initialMapCenter = [firstMarker.lat, firstMarker.lon];
+	onMount(async () => {
+		mapMarkers = MapMarkerResponseDtoDecorator.map(data.mapMarkers);
```
jrasm91, line 21: Map is a JavaScript standard array function and calling this custom method the same thing is a bit misleading/confusing. Maybe we should use an actual array.map here. So like response.map(item => Custom class.from(item)) and call the method that converts from plain to instance `from` instead of `map`. Or, just do new CustomClass(item).

server/apps/immich/src/api-v1/asset/response-dto/map-marker-response.dto.ts
matthinc: This way the array has always the same length, no matter if `preload` is true or not. This way we need just one DTO and 3 characters `"",` are IMO not worth removing.

web/src/api/api.ts
matthinc: The problem is that OpenAPI generates `Array<object>` here. I don't think there is a way of casting that to an array without `unknown`?

server/apps/immich/src/api-v1/asset/response-dto/map-marker-response.dto.ts
jrasm91: I don't understand the benefit of the array always being the same length. You have to know if the value is going to be there or not still. 

What do you think about making assetId the third field and making it optional?

server/apps/immich/src/api-v1/asset/asset.service.ts
```diff
@@ -144,10 +143,12 @@ export class AssetService {
...
 
-  public async getMapMarkers(authUser: AuthUserDto, dto: AssetSearchDto): Promise<MapMarkerResponseDto[]> {
-    const assets = await this._assetRepository.getAllByUserId(authUser.id, dto);
+  public async getMapMarkers(authUser: AuthUserDto, preload: boolean): Promise<MapMarkerResponseDto[]> {
+    const assets = await this._assetRepository.getMapMarkerByUserId(authUser.id);
 
-    return assets.map((asset) => mapAssetMapMarker(asset)).filter((marker) => marker != null) as MapMarkerResponseDto[];
+    return assets
+      .map((asset) => mapAssetMapMarker(asset, preload))
+      .filter((marker) => marker != null) as MapMarkerResponseDto[];
```
michelheusschen, line 151: Why is the filter needed? The type assertion also seems unnecessary

server/apps/immich/src/api-v1/asset/response-dto/map-marker-response.dto.ts
```diff
@@ -0,0 +1,24 @@
...
+  number,
+  // assetId
+  string?,
+];
+
+export function mapAssetMapMarker(asset: AssetEntity, preload: boolean): MapMarkerResponseDto {
+  const lat = asset.exifInfo?.latitude || 0;
+  const lon = asset.exifInfo?.longitude || 0;
+
+  const response = [roundToDecimals(lat, 5), roundToDecimals(lon, 5)] as MapMarkerResponseDto;
```
michelheusschen, line 17: This should be a regular type annotation instead of a cast. And for rounding we can also use Lodash, instead of having to create our own implementation

web/src/api/api.ts
```diff
@@ -86,4 +86,41 @@ export class ImmichApi {
...
+export type MapMarkerResponseDto = [
+	// latitude
+	number,
+	// longitude
+	number,
+	// assetId
+	string
+];
+
+export class MapMarkerResponseDtoDecorator {
```
michelheusschen, line 98: That's an unusual way of doing this conversion. What do you think about the following implementation?

```typescript
export type MapMarkerResponseDto = [
	// latitude
	number,
	// longitude
	number,
	// assetId
	string?
];

export type MapMarker = {
	assetId?: string;
	location: [number, number];
};

export function convertObjectToMarker(obj: object): MapMarker {
	const [lat, lon, assetId] = obj as MapMarkerResponseDto;

	return {
		assetId,
		location: [lat, lon]
	};
}

// Then to use it on the map page:
mapMarkers = data.mapMarkers.map(convertObjectToMarker);
```

matthinc: I personally prefer having everything in a contained class instead of separate type definitions and methods tbh. Classes also provide more flexibility in case more attributes get added to the array in the future.

michelheusschen: Hmm I mainly dislike the `from` and `fromMany` methods, because they're pretty unusual. Besides, this code isn't directly related to the API and may as well be in it's own file. And maybe remove decorator part from the name, since it's not a decorator.

In my opinion the following approach is more idiomatic, but all good if you keep it as is:

```typescript
export class MapMarker {
	private lat: number;
	private lon: number;
	public readonly assetId?: string;

	constructor(obj: object) {
		const [lat, lon, assetId] = obj as MapMarkerResponseDto;

		this.lat = lat;
		this.lon = lon;
		this.assetId = assetId;
	}

	// Can be accessed with marker.location to make destructuring easier
	get location(): [number, number] {
		return [this.lat, this.lon];
	}
}

// markers.map(obj => new MapMarker(obj))
```

matthinc: I agree that the term `decorator` is misleading here, therefore I renamed the class.

server/apps/immich/src/api-v1/asset/asset-repository.ts
```diff
@@ -238,6 +239,30 @@ export class AssetRepository implements IAssetRepository {
     });
   }
 
+  /**
+   * Get all assets belong to the user on the database and that contain exif location information
+   * @param ownerId
+   */
+  async getMapMarkerByUserId(ownerId: string): Promise<AssetEntity[]> {
```
jrasm91, line 246: What do you think about excluding archived assets?

web/src/routes/(user)/map/+page.svelte
```diff
@@ -8,17 +8,25 @@
...
-			initialMapCenter = [firstMarker.lat, firstMarker.lon];
+	onMount(async () => {
+		mapMarkers = MapMarkerResponse.fromMany(data.mapMarkers);
+
+		if (mapMarkers.length) {
+			initialMapCenter = mapMarkers[0].getLatLon();
 		}
-	}
+
+		const markersResponse = (await api.assetApi.getMapMarkers()).data;
```
jrasm91, line 27: Everywhere else in the code we use destructuring. Eg `const { data: markersResponse } = await ...`

jrasm91 (approved):

server/apps/immich/src/api-v1/asset/asset-repository.ts
```diff
@@ -238,6 +239,30 @@ export class AssetRepository implements IAssetRepository {
...
+    return this.assetRepository.find({
+      where: {
+        ownerId,
+        resizePath: Not(IsNull()),
+        isVisible: true,
+        exifInfo: {
+          latitude: Not(IsNull()),
+          longitude: Not(IsNull()),
```
bo0tzz, line 254: Should lat/long have an index on them for this query?

server/apps/immich/src/api-v1/asset/asset-repository.ts
```diff
@@ -238,6 +239,30 @@ export class AssetRepository implements IAssetRepository {
     });
   }
 
+  /**
+   * Get all assets belong to the user on the database and that contain exif location information
+   * @param ownerId
+   */
+  async getMapMarkerByUserId(ownerId: string): Promise<AssetEntity[]> {
+    return this.assetRepository.find({
+      where: {
```
michelheusschen, line 248: This query selects all fields of the asset and exif entities by default, which takes more time and can lead to high memory consumption. I would recommend to only select the fields that are actually used:

```suggestion
    return this.assetRepository.find({
      select: {
        id: true,
        exifInfo: {
          latitude: true,
          longitude: true,
        },
      },
      where: {
```

lefuturiste:
@matthinc if you don't mind, what's the reason that you've closed this PR?

jrasm91:
Superseded by #2498

PR #2388, opened 2023-05-06
feat (server, web): Share with partner

brighteyed:
This PR implements a minimal version of Share with partner feature (for web UI only). Similar to Google Photos implementation

- Added `Sharing` section to user account page:
user-settings

- Here I can add users that will be able to access my photos:
add-partner-modal-2

- I can see partners that have shared their assets with me at the `Sharing` page above shared albums:
sharing-page

- `Sharing` page uses timeline view:
assets

I have added new `partners` table with `sharedBy` and `sharedWith` columns. In the future I think this table could store some sharing options: share archived assets or not, starting day, share only photos with specified persons detected, etc.

- Multiple selection is supported. Selected partner's photos can be downloaded
assets_selected

Endpoints:
- `POST /partner/:id` - add partner who will be able to see my assets
- `DELETE /partner/:id` - delete partner who will be able to see my assets
- `GET /partner?direction=shared-by` - get all partners who are able to see my assets
- `GET /partner?direction=shared-with` - get all partners who share assets with me


alextran1502:
Thank you for the PR. I think to make this work, we have to refactor the timeline into components and then use it for this feature. Otherwise, it wouldn't work for the majority of users with more than 5000 assets on their timeline

brighteyed:
I agree that using timeline view instead of the `GalleryViewer` is a preferred solution. I would love to see timeline here. But I haven't tested `GalleryViewer` with thousands of photos so it's hard for me to say if using timeline view is a blocking requirement for this PR.
Not sure if I am capable to make that refactoring myself. Need to look deep into code.

alextran1502:
@brighteyed from my previous benchmark, about 4000-5000 assets would start to cause issue and the browser would eventually crash

brighteyed:
Oh, that's bad. I will try to refactor the `AssetGrid` and server API so that other pages can use the timeline

brighteyed:
> I will try to refactor the `AssetGrid` and server API so that other pages can use the timeline

Done. `AssetGrid` now has an optional `user` parameter. Modified `getAssetCountByTimeBucket` and `getAssetByTimeBucket` methods of the `AssetService`

server/apps/immich/src/api-v1/asset/asset.service.ts
```diff
@@ -492,6 +514,20 @@ export class AssetService {
...
+          if ((await this._assetRepository.countByIdAndUser(assetId, partnerId)) == 1) {
+            owner = true;
+            break;
+          }
+        }
+
+        if (owner) {
+          continue;
+        }
+
```
jrasm91, line 530: Ideally this is added to partnerCore and would just be something like `await this.partnerCore.hasAssetAccess(assetId, authUser.id)`. I think it could be a single query as well. You can join asset to user to partner in a single request.

server/apps/immich/src/api-v1/asset/asset.service.ts
```diff
@@ -505,6 +541,16 @@ export class AssetService {
     }
   }
 
+  private async checkUserAccess(authUser: AuthUserDto, userId: string) {
+    // Check if userId shares assets with authUser
+    const hasAccess = (await this.partnerCore.getAll(authUser.id, 'shared-with')).some(
+      (item) => item.sharedBy == userId,
```
jrasm91, line 547: It'd be great to filter this in the database itself, rather than loading it into memory and then filtering.

server/apps/immich/src/controllers/user.controller.ts
```diff
@@ -106,4 +107,22 @@ export class UserController {
...
+  @Post('partner/:id')
+  addPartner(@GetAuthUser() authUser: AuthUserDto, @Param() { id }: UUIDParamDto): Promise<void> {
+    return this.service.addPartner(authUser, id);
+  }
+
+  @Authenticated()
+  @Delete('partner/:id')
+  removePartner(@GetAuthUser() authUser: AuthUserDto, @Param() { id }: UUIDParamDto): Promise<void> {
+    return this.service.removePartner(authUser, id);
+  }
```
jrasm91, line 127: This looks great, but it would be better IMO for this to be in a separate controller, `PartnerController` with the url `/partner/...` instead of `/user/partner/...`

server/libs/domain/src/partner/partner.core.ts
```diff
@@ -0,0 +1,38 @@
...
+    try {
+      return this.repository.create({
+        sharedBy: sharedBy,
+        sharedWith: sharedWith,
+      });
+    } catch (error: any) {
+      this.logger.error(error, error.stack);
+      throw new InternalServerErrorException('failed to create partner');
+    }
+  }
```
jrasm91, line 28: I don't know if I care about catching and rethrowing an internal exception here or not. We don't really do it most places. I might prefer not re-throwing, but just letting the original error bubble and handle it somewhere else.

server/libs/infra/src/entities/partner.entity.ts
```diff
@@ -0,0 +1,21 @@
...
+@Entity('partners')
+export class PartnerEntity {
+  @PrimaryGeneratedColumn('uuid')
+  id!: string;
+
+  @ManyToOne(() => UserEntity, { onDelete: 'CASCADE' })
+  sharedBy!: string;
+
+  @ManyToOne(() => UserEntity, { onDelete: 'CASCADE' })
+  sharedWith!: string;
```
jrasm91, line 14: This is basically an many to many intersection table between user and user. I don't think we need the id column here, and I don't think it is used anywhere, is it? The shared with/by combination could be a joint primary column, which is how intersection tables often are configured (both columns have `@PrimaryColumn()` decorators). I believe right now `UserA` could re-share with `UserB` twice and the database would have no problem with it. Since we don't need that specific functionality I think I'd prefer using two columns for the primary key instead of a uuid.

I've never used`@ManyToX` relationships with the type `string` oppose to `Entity`. Does this just return the uuid of the user oppose to the user entity, or does it return the user and the type is just wrong?

server/libs/infra/src/entities/user.entity.ts
```diff
@@ -53,4 +54,10 @@ export class UserEntity {
 
   @OneToMany(() => AssetEntity, (asset) => asset.owner)
   assets!: AssetEntity[];
+
+  @OneToMany(() => PartnerEntity, (partner) => partner.sharedBy)
+  sharedWith!: PartnerEntity[];
+
+  @OneToMany(() => PartnerEntity, (partner) => partner.sharedWith)
+  sharedBy!: PartnerEntity[];
```
jrasm91, line 62: The wording is a bit tricky because users have `shared` their libraries `with` me, but I've also `shared` my library `with` others.

`user0.sharedWith=user 1, user 2` means user 0 has shared their user 1 and user 2, right?

server/libs/infra/src/repositories/partner.repository.ts
```diff
@@ -0,0 +1,63 @@
...
+    } else {
+      return this.repository
+        .createQueryBuilder('partners')
+        .select('partners.sharedBy', 'sharedBy')
+        .addSelect('partners.sharedWith', 'sharedWith')
+        .addSelect('partners.createdAt', 'createdAt')
+        .addSelect('partners.updatedAt', 'updatedAt')
+        .addSelect('partners.id', 'id')
+        .where('"sharedWithId" = :sharedWith', { sharedWith: userId })
+        .getRawMany();
```
jrasm91, line 36: These two conditions are exactly the same except the where clause. Can you re-write it like this instead?
```typescript
return this.repository.find({ where: { [direction === 'shared-by'? 'sharedById' : 'sharedWithId']: userId })
```

I'd also recommend changing the relationship on the entity by include the id columns as strings and the relations as `UserEntity`.

server/libs/infra/src/repositories/partner.repository.ts
```diff
@@ -0,0 +1,63 @@
...
+        .select('partners.sharedBy', 'sharedBy')
+        .addSelect('partners.sharedWith', 'sharedWith')
+        .addSelect('partners.createdAt', 'createdAt')
+        .addSelect('partners.updatedAt', 'updatedAt')
+        .addSelect('partners.id', 'id')
+        .where('"sharedById" = :sharedBy', { sharedBy: sharedBy })
+        .andWhere('"sharedWithId" = :sharedWith', { sharedWith: sharedWith })
+        .getRawOne<PartnerEntity>()) ?? null;
+
+    return entity;
```
jrasm91, line 53: We should use the normal `this.repository.find` instead of the query builder. I think you are using this because of only having a single property on the entity for the relation/id, but there are ways around that, which still make it possible to avoid the query builder.

jrasm91 (commented):
This is looking really good, thanks for taking it on!

server/libs/domain/src/partner/partner.core.ts
brighteyed: Removed the same rethrowing in ShareCore also

server/apps/immich/src/api-v1/asset/asset.service.ts
brighteyed: Ah, sure. `PartnerCore` already has the needed method

server/libs/infra/src/entities/user.entity.ts
brighteyed: There was a mistake in the field names (`shareBy/shareWith` ). Fixed while I was working on the relations for `PartnerEntity`

brighteyed:
@jrasm91 Thanks for the detailed review! Changed relations from `string` to `UserEntity` in `PartnerEntity` and created composite primary keys for the `partners` table

server/libs/domain/src/partner/partner.core.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  create(sharedBy: string, sharedWith: string): Promise<PartnerEntity> {
+    return this.repository.create({
+      sharedBy,
+      sharedWith,
+    });
+  }
+
+  async remove(sharedBy: string, sharedWith: string): Promise<PartnerEntity> {
```
jrasm91, line 25: Just a recommendation here - I have personally messed up a few times where I called a method like this with the arguments in the wrong order. It may be worth making an interface for this and taking advantage of the difference in key names, opposed to relying solely on the argument order. Especially since they (for me at least) are already somewhat easy to get mixed up.

```typescript
export interface PartnerId {
  sharedBy: string;
  sharedWith: string;
}

get({ sharedBy, sharedWith }: PartnerIds): ...
create({ sharedBy, sharedWith }: PartnerIds): ...
```

server/libs/infra/src/migrations/1683734723680-RemovePartnerIdColumn.ts
```diff
@@ -0,0 +1,30 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class RemovePartnerIdColumn1683734723680 implements MigrationInterface {
+    name = 'RemovePartnerIdColumn1683734723680'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
```
jrasm91, line 6: Since this is within the same PR, we should be able to revert/drop both of these migrations, and then re-generate a single/unified one where the table is created how we want it from the get-go. There are some scripts in the server/package.json for undoing migrations and regenerating them.

TLDR - this PR should have a single migration file that creates the partner table.

server/apps/immich/src/api-v1/asset/asset.controller.ts
```diff
@@ -251,6 +251,21 @@ export class AssetController {
...
+   */
+  @Authenticated()
+  @Get('/partner/:id')
+  @ApiHeader({
+    name: 'if-none-match',
+    description: 'ETag of data already cached on the client',
+    required: false,
+    schema: { type: 'string' },
+  })
+  getUserAssets(@GetAuthUser() authUser: AuthUserDto, @Param('id') id: string): Promise<AssetResponseDto[]> {
```
michelheusschen, line 265: This endpoint can be called with any user ID and it will return their assets, because there's no partner validation. Also the `id` param could use better validation: `@Param() { id }: UUIDParamDto`

web/src/lib/components/user-settings-page/partner-settings.svelte
```diff
@@ -0,0 +1,107 @@
...
+
+	$: {
+		if (browser) {
+			if (showAddUserModal) {
+				document.body.style.overflow = 'hidden';
+			} else {
+				document.body.style.overflow = 'auto';
+			}
+		}
+	}
```
michelheusschen, line 22: Scroll will stay disabled if the user opens the modal and goes to another page without closing it. This can probably be added to the modal component using lifecycle functions instead

web/src/lib/components/user-settings-page/partner-settings.svelte
```diff
@@ -0,0 +1,107 @@
...
+	const loadPartners = async () => {
+		const { data: sharedWith } = await api.partnerApi.getAllPartners();
+
+		const partners: UserResponseDto[] = [];
+		for (const partner of sharedWith) {
+			const { data: user } = await api.userApi.getUserById(partner.sharedWith);
+			if (user) {
+				partners.push(user);
+			}
+		}
```
michelheusschen, line 65: Can't the user data be included in the response of the `getAllPartners()` request? I've noticed this looped data fetching in a few other places as well, it would be preferred if the server could send this in one request. But not a big issue, since the number of partners is likely to be small.

web/src/routes/(user)/sharing/[userId]/+page.svelte
```diff
@@ -0,0 +1,24 @@
...
+	let partnerAssets: AssetResponseDto[] = [];
+
+	onMount(async () => {
+		try {
+			const { data: assets } = await api.assetApi.getUserAssets(data.partner.id);
+			partnerAssets = assets;
+		} catch {
+			handleError(Error, 'Unable to load assets');
+		}
+	});
```
Comment (michelheusschen) [inline web/src/routes/(user)/sharing/[userId]/+page.svelte:19]: The `partnerAssets` data seems to be never used

server/apps/immich/src/controllers/share.controller.ts
```diff
@@ -24,6 +25,12 @@ export class ShareController {
     return this.service.getMine(authUser);
   }
 
+  @Authenticated()
+  @Get('partner')
+  getPartners(@GetAuthUser() authUser: AuthUserDto): Promise<PartnerResponseDto[]> {
+    return this.service.getAllPartners(authUser);
+  }
+
```
jrasm91, line 33: This is really tricky, because if it is below the next get request it'll never get hit. This is just asking for trouble IMO.

One solution would be to update the link urls to have a `link` prefix:
```
GET /share/link
GET /share/link/me
GET /share/link/:id
GET /share/partner
```

However, I'd probably prefer to just keep all the partner sharing in the `PartnerController` and all of the shared links in this controller, which might be better named `SharedLinkController`. If each controller handles CRUD operations for the entity it is easier to follow.

We could move this endpoint to the `PartnerController` and use `GET /partner?sent=1` or similar to get this list back. 

The actual implementation of this method is identical to the one in partner controller, it just uses a different `PartnerDirection`, so I think it might be better for the url to exist inside of that controller.

server/libs/domain/src/share/share.service.ts
```diff
@@ -57,4 +61,9 @@ export class ShareService {
     const link = await this.shareCore.save(authUser.id, id, dto);
     return mapSharedLink(link);
   }
+
+  async getAllPartners(authUser: AuthUserDto): Promise<PartnerResponseDto[]> {
+    const partners = await this.partnerCore.getAll(authUser.id, 'shared-with');
+    return partners.map(mapPartner);
+  }
```
jrasm91, line 68: Just a note - if this was moved to the `PartnerService`, all the partner logic would be accessed from a single controller/service and `PartnerCore` could be merged/flattened back into the `PartnerService`.

server/libs/infra/src/entities/partner.entity.ts
```diff
@@ -0,0 +1,26 @@
...
+  @PrimaryColumn('uuid')
+  sharedWithId!: string;
+
+  @ManyToOne(() => UserEntity, (user) => user.sharedBy, { onDelete: 'CASCADE' })
+  @JoinColumn({ name: 'sharedById' })
+  sharedBy!: UserEntity;
+
+  @ManyToOne(() => UserEntity, (user) => user.sharedWith, { onDelete: 'CASCADE' })
+  @JoinColumn({ name: 'sharedWithId' })
+  sharedWith!: UserEntity;
```
jrasm91, line 19: This looks perfect :pray:

server/libs/infra/src/repositories/partner.repository.ts
```diff
@@ -0,0 +1,62 @@
...
+  create(createPartnerDto: CreatePartnerDto): Promise<PartnerEntity> {
+    const partner = new PartnerEntity();
+
+    partner.sharedWith = new UserEntity();
+    partner.sharedWith.id = createPartnerDto.sharedWith;
+
+    partner.sharedBy = new UserEntity();
+    partner.sharedBy.id = createPartnerDto.sharedBy;
+
+    return this.repository.save(partner);
```
jrasm91, line 56: This works fine, but I think you could also do

```typescript
return this.repository.save({ sharedWithId: dto.sharedWithId: sharedById: dto.sharedBy });
```

AFAIK setting `sharedWithId` is the equivalent to setting `sharedWith: { id: sharedWithId }`. At the end of the day TypeORM is just going to write values to those two columns (plus the computed ones).

web/src/lib/components/user-settings-page/partner-settings.svelte
jrasm91: We actually load from the database the `UserEntity` and then just map it to the id. We could definitely just update the response to send the `UserResponseDto` as a list for sharedWith and sharedBy instead of the id + client side querying.

server/apps/immich/src/api-v1/asset/asset.controller.ts
brighteyed: Removed `getUserAssets` method. I forgot to delete it while I was moving from `GalleryViewer` to `AssetGrid`

server/apps/immich/src/api-v1/asset/asset.controller.ts
```diff
@@ -251,6 +251,21 @@ export class AssetController {
...
+    return this.assetService.getUserAssets(authUser, id);
```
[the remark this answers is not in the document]
brighteyed: Removed this method

web/src/lib/components/user-settings-page/partner-settings.svelte
brighteyed: Thanks for pointing this out. Now code looks much better

jrasm91:
Are you on the immich discord server by chance?

server/apps/immich/src/controllers/share.controller.ts
brighteyed: Moved partner sharing endpoints to `PartnerController`. Also renamed `ShareController` to `SharedLinkController`

server/libs/domain/src/share/share.service.ts
brighteyed: `PartnerCore` seems to be necessary by `AssetService` to check asset access

jrasm91: Yes, this looks like one method which makes sense for another controller to potentially handle.

web/src/lib/components/user-settings-page/partner-settings.svelte
brighteyed: I moved this code to `BaseModal`, but then after some testing and thinking I decided to revert this scrollbar style changes in this PR at all. I don't like how centered content of the page (user account settings, for example) immediately jumps to the right after I hide a scrollbar

server/libs/infra/src/entities/partner.entity.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  @CreateDateColumn({ type: 'timestamptz' })
+  createdAt?: string;
+
+  @UpdateDateColumn({ type: 'timestamptz' })
+  updatedAt?: string;
```
michelheusschen, line 25: Both these fields are *not* nullable in the database and that should be reflected here. I'm also not too sure if the `string` type is correct, from the [TypeORM docs](https://typeorm.io/decorator-reference#createdatecolumn) it looks to be `Date`

web/src/lib/components/user-settings-page/partner-settings.svelte
```diff
@@ -0,0 +1,99 @@
...
+
+					<div class="text-left">
+						<p class="text-immich-fg dark:text-immich-dark-fg">
+							{partner.firstName}
+							{partner.lastName}
+						</p>
+						<p class="text-xs ">
+							{partner.email}
+						</p>
+					</div>
```
michelheusschen, line 81: Email has poor contrast in dark mode

web/src/lib/components/user-settings-page/partner-settings.svelte
```diff
@@ -0,0 +1,99 @@
...
+		partners = await loadPartners();
+	});
+</script>
+
+<section class="my-4">
+	{#if partners.length > 0}
+		<div class="flex flex-row gap-4">
+			{#each partners as partner}
+				<div class="flex rounded-lg gap-4 py-4 px-5 transition-all">
+					<CircleAvatar user={partner} />
```
michelheusschen, line 71: The `CircleAvatar` is a button and clickable, but there's no action associated with clicking it

web/src/lib/components/user-settings-page/partner-selection-modal.svelte
```diff
@@ -0,0 +1,84 @@
...
+	import BaseModal from '../shared-components/base-modal.svelte';
+	import CircleAvatar from '../shared-components/circle-avatar.svelte';
+	import ImmichLogo from '../shared-components/immich-logo.svelte';
+	import Button from '../elements/buttons/button.svelte';
+	import { createEventDispatcher, onMount } from 'svelte';
+
+	let availableUsers: UserResponseDto[] = [];
+	let selectedUsers: UserResponseDto[] = [];
+
+	const dispatch = createEventDispatcher();
```
michelheusschen, line 12: Should include types for all events that can be dispatched, so in this component that would be for `close` and `add-user`

web/src/lib/components/user-settings-page/partner-settings.svelte
```diff
@@ -0,0 +1,99 @@
...
+						<p class="text-xs ">
+							{partner.email}
+						</p>
+					</div>
+					<button
+						on:click={() => handleRemovePartner(partner)}
+						class="immich-circle-icon-button hover:bg-immich-primary/10 dark:text-immich-dark-fg hover:dark:bg-immich-dark-primary/20  rounded-full p-3 flex place-items-center place-content-center transition-all"
+					>
+						<Close size="16" />
+					</button>
```
michelheusschen, line 87: Can use the `CircleIconButton` component and maybe add `title="Remove partner"`

web/src/lib/components/user-settings-page/partner-settings.svelte
brighteyed: Hm, in the future I believe it should be the action to show sharing settings (dialog or even separate page) for the partner. But now I can imagine only remove partner action. Not sure it is better than no action. 

What do you think?

server/libs/infra/src/entities/partner.entity.ts
brighteyed: Thank you! I see several other entities also use `string` type. Looks like some kind of tech-debt

web/src/routes/(user)/partners/[userId]/+page.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
+	import type { PageData } from './$types';
+	import AssetGrid from '$lib/components/photos-page/asset-grid.svelte';
+	import UserPageLayout from '$lib/components/layouts/user-page-layout.svelte';
+
+	export let data: PageData;
+</script>
+
+<UserPageLayout user={data.user} title="{data.partner.firstName} {data.partner.lastName}'s Photos">
+	<AssetGrid user={data.partner} />
```
Comment (michelheusschen) [inline web/src/routes/(user)/partners/[userId]/+page.svelte:10]: The thumbnails in the asset grid can be selected, but the action bar doesn't show up and the scrubbable scollbar isn't present like in the main timeline. Would be good to keep the experience the same regardless if users are viewing their own timeline or that of others.

web/src/routes/(user)/partners/[userId]/+page.svelte
Comment (michelheusschen) [inline web/src/routes/(user)/partners/[userId]/+page.svelte:1]: Ran into a weird situation. If I go from the main timeline `/photos` and then navigate to a shared timeline `/partner/{id}` the shared timeline doesn't show any assets. Do you also have this behavior?

server/apps/immich/src/controllers/partner.controller.ts
```diff
@@ -0,0 +1,36 @@
...
+  getPartners(
+    @GetAuthUser() authUser: AuthUserDto,
+    @Query('direction') direction: PartnerDirection,
+  ): Promise<UserResponseDto[]> {
+    return this.service.getAll(authUser, direction);
+  }
+
+  @Authenticated()
+  @Post(':id')
+  createPartner(@GetAuthUser() authUser: AuthUserDto, @Param() { id }: UUIDParamDto): Promise<UserResponseDto> {
```
michelheusschen, line 27: This should have `UserResponseDto` as response, but only `{ id }` gets returned

michelheusschen (commented):
Looking really good now! Just some last few things and this should be good to go

server/apps/immich/src/controllers/partner.controller.ts
jrasm91: Hmmk, might need to reload the entity in the repository to populate the relations.

server/libs/infra/src/repositories/partner.repository.ts
```diff
@@ -0,0 +1,52 @@
...
+  getAll(userId: string): Promise<PartnerEntity[]> {
+    return this.repository.find({ where: [{ sharedWithId: userId }, { sharedById: userId }] });
+  }
+
+  get({ sharedWithId, sharedById }: PartnerIds): Promise<PartnerEntity | null> {
+    return this.repository.findOne({ where: { sharedById, sharedWithId } });
+  }
+
+  create({ sharedById, sharedWithId }: PartnerIds): Promise<PartnerEntity> {
+    return this.repository.save({
```
jrasm91, line 20: Turns out this doesn't load relations for the users in the newly created entity.

We should change this to

```typescript
await this.repository.save(...);
return this.reposotory.findOrFail(...);
```

web/src/routes/(user)/partners/[userId]/+page.svelte
Comment (jrasm91) [reply web/src/routes/(user)/partners/[userId]/+page.svelte:10]: That might be a bit complicated. What actions would need to be enabled for the selected assets? Do we support adding partner assets to albums, etc.? Might be non trivial to have that correctly integrated. It seems like there will be some nuances with it. I'd be happy with read only for now and another PR for actions later.

Comment (brighteyed) [reply web/src/routes/(user)/partners/[userId]/+page.svelte:10]: I see Google Photos allows a limited set of actions for partner's photos (view, info and save to the user's library only). At the moment actions are only available while viewing assets in `AssetViewer` (copy, download, info)

I haven't noticed the scrollbar issue myself while testing. I don't like that this page doesn't have the same scroller as the main timeline, but looking into the code, I unfortunately can't understand why. I could add `slot="content"` attribute to the `AssetGrid` to fix it, but in this case the page loses its header

web/src/routes/(user)/partners/[userId]/+page.svelte
Comment (brighteyed) [reply web/src/routes/(user)/partners/[userId]/+page.svelte:1]: I can't reproduce it. Tested in Chrome and Edge

Comment (michelheusschen) [reply web/src/routes/(user)/partners/[userId]/+page.svelte:1]: Only happens in Firefox for me. It might be because the store retains the data of the main timeline and that messes up something when switching to the shared timeline. Haven't looked into why, but resetting the store seems to fix the issue:

```ts
// asset-grid.svelte
onDestroy(() => {
	assetStore.setInitialState(0, 0, { totalCount: 0, buckets: [] }, undefined);
});
```

Comment (michelheusschen) [reply web/src/routes/(user)/partners/[userId]/+page.svelte:1]: When using `<AssetGrid user={data.partner} slot="content" />` the scrubbable scrollbar shows up and the issue I described is also gone.

jrasm91 (approved):

web/src/routes/(user)/partners/[userId]/+page.svelte
Comment (brighteyed) [reply web/src/routes/(user)/partners/[userId]/+page.svelte:1]: I like `assetStore` resetting in `onDestroy` because this solution also removes minor visual glitches when switching between main timeline and partners' timelines (old content is shown for about 100-200ms before it is replaced with the new one)

alextran1502:
I just did some tests with the PR, thank you for the fantastic work. I notice I few things I would like to address, please let me know your plan for each item.

1. Scrolling using the mouse wheel is very jerky, I believe this issue has been addressed by Michel. I think this is due to the AssetGrid being wrapper into `UserPageLayout`. Following up on point 3 below.
2. I would like to include the timeline scrollbar similar to the main timeline because when the user has a lot of assets, the scroll thumb will become very small and become impractical to click on it and drag.
3. Please feel free to not use the UserPageLayout for showing the assets of the partner user, you can create a new page with a proper header so that integrating the timeline scrollbar would be easier.

brighteyed:
> I just did some tests with the PR, thank you for the fantastic work. I notice a few things I would like to address, please let me know your plan for each item.
> 
> 1. Scrolling using the mouse wheel is very jerky, I believe this issue has been addressed by Michel. I think this is due to the AssetGrid being wrapped into `UserPageLayout`. Following up on point 3 below.
> 2. I would like to include the timeline scrollbar similar to the main timeline because when the user has a lot of assets, the scroll thumb will become very small and become impractical to click on it and drag.
> 3. Please feel free to not use the UserPageLayout for showing the assets of the partner user, you can create a new page with a proper header so that integrating the timeline scrollbar would be easier.


Yesterday I had been experimenting with scrollbar but had no success.
Transition to a completely new page without `UserPageLayout` seems a good idea. It should fix all the issues above. I'll try it later today

brighteyed:
> Transition to a completely new page without `UserPageLayout` seems a good idea. It should fix all the issues above. I'll try it later today

Done. All 3 issues seem to have fixed now

web/src/routes/(user)/partners/[userId]/+page.svelte
Comment (brighteyed) [reply web/src/routes/(user)/partners/[userId]/+page.svelte:24]: I've changed page layout completely and the scrollbar is present and works as expected now

Carbur8tr:
Just checking back in on this functionality. I'm on 1.91.4 and when trying to add a partner it appears as though there are no photos shared.  In addition there does not appear to be a "partners" tab under sharing. 

Apologies if this is not the appropriate avenue.

image
image

alextran1502:
@Carbur8tr The other user will also have to share with you so that you can look at their photos. 

We don't have an automatic 2-way share because we don't have a notification/accepting system yet, so, for privacy and security purposes, the other user will need to initialize the partners sharing with you.

Carbur8tr:
> @Carbur8tr The other user will also have to share with you so that you can look at their photos.
> 
> We don't have an automatic 2-way share because we don't have a notification/accepting system yet, so, for privacy and security purposes, the other user will need to initialize the partners sharing with you.

Appreciate it, I was not aware that by default the shared user needs to utilize the slider under their library settings to show photos. 

Took care of the main issue, but question still remains on the "partners" tab.  Is this still a feature?

DRIgnazGortngschirl:
@jrasm91 Sorry for ping. Awesome sharing function would be amazing to share also the "X years since" photos to get flashbacks for all the shared partners. I was missing this at OneDrive and switched to immich for a self-hosted and AI detection, which works amazing! Thank you so much for working on this project!

Maybe I should open a feature request ?

PR #2453, opened 2023-05-17
feat(web): improve page header + scrolling

jrasm91:
Extend line to edge of the screen between page title and content. Also, the content scrolls _after_ the title. The sidemenu also now scrolls similarly. 

image
image

image
image

image
image


jrasm91:
> This moves the whole user layout a bit to the right, not only the line. The map view now touches the right edge for example. Is that intended?

No, it turns out I wanted the line to extend, the header to be sticky, and the content to have it's own scroll context. I've updated the user layout component to work like this now.

alextran1502 (approved):
Great fix, thank you

PR #2464, opened 2023-05-17
feat(web,server): use time-buckets on more pages (albums, shared links, favorites, archived)

jrasm91:
TODO:
- [x] Timeline scroll on Archive page
- [x] Timeline scroll on Favorites page
- [ ] Access control for shared links and shared albums
- [ ] Album thumbnail selection
- [ ] Test album page actions
- [ ] Test shared link page actions
- [x] Test favorite page actions
- [x] Test archive page actions
- [ ] Large album test
- [ ] Test coverage
- [x] Timeline scroll on Person detail page


jrasm91:
@michelheusschen - If you are interested, feel free to jump in and fix anything that's not working with the actions, album page, etc.

web/src/routes/(user)/albums/[albumId]/+page.server.ts
```diff
@@ -11,8 +11,16 @@ export const load = (async ({ params, locals: { api, user } }) => {
 
 	try {
 		const { data: album } = await api.albumApi.getAlbumInfo(albumId);
+
+		const startDate = new Date(album.assets[0].fileCreatedAt);
+		const endDate = new Date(album.assets[album.assetCount - 1].fileCreatedAt);
```
Comment (brighteyed) [inline web/src/routes/(user)/albums/[albumId]/+page.server.ts:16]: For just created empty albums `album.assets` is an empty array

web/src/lib/components/album-page/asset-selection.svelte
```diff
@@ -75,6 +75,12 @@
 		</svelte:fragment>
 	</ControlAppBar>
 	<section class="pt-[100px] pl-[70px] grid h-screen bg-immich-bg dark:bg-immich-dark-bg">
-		<AssetGrid isAlbumSelectionMode={true} />
+		<AssetGrid
+			isAlbumSelectionMode={true}
+			options={{
+				isArchived: false,
```
brighteyed, line 81: I think we should remove `isArchived: false` here since user should be able to add archived photos to the album

michelheusschen:
Fixed some small issues, but I noticed a bigger problem. `AssetGrid` can only be mounted once, because the state is managed in global stores. On the album page, an additional asset grid is mounted when adding new assets to the album which causes the page to freeze. I've implemented a hacky workaround for now, but we should probably find a better solution.

samip5:
While this is getting work done, maybe also include an toggle for the web to have time buckets by month rather than day. :)

jrasm91:
Superseded by #3604, #3557, #3554, #3537.

Time buckets are now used on:
- Archive
- Favorites
- Timeline
- Partner Sharing Timeline
- Albums
- Photos of a person (person detail)

Access control has been updated in #2910, #2776, #2680.

Future work will need to be done for shared links and search. Search in particular will require more sweeping design changes to before timeline buckets can be used there.

PR #2466, opened 2023-05-17
feat(server): xmp sidecar metadata

alex-phillips:
This PR adds support for ingesting metadata from XMP sidecar files.

- If an XMP sidecar exists with a media file, it is assumed that the XMP file takes priority and all metadata is read from the file.
- XMP sidecars can now be uploaded with the initial media files via the API.
- "All" metadata scans check for the presence of new sidecar files if the media file does not have a sidecar file associated with it. This allows the creation of metadata files from third party software such as Digikam or Adobe Lightroom.
- "Missing" metadata scans include media assets that do not have a sidecar file associated with them. It attempts to find the sidecar in the filesystem and associates that sidecar file with the asset in the future.
- The XMP sidecar path is stored along with the media asset with a reference to its filepath.

image
image
image


server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -181,7 +185,11 @@ export class MetadataExtractionProcessor {
       }
 
       await this.exifRepository.upsert(newExif, { conflictPaths: ['assetId'] });
-      asset = await this.assetCore.save({ id: asset.id, fileCreatedAt: fileCreatedAt?.toISOString() });
+      asset = await this.assetCore.save({
+        id: asset.id,
+        fileCreatedAt: fileCreatedAt?.toISOString(),
+        ...(!asset.sidecarPath && sidecarPath && { sidecarPath }),
```
jrasm91, line 191: I think it is OK to just set it every time. Eg.

```typescript
.save({ 
  ...
  sidecarPath
  ...
})
```

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -91,7 +92,8 @@ export class MetadataExtractionProcessor {
     let asset = job.data.asset;
 
     try {
-      const exifData = await exiftool.read<ImmichTags>(asset.originalPath).catch((error: any) => {
+      const sidecarPath = asset.sidecarPath || (fs.existsSync(`${asset.originalPath}.xmp`) ? `${asset.originalPath}.xmp` : null)
```
jrasm91, line 95: This works, but I'd prefer using the async method here.

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -213,7 +221,8 @@ export class MetadataExtractionProcessor {
         }
       }
 
-      const exifData = await exiftool.read<ImmichTags>(asset.originalPath).catch((error: any) => {
+      const sidecarPath = asset.sidecarPath || (fs.existsSync(`${asset.originalPath}.xmp`) ? `${asset.originalPath}.xmp` : null)
```
jrasm91, line 224: Same logic again here again. Maybe we should make a private method `getSidecarPath(asset: AssetEntity)` that handles this logic`

server/libs/domain/src/storage-template/storage-template.service.ts
```diff
@@ -77,9 +77,22 @@ export class StorageTemplateService {
 
       try {
         await this.storageRepository.moveFile(asset.originalPath, destination);
+
+        let sidecarDestination;
+        if (asset.sidecarPath != null) {
```
jrasm91:
```suggestion
        if (!asset.sidecarPath) {
```

alex-phillips: This actually creates a typescript error since asset.sidecarPath is string or null, the `!` appears to also include empty string.

image

jrasm91: Sorry, I meant to suggest `if(asset.sidecarPath)` (negated)

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -338,3 +341,86 @@ export class MetadataExtractionProcessor {
     return Duration.fromObject({ seconds: videoDurationInSecond }).toFormat('hh:mm:ss.SSS');
   }
 }
+
+@Processor(QueueName.SIDECAR)
+export class SidecarProcessor {
+  private logger = new Logger(SidecarProcessor.name);
+  private assetCore: AssetCore;
+
```
jrasm91, line 349: The code in `domain/` is basically refactored code, this code here hasn't been updated and doesn't have any tests for it yet. Eventually the thought is to move all the logic out of this file into a `domain/media/media.service.ts` file. If you're up for that, it would be the place to put it. Otherwise, I can move it after it's merged (refactored processors are in `processors.ts` and a new service would need to be exported in `domain.module.ts`). If not, no worries.

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -338,3 +341,86 @@ export class MetadataExtractionProcessor {
...
+  @Process(JobName.QUEUE_SIDECAR)
+  async handleQueueSidecar(job: Job<IBaseJob>) {
+    try {
+      const { force } = job.data;
+      const assets = force ? await this.assetRepository.getWith(WithProperty.SIDECAR) : await this.assetRepository.getWithout(WithoutProperty.SIDECAR);
+
+      for (const asset of assets) {
+        const fileName = asset.originalFileName
+        const name = force ? JobName.SIDECAR_SYNC : JobName.SIDECAR_DISCOVERY;
+        await this.jobRepository.queue({ name, data: { asset, fileName } });
```
jrasm91, line 370: You just need to pass asset here.

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -338,3 +341,86 @@ export class MetadataExtractionProcessor {
...
+      }
+    } catch (error: any) {
+      this.logger.error(`Unable to queue sidecar scanning`, error?.stack);
+    }
+  }
+
+  @Process(JobName.SIDECAR_DISCOVERY)
+  async handleQueueSidecarDiscovery(job: Job<IAssetUploadedJob>) {
```
jrasm91, line 378: This can just be `handleSidecarDiscovery`

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -338,3 +341,86 @@ export class MetadataExtractionProcessor {
...
+    if (!asset.isVisible) {
+      return;
+    }
+
+    try {
+      const sidecarPath = await this.getSidecarPath(asset);
+      if (sidecarPath) {
+        asset = await this.assetCore.save({ id: asset.id, sidecarPath });
+        const name = asset.type === AssetType.VIDEO ? JobName.EXTRACT_VIDEO_METADATA : JobName.EXIF_EXTRACTION;
+        await this.jobRepository.queue({ name, data: { asset, fileName } });
```
jrasm91, line 389: the metadata job, the filename is coming from asset.originalFileName. That is semi new, but it makes this extra filename param obsolete. It can be removed everywhere, we just haven't gotten around to it yet, but you definitely don't need to use it for any new jobs.

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -338,3 +341,86 @@ export class MetadataExtractionProcessor {
...
+
+  @Process(JobName.QUEUE_SIDECAR)
+  async handleQueueSidecar(job: Job<IBaseJob>) {
+    try {
+      const { force } = job.data;
+      const assets = force ? await this.assetRepository.getWith(WithProperty.SIDECAR) : await this.assetRepository.getWithout(WithoutProperty.SIDECAR);
+
+      for (const asset of assets) {
+        const fileName = asset.originalFileName
+        const name = force ? JobName.SIDECAR_SYNC : JobName.SIDECAR_DISCOVERY;
```
jrasm91, line 369: force: true = sidecar sync
force: false = sidecar discover

In the UI force (sync) is on top and discover is on bottom. 

The order here is sync then discover.

Would you mind having the sync method implementation come before/above the discover one below? :stuck_out_tongue:

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -338,3 +341,86 @@ export class MetadataExtractionProcessor {
...
+
+  @Process(JobName.SIDECAR_DISCOVERY)
+  async handleQueueSidecarDiscovery(job: Job<IAssetUploadedJob>) {
+    let { asset, fileName } = job.data
+    if (!asset.isVisible) {
+      return;
+    }
+
+    try {
+      const sidecarPath = await this.getSidecarPath(asset);
```
jrasm91, line 385: I think this is the only place the method is used and we'd actually prefer to just skip assets that are already linked, even though they should theoretically never get processed in the first place. So, can you inline it here, but change it so that if an asset already has a sidecar path it is skipped? If there isn't one found we should return, otherwise queue exif.

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -338,3 +341,86 @@ export class MetadataExtractionProcessor {
...
+        const name = asset.type === AssetType.VIDEO ? JobName.EXTRACT_VIDEO_METADATA : JobName.EXIF_EXTRACTION;
+        await this.jobRepository.queue({ name, data: { asset, fileName } });
+      }
+    } catch (error: any) {
+      this.logger.error(`Unable to queue metadata extraction`, error?.stack);
+    }
+  }
+
+  @Process(JobName.SIDECAR_SYNC)
+  async handleQueueSidecarSync(job: Job<IAssetUploadedJob>) {
```
jrasm91, line 397: `handleSidecarSync`

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -338,3 +341,86 @@ export class MetadataExtractionProcessor {
...
+      this.logger.error(`Unable to queue metadata extraction`, error?.stack);
+    }
+  }
+
+  private async getSidecarPath(asset: AssetEntity): Promise<string|null> {
+    if (asset.sidecarPath) {
+      return asset.sidecarPath
+    }
+
+    return await new Promise((resolve) => {
```
jrasm91, line 416: The method doesn't need to be `async` if it returns a promise and you can return a promise without awaiting it.

Also, is there no async/promise based method in `fs` to check if a file exists?

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -338,3 +341,86 @@ export class MetadataExtractionProcessor {
...
+
+  @Process(JobName.SIDECAR_SYNC)
+  async handleQueueSidecarSync(job: Job<IAssetUploadedJob>) {
+    const { asset, fileName } = job.data
+    if (!asset.isVisible) {
+      return;
+    }
+
+    try {
+      const name = asset.type === AssetType.VIDEO ? JobName.EXTRACT_VIDEO_METADATA : JobName.EXIF_EXTRACTION;
```
jrasm91, line 404: This method is very trivial right now because it just queues all assets, but we should re-visit optimizing this to only detected changed files. We can add that in this PR, or I'm fine merging it as is and revisiting that in a subsequent PR. If that's the case, you should add a `// TODO: optimize to only queue assets with recent xmp changes` note.

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -338,3 +341,86 @@ export class MetadataExtractionProcessor {
...
+      await this.jobRepository.queue({ name, data: { asset, fileName } });
+    } catch (error: any) {
+      this.logger.error(`Unable to queue metadata extraction`, error?.stack);
```
jrasm91, line 407: Unable to sync sidecar

jrasm91 (commented):
Looking good! Have you run any of the linting checks? You should look at server/package.json or the contributing docs and run all the check commands to lint and format, and test the code (both web and UI).

jrasm91:
Can you attach a few screenshots? One of the files in `library/` (with xmp) and one of the new job? Thanks.

server/libs/domain/src/storage-template/storage-template.service.ts
```diff
@@ -77,9 +77,17 @@ export class StorageTemplateService {
...
+
         try {
-          await this.assetRepository.save({ id: asset.id, originalPath: destination });
+          let sidecarDestination;
+          if (asset.sidecarPath != null) {
+            sidecarDestination = `${destination}.xmp`
+            await this.storageRepository.moveFile(asset.sidecarPath, sidecarDestination);
+          }
+
+          await this.assetRepository.save({ id: asset.id, originalPath: destination, sidecarPath: sidecarDestination });
```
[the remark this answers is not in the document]
alex-phillips: Wouldn't this be an issue if the media file was also moved but failed to save? Shouldn't we check if either was moved and revert the move if that's the case?

server/apps/microservices/src/processors/metadata-extraction.processor.ts
alex-phillips: Right, forgot that fs calls where now async/await compatible.

 Per the node docs, fs.exists is deprecated in favor of `.stat` and `.access`

server/apps/microservices/src/processors/metadata-extraction.processor.ts
alex-phillips: Added a todo here.

server/apps/microservices/src/processors/metadata-extraction.processor.ts
alex-phillips: Added it inline, also performed error checking since we have to use fs.access anyway. The error check just returns if the file doesn't exist but also outputs an error log if the file does exist but is not writable.

server/apps/microservices/src/processors/metadata-extraction.processor.ts
alex-phillips: Unfortunately this seems required. The `JobName.EXIF_EXTRACTION` requires a parameter of `IAssetUploadedJob` and does not accept `IAssetJob`. Due to this, it's required I pass in the filename. Unless you want me to refactor this now, it is throwing transpiling errors.

alex-phillips:
Screenshot of sidecars in place after library upload:
image

Screenshot of the new jobs section:
image

image

server/libs/domain/src/storage-template/storage-template.service.ts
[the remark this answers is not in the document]
jrasm91: Yeah, we could set original moved = true and check that in the catch.

server/apps/immich/src/api-v1/asset/asset.core.ts
```diff
@@ -39,6 +40,7 @@ export class AssetCore {
       sharedLinks: [],
       originalFileName: parse(file.originalName).name,
       faces: [],
+      sidecarPath: sidecarFile != null ? sidecarFile.originalPath : null,
```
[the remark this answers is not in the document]
alex-phillips: This was resolved with  `sidecarPath: sidecarFile?.originalPath || null,`

server/apps/microservices/src/processors/metadata-extraction.processor.ts
alex-phillips: This was 'resolved' as the saving of the asset was moved into the SYNC_DISCOVER job, so it always saves the `sidecarPath` property.

server/apps/microservices/src/processors/metadata-extraction.processor.ts
alex-phillips: We use this elsewhere now, but this was resolved using the async `fs.promises.access` method. Using access instead of exists as exists is now deprecated in the node docs and recommends using access if checking file access without making modifications.

server/libs/domain/src/storage-template/storage-template.service.ts
alex-phillips: @jrasm91  I'm not sure what you mean here by negating it.

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -338,3 +341,86 @@ export class MetadataExtractionProcessor {
...
+
+  constructor(
+    @Inject(IAssetRepository) private assetRepository: IAssetRepository,
+    @Inject(IJobRepository) private jobRepository: IJobRepository,
+  ) {
+    this.assetCore = new AssetCore(assetRepository, jobRepository);
+    this.init();
+  }
+
+  private async init() {
```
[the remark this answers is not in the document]
alex-phillips: I've removed this code.

server/libs/domain/src/job/job.repository.ts
```diff
@@ -51,6 +51,11 @@ export type JobItem =
   | { name: JobName.EXIF_EXTRACTION; data: IAssetUploadedJob }
   | { name: JobName.EXTRACT_VIDEO_METADATA; data: IAssetUploadedJob }
 
+  // Sidecar Scanning
+  | { name: JobName.QUEUE_SIDECAR; data: IBaseJob }
+  | { name: JobName.SIDECAR_DISCOVERY; data: IAssetUploadedJob }
+  | { name: JobName.SIDECAR_SYNC; data: IAssetUploadedJob }
```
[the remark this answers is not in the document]
alex-phillips: This is resolved.

server/libs/domain/src/storage-template/storage-template.service.ts
```diff
@@ -77,9 +77,17 @@ export class StorageTemplateService {
 
       try {
         await this.storageRepository.moveFile(asset.originalPath, destination);
+
         try {
-          await this.assetRepository.save({ id: asset.id, originalPath: destination });
+          let sidecarDestination;
+          if (asset.sidecarPath != null) {
```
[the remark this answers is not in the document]
alex-phillips: Unresolving this one, thought I understood above (same issue) but not sure by what you mean to 'negate' this.

server/apps/microservices/src/processors/metadata-extraction.processor.ts
alex-phillips: Renamed this accordingly.

server/libs/domain/src/storage-template/storage-template.service.ts
alex-phillips: Nevermind, I was thinking about it backwards. This is resolved.

server/libs/domain/src/storage-template/storage-template.service.ts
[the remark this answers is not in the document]
alex-phillips: This is resolved.

server/libs/domain/src/storage-template/storage-template.service.ts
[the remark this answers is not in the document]
alex-phillips: Implemented this, should be allset.

jrasm91 (approved):

alextran1502:
Can you help update the documentation of how we can use this feature?

alex-phillips:
> Can you help update the documentation on how we can use this feature?

Added documentation and included some of the images in it as well. Let me know if I should elaborate on anything or include more details.

alextran1502:
@alex-phillips Thank you. There are some typo in the documentation, you can copy and paste the paragraph below to fix

```
There are 2 administrator jobs associated with sidecar files: `SYNC` and `DISCOVER`. The sync job will re-scan all media with existing sidecar files and queue them for a metadata refresh. This is a great use case when third-party applications are used to modify the metadata of media. The discover job will attempt to scan the filesystem for new sidecar files for all media that does not currently have a sidecar file associated with it.
```

Can you also help resolve the conflicts? 

Thank you

alextran1502:
Is there something wrong with the merge conflict? Looks like there are problems after the conflicts are resolved.

web/src/lib/components/admin-page/jobs/jobs-panel.svelte
```diff
@@ -125,7 +126,11 @@
 {/if}
 
 <div class="flex flex-col gap-7">
+<<<<<<< HEAD
 	{#each jobDetailsArray as [jobName, { title, subtitle, allText, missingText, allowForceCommand, component, handleCommand: handleCommandOverride }]}
+=======
+	{#each jobDetailsArray as [jobName, { title, subtitle, allowForceCommand, icon, component, handleCommand: handleCommandOverride }]}
+>>>>>>> upstream/main
```
jrasm91: Missed merge conflict here btw.

alex-phillips: Ah, I did resolve this but it didn't get pushed up. Thanks/.

server/apps/microservices/src/processors/metadata-extraction.processor.ts
```diff
@@ -345,3 +371,85 @@ export class MetadataExtractionProcessor {
...
+
+    try {
+      await fs.promises.access(`${asset.originalPath}.xmp`, fs.constants.W_OK);
+
+      try {
+        asset = await this.assetCore.save({ id: asset.id, sidecarPath: `${asset.originalPath}.xmp` });
+        const fileName = asset.originalFileName;
+        // TODO: optimize to only queue assets with recent xmp changes
+        const name = asset.type === AssetType.VIDEO ? JobName.EXTRACT_VIDEO_METADATA : JobName.EXIF_EXTRACTION;
+        await this.jobRepository.queue({ name, data: { asset, fileName } });
```
jrasm91, line 443: `fileName` isn't included in this job anymore.

web/src/lib/components/admin-page/jobs/jobs-panel.svelte
```diff
@@ -56,6 +58,12 @@
 			title: 'Extract Metadata',
 			subtitle: 'Extract metadata information i.e. GPS, resolution...etc'
 		},
+		[JobName.SidecarQueue]: {
+			title: 'Sidecar Metadata',
+			subtitle: 'Discover or synchronize sidecar metadata from the filesystem',
+			allText: 'SYNC',
+			missingText: 'DISCOVER'
```
jrasm91, line 65: This is missing a required icon.

alex-phillips: Got it!

server/apps/microservices/src/processors/metadata-extraction.processor.ts
alex-phillips: Cleaned up.

alextran1502 (approved):
Thank you for the high-quality PR! And sorry for the constant new changes to the main branch :P

alextran1502 (approved):
Thank you for the high-quality PR! And sorry for the constant new changes to the main branch :P

PR #2488, opened 2023-05-20
feat(server): transcode bitrate and thread settings

mertalev:
This PR allows users to set the max bitrate and thread count of transcoded videos through the admin transcoding page. This addresses two pain points with transcoding: high CPU usage and large file sizes.

**Max Bitrate**
This option is used for constrained quality (CQ) mode in FFmpeg. This mode respects the chosen CRF so long as it doesn't go beyond this bitrate, in which case it lowers quality dynamically. Setting an upper bound allows for predictable file sizes and prevents the issue of transcoded videos being much larger than the original.

**Threads**
Transcoding is naturally very CPU-intensive, and this can have an effect on other tasks and apps running on a device. Setting a thread count ensures that transcoding can only occupy a certain level of CPU resources. 

However, the exact effect of this depends on the exact library used. While libvpx-vp9 respects FFmpeg's `-threads` flag, x264 and x265 create their own thread pools internally with heuristics set to max utilization. Disabling thread pooling led to a much closer relationship between `-threads` and actual utilization, though there seems to be something else that isn't being affected by this option.

**Setting Descriptions**
With the growth of the transcoding page over time, I think it's helpful to give some pointers on what these settings mean and what they can be useful for.

**Extra**
I also added support for two-pass transcoding, which does a fast transcode to generate statistics and uses those statistics to produce a better encoded video. CRF is ignored in this setting, and the actual quality is derived from the max bitrate set (following [these](https://developers.google.com/media/vp9/settings/vod#recommended_settings) recommendations). This setting is particularly good for VP9, as libvpx is optimized to take advantage of it for both quality and efficiency. The downside, of course, is that it takes more time to transcode in two passes than one.

It can be enabled by setting the environmental variable `ENABLE_TWO_PASS` to `true`.


alextran1502:
Tagging @zackpollard since this is his baby

alextran1502:
> It can be enabled by setting the environmental variable ENABLE_TWO_PASS to true.

Can we move this option to the setting page as well? and have it defaulted to `false`?

server/libs/domain/src/system-config/system-config.core.ts
```diff
@@ -10,10 +10,12 @@ export type SystemConfigValidator = (config: SystemConfig) => void | Promise<voi
 const defaults: SystemConfig = Object.freeze({
   ffmpeg: {
     crf: '23',
+    threads: '2',
```
michelheusschen, line 13: Why is `2` the default? Wouldn't it be better to use ffmpeg's default of `0` which automatically determines the number of threads to use?

web/src/lib/components/admin-page/settings/setting-input-field.svelte
```diff
@@ -39,8 +40,14 @@
 			</div>
 		{/if}
 	</div>
+
+	<div class={`flex place-items-center gap-1`}>
+		<label class={`immich-form-label text-xs`} for="{label}-desc">{desc}</label>
+	</div>
```
michelheusschen, line 46: Also for `SettingSelect` component

```suggestion
	{#if desc}
		<p class="immich-form-label text-xs pb-2" id="{name}-desc">
			{desc}
		</p>
	{/if}
```

web/src/lib/components/admin-page/settings/setting-input-field.svelte
```diff
@@ -39,8 +40,14 @@
...
+
 	<input
-		class="immich-form-input w-full"
+		style="position:relative;top:10px"
+		class="immich-form-input gap-2 w-full"
```
michelheusschen, line 50: Also for `SettingSelect` component

```suggestion
		class="immich-form-input w-full"
		aria-describedby={desc ? `${label}-desc` : undefined}
```

server/libs/domain/src/media/media.service.ts
```diff
@@ -241,6 +240,47 @@ export class MediaService {
...
+    const validMaxRate = maxBitrateValue && maxBitrateValue > 0;
+    const bitrateUnit = ffmpeg.maxBitrate.substring(maxBitrateValue.toString().length) || 'k'; // use inputted unit if provided, else default to kbps
+
+    if (validMaxRate && twoPass) {
+      const targetBitrateValue = maxBitrateValue / 1.45; // recommended by https://developers.google.com/media/vp9/settings/vod
+      const minBitrateValue = targetBitrateValue / 2;
+
+      options.push(`-b:v ${targetBitrateValue}${bitrateUnit}`);
+      options.push(`-minrate ${minBitrateValue}${bitrateUnit}`);
+      options.push(`-maxrate ${maxBitrateValue}${bitrateUnit}`);
```
brighteyed, line 278: Why do these options relate only to two pass transcoding?

server/libs/domain/src/system-config/system-config.core.ts
mertalev: There are many tasks running when photos and videos are added, so while FFmpeg chooses the best setting to max out transcoding, it can lead to thrashing between ML inference and thumbnail generation. Two threads seemed like a reasonable balance, but I can set the default to 0 if preferred.

server/libs/domain/src/media/media.service.ts
mertalev: Two-pass encoding is generally done with bitrate ranges. If two-pass encoding isn't enabled, then the video bitrate is chosen dynamically based on CRF so the minimum and target bitrates aren't useful.

server/libs/domain/src/system-config/dto/system-config-ffmpeg.dto.ts
```diff
@@ -5,6 +5,9 @@ export class SystemConfigFFmpegDto {
   @IsString()
   crf!: string;
 
+  @IsString()
+  threads!: string;
```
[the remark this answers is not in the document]
mertalev: Sure! I can set the min to 0, but setting a max is tricky since everyone has different specs. Should `crf` also be changed to a number, or just `threads`?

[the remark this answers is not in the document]
brighteyed: Ah, we bind this option to `string` value in the `SettingInputField`. Let's leave it as is now

mertalev:
> > It can be enabled by setting the environmental variable ENABLE_TWO_PASS to true.
> 
> Can we move this option to the setting page as well? and have it defaulted to `false`?

Yup, I added it as a slider in the admin page.

server/libs/domain/src/media/media.service.ts
```diff
@@ -241,6 +240,47 @@ export class MediaService {
...
+    } else if (validMaxRate) {
+      options.push(`-crf ${ffmpeg.crf}`);
+      options.push(`${isVP9 ? '-b:v' : '-maxrate'} ${maxBitrateValue}${bitrateUnit}`);
+    }
```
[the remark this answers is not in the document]
mertalev: Good catch! Fixed.

server/libs/domain/src/system-config/dto/system-config-ffmpeg.dto.ts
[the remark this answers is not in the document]
mertalev: I actually went ahead and changed `threads` and `crf` to be numbers, though I had to change the `string` value to `string | number` for TS to not complain.

server/libs/domain/src/media/media.service.ts
```diff
@@ -221,26 +221,75 @@ export class MediaService {
...
+      const targetBitrateValue = Math.ceil(maxBitrateValue / 1.45); // recommended by https://developers.google.com/media/vp9/settings/vod
+      const minBitrateValue = targetBitrateValue / 2;
+
+      options.push(`-b:v ${targetBitrateValue}${bitrateUnit}`);
+      options.push(`-minrate ${minBitrateValue}${bitrateUnit}`);
+      options.push(`-maxrate ${maxBitrateValue}${bitrateUnit}`);
+    } else if (constrainMaximumBitrate) {
+      options.push(`-crf ${ffmpeg.crf}`);
+      options.push(`${isVP9 ? '-b:v' : '-maxrate'} ${maxBitrateValue}${bitrateUnit}`);
+    } else {
```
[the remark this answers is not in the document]
brighteyed: I mean that we must use a combination of `-crf` and `-b:v 0` for VP9 for constant quality mode (the same is for single pass encoding)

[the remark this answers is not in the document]
mertalev: Good point! I added constant quality mode for VP9. Since the default max bitrate is 0, this is now the default behavior for vp9 (as with h264 and hevc).

brighteyed:
2-pass-switch

Could you try to fix switch position?

Sherlock1979:
Hi,

I am a complete noob when it comes to transcoding. But as a user, I need to tweak ffmpeg settings as the defaults just did not work for me. I need smaller files so that there is no buffering in low bandwidth scenarios while finding a balance with acceptable quality.

Hence, I started reading about vp9 recommendations and I realised that in Immich I cannot apply some settings that are recommended by many articles, including ffmpeg website and Google. I found this article particularly useful: https://streaminglearningcenter.com/codecs/encoding-vp9-in-ffmpeg-an-update.html#The_Recommended_Command_String

I am very pleased to see this PR and thank you for solving some of these challenges.

However, I feel that a couple of other options (or hardcoded defaults? for vp9)  would make sense. E.g. "-row-mt 1" is highly recommended.

I can also imagine that hundreds of other users may come up with hundreds of other needs. I have seen in some applications that they solve such problems by letting the users edit any ffmpeg options as they will. Maybe as an 'advanced option' on the UI?

Thanks for considering my use case and ideas.

Note, as per the linked article and Google's m vp9 guide (https://developers.google.com/media/vp9/settings/vod) I think that your below statement may not be fully correct because they use CRF with 2-pass encoding. But again, I am a noob, so it is very likely that I am wrong:

> I also added support for two-pass transcoding, which does a fast transcode to generate statistics and uses those statistics to produce a better encoded video. CRF is ignored in this setting, and the actual quality is derived from the max bitrate set (following these recommendations).

Thanks again! 🙏

server/libs/domain/src/system-config/dto/system-config-ffmpeg.dto.ts
```diff
@@ -1,9 +1,18 @@
...
+  @IsInt()
+  @Min(0)
+  @Max(51)
+  @Type(() => Number)
+  crf!: number;
+
+  @IsInt()
+  @Min(0)
+  @Type(() => Number)
+  threads!: number;
```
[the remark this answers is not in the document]
mertalev: Added

server/libs/domain/src/system-config/dto/system-config-ffmpeg.dto.ts
```diff
@@ -1,9 +1,18 @@
...
+import { Type } from 'class-transformer';
 
 export class SystemConfigFFmpegDto {
-  @IsString()
-  crf!: string;
+  @IsInt()
+  @Min(0)
+  @Max(51)
+  @Type(() => Number)
+  crf!: number;
```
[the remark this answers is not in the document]
mertalev: Added

mertalev:
> 2-pass-switch
> 
> Could you try to fix switch position?

CSS is not my strong suit but it should be fixed now

mertalev:
> However, I feel that a couple of other options (or hardcoded defaults? for vp9) would make sense. E.g. "-row-mt 1" is highly recommended.

Yup, this PR adds that flag for all VP9 transcodes.

> Note, as per the linked article and Google's m vp9 guide (https://developers.google.com/media/vp9/settings/vod) I think that your below statement may not be fully correct because they use CRF with 2-pass encoding.

You're right, VP9 allows for two-pass transcoding with constant quality based on CRF. The original PR didn't support this since the other two codecs (h264 and hevc) want bitrate ranges for two-pass encoding. 

That said, I later added support for VP9's CRF-based two-pass mode. The way two-pass mode works now is that if the max bitrate is 0, VP9 will use the provided CRF, else it will use the max bitrate specified. (For h264 and h265, setting the max bitrate to 0 will still disable two-pass mode even if it's enabled.)

Sherlock1979:
Thanks for the clarifications. Music to my ears. :) I am very excited to see this PR getting accepted and released.

> That said, I later added support for VP9's CRF-based two-pass mode. The way two-pass mode works now is that if the max bitrate is 0, VP9 will use the provided CRF, else it will use the max bitrate specified. (For h264 and h265, setting the max bitrate to 0 will still disable two-pass mode even if it's enabled.)

Hmmm, I might have read that VP9 can do max bitrate AND CRF combined AND two-pass. Or have I misinterpreted something?

mertalev:
> Hmmm, I might have read that VP9 can do max bitrate AND CRF combined AND two-pass. Or have I misinterpreted something?

According to the [FFmpeg wiki](https://trac.ffmpeg.org/wiki/Encode/VP9#twopass), two-pass mode can be done with either bitrates or CRF, but it doesn't mention using both. Some of the commands from the Google page actually do have both bitrate ranges and CRF, but the article you shared suggests that the CRF in them was ignored (in [this ](https://streaminglearningcenter.com/codecs/encoding-vp9-in-ffmpeg-an-update.html#isolating-the-key-differences)section). It's possible that just setting a max rate would use both, but the resources I've seen only mention that combination in the context of one-pass encoding.

Sherlock1979:
Thanks for the clarification. You are right, I think the Google page confused me. Sorry for the trouble.

brighteyed:
I have tested all possible combinations of `constrainMaximumBitrate` and `ffmpeg.twoPass`:

| `constrainMaximumBitrate` | `ffmpeg.twoPass` | h264/h265 | vp9 |
|--------|--------|--------|--------|
| `true` | `true` | ✔️  | ✔️  |
| `false` | `true` |  ⛔ for h264 second pass fails, ✔️ h265 is fine | ✔️  |
| `true` | `false` | ✔️  | ✔️  |
| `false` | `false` | ✔️  | ✔️  | 

After some googling it looks like h264 doesn't support 2-pass encoding with `crf` option specified. I believe we should ignore `ffmpeg.twoPass` in this case. What do you think?

server/libs/domain/src/media/media.repository.ts
```diff
@@ -47,5 +47,5 @@ export interface IMediaRepository {
   // video
   extractVideoThumbnail(input: string, output: string, size: number): Promise<void>;
   probe(input: string): Promise<VideoInfo>;
-  transcode(input: string, output: string, options: any): Promise<void>;
+  transcode(input: string, output: string, options: any, twoPass: boolean): Promise<void>;
```
jrasm91, line 50: Not critical, but I don't like having too many arguments in an interface. Everything else uses input/output + options I believe. Could we merge the last to into something like this? `options: TranscodeOptions`

Eg:
```typescript
export interface TranscodeOptions {
  outputOptions: string[];
  twoPass: boolean;
}


```

server/libs/infra/src/repositories/media.repository.ts
```diff
@@ -3,6 +3,7 @@ import { exiftool } from 'exiftool-vendored';
 import ffmpeg, { FfprobeData } from 'fluent-ffmpeg';
 import sharp from 'sharp';
 import { promisify } from 'util';
+import * as fs from 'fs/promises';
```
jrasm91, line 6: I don't believe we need `* as fs`. `import fs from 'fs/promises` should work just fine.

web/src/lib/components/admin-page/settings/setting-input-field.svelte
```diff
@@ -39,8 +40,16 @@
...
+	{#if desc}
+		<p class="immich-form-label text-xs pb-2" id="{label}-desc">
+			{desc}
+		</p>
+	{/if}
+
 	<input
-		class="immich-form-input w-full"
+		class="immich-form-input pb-2 w-full"
+		aria-describedby={desc ? `${label}-desc` : undefined}
```
jrasm91, line 52: Might as well add `aria-labelledby="{label}-label` too (need to to add `id="{label}-label"` to the label)

brighteyed:
> Ah, my mistake. I handled that case in `getFfmpegOptions` by ignoring two-pass mode and falling back to CRF, but I didn't actually disable the `ffmpeg.twoPass` flag. It should be fixed now.

All encoding options are fine now. It is definitely a notable step for video transcoding. Thank you for the PR!
Btw, great descriptions for options!

server/libs/domain/src/media/media.repository.ts
mertalev: Good call, added

mertalev:
Thanks for the detailed reviews! @brighteyed @jrasm91 @michelheusschen

alextran1502 (approved):

Sherlock1979:
Hi @mertalev,

Once again, thank you for this change. The improvements are fantastic.

I am doing vp9/opus transcoding and I have also noticed that only about 30-40% of my CPU capacity is utilised and meanwhile transcoding feels to take significantly slower than before. So I am trying to investigate the situation and better understand what is going on behind the scenes.

My ffmpeg Immich UI settings:
```
- CFR: 34
- PRESET: faster
- AUDIO CODEC: opus
- VIDEO CODEC: vp9
- TARGET RESOLUTION: 720p
- MAX BITRATE: 0
- THREADS: 0
- TRANSCODE: Videos higher than target resolution or not in the desired format
- TWO-PASS ENCODING: yes
```

Actual ffmeg paramteres while running:
`ffmpeg -i upload/library/admin/[original file name].mp4 -y -vcodec vp9 -acodec opus -strict experimental -movflags faststart -vf scale=-2:720 -cpu-used 5 -row-mt 1 -crf 34 -b:v 0 -pass 2 -passlogfile upload/encoded-video/[some-uuid] upload/encoded-video/[some-uuid2]/[some-uuid3]`

So, I have noticed that there is a "-cpu-used 5" parameter. Is this hardcoded? Or is this a representation of the preset that I set in the Immich UI?
- If yes then what is the relationship / mapping between the two?
- If no then is it possible that the preset UI setting is actually not having any effect? (as I don't see any other related parameters while ffpmeg is running)

Also, can you please point me in the right direction so that I can better understand the logic behind presets vs quality/deadline vs cpu-used/speed? I did some googling but could not find a source that explains it well and is trustworthy.

Thanks.

mertalev:
@Sherlock1979 Great questions!

>  I have also noticed that only about 30-40% of my CPU capacity is utilised

If you set `threads` to 0, the threading is actually the same as the old behavior. The difference comes from a recent change in the queue to transcode videos one at a time instead of two at a time. If you want higher utilization, you can try setting it to a number yourself since FFmpeg doesn't always choose the most optimal number.

> So, I have noticed that there is a "-cpu-used 5" parameter. Is this hardcoded? Or is this a representation of the preset that I set in the Immich UI?

The latter. The preset options map from 0-5 with `veryslow` being 0 and `faster` being 5. This was added with the PR so the presets were meaningful for all three supported codecs. Before this, the ffmpeg default would have been 1.

Another change was to mirror H.264 and HEVC in making constant quality the default when only using CRF. Omitting `b:v 0` still constrains bitrate if I understand the wiki correctly. But I don't know whether this affects performance, or if it may only affect it in two-pass mode. 

Speaking of two-pass mode, I think the slower speed you noticed comes primarily from that. It just fundamentally adds a good chunk of work for each video.

Sherlock1979:
@mertalev 

> If you set `threads` to 0, the threading is actually the same as the old behavior. The difference comes from a recent change in the queue to transcode videos one at a time instead of two at a time. If you want higher utilization, you can try setting it to a number yourself since FFmpeg doesn't always choose the most optimal number.

I tried with 0, 5, and 10 (I have 12 cores). Overall utilisation was below 50% in all cases.

> Speaking of two-pass mode, I think the slower speed you noticed comes primarily from that. It just fundamentally adds a good chunk of work for each video.

Actually, first pass seems to be very fast so it should not have a significant impact on my overall experience. I will try to time it and come back with some results soon.

mertalev:
Hmm, given that, it might be good to make the job queue concurrency user-configurable. Encoding more than one video in parallel seems to be the only way to scale vp9 further.

Sherlock1979:
> Hmm, given that, it might be good to make the job queue concurrency user-configurable. Encoding more than one video in parallel seems to be the only way to scale vp9 further at lower resolutions.

Thanks for looking into it.

Yes, I would love to be able to change concurrency.

PR #2498, opened 2023-05-20
feat(web+server): map improvements

michelheusschen:
- Hide archived assets, closes #2403
- Add settings menu for:
   - Disabling map dark mode, closes #2426
   - Filtering for favorite assets only (needs better refresh method first before expanding)
- Constrain map view when zooming out
- More efficient query method for map markers (~3x response time improvement)
- Moved [global stylesheets](https://github.com/sveltejs/vite-plugin-svelte/blob/main/docs/faq.md#where-should-i-put-my-global-styles) to a separate file

Previously @matthinc attempted to implement array responses like `[id, lat, lon]` instead of objects but I've decided against that, because it has worse type safety and gzipped API response sizes are pretty much equal (~2% difference)

image


jrasm91 (approved):
Nice! These settings are just persisted client side right?

michelheusschen:
> Nice! These settings are just persisted client side right?

Yes in localStorage

alextran1502 (approved):

PR #2508, opened 2023-05-21
chore(server): queue handlers shouldn't increase concurrency

mertalev:
Currently, queue handlers add an additional level of concurrency to each job despite taking up an insignificant amount of time. Once queueing is complete, this results in the remaining slower handlers of each processor to have levels of concurrency higher than intended. For instance, object tagging runs 3 jobs at a time despite object detection and image classification each being assigned a concurrency of 1, and 2 videos are transcoded at a time despite a transcode concurrency of 1.

With this PR, queue handlers will no longer offset concurrency.

### DIFF
--- a/server/apps/microservices/src/processors.ts
+++ b/server/apps/microservices/src/processors.ts
@@ -68,7 +68,7 @@ export class BackgroundTaskProcessor {
 export class ObjectTaggingProcessor {
   constructor(private smartInfoService: SmartInfoService) {}
 
-  @Process({ name: JobName.QUEUE_OBJECT_TAGGING, concurrency: 1 })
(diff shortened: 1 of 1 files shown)


jrasm91 (commented):
Can you instead omit the concurrency argument? I think you can actually remove the whole object and just use a single argument for job name too

mertalev:
> Can you instead omit the concurrency argument? I think you can actually remove the whole object and just use a single argument for job name too

Not including it seems to have the default behavior of one level of concurrency per consumer. Setting it to 0 is a workaround suggested from [this](https://github.com/nestjs/bull/issues/258) issue page.

jrasm91 (approved):

PR #2529, opened 2023-05-23
docker: use default entrypoint and command where applicable

uhthomas:
A default entrypoint and command make it just a bit easier to use the images as there is no longer a need for an explicit entrypoint. The exception is the server image, which still requires the shell script to be specified.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   immich-server:
     container_name: immich_server
     image: ghcr.io/immich-app/immich-server:release
-    entrypoint: ["/bin/sh", "./start-server.sh"]
+    command: ["start-server.sh"]
     volumes:
       - ${UPLOAD_LOCATION}:/usr/src/app/upload
     env_file:
@@ -18,7 +18,7 @@ services:
   immich-microservices:
     container_name: immich_microservices
     image: ghcr.io/immich-app/immich-server:release
-    entrypoint: ["/bin/sh", "./start-microservices.sh"]
+    command: ["start-microservices.sh"]
     volumes:
       - ${UPLOAD_LOCATION}:/usr/src/app/upload
     env_file:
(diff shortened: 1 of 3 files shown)


bo0tzz (approved):

jrasm91 (approved):

zatricky:
Does this depend on something else to function?

If I update my compose to match the new version the containers fail to start:
immich_server with image `ghcr.io/immich-app/immich-server:release` with `command: ["start-server.sh"]`:

    Node.js v18.16.0
    /usr/src/app/start-server.sh:3
    if [ "$DB_HOSTNAME_FILE" ]; then
       ^
    
    SyntaxError: Unexpected token '['
        at internalCompileFunction (node:internal/vm:73:18)
        at wrapSafe (node:internal/modules/cjs/loader:1176:20)
        at Module._compile (node:internal/modules/cjs/loader:1218:27)
        at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
        at Module.load (node:internal/modules/cjs/loader:1117:32)
        at Module._load (node:internal/modules/cjs/loader:958:12)
        at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
        at node:internal/main/run_main_module:23:47

It works again when reverting back to specifying the full `entrypoint`.

I had an identical issue with `immich_microservices`. Further, `immich-web` also failed to run but had no corresponding error message ; it simply quits without any output.

jrasm91:
Well, it would require a release first.

uhthomas:
@zatricky You should use the compose file attached to the release, or the file associated with the release tag.

https://github.com/immich-app/immich/releases/tag/v1.57.1

https://github.com/immich-app/immich/blob/v1.57.1/docker/docker-compose.yml

There's nothing we can do if people are mixing released artefacts with non-release arefacts.

PR #2553, opened 2023-05-24
feat(web): improved action bar actions

jrasm91:
### Photos page 
Favorite is moved to vertical dots menu, along with download and archive

Before
image
After
image

### Favorite page
Added add to albums, download, remove from favorites, and archive/unarchive

Before
image
After
image

### Archive page
Archive, download, and favorite moved to vertical dots menu

Before
image
After
image

### Person page
Added add to albums, download, remove from favorites, and archive/unarchive

Before
image
After
image

### Partner timeline page

Before
image
After
image

### Search page
Favorite is moved to vertical dots menu, along with download and archive

Before
image
After
image

### Confirm delete
Before
image
After
image


web/src/lib/components/photos-page/actions/archive-action.svelte
michelheusschen, line 1: - The `CircleIconButton` has the action reversed compared to `MenuItem` and always calls `handleArchive` with true
- The `isAll` prop is pretty vague and doesn't make much sense in the context of this component

What do you think of something like this:

```svelte
<script lang="ts">
	export let unarchive = false;

	$: text = unarchive ? 'Unarchive' : 'Archive';
	$: logo = unarchive ? ArchiveArrowUpOutline : ArchiveArrowDownOutline;
	
	// handleArchive can directly access unarchive
</script>

{#if menuItem}
	<MenuOption {text} on:click={handleArchive} />
{:else}
	<CircleIconButton title={text} {logo} on:click={handleArchive} />
{/if}
```

web/src/lib/components/photos-page/actions/delete-assets.svelte
```diff
@@ -7,39 +7,45 @@
...
 </script>
 
-<CircleIconButton title="Delete" logo={DeleteOutline} on:click={deleteSelectedAssetHandler} />
+<CircleIconButton title="Delete" logo={DeleteOutline} on:click={() => (confirm = true)} />
+
+{#if confirm}
+	<ConfirmDialogue
+		prompt="Are you sure you want to delete {getAssets()
+			.size} assets? This step also deletes assets in the album(s) to which they belong. You can not undo this action!"
+		on:confirm={handleDelete}
```
michelheusschen, line 48: For the confirm button it's generally recommended to choose a more specific label like delete or update over generic ones like save, yes and confirm

```suggestion
		title="Delete assets?"
		confirmText="Delete"
		on:confirm={handleDelete}
```

web/src/lib/components/photos-page/actions/favorite-action.svelte
```diff
@@ -0,0 +1,54 @@
...
+		<MenuOption text="Remove from Favorites" on:click={() => handleFavorite(false)} />
+	{:else}
+		<MenuOption text="Favorite" on:click={() => handleFavorite(true)} />
+	{/if}
+{:else}
+	<!-- <CircleIconButton
+		title="Remove Favorite"
+		logo={HeartMinusOutline}
+		on:click={() => handleFavorite(false)}
+	/> -->
```
michelheusschen, line 53: Could use the same approach as suggested for the archive action, or just remove it

web/src/lib/components/photos-page/actions/archive-action.svelte
jrasm91: That works for me. The idea was just to be able to use the same button for both actions and needing some easy to know which action it should use. I like this naming better though. Do you mind pushing the changes?

web/src/lib/components/photos-page/actions/delete-assets.svelte
jrasm91: This is used several places. It may be worth revisiting then as well.

alextran1502:
Maybe in the Archive page, the icon of unarchived should be presented in the app bar instead of in the menu option?

web/src/routes/(user)/photos/+page.svelte
```diff
@@ -47,7 +48,7 @@
 					<DownloadAction menuItem />
 					<ArchiveAction
 						menuItem
-						unarchive={true}
+						unarchive={isAllArchived}
```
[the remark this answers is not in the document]
michelheusschen: Oh yeah true, however it should be `unarchive={false}` because you would always want to perform the regular archive action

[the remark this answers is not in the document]
jrasm91: oh duh. I don't think it matters too much either way.

alextran1502 (approved):

PR #2554, opened 2023-05-24
fix: migrate admin layout to a component

jrasm91:
I got the layout bug (from awhile ago) after #2550. This PR migrates the admin pages to use a component to achieve a shared layout, instead of a `layout.svelte` file.


jrasm91:
Actually, this doesn't seem to address the issue, if anything it makes it happen more often.

PR #2560, opened 2023-05-24
feat: disk stats for library folder

jrasm91:
Change the disk statistics folder from `UPLOAD_LOCATION` to `UPLOAD_LOCATION/library`

This should only impact users mounting separate volumes for library, thumbs, etc. In the case that each folder is a separate mount, I think it makes sense to show the disk statistics for library.


jrasm91:
Yeah, I believe so. AFAICT it returns the stats for the mounted _filesystem_ where that folder lives, not the stats of the folder itself per se. So, if the new path is mounted to the same filesystem as the old path you would see the same values.

alextran1502 (approved):

PR #2562, opened 2023-05-25
refactor(server): update album

jrasm91:
Move update album to the domain.

### DIFF
--- a/server/apps/immich/src/controllers/album.controller.ts
+++ b/server/apps/immich/src/controllers/album.controller.ts
@@ -1,10 +1,11 @@
-import { AlbumService, AuthUserDto, CreateAlbumDto } from '@app/domain';
+import { AlbumService, AuthUserDto, CreateAlbumDto, UpdateAlbumDto } from '@app/domain';
 import { GetAlbumsDto } from '@app/domain/album/dto/get-albums.dto';
-import { Body, Controller, Get, Post, Query } from '@nestjs/common';
+import { Body, Controller, Get, Param, Patch, Post, Query } from '@nestjs/common';
 import { ApiTags } from '@nestjs/swagger';
 import { GetAuthUser } from '../decorators/auth-user.decorator';
 import { Authenticated } from '../decorators/authenticated.decorator';
 import { UseValidation } from '../decorators/use-validation.decorator';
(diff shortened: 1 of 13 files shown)


server/apps/immich/src/controllers/album.controller.ts
```diff
@@ -22,4 +23,10 @@ export class AlbumController {
   createAlbum(@GetAuthUser() authUser: AuthUserDto, @Body() dto: CreateAlbumDto) {
     return this.service.create(authUser, dto);
   }
+
+  @Authenticated()
```
michelheusschen, line 27: Can be removed, since `@Authenticated()` is already added on controller level

server/libs/infra/src/repositories/album.repository.ts
```diff
@@ -123,11 +123,31 @@ export class AlbumRepository implements IAlbumRepository {
     });
   }
 
-  create(album: Partial<AlbumEntity>): Promise<AlbumEntity> {
+  async hasAsset(id: string, assetId: string): Promise<boolean> {
+    const count = await this.repository.count({
```
michelheusschen, line 127: Why count instead of exists?

michelheusschen (approved):

server/apps/immich/src/controllers/album.controller.ts
jrasm91: Thanks, missed that copying it over.

server/libs/infra/src/repositories/album.repository.ts
jrasm91: To me, `exist` implies a boolean, where this is a number, so I went with count. I don't have strong opinions about it though.

michelheusschen: I meant for the `repository.count` instead of `repostitory.exist`, should've said that clearer. But either way, it doesn't really matter

alextran1502 (approved):

server/libs/infra/src/repositories/album.repository.ts
jrasm91: Oh. For that, I did not know it existed :smile:. I'll update it in my next PR.

PR #2565, opened 2023-05-25
feat(web+server): map date filters + small changes

michelheusschen:
Adds support for date filtering to the map. Can be used with a relative time (past 24 hours, 30 days, etc.) and with manual before/after dates. Also quite some other changes to the map, mainly related to clustering of asset markers:
- Clicking a cluster now activates spiderfy instead of hovering, because I've found hovering to work pretty inconsistent and because of
- Selecting a cluster allows you to click individual assets and opens the gallery viewer for that asset
- Made clusters that cannot spiderfy (>10 assets) bigger to indicate that they go directly to the gallery viewer
- Clustering is now based on zoom level which means less clusters when zoomed out and more when zooming in
- Map is now completely zoomed out by default and no longer zooms to the first asset
- Better performance and memory utilization of `AssetMarker` by deferring icon creation

map-date-filter-2map-date-filter-1


web/src/lib/components/map-page/map-settings-modal.svelte
```diff
@@ -27,9 +35,90 @@
...
+			{:else}
+				<div in:fly={{ y: -10, duration: 200 }} class="flex flex-col gap-1">
+					<SettingSelect
+						label="Date range"
+						name="date-range"
+						bind:value={settings.relativeDate}
+						options={[
+							{
+								value: '',
+								text: '---'
```
jrasm91, line 86: Would it make sense to put the text `All` here or similar?

web/src/routes/(user)/map/+page.svelte
```diff
@@ -29,22 +31,55 @@
...
+		if (abortController) abortController.abort();
 		assetInteractionStore.clearMultiselect();
 		assetInteractionStore.setIsViewingAsset(false);
 	});
 
 	async function loadMapMarkers() {
-		const { data } = await api.assetApi.getMapMarkers($mapSettings.onlyFavorites || undefined, {
-			signal: abortController.signal
-		});
+		if (abortController) abortController.abort();
```
jrasm91, line 40: Not a huge fan of the brace-less if statements myself.

jrasm91 (approved):

server/libs/domain/src/asset/dto/map-marker.dto.ts
```diff
@@ -1,10 +1,22 @@
+import { ApiProperty } from '@nestjs/swagger';
 import { toBoolean } from 'apps/immich/src/utils/transform.util';
 import { Transform } from 'class-transformer';
-import { IsBoolean, IsOptional } from 'class-validator';
+import { IsBoolean, IsISO8601, IsOptional } from 'class-validator';
 
 export class MapMarkerDto {
+  @ApiProperty()
```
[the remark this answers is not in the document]
michelheusschen: Parameters without `@ApiProperty()` are sorted first so this just keeps the order the same

web/src/lib/components/map-page/map-settings-modal.svelte
michelheusschen: Yeah can do, the other values are also very subjective so feel free to suggest other date ranges

server/libs/domain/src/asset/dto/map-marker.dto.ts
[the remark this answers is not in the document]
jrasm91: OK cool, thanks.

web/src/lib/components/map-page/map-settings-modal.svelte
jrasm91: What do you think about these:
- 24 hours
- 7 days
- 14 days
- 1 month
- 3 months
- 6 months
- 1 year
- All

PR #2566, opened 2023-05-25
fix(web): back button on person page

michelheusschen:
Set default route of the back button on the `people/[personId]` page to the previous route, with a fallback to `/explore` instead of always going to that page


jrasm91 (approved):
Thanks for fixing this!

PR #2570, opened 2023-05-25
refactor(server): delete album

jrasm91:
- Move delete album to the domain
- Cascade delete shared links when an album is deleted


michelheusschen (approved):

alextran1502 (approved):
- [x] Delete owned album
- [x] Delete local shared album
- [x] Delete album shared with public users

PR #2582, opened 2023-05-26
feat(server): return asset checksum

jrasm91:
- Add checksum (base64)
- Add `resize: boolean`
- Remove webp, encoded video path, resize path

### DIFF
--- a/mobile/openapi/doc/AssetBulkUploadCheckItem.md
+++ b/mobile/openapi/doc/AssetBulkUploadCheckItem.md
@@ -9,7 +9,7 @@ import 'package:openapi/api.dart';
 Name | Type | Description | Notes
 ------------ | ------------- | ------------- | -------------
 **id** | **String** |  | 
-**checksum** | **String** |  | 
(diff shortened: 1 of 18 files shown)


michelheusschen:
Encoding the checksum as base64 is inconsistent with `bulkUploadCheck` where they're handled as hexadecimal, it would be good to keep that consistent

fyfrey:
> Encoding the checksum as base64 is inconsistent with `bulkUploadCheck` where they're handled as hexadecimal, it would be good to keep that consistent

base64 is shorter and we already have HUGE requests/responses. Maybe we can change `bulkUploadCheck` in follow-up PR to also use base64 (or anything even shorter if we have that)

fyfrey (approved):
thank you! checksum works flawless in the app

jrasm91:
I updated the bulk endpoint to accept both `hex` and `base64` values in the checksum field. I think that provides both flexibility and consistency. We can return checksum as `base64` in the asset response dto though, since it is more compact.

PR #2585, opened 2023-05-27
refactor(web): user avatar

michelheusschen:
Use `UserAvatar` everywhere to show the user's profile image with fallback to initials. Based on the old `CircleAvatar` with the following changes:
- Only try to load image when `profileImagePath` is set
- Use a color based on the user UUID instead of being random
- Remove wrapping button, because there are situations where the avatar shouldn't be clickable


web/src/lib/components/shared-components/navigation-bar/account-info-panel.svelte
```diff
@@ -1,78 +1,52 @@
...
-				</div>
-			</Button>
-		</div>
+		<Button
+			color="dark-gray"
+			size="sm"
+			shadow={false}
+			border
+			on:click={() => {
+				goto('/user-settings');
```
jrasm91, line 41: Would you mind making this route path a constant/enum?

jrasm91 (approved):

PR #2589, opened 2023-05-28
refactor(server): tags

jrasm91:
Move tags to the domain. Haven't tested it yet, since I don't think it's used in the UI anywhere. Would be good to validate via rest calls though.


michelheusschen (approved):
Tag controller works, with the following findings:

- `renameTagId` of the tag entity seems unused
- Creating duplicate tags results in error 500 because of unique constraint (marked as TODO)
- `UpdateAssetDto` accepts `tagIds` but `AssetRepository.update` doesn't, so linking tags to assets isn't possible

jrasm91:
I think rename was added to be used for merging in the future though I'm not sure how.

I think you used to be able to add tags to assets, but I'm not sure why it was removed.

Tags seem like it was not 100% finished and not currently used anywhere either.

Perhaps we can add a new endpoint in the tag controller to add/remove assets using asset IDs, similar to how albums work.

jrasm91:
I think rename was added to be used for merging in the future though I'm not sure how.

I think you used to be able to add tags to assets, but I'm not sure why it was removed.

Tags seem like it was not 100% finished and not currently used anywhere either.

Perhaps we can add a new endpoint in the tag controller to add/remove assets using asset IDs, similar to how albums work.

PR #2605, opened 2023-05-29
feat(web): sticky asset date group title

sliterok:
Label for date is fixed horizontally to reduce jitter and made sticky vertically, so you can navigate easier when you have a lot of photos per group.
Maybe there's something that can be done to increase contrast? I wasn't sure if it's a good idea to add a background.
Let me know what you think.

### DIFF
--- a/web/src/lib/components/photos-page/asset-date-group.svelte
+++ b/web/src/lib/components/photos-page/asset-date-group.svelte
@@ -133,12 +133,12 @@
 		>
 			<!-- Date group title -->
 			<p
-				class="font-medium text-xs md:text-sm text-immich-fg dark:text-immich-dark-fg mb-2 flex place-items-center h-6"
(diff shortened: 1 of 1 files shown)


alextran1502:
Hello, thank you for the PR. Can you help adding a screenshots?

michelheusschen:
Don't really like the look of this to be honest, there also seem to be some issues:
- Margin to the title is always applied
- The checkbox on title hover still shows on Firefox, but can't be clicked
- When hovering over the thumbnail, the checkbox and shadow of it are shown on top of the date title

We could perhaps show a centered date like WhatsApp does or maybe show the date while scrolling in the scrollbar. What do you think of these ideas?

alextran1502:
Hello, I will have to reject this proposal because I think it adds noise when viewing the assets. 

image


Instead, we can toggle the thumb scroll info to display when scrolling (shown above), even with the mouse scroll's wheel, which is similar to how GPhotos is currently working. It would serve the purpose of letting you know where you are at on the timeline while hiding away from the main center view.

sliterok:
I see, closing.

PR #2640, opened 2023-06-02
refactor(server): move asset checks to service

jrasm91:


server/apps/immich/src/api-v1/asset/asset.service.ts
```diff
@@ -186,6 +188,8 @@ export class AssetService {
   }
 
   public async updateAsset(authUser: AuthUserDto, assetId: string, dto: UpdateAssetDto): Promise<AssetResponseDto> {
+    await this.checkAssetsAccess(authUser, [assetId], true);
+
     const asset = await this._assetRepository.getById(assetId);
     if (!asset) {
```
michelheusschen, line 194: Not caused by this PR, but this check can never be reached, right? The `checkAssetsAccess` also makes sure the asset exists and `_assetRepository.getById` uses `findOneOrFail` which would throw it's own error. This happens a few more times, if you're willing to change it.

michelheusschen (approved):

server/apps/immich/src/api-v1/asset/asset.service.ts
jrasm91: That's a good point. I think I will handle it in a follow up PR.

jrasm91: To circle back to this, we actually have a more general inconsistency I'd like to address. Sometimes we throw a bad request when an entity isn't found. But the query uses a where clause that includes the user id. So like `{ where: { userId, id } }`. I think it might be better to always query by `id`, and then have a subsequent check to verify access. This would mean we throw separate errors between truly not found id vs an id the user doesn't have access to.

alextran1502 (approved):

server/apps/immich/src/api-v1/asset/asset.service.ts
michelheusschen: A bad request error should only be returned if there's something wrong with the request itself, not for unauthorized or not found situations.

Instead of separate errors it's also pretty common for APIs to always return a not found error, because it doesn't provide any information to unauthorized users.

jrasm91: IMO 404 implies the request url is wrong. If you make a request to `asset/:id` with an id that doesn't exist the path is right, the request is wrong, so it's a 400.

I've also seen and am also fine with non 403 response codes when the user can't access/see some data. Either way, in our API we currently have a mix of both of these. It would be good to make it consistent one way or the other.

PR #2643, opened 2023-06-02
fix(web): use id instead of assetId

michelheusschen:
#2641 changed `assetId` to `id` but some places on web didn't get changed, leading to an error when making API calls. Also fixed invalid types in `+page.server.ts`.

These type of errors should get caught automatically, I'll create a separate PR to add that.


jrasm91 (approved):
Hmm, why did these not cause a compilation error?

alextran1502 (approved):

PR #2651, opened 2023-06-02
refactor(server): remove invalid exif coordinates

michelheusschen:
Use a migration to remove existing invalid data, instead of filtering invalid data out with a query like I did in #2648


jrasm91 (approved):
Perfect, thanks!

alextran1502 (approved):

PR #2652, opened 2023-06-03
fix(web+server): showing assets without thumbnail

michelheusschen:
#2561 added support for showing assets without thumbnails, but some edge cases were missed:
1. When a bucket only contains assets without thumbnails, the bucket won't be returned by `/asset/count-by-time-bucket` and won't show up in the timeline
2. If you archive an asset without thumbnail, the asset will be become hidden

Fixed by adding `withoutThumbnails` to these endpoints and showing assets without thumbnails on the archive and favorites page


jrasm91 (approved):

alextran1502 (changes requested):
I ran into this error in development.

image

Have you tested this change in production build?

alextran1502 (approved):
Wrong comment for a different PR :P

PR #2666, opened 2023-06-05
feat(web, server): Implement justified layout for AssetGrid

brighteyed:
This PR implements justified layout for `AssetGrid`

Photos (wide):

Photos (narrow): 

Add to album:


server/apps/immich/src/api-v1/asset/asset-repository.ts
```diff
@@ -150,6 +154,27 @@ export class AssetRepository implements IAssetRepository {
...
+      .createQueryBuilder('asset')
+      .leftJoin('asset.exifInfo', 'ei')
+      .select(`date_trunc('month', "fileCreatedAt")`, 'timeBucket')
+      .addSelect('ei.exifImageHeight', 'exifImageHeight')
+      .addSelect('ei.exifImageWidth', 'exifImageWidth')
+      .addSelect('ei.orientation', 'orientation')
+      .where('"ownerId" = :userId', { userId })
+      .andWhere('asset.isVisible = true')
+      .andWhere('asset.isArchived = false')
+      .andWhere('asset.resizePath is not NULL')
```
jrasm91, line 168: I think you need to remove this one, since you have it conditionally in the if statement below now.

jrasm91 (commented):
Do we need a separate endpoint for this. Can we now just add the aspect ratio property to the original time bucket response and re-use the existing endpoint?

brighteyed:
> Do we need a separate endpoint for this. Can we now just add the aspect ratio property to the original time bucket response and re-use the existing endpoint?

Removed `/asset/timeline-layout` endpoint. Looks like rough timeline height estimation is enough

alextran1502 (approved):

PR #2668, opened 2023-06-05
use imagemagick and libraw for raw image support

### LINKED ISSUE #2156: [BUG] Sony RAW (.ARW) unsupported
### The bug

I found https://github.com/immich-app/immich/issues/1109, but do not believe the resolution is fair. The Sony RAW format is very popular and is a major blocker for adoption.

The linked [libvips issue](https://github.com/libvips/libvips/issues/1304) suggests there is support for the file format, but `magickload` must be used instead.

N/A

### Version of Immich Server

v1.52.1

N/A

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
N/A
```

### Additional information

N/A

uhthomas:
imagemagick and libraw have generally good support for raw images, including
Sony's ARW format. These tools should also allow Immich to support many more
image formats in future without any major code changes.

https://www.libraw.org/supported-cameras

I've tested and verified this change with .ARW files and other standard formats.

Fixes: https://github.com/immich-app/immich/issues/2156

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -2,7 +2,7 @@ FROM node:18.16.0-alpine3.18@sha256:f41850f74ff16a33daff988e2ea06ef8f5daeb6fb849
 
 WORKDIR /usr/src/app
 
-RUN apk add --update-cache build-base python3 vips-heif vips-dev ffmpeg perl
+RUN apk add --update-cache build-base imagemagick-dev python3 ffmpeg libraw-dev perl vips-dev vips-heif vips-magick
 
 COPY package.json package-lock.json ./
 
@@ -23,7 +23,7 @@ ENV NODE_ENV=production
 
 WORKDIR /usr/src/app
 
-RUN apk add --no-cache vips-heif vips vips-cpp ffmpeg perl
+RUN apk add --no-cache ffmpeg imagemagick-dev libraw-dev perl vips-dev vips-heif vips-magick
 
 COPY --from=prod /usr/src/app/node_modules ./node_modules
 COPY --from=prod /usr/src/app/dist ./dist
--- a/server/src/domain/media/media.service.ts
+++ b/server/src/domain/media/media.service.ts
@@ -60,9 +60,9 @@ export class MediaService {
           size: JPEG_THUMBNAIL_SIZE,
           format: 'jpeg',
         });
-      } catch (error) {
+      } catch (error: any) {
         this.logger.warn(
-          `Failed to generate jpeg thumbnail using sharp, trying with exiftool-vendored (asset=${asset.id})`,
+          `Failed to generate jpeg thumbnail using sharp, trying with exiftool-vendored (asset=${asset.id}): ${error.message}`,
         );
         await this.mediaRepository.extractThumbnailFromExif(asset.originalPath, jpegThumbnailPath);
       }
--- a/web/src/lib/utils/asset-utils.ts
+++ b/web/src/lib/utils/asset-utils.ts
@@ -125,34 +125,20 @@ export function getAssetFilename(asset: AssetResponseDto): string {
  * Returns the MIME type of the file and an empty string when not found.
  */
 export function getFileMimeType(file: File): string {
-	if (file.type !== '') {
-		// Return the MIME type determined by the browser.
-		return file.type;
-	}
-
-	// Return MIME type based on the file extension.
-	switch (getFilenameExtension(file.name)) {
-		case 'heic':
-			return 'image/heic';
-		case 'heif':
-			return 'image/heif';
-		case 'dng':
-			return 'image/dng';
-		case '3gp':
-			return 'video/3gpp';
-		case 'nef':
-			return 'image/x-nikon-nef';
-		case 'raf':
-			return 'image/x-fuji-raf';
-		case 'srw':
-			return 'image/x-samsung-srw';
-		case 'insp':
-			return 'image/jpeg';
-		case 'insv':
-			return 'video/mp4';
-		default:
-			return '';
-	}
+	const mimeTypes: Record<string, string> = {
+		'3gp': 'video/3gpp',
+		arw: 'image/x-sony-arw',
+		dng: 'image/dng',
+		heic: 'image/heic',
+		heif: 'image/heif',
+		insp: 'image/jpeg',
+		insv: 'video/mp4',
+		nef: 'image/x-nikon-nef',
+		raf: 'image/x-fuji-raf',
+		srw: 'image/x-samsung-srw'
+	};
+	// Return the MIME type determined by the browser or the MIME type based on the file extension.
+	return file.type || (mimeTypes[getFilenameExtension(file.name)] ?? '');
 }
 
 /**
--- a/web/src/lib/utils/asset-utils.ts
+++ b/web/src/lib/utils/asset-utils.ts
@@ -127,7 +127,7 @@ export function getAssetFilename(asset: AssetResponseDto): string {
 export function getFileMimeType(file: File): string {
 	const mimeTypes: Record<string, string> = {
 		'3gp': 'video/3gpp',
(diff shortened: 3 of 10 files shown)


uhthomas:
I've tested this, and am awaiting a response from the sharp author regarding https://github.com/lovell/sharp/issues/3612#issuecomment-1576714726. Please don't merge yet.

samip5:
I think the idea was to use darktable.

See https://github.com/immich-app/immich/pull/2198

uhthomas:
> I think the idea was to use darktable.
> 
> See #2198

Also see https://github.com/immich-app/immich/issues/2156#issuecomment-1561955606.

uhthomas:
It would be nice to merge https://github.com/immich-app/immich/pull/2736 before this.

server/Dockerfile
```diff
@@ -17,13 +17,13 @@ RUN npm run build
...
 
-FROM node:18-alpine3.17
+FROM node:18.16.0-alpine3.18@sha256:f41850f74ff16a33daff988e2ea06ef8f5daeb6fb84913c7df09552a98caba09
 
 ENV NODE_ENV=production
 
 WORKDIR /usr/src/app
 
-RUN apk add --no-cache vips-heif vips vips-cpp ffmpeg perl
+RUN apk add --no-cache ffmpeg imagemagick-dev libraw-dev perl vips-dev vips-heif vips-magick
```
uhthomas, line 26: `-dev` packages may not be needed here.

uhthomas: I tested without `vips-dev` and it wasn't happy at all. I presume all it wants is `vips-cpp`.

uhthomas: Nope, not happy with `vips-cpp` either. I'll leave the `-dev` packages as is :)

server/src/domain/media/media.service.ts
```diff
@@ -60,9 +60,9 @@ export class MediaService {
           size: JPEG_THUMBNAIL_SIZE,
           format: 'jpeg',
         });
-      } catch (error) {
+      } catch (error: any) {
         this.logger.warn(
-          `Failed to generate jpeg thumbnail using sharp, trying with exiftool-vendored (asset=${asset.id})`,
+          `Failed to generate jpeg thumbnail using sharp, trying with exiftool-vendored (asset=${asset.id}): ${error.message}`,
```
uhthomas, line 65: Q for reviewers: Should we just remove this? vips/imagemagick should do this automatically. Might be nice to reduce code footprint and complexity. What do you think?

jrasm91: This was a work around (fallback) for raw images that had an embedded thumbnail. If raw photos work without this I think we can definitely remove it.

uhthomas: https://github.com/immich-app/immich/issues/2744

jrasm91 (commented):
How big of an image size jump does moving away from alpine cause?

uhthomas:
> How big of an image size jump does moving away from alpine cause?

Hm? We are using alpine.

web/src/lib/utils/asset-utils.ts
```diff
@@ -131,28 +131,20 @@ export function getFileMimeType(file: File): string {
...
+			'.arw': 'image/x-sony-arw',
+			'.dng': 'image/dng',
+			'.heic': 'image/heic',
+			'.heif': 'image/heif',
+			'.insp': 'image/jpeg',
+			'.insv': 'video/mp4',
+			'.nef': 'image/x-nikon-nef',
+			'.raf': 'image/x-fuji-raf',
+			'.srw': 'image/x-samsung-srw'
+		}[getFilenameExtension(file.name)] ?? ''
```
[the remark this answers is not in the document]
uhthomas: You're right, I'm not sure how I did that. Thank you.

jrasm91:
> > How big of an image size jump does moving away from alpine cause?
> 
> Hm? We are using alpine.

Ah, sorry I mis-read the change. `alpine` got bumped to the next line with the addition of the hash.

web/src/lib/utils/asset-utils.ts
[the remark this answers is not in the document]
uhthomas: @jrasm91 I did one even better and added a test for this. The test also revealed that null or undefined values were not handled correctly, and has been fixed.

elliotclee (commented):
Hi,

Thanks for working on this. I like the way you changed to using arrays instead of endless if() statements in a number of areas.

I think it would make sense to add support for all the filetypes that libraw can handle. I came looking for .cr2 (image/x-canon-cr2), and I gather there are several other types that it would be helpful to accept. See https://invent.kde.org/graphics/digikam/-/blob/master/core/libs/rawengine/drawfiles.cpp for a fairly complete list of the file types supported by libraw, as well as https://github.com/angryziber/gnome-raw-thumbnailer/blob/master/data/raw-thumbnailer.xml for info on the MIME types for them.

Hope this helps,

uhthomas:
> Hi,
> 
> Thanks for working on this. I like the way you changed to using arrays instead of endless if() statements in a number of areas.
> 
> I think it would make sense to add support for all the filetypes that libraw can handle. I came looking for .cr2 (image/x-canon-cr2), and I gather there are several other types that it would be helpful to accept. See https://invent.kde.org/graphics/digikam/-/blob/master/core/libs/rawengine/drawfiles.cpp for a fairly complete list of the file types supported by libraw, as well as https://github.com/angryziber/gnome-raw-thumbnailer/blob/master/data/raw-thumbnailer.xml for info on the MIME types for them.
> 
> Hope this helps,

I would love to do this, but I am afraid for the scope of this PR becoming too large, which is why I separated some changes into #2736. I think the best thing to do would be to merge this as-is, closing the original Sony ARW issue and create another issue like #2743 to support the remaining libraw formats. Does that sound good to you? I can create that issue now.

jrasm91:
I think it is OK to add, test, and verify support for specific formats. With the change to the base image it'll be pretty trivial to add support for additional formats. Whether we do all that testing in this PR or it follow-up PRs seems less important. I don't think @uhthomas needs to commit to testing every single format type before we merge this support for ARW.

uhthomas:
I've created #2745 to track the implementation of all libraw formats.

jrasm91 (approved):

uhthomas:
It looks like this job was cancelled and is confusing GitHub. Any idea what happened, or if it's intentional?

https://github.com/immich-app/immich/actions/runs/5248322532/jobs/9479643740?pr=2668

uhthomas:
Strange, I made force pushed and now a different build was 'cancelled' https://github.com/immich-app/immich/actions/runs/5248561828/jobs/9480220088?pr=2668.

jrasm91:
I have had some issues of late with force-pushing having weird impact on the running jobs.

uhthomas:
Ah, https://www.githubstatus.com/incidents/csnbwc85chp2.

web/src/lib/utils/asset-utils.ts
```diff
@@ -125,34 +125,20 @@ export function getAssetFilename(asset: AssetResponseDto): string {
...
-		case 'insp':
-			return 'image/jpeg';
-		case 'insv':
-			return 'video/mp4';
-		default:
-			return '';
-	}
+	const mimeTypes: Record<string, string> = {
+		'3gp': 'video/3gpp',
+		arw: 'image/x-sony-arw' || 'image/awr',
```
uhthomas, line 130: I don't think that will work?

uhthomas:
It seems like this image is missing quite a few runtime dependencies. Namely imagemagick-dev and vips-magick.

https://github.com/immich-app/immich/blob/af86712df3b093cc2a342cff3279833ba055d520/server/Dockerfile#L26

PR #2693, opened 2023-06-08
feat(ml): batched model inference

mertalev:
This is a draft PR that makes ML inference requests and responses support a batch dimension. Models are optimized to run inference in batches for a number of reasons. However, until now, every model input had to be sent as a separate request and handled alone. This adds considerable overhead and is a major performance bottleneck for GPUs from testing.

While this PR (seems to) work, it's not quite ready for merging yet. 

- Only explicitly started jobs are batched; jobs that happen after thumbnail generation aren't affected since they're handled separately. 
- The batch size is currently a hardcoded constant and separate from the models' job concurrencies. This means with a batch size of 4 and a concurrency of 2, 2 requests are sent with 4 images each. 
  - It's possible to instead have the job concurrency act as the batch size and send 1 request at a time, but I don't know which is better.
- I also want to test it a bit more to make sure the batching logic is completely sound.

### DIFF
--- a/machine-learning/app/cache.py
+++ b/machine-learning/app/cache.py
@@ -2,6 +2,7 @@ from aiocache.plugins import TimingPlugin, BasePlugin
 from aiocache.backends.memory import SimpleMemoryCache
 from aiocache.lock import OptimisticLock
 from typing import Any
+from schemas import ModelType
 from models import get_model
 
 
@@ -36,7 +37,7 @@ class ModelCache:
         )
 
     async def get_cached_model(
-        self, model_name: str, model_type: str, **model_kwargs
+        self, model_name: str, model_type: ModelType, **model_kwargs
     ) -> Any:
         """
         Args:
--- a/machine-learning/app/main.py
+++ b/machine-learning/app/main.py
@@ -1,8 +1,9 @@
 import os
-from typing import Any
-
+from pathlib import Path
+from typing import Any, Callable
 from cache import ModelCache
 from schemas import (
+    R,
     EmbeddingResponse,
     FaceResponse,
     TagResponse,
@@ -12,10 +13,10 @@ from schemas import (
     VisionModelRequest,
 )
 import uvicorn
-
-from PIL import Image
+from PIL import Image, UnidentifiedImageError
 from fastapi import FastAPI, HTTPException
 from models import get_model, run_classification, run_facial_recognition
+import cv2 as cv
 
 classification_model = os.getenv(
     "MACHINE_LEARNING_CLASSIFICATION_MODEL", "microsoft/resnet-50"
@@ -73,7 +74,8 @@ async def image_classification(payload: VisionModelRequest) -> list[str] | list[
     model = await _model_cache.get_cached_model(
         classification_model, "image-classification"
     )
-    labels = run_classification(model, payload.image_paths, min_tag_score)
+    images = _load_images(payload.image_paths, Image.open)
+    labels = run_classification(model, images, min_tag_score)
     return labels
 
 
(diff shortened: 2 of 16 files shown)


jrasm91:
This adds quite a bit of complexity to these jobs. Have you measured the performance gains that these supposedly bring? If we do indeed want to batch it we may want to have a different architecture to accomplish it, like having ML pull jobs directly from Redis and do the batching in ML directly. We lose the ability to efficiently report problems and re-try failed jobs when we use batches. This also goes back to weighing the benefit of optimizing for bulk processing vs single items. Bulk processing is beneficial in some situations, like immich onboarding, but decreases after that as assets are usually backed up on smaller numbers afterwards. It would have been good to create a focus topic in discord to discuss these changes ahead of time btw.

mertalev:
These changes are mostly from an earlier stash. I wasn't sure where would be the best place to discuss the implementation details, which is partly why I made this a draft. I'll keep the discord in mind for that. 

In terms of performance, I'll be running some benchmarks later today.

>  If we do indeed want to batch it we may want to have a different architecture to accomplish it, like having ML pull jobs directly from Redis and do the batching in ML directly.

That's an interesting idea, but I feel it would straddle the line between the roles of `server` and `machine-learning`. Would the main benefit be knowing, e.g., when a particular image failed to load? I wonder if that could be handled with more error handling. It could respond with a particular HTTP status code like `415 Unsupported Media Type` to signal the server to re-queue that batch in the form of individual images to find the offender.

> Bulk processing is beneficial in some situations, like immich onboarding, but decreases after that as assets are usually backed up on smaller numbers afterwards.

Since my code doesn't handle inference from added images, does this mean this bug is a feature?  😄

mertalev:
Results for CPU inference
cpu_stats

While the biggest benefit to batching is on the GPU, these results suggest that batching improves performance for CPU usage as well. CLIP and facial recognition see improvements of 30-40%, while object tagging is in the range of 13-18%. Interestingly facial recognition sees a regression at a batch size of 16. I'm not sure why; I'll compare it with using a non-batched version of `run_facial_recognition` to see if it has to do with that.

mertalev:
For comparison, running the older version of `run_facial_recognition` in a list comprehension without taking advantage of batching takes 389.84s with a batch size of 16, versus 232.02s when properly batched.

jrasm91:
Clip generates a single embedding per image, so a batch of 16 is that many embeddings. FR generates an embedding per face, so a batch of 16 images could have anywhere from 0 to 50+. I wonder if that contributes to the decreased throughput. Perhaps the overhead and storing, encoding, sending, decoding, inserting, etc. that much information causes it to not work as well as say batch of size 8. Overall, I think the testing looks really good and it seems like something we should definitely consider implementing.

mertalev:
Ohhh, that makes a lot of sense. There's a point of diminishing returns with batching on CPU, so the benefit of generating more embeddings at once plateaus while the overhead surrounding it still adds up. Even for GPU usage, the fact that the number of faces can vary is a good reason to stick with something lower than 16 since it can add up to a much larger number of embeddings that might not fit in the GPU.

jrasm91:
Can you change the interfaces to always use lists? I think the code would be a bit simpler (less if/else statements) if you could assume it as always a list coming in and always a list going out.

mertalev:
Agreed, it's a lot of clutter for something that's not actually being used.

mertalev:
In terms of batching without explicitly starting a job, is there a particular approach that comes to mind? The batching currently just breaks up the pages pulled by the queue handlers, but those handlers don't get used for the `onDone` jobs. That's fine for daily uploads, but for the initial import it'd be nice for that to be batched as well.

jrasm91:
I'll have to think about that. Maybe what we do for search. The search index jobs add the IDs to a list and then they are indexed in a batch. The tricky part is just figuring out how to chain jobs correctly, specifically the on done portion. It may be worthwhile to just focus on the manually invoked jobs first and update those to work correctly.

mertalev:
Sounds like it'll take some work to support that. I agree it should be fine to focus on explicit starts for this PR, it's big enough as it is haha. 

I'm thinking maybe the ML jobs can just be nightly cron jobs, so there's only one case to handle regardless. It'd also be good in terms of minimizing RAM/VRAM usage since only one model would need to be used at a time.

mertalev:
Closing this PR for now since #2758 addresses batching in a way that's simpler for the server. Might re-approach it later if the need arises.

PR #2705, opened 2023-06-09
feat(mobile): sync remote assets without thumbs

fyfrey:
syncs all remote assets to the app - even if the assets do not yet have a thumbnail image on the server.

This prevents users from thinking their backup or sync is not working properly (while only the server has not yet produced a thumbnail image for whatever reason)

### DIFF
--- a/mobile/lib/shared/services/asset.service.dart
+++ b/mobile/lib/shared/services/asset.service.dart
@@ -66,8 +66,11 @@ class AssetService {
     try {
       final etag = hasCache ? _db.eTags.getByIdSync(user.id)?.value : null;
       final (List<AssetResponseDto>? assets, String? newETag) =
-          await _apiService.assetApi
-              .getAllAssetsWithETag(eTag: etag, userId: user.id);
+          await _apiService.assetApi.getAllAssetsWithETag(
+        eTag: etag,
+        userId: user.id,
+        withoutThumbs: true,
+      );
       if (assets == null) {
         return null;
       } else if (assets.isNotEmpty && assets.first.ownerId != user.id) {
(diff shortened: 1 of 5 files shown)


jrasm91 (approved):

mobile/lib/shared/ui/immich_image.dart
```diff
@@ -110,8 +111,12 @@ class ImmichImage extends StatelessWidget {
         );
       },
       errorWidget: (context, url, error) {
-        debugPrint("Error getting thumbnail $url = $error");
-        CachedNetworkImage.evictFromCache(thumbnailRequestUrl);
+        if (error is HttpExceptionWithStatus &&
+            error.statusCode >= 400 &&
```
uhthomas, line 115: nit: not sure if it's valid in dart, but this might be more concise:

```dart
error.statusCode / 100 == 4
```

jrasm91: Http status codes are pretty common and I think referencing 400 is more readable than just a 4.

uhthomas: Yeah, either way is fine. Mostly just interested to offer different perspectives.

mobile/lib/shared/services/asset.service.dart
```diff
@@ -66,8 +66,11 @@ class AssetService {
     try {
       final etag = hasCache ? _db.eTags.getByIdSync(user.id)?.value : null;
       final (List<AssetResponseDto>? assets, String? newETag) =
-          await _apiService.assetApi
-              .getAllAssetsWithETag(eTag: etag, userId: user.id);
+          await _apiService.assetApi.getAllAssetsWithETag(
```
uhthomas, line 69: strange formatting?

fyfrey: all good, it's dart's auto-formatter and you find it everywhere in the dart codebase :-)

PR #2709, opened 2023-06-09
feat(server): option to transcode to original resolution

mertalev:
Implements #2707, allowing users to transcode without setting a target resolution. Specifically, a new option is added for target resolution: "original".

### DIFF
--- a/server/src/domain/media/media.service.spec.ts
+++ b/server/src/domain/media/media.service.spec.ts
@@ -259,6 +259,23 @@ describe(MediaService.name, () => {
       );
     });
 
+    it('should not scale resolution if no target resolution', async () => {
+      mediaMock.probe.mockResolvedValue(probeStub.videoStream2160p);
+      configMock.load.mockResolvedValue([
+        { key: SystemConfigKey.FFMPEG_TRANSCODE, value: 'all' },
+        { key: SystemConfigKey.FFMPEG_TARGET_RESOLUTION, value: '' },
+      ]);
+      await sut.handleVideoConversion({ id: assetEntityStub.video.id });
+      expect(mediaMock.transcode).toHaveBeenCalledWith(
+        '/original/path.ext',
+        'upload/encoded-video/user-id/asset-id.mp4',
+        {
+          outputOptions: ['-vcodec h264', '-acodec aac', '-movflags faststart', '-preset ultrafast', '-crf 23'],
+          twoPass: false,
+        },
+      );
+    });
+
     it('should transcode with alternate scaling video is vertical', async () => {
       mediaMock.probe.mockResolvedValue(probeStub.videoStreamVertical2160p);
       configMock.load.mockResolvedValue([{ key: SystemConfigKey.FFMPEG_TRANSCODE, value: 'optimal' }]);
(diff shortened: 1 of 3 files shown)


server/src/domain/media/media.service.spec.ts
```diff
@@ -259,6 +259,23 @@ describe(MediaService.name, () => {
       );
     });
 
+    it('should not scale resolution if no target resolution', async () => {
+      mediaMock.probe.mockResolvedValue(probeStub.videoStream2160p);
+      configMock.load.mockResolvedValue([
+        { key: SystemConfigKey.FFMPEG_TRANSCODE, value: 'all' },
+        { key: SystemConfigKey.FFMPEG_TARGET_RESOLUTION, value: '' },
```
jrasm91, line 266: This would actually be the value `original`, right?

server/src/domain/media/media.service.ts
```diff
@@ -180,8 +180,8 @@ export class MediaService {
 
     const allTargetsMatching = isTargetVideoCodec && isTargetAudioCodec && isTargetContainer;
 
-    const targetResolution = Number.parseInt(ffmpegConfig.targetResolution);
-    const isLargerThanTargetResolution = Math.min(videoStream.height, videoStream.width) > targetResolution;
+    const targetRes = Number.parseInt(ffmpegConfig.targetResolution) || Math.min(videoStream.height, videoStream.width);
+    const isLargerThanTargetRes = Math.min(videoStream.height, videoStream.width) > targetRes;
```
jrasm91, line 184: This is not very clear to me. You are relying on parseInt to fail to parse `original` for this code to work properly. Can we add a more clear/obvious check for the text `original` instead? Specifically, leave the existing code as is and add a new variable is original then set isLargerThanTargetRes to the original condition  and not isOriginal. Maybe worth extracting get target resolution into a separate method for clarity too.

mertalev: The issue I saw was that `targetResolution` isn't actually a valid target resolution if `original` is set. I guess it's fine though since it's a dropdown and not user input.

server/src/domain/media/media.service.spec.ts
mertalev: Yup, the text value was originally just an empty string but I forgot to update the test when I changed it..

server/src/domain/media/media.service.ts
jrasm91: Validating user input for that field is more of an unrelated issue IMO. In this PR we're adding a known value of "original", which I think should have specific behavior linked to it, which should be easy to follow in the code here. I think unrelated we can add some logic for what to do when the `targetResolution` is invalid, which _could_ also be to not adjust it. Although, the better behavior might be to fallback to the default value (`'720'`) instead.

jrasm91 (approved):
Looks great. I'd be fine also adding a fallback value to the `targetResolution` in the case that the value is invalid for whatever reason. But we could also do that in a separate PR.

server/src/domain/media/media.service.spec.ts
jrasm91: Ah, makes sense. Looks good now!

mertalev:
For now it's probably best to leave it as-is. If it's NaN but also not `original`, then something clearly went very wrong so it shouldn't get hidden with a fallback value.

PR #2715, opened 2023-06-10
feat(server): support for read-only assets  and importing existing items in the filesystem

alex-phillips:


server/src/domain/storage-template/storage-template.service.ts
```diff
@@ -76,6 +76,11 @@ export class StorageTemplateService {
 
   // TODO: use asset core (once in domain)
   async moveAsset(asset: AssetEntity, metadata: MoveAssetMetadata) {
+    if (asset.isReadOnly) {
+      this.logger.debug(`Not moving read-only asset: ${asset.originalPath}`);
```
jrasm91, line 80: Maybe this should be a verbose log actually

server/src/immich/api-v1/asset/asset.controller.ts
```diff
@@ -118,6 +119,38 @@ export class AssetController {
     return responseDto;
   }
 
+  @SharedLinkRoute()
+  @Post('import')
+  @ApiBody({
+    description: 'Asset Import Information',
+    type: ImportAssetDto,
```
jrasm91, line 126: This API body decorator should not be needed. It can be inferred from the `@Body` type in the method definition, which actually isn't the same type. Is that on purpose?

server/src/immich/api-v1/asset/dto/import-asset.dto.ts
```diff
@@ -0,0 +1,48 @@
...
+export interface ImportFile {
+  mimeType: string;
+  checksum: Buffer;
+  originalPath: string;
+  originalName: string;
+}
+
+export async function mapToImportFile(filepath: string): Promise<ImportFile> {
+  const pathInfo = path.parse(filepath);
+  const fileBuffer = await fs.readFile(filepath);
```
jrasm91, line 38: This will load the whole file into memory at once. It should be streamed instead.

server/src/immich/api-v1/asset/dto/import-asset.dto.ts
```diff
@@ -0,0 +1,48 @@
...
+  @ApiProperty()
+  @IsString()
+  assetData!: string;
+
+  @ApiProperty()
+  @IsString()
+  @IsOptional()
+  livePhotoData?: string;
+
+  @ApiProperty()
```
jrasm91, line 23: These API property decorators are not needed

server/src/immich/api-v1/asset/dto/import-asset.dto.ts
```diff
@@ -0,0 +1,48 @@
...
+  mimeType: string;
+  checksum: Buffer;
+  originalPath: string;
+  originalName: string;
+}
+
+export async function mapToImportFile(filepath: string): Promise<ImportFile> {
+  const pathInfo = path.parse(filepath);
+  const fileBuffer = await fs.readFile(filepath);
+  const hashSum = crypto.createHash('md5');
```
jrasm91, line 39: We use sha1 hash

server/src/immich/api-v1/asset/dto/import-asset.dto.ts
```diff
@@ -0,0 +1,48 @@
...
+  @IsOptional()
+  sidecarData?: string;
+}
+
+export interface ImportFile {
+  mimeType: string;
+  checksum: Buffer;
+  originalPath: string;
+  originalName: string;
+}
```
jrasm91, line 34: Is there not already an immich file interface we can use or that is already used in asset core create?

server/src/immich/api-v1/asset/asset.service.ts
```diff
@@ -105,6 +105,21 @@ export class AssetService {
...
+        if (fileToCheck) {
+          try {
+            await fs.access(file.originalPath, constants.R_OK);
+          } catch (error: any) {
+            this.logger.error(`Error importing file`, `${error}`);
+            throw new BadRequestException(`Error importing file`, `${error}`);
+          }
+        }
+      }
+    }
```
jrasm91, line 121: Read only can be used independently of specifying an existing file that needs to be verified or not. Can you add a property to `UploadFile` that tracks if it is verified or not? Or if it is type upload or import? Then we can verify based on that property instead and keep it separate from RO functionality, which can be used in both situations.

I'd actually prefer a separate method here called importFile. You could put the import specific code there and then call uploadFile afterwards.

server/src/immich/api-v1/asset/asset.service.ts
```diff
@@ -386,13 +410,16 @@ export class AssetService {
...
-          asset.sidecarPath,
-        );
+
+        if (!asset.isReadOnly) {
+          deleteQueue.push(
+            asset.originalPath,
+            asset.webpPath,
+            asset.resizePath,
+            asset.encodedVideoPath,
+            asset.sidecarPath,
```
jrasm91, line 420: When an asset is deleted through the UI we need to skip read only ones there as well.

server/src/domain/storage-template/storage-template.service.ts
alex-phillips: Changed this to verbose in my next code push.

server/src/immich/api-v1/asset/asset.controller.ts
alex-phillips: Removed this in next code push.

server/src/immich/api-v1/asset/dto/import-asset.dto.ts
alex-phillips: Converted this into a stream.

server/src/immich/api-v1/asset/dto/import-asset.dto.ts
alex-phillips: Changed this to sha1 with the stream changes.

server/src/immich/api-v1/asset/dto/import-asset.dto.ts
alex-phillips: I went with ImportFile for consistency in case there were any future cases, but that can always be modified later. Removed this in favor of UploadFile.

server/src/immich/api-v1/asset/asset.service.ts
alex-phillips: Went with the last suggestion of having an importFile method which calls uploadFile afterwards, keeping the import logic separate.

server/src/immich/api-v1/asset/asset.service.ts
alex-phillips: I didn't see any other calls to queue a delete in the code. The API endpoint for `/delete` just appears to call deleteAll. Could you point me in the direction of where the others may be?

alex-phillips: Also, I initially opted to put this into the job itself, but the only data currently in the queue are actual filepaths, there is no asset information passed into this job.

server/src/immich/api-v1/asset/dto/import-asset.dto.ts
alex-phillips: Removed these.

server/src/immich/api-v1/asset/asset.service.ts
jrasm91: I think we have some work to do to refactor how assets are deleted. Ideally using a queue by asset id.

server/src/immich/api-v1/asset/asset-repository.ts
```diff
@@ -360,4 +361,23 @@ export class AssetRepository implements IAssetRepository {
...
+   */
+  getAssetByOriginalPath(ownerId: string, originalPath: string): Promise<AssetCheck[]> {
+    return this.assetRepository.find({
+      select: {
+        id: true,
+        checksum: true,
+      },
+      where: {
+        ownerId,
+        originalPath,
```
jrasm91, line 379: If we are going to query by this i wonder if we need an index in this column. If we want it to be unique, we should consider a unique constraint.

server/src/immich/api-v1/asset/dto/import-asset.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+  @IsOptional()
+  sidecarData?: string;
+}
+
+export async function mapToImportFile(filepath: string): Promise<UploadFile> {
+  const pathInfo = path.parse(filepath);
+
+  const checksum = await new Promise<string>((resolve, reject) => {
+    const hash = crypto.createHash('sha1');
+    const stream = fs.createReadStream(filepath);
```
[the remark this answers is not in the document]
jrasm91: I think you can use sanitize filename package here to fix this.

server/src/immich/api-v1/asset/asset.service.ts
jrasm91: Oh you're right it does always call this delete all endpoint

jrasm91:
Should we add a check that prevents a user from importing a path that is in another users folder? The existing structure uses /usr/src/app/upload/library/label or UUID/...

There could be some security issue where a user could gain access to another file using this import method with a valid path of another users file and then accessing or deleting it.

alex-phillips:
> 

Hmm, not sure here. It could be a security issue, but if the person doing the importing also has access to the files, wouldn't it be ok?

jrasm91:
Well they don't have access to them per se, hence the security issue. Relying on them just not knowing the path isn't security imo

server/src/immich/api-v1/asset/asset-repository.ts
alex-phillips: Added an index

server/src/immich/api-v1/asset/dto/import-asset.dto.ts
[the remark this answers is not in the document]
alex-phillips: Added this sanitize call

server/src/immich/api-v1/asset/asset.service.ts
alex-phillips: Marking this as resolved unless you want the refactor to happen in this PR.

alex-phillips:
> Well they don't have access to them per se, hence the security issue. Relying on them just not knowing the path isn't security imo

Good point - added in a new method and duplicate check in the importFile method call.

alextran1502:
Can you please help outline the steps on how to test the PR?

etnoy:
Could you add a bit more description to the PR? I want to try it but it's not obvious what paths can be imported

server/src/immich/api-v1/asset/asset.service.ts
```diff
@@ -142,6 +145,73 @@ export class AssetService {
...
   }
 
+  public async importFile(authUser: AuthUserDto, dto: ImportAssetDto): Promise<AssetFileUploadResponseDto> {
+    dto = {
+      ...dto,
+      assetPath: path.resolve(dto.assetPath),
+      sidecarPath: dto.sidecarPath ? path.resolve(dto.sidecarPath) : undefined,
+    };
+
+    const assetPathType = mime.lookup(dto.assetPath) as string
```
jrasm91, line 155: We may want/need to test this lookup library for the mime types we support to make sure it correctly sends us the right one.

jrasm91 (approved):

server/src/infra/migrations/1686668349751-add-user-externalpath.ts
```diff
@@ -0,0 +1,14 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class AddUserExternalpath1686668349751 implements MigrationInterface {
+    name = 'AddUserExternalpath1686668349751'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "users" ADD "externalPath" character varying`);
```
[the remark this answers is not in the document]
alex-phillips: Done.

server/src/immich/api-v1/asset/asset.service.ts
```diff
@@ -142,6 +145,73 @@ export class AssetService {
...
+      }
+
+      if (error instanceof QueryFailedError && (error as any).constraint === 'UQ_4ed4f8052685ff5b1e7ca1058ba') {
+        const duplicate = await this._assetRepository.getByOriginalPath(dto.assetPath);
+        if (duplicate) {
+          if (duplicate.ownerId === authUser.id) {
+            return { id: duplicate.id, duplicate: true };
+          }
+
+          throw new BadRequestException('Path in user by another user');
```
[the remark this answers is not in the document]
alex-phillips: Fixed typo.

etnoy:
It appears the code doesn't build right now so I can't test it

server/src/domain/shared-link/shared-link.core.ts
```diff
@@ -71,6 +71,7 @@ export class SharedLinkCore {
             isAllowUpload: link.allowUpload,
             isAllowDownload: link.allowDownload,
             isShowExif: link.showExif,
+            externalPath: user.externalPath,
```
alextran1502, line 74: Why does shared link is related in this PR?

alextran1502:
@jrasm91 When you have time, can you help me look at the failed tests?

web/src/lib/components/forms/edit-user-form.svelte
```diff
@@ -131,6 +132,17 @@
 			</p>
 		</div>
 
+		<div class="m-4 flex flex-col gap-2">
```
alextran1502, line 135: We should add some description here of how the file path would be i.e Is it relative or absolute path? If it is relative, where can it be reference from?

alextran1502:
We should also include a column with checkmark of user that has import path in this table
image

server/src/domain/shared-link/shared-link.core.ts
alex-phillips: I think this was accidental, me attempting to get externalPath everywhere the user was exported. Will remove this.

web/src/lib/components/forms/edit-user-form.svelte
alex-phillips: Adding a note here.  Let me know what you think of the  phrasing.

alextran1502:
My first round of testing resulted in the following behavior. Let me know what did I do wrong here.

In the `Administration` page, I have the External Path for the user as the following image.

image

And this is how I created the folder in the location of UPLOAD_LOCATION

image

image

This is the CLI command I use the error I get

image

alex-phillips:
> 

First, problem is that external path for the user must be an absolute path. If you are mounting that folder into the container at `/mnt/immich_data/aleximport` then your external path must be `/mnt/immich_data/aleximport`. Alternatively, if you were creating a new mountpoint inside the container at `/aleximport`, then the external path would be `/aleximport`. The external path needs to be absolute inside the container, beginning with a `/`.

Second, could you share your docker-compose for adding this new mount? The error says file doesn't exist, so I would be interested if it's not mounted properly. You can also create a container in the shell and do an `ls` on one of the expected files. Since you are running the CLI agains the `/mnt/immich_data/aleximport` directory, you should be able to ls that path inside the container as well.

server/src/immich/config/asset-upload.config.ts
```diff
@@ -49,68 +50,18 @@ export const multerUtils = { fileFilter, filename, destination };
...
-  }
-
-  // Additionally support XML but only for sidecar files.
-  if (file.fieldname === 'sidecarData' && ['application/xml', 'text/xml'].includes(file.mimetype)) {
-    return cb(null, true);
+  } else {
+    // Additionally support XML but only for sidecar files.
+    if (file.fieldname === 'sidecarData' && ['application/xml', 'text/xml'].includes(file.mimetype)) {
+      return cb(null, true);
+    }
```
uhthomas, line 64: This should be reverted.

server/src/domain/domain.constant.ts
```diff
@@ -27,3 +27,59 @@ export function assertMachineLearningEnabled() {
     throw new BadRequestException('Machine learning is not enabled.');
   }
 }
+
+const validMimeTypes = [
```
uhthomas, line 31: This should be updated to match what's in main.

image

server/src/immich/api-v1/asset/asset.service.ts
```diff
@@ -129,6 +135,73 @@ export class AssetService {
...
+    };
+
+    const assetPathType = mime.lookup(dto.assetPath) as string;
+    if (!isSupportedFileType(assetPathType)) {
+      throw new BadRequestException(`Unsupported file type ${assetPathType}`);
+    }
+
+    if (dto.sidecarPath) {
+      const sidecarType = mime.lookup(dto.sidecarPath) as string;
+      if (!sidecarType.match(/\/xml$/)) {
```
uhthomas, line 152:
```suggestion
      if (!['application/xml', 'text/xml'].includes(file.mimetype)) {
```

It might be better to make a function for this like was done for `isSupportedFileType`.

server/src/domain/domain.constant.ts
alextran1502: reverted this file to match in `main` = remove changes in this file

PR #2725, opened 2023-06-11
feat(web): support uploading Insta360 file format

truongsinh:
Insta360 "raw" formats `insv` and `insp` are actually mp4 (video) and jpeg (picture) respectively.
However, we don't want user to rename the original files, because they follow Insta360 convention, which is required by Insta360 Studio.

### DIFF
--- a/mobile/lib/utils/files_helper.dart
+++ b/mobile/lib/utils/files_helper.dart
@@ -47,6 +47,12 @@ class FileHelper {
       case 'webm':
         return {"type": "video", "subType": "webm"};
 
+      case 'insp':
+        return {"type": "image", "subType": "jpeg"};
+
+      case 'insv':
+        return {"type": "video", "subType": "mp4"};
+
       default:
         return {"type": "unsupport", "subType": "unsupport"};
     }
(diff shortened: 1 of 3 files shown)


jrasm91 (commented):
Everything else works correctly for these types of files? Transcoding, thumbnails, exif extraction, etc.?

truongsinh:
> Everything else works correctly for these types of files? Transcoding, thumbnails, exif extraction, etc.?

Yes, because I did attempt to rename `insp` to `jpg` and upload to Immich. Immich can get thumbnails and get camera info (EXIF). For `insv`, ffmpeg can do transcoding.

alextran1502 (approved):

jagjordi:
Is this supposed to be implemented already? I am trying to upload some files using both CLI and drag and drop in browser, however insv files are ignored.

jrasm91:
This might be a regression.

albatrosify:
Uploading the Insta360 Files work, however its not beeing displayed correctly in immich:
Screenshot

jrasm91:
Yes there are variations of 360 and panoramas and we haven't been able to add support for all of them yet.

PR #2753, opened 2023-06-13
run containers as non-root

### LINKED ISSUE #2750: [BUG] Containers should run as non-root
### The bug

Containers currently run as root, which is bad container security practise and causes problems like mentioned [here](https://github.com/immich-app/immich/issues/2740#issuecomment-1588157854).

```sh
❯ docker run -it ghcr.io/immich-app/immich-server:v1.60.0
/usr/src/app # id
uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
```

Immich is perfectly capable of running as non-root, as I do so with security contexts in my [Kubernetes cluster](https://github.com/uhthomas/automata/blob/6ec58b723bb6a34f005066e4546f25f344a54e09/k8s/unwind/immich/immich_server/deployment_list.cue#L115-L128). I believe the nginx container is already unprivileged, so it should just be a case of following the official [non-root guide](https://github.com/nodejs/docker-node/blob/1b2e8eb26c393986426ad5d88a1e190b93534906/docs/BestPractices.md#non-root-user) for node images.

N/A

### Version of Immich Server

v1.60.0

N/A

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
N/A
```

### Additional information

_No response_

uhthomas:
Fixes #2750

### DIFF
--- a/machine-learning/Dockerfile
+++ b/machine-learning/Dockerfile
@@ -1,19 +1,34 @@
 FROM python:3.11 as builder
+
+RUN adduser python
+
+USER python
+
 ENV PYTHONDONTWRITEBYTECODE=1 \
   PYTHONUNBUFFERED=1 \
   PIP_NO_CACHE_DIR=true
 
 RUN pip install --upgrade pip && pip install poetry
+
 RUN poetry config installer.max-workers 10 && \
   poetry config virtualenvs.create false
+
 RUN python -m venv /opt/venv
+
 ENV VIRTUAL_ENV="/opt/venv" PATH="/opt/venv/bin:${PATH}"
 
 COPY poetry.lock pyproject.toml ./
+
 RUN poetry install --sync --no-interaction --no-ansi --no-root --only main
 
 FROM python:3.11-slim
+
+RUN adduser python
+
+USER python
+
 WORKDIR /usr/src/app
+
 ENV NODE_ENV=production \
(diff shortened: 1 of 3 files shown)


bo0tzz:
Will this cause permission problems with the library folder for existing installations?

uhthomas:
> Will this cause permission problems with the library folder for existing installations?

I was just about to comment the same thing. I think it's possible, though I also don't think that should block this PR. We should have an advisory in the release notes for users to "re-chown" their libraries in-line with this change.

Fortunately for users where the security context is enforced, like [here](https://github.com/uhthomas/automata/blob/6ec58b723bb6a34f005066e4546f25f344a54e09/k8s/unwind/immich/immich_server/deployment_list.cue#L115-L128), it will have no effect.

jrasm91:
Outdated.

fourstepper:
> Outdated.

Sorry if this is going to sound ignorant, but based on what is this PR outdated? Was there different work done to make this work?

Thanks for the reply.

jrasm91:
It has been open for 5 months and there are merge conflicts in the dockerfile.

PR #2766, opened 2023-06-14
fix: increase request timeout from 5 minutes to 30 minutes

zackpollard:
Since node 18 the default request timeout was changed from unlimited to 5 minutes, this means large files get timed out and fail to upload

### DIFF
--- a/server/src/immich/main.ts
+++ b/server/src/immich/main.ts
@@ -28,7 +28,8 @@ export async function bootstrap() {
   useSwagger(app, isDev);
 
   await app.get(AppService).init();
-  await app.listen(port);
+  const server = await app.listen(port);
+  server.setTimeout(1800000);
 
   logger.log(`Immich Server is listening on ${port} [v${SERVER_VERSION}] [${envName}] `);
 }


uhthomas:
What is the purpose of this change? Is there an open issue to track this? I presume it's to prevent long uploads from timing out?

zackpollard:
> What is the purpose of this change? Is there an open issue to track this? I presume it's to prevent long uploads from timing out?

There is no open issue, I have only just discovered this bug and we were discussing the fix internally, turns out it was cuased by our upgrade to node 18.

server/src/immich/main.ts
```diff
@@ -28,7 +28,8 @@ export async function bootstrap() {
   useSwagger(app, isDev);
 
   await app.get(AppService).init();
-  await app.listen(port);
+  const server = await app.listen(port);
+  server.setTimeout(180000);
```
uhthomas, line 32: I believe 180000ms is 180 seconds (3 minutes)? The docs say this number is definitely ms.

https://nodejs.org/api/http.html#serversettimeoutmsecs-callback

Might be clearer to do:

```ts
server.setTimeout(30 * 1000); // 30 minutes
```

uhthomas: I understand the value is now correct, though I would still suggest implementing this specific change. Any arbitrary number like 1.8 million is hard to read and understand, as proven by the mistake of using 3 minutes earlier.

jrasm91 (approved):

server/src/immich/main.ts
zackpollard: Sure thing, not really sure how I messed up the time, was certain I copied and pasted it 😆 Anyway, yea I will make the change you suggest using the multiplication in code, the PR isn't ready yet I am just testing it, and need a PR to make it build the images 😄

uhthomas: Thanks :) Sorry to be annoying.

zackpollard: Feedback is always appreciated 😄 We are actually thinking of removing the timeout entirely, and people can configure any timeouts they would want on their reverse proxy, as it was before the node 18 upgrade.

uhthomas: Inline with https://blog.cloudflare.com/exposing-go-on-the-internet/, it's always a good idea to set timeouts.

zackpollard: Our approach with the server is that it shouldn't be exposed directly to the open internet, and instead should be behind a reverse proxy where these kind of things should be applied. If the timeout is already at 30 minutes for file uploads, then a file descriptor exhaustion attack is already very easily possible.

uhthomas: It's not infeasible that a connection could be interrupted and hang around forever, even with a reverse proxy. Having any timeout at all protects against long term resource leaks.

alextran1502 (approved):

PR #2769, opened 2023-06-15
fix: increase request timeout from 5 minutes to 30 minutes

zackpollard:
since node 18 the default request timeout was changed from unlimited to 5 minutes, causing large uploads to fail on slow connections


alextran1502 (approved):

Asher9971:
Hello,
how can i set this timeout manually?
Same problem when downloading large Files from slow remote connection.
After ~ 15 minutes the download stops.

PR #2772, opened 2023-06-15
fix(server): memory lane title

jrasm91:
- Update the API to include the requested `date` for memories. The db query is this date + 24 hours going back x years
- Update the API to include an option `years` for how many years to go back
- Update the title to correctly set `s` (`2 years since` and `1 year since`)
- Use luxon for relative date calculations
- Add unit tests
- Run db queries in parallel

### DIFF
--- a/server/immich-openapi-specs.json
+++ b/server/immich-openapi-specs.json
@@ -1539,7 +1539,7 @@
         "operationId": "getMemoryLane",
         "parameters": [
           {
-            "name": "date",
+            "name": "timestamp",
             "required": true,
             "in": "query",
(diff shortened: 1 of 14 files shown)


server/immich-openapi-specs.json
```diff
@@ -1539,10 +1539,12 @@
         "operationId": "getMemoryLane",
         "parameters": [
           {
-            "name": "timezone",
+            "name": "date",
```
uhthomas, line 1542: timestamp might be nicer? a date typically suggests a date without a time.

alextran1502 (approved):

PR #2821, opened 2023-06-16
chore(web): standardize linting

jrasm91:


uhthomas (approved):
I think I'm just going to have to trust you haven't included a crypto miner in this change.

PR #2856, opened 2023-06-19
chore(web): fade between thumbhash and thumbnail

uhthomas:
Thumbnails would previously fade in, but that behaviour was removed when thumbhashes were introduced in #2649. This change brings that back.

Demo (with cache disabled):

[Screencast from 2023-06-19 19-01-09.webm](https://github.com/immich-app/immich/assets/9749173/ba568e22-af4a-49ae-ad17-653ba5de605e)

### DIFF
--- a/web/src/lib/components/assets/thumbnail/image-thumbnail.svelte
+++ b/web/src/lib/components/assets/thumbnail/image-thumbnail.svelte
@@ -1,7 +1,8 @@
 <script lang="ts">
-	import { onMount } from 'svelte';
-	import { lazyLoad } from 'unlazy';
 	import { imageLoad } from '$lib/utils/image-load';
+	import { fade } from 'svelte/transition';
+	import { thumbHashToDataURL } from 'thumbhash';
+	import { Buffer } from 'buffer';
 
 	export let url: string;
 	export let altText: string;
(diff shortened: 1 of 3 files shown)


jrasm91 (approved):

web/src/lib/components/assets/thumbnail/image-thumbnail.svelte
```diff
@@ -1,7 +1,7 @@
 <script lang="ts">
-	import { onMount } from 'svelte';
-	import { lazyLoad } from 'unlazy';
```
jrasm91, line 3: Can we remove this dependency from package.json now?

uhthomas: Guess so, let me check

web/src/lib/components/assets/thumbnail/image-thumbnail.svelte
```diff
@@ -11,48 +11,38 @@
...
 
-	onMount(() => {
-		if (thumbhash) {
-			lazyLoad(imageElement, {
-				hash: thumbhash,
-				hashType: 'thumbhash'
-			});
-		}
-	});
+	const decodeBase64 = (s: string) => Uint8Array.from(atob(s), (c) => c.charCodeAt(0));
```
michelheusschen, line 17: Have you tested this with SSR? `atob` can't be used by Node.js if I remember correctly

web/src/lib/components/assets/thumbnail/image-thumbnail.svelte
```diff
@@ -11,48 +11,38 @@
...
+	class:rounded-full={circle}
+	class:opacity-0={!thumbhash && !complete}
+	draggable="false"
+	use:imageLoad
+	on:image-load|once={() => (complete = true)}
+/>
 
-	<!-- not everthing yet has thumbhash support so the old method is kept -->
-{:else}
+{#if thumbhash && !complete}
```
michelheusschen, line 35: Why the second img element? Can't the thumbnail be loaded in onMount and replace `src`?

web/src/lib/components/assets/thumbnail/image-thumbnail.svelte
uhthomas: I think `atob` is still in NodeJS, but considered deprecated. We had a chat in Discord about how to use `Buffer`, but couldn't get it to work. If you have any insight, I'd be grateful.

```ts
import { Buffer } from 'buffer';

Buffer.from('', 'base64');
```

web/src/lib/components/assets/thumbnail/image-thumbnail.svelte
jrasm91: I don't think the transition works when it is the same element

uhthomas: It's not possible to transition images with just a single image element. There is no control over how the underlying API handles changes to the `src` attribute.

web/src/lib/components/assets/thumbnail/image-thumbnail.svelte
uhthomas: Fixed, I just needed to rebuild the containers for it to work.

web/src/lib/components/assets/thumbnail/image-thumbnail.svelte
uhthomas: FWIW though, I do agree with you. It would be nice if svelte could just swap them dynamically or something. I initially started with something like:

```ts
const img = new Image();
img.src = url;
```

and

```svelte
{#await img.decode()}
<img thumbhash>
{:then}
<img src:{url}>
{/await}
```

In practise, it's just not very effective. It loads the target image multiple times and can get a little messy. I think this implementation is ultimately quite nice. It allows really easy compatibility for assets which don't have thumbhashes, using the old behaviour, it's quite simple and doesn't end up loading the target image twice. It also will eventually remove the thumbhash from the DOM entirely, and so the result is just one image anyway.

PR #2876, opened 2023-06-20
fix(server): use HTTP status OK instead of CREATED for assets

uhthomas:
The NGINX gzip module does not compress responses with a status of 201, which is
a major issue specifically for the /api/asset/time-bucket endpoint where
responses can be upwards of 5Mi. The size of the response is dramatically
reduced with gzip to 500Ki in some cases.

https://trac.nginx.org/nginx/ticket/471
https://trac.nginx.org/nginx/ticket/394

The signature of these endpoints should be GET rather than POST anyway, but that
is a bigger discussion.

### DIFF
--- a/server/src/immich/api-v1/asset/asset.controller.ts
+++ b/server/src/immich/api-v1/asset/asset.controller.ts
@@ -7,6 +7,7 @@ import {
   Header,
   Headers,
   HttpCode,
+  HttpStatus,
   Param,
   ParseFilePipe,
   Patch,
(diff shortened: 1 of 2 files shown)


server/src/immich/api-v1/asset/asset.controller.ts
```diff
@@ -71,7 +72,7 @@ interface UploadFiles {
 @Controller('asset')
 @Authenticated()
 export class AssetController {
-  constructor(private assetService: AssetService) {}
+  constructor(private assetService: AssetService) { }
```
uhthomas, line 75: idk?

jrasm91 (approved):

server/src/immich/api-v1/asset/asset.controller.ts
jrasm91: That is a linting error.

PR #2879, opened 2023-06-20
refactor(server, web): create shared link

jrasm91:


web/src/lib/components/shared-components/create-share-link-modal/create-shared-link-modal.svelte
```diff
@@ -248,7 +233,7 @@
 	<hr />
 
 	<section class="m-6">
-		{#if !isShowSharedLink}
+		{#if !sharedLink}
```
uhthomas, line 236: Would a flattened chain like this make more sense?

```svelte
{#if sharedLink}
{:else if editingLink}
{:else}
{/if}
```

jrasm91: Maybe. I think there should be separate components for create vs edit myself.

alextran1502 (approved):

PR #2896, opened 2023-06-21
feat: M2TS

uhthomas:
Support the Blu-ray disc Audio-Video (BDAV) MPEG-2 Transport Stream (M2TS) format.

https://en.wikipedia.org/wiki/.m2ts

image

Fixes: #2350


jrasm91 (approved):
You should fix content-type checking already so you don't have to keep adding new types in three places :-)

uhthomas:
> You should fix content-type checking already so you don't have to keep adding new types in three places :-)

:p

PR #2897, opened 2023-06-21
fix(web): keep video volume

### LINKED ISSUE #2885: [BUG] video player loses the set volume on the next video
### The bug

When I'm playing a video on the Web, and then use mouse click or arrow keys to go to the next one, it loses the audio volume slider's position. I believe it should persist.

Kubernetes (k3s)

### Version of Immich Server

v1.62.1

not relevant

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

```bash
1. Choose any video
2. Set it's volume to something else than 100%
3. Go to the next video
4. See that it has not kept the slider posiition.
```

### Additional information

_No response_

faupau03:
save video volume in asset-interaction.store.ts, fixes #2885
google saves the video volume in localstorage, which i think would be the better implementation?

`yt-player-volume:"{"data":"{\"volume\":37,\"muted\":false}","expiration":1689945428802,"creation":1687353428802}"`

### DIFF
--- a/web/src/lib/components/asset-viewer/video-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/video-viewer.svelte
@@ -2,7 +2,7 @@
 	import { api } from '@api';
 	import { fade } from 'svelte/transition';
 	import { createEventDispatcher } from 'svelte';
-	import { videoViewerVolume } from '$lib/stores/asset-interaction.store';
+	import { videoViewerVolume } from '$lib/stores/preferences.store';
 	import LoadingSpinner from '../shared-components/loading-spinner.svelte';
 
 	export let assetId: string;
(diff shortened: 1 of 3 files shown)


web/src/lib/components/asset-viewer/video-viewer.svelte
```diff
@@ -30,6 +39,7 @@
 		class="h-full object-contain"
 		on:canplay={handleCanPlay}
 		on:ended={() => dispatch('onVideoEnded')}
+		bind:volume
```
jrasm91, line 42: You can bind to the store directly by using $videoViewerVolume

jrasm91 (commented):
We have some stores that write to local storage maybe you could look at one of those

jrasm91 (approved):

PR #2898, opened 2023-06-21
feature(server): Add endpoint that returns bearer token from api key

etnoy:
Adds a /auth/keyLogin endpoint that allows a cli client to get a bearer token (and user data) by logging in with an api key


etnoy:
Closing this since bearer tokens are long lived anyway, CLI can use api keys for now

PR #2922, opened 2023-06-23
chore(server): add Canon CR3 mime type

416c616e:
Allows for successful upload for Canon CR3 format as it reports `image/cr3`

I will note that full support for all Canon cameras will depend on ImageMagick support, for example my Canon EOS RP does work, but my more recent Canon EOS R5C does not work as of yet.

### DIFF
--- a/server/src/domain/domain.constant.ts
+++ b/server/src/domain/domain.constant.ts
@@ -30,6 +30,7 @@ export function assertMachineLearningEnabled() {
 
 const validMimeTypes = [
   'image/avif',
+  'image/cr3',
   'image/gif',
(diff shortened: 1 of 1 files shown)


alextran1502 (approved):

elliotclee:
This needs to be addressed on a more systematic level. The basic problem is that Windows' registry has the mime types for all the RAW filetypes as image/cr3, image/k25, image/raf, etc. and when someone uploads from Windows it sends those mime types. However, other systems still use image/x-canon-cr3 et al. I have a patch that I have not committed that takes care of all of them.

PR #2938, opened 2023-06-24
fix(deps): install poetry in pump workflow

mertalev:
This fix makes version pumping apply to the ML `pyproject.toml`. The pump script itself was updated with the inclusion of Poetry, but Poetry wasn't installed in the pump workflow.


jrasm91 (approved):
Nice catch. Can you put new lines between each step?

mertalev:
Sure thing!

PR #2945, opened 2023-06-25
Allow Windows' version of the MIME types for raw photos.

elliotclee:
On Windows, the registry sets MIME types for raw images that are different than the ones used on other operating systems, which are then used by the browser when uploading. This patch accepts both variants for upload.


server/src/domain/domain.constant.ts
```diff
@@ -67,6 +67,33 @@ const validMimeTypes = [
   'image/x-sony-arw',
   'image/x-sony-sr2',
   'image/x-sony-srf',
+  'image/dng',
```
uhthomas, line 70: This is a duplicate entry I believe.

Can we also sort this list alphabetically?

server/src/immich/config/asset-upload.config.spec.ts
```diff
@@ -87,6 +87,34 @@ describe('assetUploadOption', () => {
       { mimetype: 'image/x-sony-arw', extension: 'arw' },
       { mimetype: 'image/x-sony-sr2', extension: 'sr2' },
       { mimetype: 'image/x-sony-srf', extension: 'srf' },
+
+      { mimetype: 'image/dng', extension: 'dng' },
```
uhthomas, line 91: can we also sort this list?

web/src/lib/utils/asset-utils.spec.ts
```diff
@@ -98,6 +98,35 @@ describe('get file mime type', () => {
 		{ mimetype: 'image/x-sony-arw', extension: 'arw' },
 		{ mimetype: 'image/x-sony-sr2', extension: 'sr2' },
 		{ mimetype: 'image/x-sony-srf', extension: 'srf' },
+		/*** The following MIME types are allowed for upload but not returned by getFileMimeType() ***
```
uhthomas, line 101: No need to have this here. Will be quite burdensome to maintain correctly.

server/src/domain/domain.constant.ts
elliotclee: Fixed on both counts.

server/src/immich/config/asset-upload.config.spec.ts
elliotclee: Done.

web/src/lib/utils/asset-utils.spec.ts
elliotclee: Fixed

alextran1502 (approved):

server/src/domain/domain.constant.ts
uhthomas: Am I missing something? Nothing changed?

elliotclee: You're not missing something, I was. I forgot to push the updates from my local tree to github. Since Alex already pushed this to main, do I need to open a separate pull request to fix your concerns, or is there a way to re-open a pull request?

uhthomas: You will need to open a follow-up PR. Thank you 🙂

PR #2954, opened 2023-06-25
fix(web): Add m: to search query upon loading results.

aero31aero:
Previously, we'd drop the m: from non-clip searches entirely. This behavior incorrectly represents the page's status (results from non-clip search but query implies a clip search). Also, any follow-up searches change to clip searches, which feels like a jarring UX if you have to add m: every time in a 'search-session'.

### DIFF
--- a/web/src/routes/(user)/search/+page.svelte
+++ b/web/src/routes/(user)/search/+page.svelte
@@ -43,7 +43,14 @@
 		}
 	});
 
-	$: term = $page.url.searchParams.get('q') || data.term || '';
+	$: term = (() => {
+		let term = $page.url.searchParams.get('q') || data.term || '';
+		const isMetadataSearch = $page.url.searchParams.get('clip') === 'false';
+		if (isMetadataSearch && term !== '') {
+			term = `m:${term}`;
+		}
+		return term;
+	})();
 
 	let selectedAssets: Set<AssetResponseDto> = new Set();
 	$: isMultiSelectionMode = selectedAssets.size > 0;


web/src/routes/(user)/search/+page.svelte
```diff
@@ -43,7 +43,8 @@
 		}
 	});
 
-	$: term = $page.url.searchParams.get('q') || data.term || '';
+	$: clip_status = $page.url.searchParams.get('clip') !== 'false';
+	$: term = `${clip_status ? '' : 'm:'}${$page.url.searchParams.get('q')}` || data.term || '';
```
uhthomas, line 47: What happens if you get a query like `/search?clip=true`? The search bar will just look like `m:`. Is this what we want?

aero31aero: It'd stay empty. But yeah, `?clip=false` would populate an empty `m:` .

uhthomas: That doesn't seem to pass validation.

image

aero31aero: Which means this edgecase input is already not possible. It isn't possible to send an input such that the resulting page would have `clip=false` and an empty query.

uhthomas: If I were to navigate to `https://example.com/search?clip=false` then it would produce `m:` which isn't valid,

aero31aero: Gotcha. Fixed.

web/src/routes/(user)/search/+page.svelte
```diff
@@ -44,6 +44,8 @@
 	});
 
 	$: term = $page.url.searchParams.get('q') || data.term || '';
+	$: is_metadata_search = $page.url.searchParams.get('clip') === 'false';
```
alextran1502, line 47: Can you use camelCase for the variable?

web/src/routes/(user)/search/+page.svelte
```diff
@@ -44,6 +44,8 @@
...
+	$: term = `${!is_metadata_search && term !== '' ? 'm:' : ''}${term}`;
```
alextran1502, line 48: Here is the error I got when trying to run this 
image

aero31aero: I see. This is my first time with reactive statements; I just figured it was a false warning since the typescript compiled properly. I just switched to an anonymous function with better legibility too.

alextran1502 (approved):

PR #2990, opened 2023-06-27
chore(server): check file extension for XMP instead of mimetype

alex-phillips:
This fixes an error server-side where a mimetype check for an XMP sidecar returns `false`. When uploading, we manually specify the mime-type in the CLI as XML, but the mime-type package server-side during the import process does not detect this. Currently, importing files with sidecars fails and is broken.

### DIFF
--- a/server/src/immich/api-v1/asset/asset.service.ts
+++ b/server/src/immich/api-v1/asset/asset.service.ts
@@ -6,7 +6,6 @@ import {
   ICryptoRepository,
   IJobRepository,
   ImmichReadStream,
-  isSidecarFileType,
   isSupportedFileType,
   IStorageRepository,
   JobName,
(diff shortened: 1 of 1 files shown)


server/src/immich/api-v1/asset/asset.service.ts
```diff
@@ -149,9 +149,8 @@ export class AssetService {
     }
 
     if (dto.sidecarPath) {
-      const sidecarType = mime.lookup(dto.sidecarPath) as string;
-      if (!isSidecarFileType(sidecarType)) {
-        throw new BadRequestException(`Unsupported sidecar file type ${assetPathType}`);
+      if (path.extname(dto.sidecarPath).toLowerCase() !== ".xmp") {
```
jrasm91, line 152: single quotes yo

jrasm91 (approved):

uhthomas:
This change looks fine, but why is this change being made? A description on the PR or a linked issue would be hugely helpful!

alex-phillips:
> This change looks fine, but why is this change being made? A description on the PR or a linked issue would be hugely helpful!

Apologies! Updated my initial comment with a description.

PR #3001, opened 2023-06-28
fix(server): add missing avi mime types and add tests

### LINKED ISSUE #2975: should support video/avi mime type
### Discussed in https://github.com/immich-app/immich/discussions/2964

<sup>Originally posted by **mkjustuk** June 26, 2023</sup>
### The feature

I saw a previous issue that looked like it had covered this, but I think it only covered a different type of .avi files.
The video type video/avi is not recognised and would be great if it could be added.

I get this in the logs on server:

[Nest] 1  - 06/26/2023, 12:23:36 PM   ERROR [AssetUploadConfig] Unsupported file type .AVI file MIME type video/avi

Happy to send a sample vid etc if needed.

### Platform

- [X] Server
- [ ] Web
- [ ] Mobile</div>

uhthomas:
See https://github.com/immich-app/immich/pull/2952#pullrequestreview-1497194041

Fixes: #2975


server/src/domain/domain.constant.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { validMimeTypes } from './domain.constant';
+
+describe('valid mime types', () => {
+  it('should be a sorted list', () => {
+    expect(validMimeTypes).toEqual(validMimeTypes.sort());
```
jrasm91, line 5: Doesn't sort mutate the array?

jrasm91 (approved):

PR #3003, opened 2023-06-28
fix(web): aspect ratio for photos with Rotate 270 CW orientation

brighteyed:
We incorrectly determined the aspect ratio for the photos with `Rotate 270 CW` orientation. This issue is mentioned in [this](https://github.com/immich-app/immich/discussions/3000) discussion. This PR fixes it.

### DIFF
--- a/web/src/lib/utils/asset-utils.ts
+++ b/web/src/lib/utils/asset-utils.ts
@@ -189,7 +189,7 @@ export function getAssetRatio(asset: AssetResponseDto) {
 	let width = asset.exifInfo?.exifImageWidth || 235;
 	const orientation = Number(asset.exifInfo?.orientation);
 	if (orientation) {
-		if (orientation == 6 || orientation == -90) {
+		if (orientation == 6 || orientation == -90 || orientation == 8) {
 			[width, height] = [height, width];
 		}
 	}
--- a/web/src/lib/utils/asset-utils.ts
+++ b/web/src/lib/utils/asset-utils.ts
@@ -189,6 +189,7 @@ export function getAssetRatio(asset: AssetResponseDto) {
 	let width = asset.exifInfo?.exifImageWidth || 235;
 	const orientation = Number(asset.exifInfo?.orientation);
 	if (orientation) {
+		// 6 - Rotate 90 CW, 8 - Rotate 270 CW
 		if (orientation == 6 || orientation == 8) {
 			[width, height] = [height, width];
 		}
--- a/web/src/lib/utils/asset-utils.ts
+++ b/web/src/lib/utils/asset-utils.ts
@@ -189,7 +189,7 @@ export function getAssetRatio(asset: AssetResponseDto) {
 	let width = asset.exifInfo?.exifImageWidth || 235;
 	const orientation = Number(asset.exifInfo?.orientation);
 	if (orientation) {
-		if (orientation == 6 || orientation == -90 || orientation == 8) {
+		if (orientation == 6 || orientation == 8) {
 			[width, height] = [height, width];
 		}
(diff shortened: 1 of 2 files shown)


web/src/lib/utils/asset-utils.ts
```diff
@@ -189,7 +189,7 @@ export function getAssetRatio(asset: AssetResponseDto) {
 	let width = asset.exifInfo?.exifImageWidth || 235;
 	const orientation = Number(asset.exifInfo?.orientation);
 	if (orientation) {
-		if (orientation == 6 || orientation == -90) {
+		if (orientation == 6 || orientation == -90 || orientation == 8) {
```
uhthomas, line 192: nit: should we have constants for `6` and `8`? I don't think those values intuitively mean anything to anyone.

uhthomas, line 192: Would the intention of this snippet be clearer?

```suggestion
		if ([-90, 6, 8].includes(orientation)) {
```

brighteyed: I would leave digits here. The EXIF tag values for orientation (1-8) are often referred by just numbers

brighteyed: I'd prefer to leave it as is. IMO, the suggestion seems a bit redundant

jrasm91: Personally I think the rotation data should be standardized during exif extraction and then we could reference a single option/value here. Doing it like this means the mobile app and the web and every other consuming client needs to handle all these edge cases and inconsistencies over and over again.

brighteyed:
Thank you for your thoughts @uhthomas and @jrasm91. Based on them, I pushed additional changes into this PR:
  * `exiftool-vendored` disables the print conversion for the `Orientation` tag. I believe that guarantees we can get only `1-8` values for the `orientation`. So I safely removed legacy (?) code in `thumbnail.svelte` that had been checking for `Rotate 90 CW` and `Horizontal (normal)` values
  * I removed the `-90` value check in `getAssetRatio` assuming we can get only `1-8` here
  * Added comment to `6` and `8`

I guess the use of the orientation tag should be pretty limited in our code. Now we use it only for the calculation of the justified layout.

web/src/lib/utils/asset-utils.ts
```diff
@@ -189,7 +189,8 @@ export function getAssetRatio(asset: AssetResponseDto) {
 	let width = asset.exifInfo?.exifImageWidth || 235;
 	const orientation = Number(asset.exifInfo?.orientation);
 	if (orientation) {
-		if (orientation == 6 || orientation == -90) {
+		// 6 - Rotate 90 CW, 8 - Rotate 270 CW
+		if (orientation == 6 || orientation == 8) {
```
jrasm91, line 193: Sweet. I agree with @uhthomas that a variable name would be better than a comment + magic number. What about enums? Something like

```typescript
enum Rotation {
  CLOCKWISE_90=6,
  CLOCKWISE_270=8,
}
```

Don't necessarily have to do it here, but we could even change the type in the exif info entity and have the response dto be an enum. Then all the clients would know what the values are and mean.

alextran1502 (approved):

web/src/lib/utils/asset-utils.ts
brighteyed: Ah, `ffprobe` returns other values for the orientation (-90, 90, etc.). #3023 fixes it.

And we definitely have to refactor `ExifEntity` and `ExifResponseDto` to change the type of the orientation field to enum. Created an issue #3022 for that.

PR #3023, opened 2023-06-29
fix(web): aspect ratio for videos with 90 CW and 270 CW orientation

brighteyed:
This is a follow up PR for #3003.

#3003 fixed the aspect ratio calculation for photos. Since `ffprobe` returns other values for the orientation tag (`-90`, `90`, etc.), `getAspectRatio` should handle both `exiftool-vendored` range (`1-8`) and `ffprobe` values (`-90`, `90`, `180`, etc.)


alextran1502 (approved):

jrasm91:
Another option is to detect a rotated image (+/- 90) and flip image height/width accordingly in the exif info. Then everything could be used as is.

PR #3027, opened 2023-06-29
fix(web): Share link multi-select download icon showing when not available #3006

### LINKED ISSUE #3006: [BUG] Share link multi-select download icon showing when not available
### The bug

The download icon shows up when multi-selecting assets in a shared link, even when downloading is disabled.

N/A

### Version of Immich Server

v1.64.0

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1. Create a shared link with download disabled
2. Open a shared link in an incognito windows
3. Multi-select a few photos
4. Verify the download icon shows up (unexpected)
5. Clicking the download icon results in an error (expected)

Expected: the download icon should not show up when multi-selecting assets
Actual: icon shows up, but doesn't work (results in server error)
```

### Additional information

_No response_

faupau03:
only show download button if allowDownload
add SelectAll to individual share

fixes #3006

### DIFF
--- a/web/src/lib/components/album-page/album-viewer.svelte
+++ b/web/src/lib/components/album-page/album-viewer.svelte
@@ -351,7 +351,7 @@
 			clearSelect={() => (multiSelectAsset = new Set())}
 		>
 			<CircleIconButton title="Select all" logo={SelectAll} on:click={handleSelectAll} />
-			{#if sharedLink?.allowDownload}
+			{#if sharedLink?.allowDownload || !isPublicShared}
 				<DownloadAction filename={album.albumName} sharedLinkKey={sharedLink?.key} />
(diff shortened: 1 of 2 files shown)


jrasm91 (commented):
This is the album viewer component, which is also used for shared albums. I believe the condition should be "no shared link or shared link and shared link allows download". Can you verify you can download a shared album with your changes?

faupau03:
> 

fixed to allow download album if not isPublicShared

confirmed download works on normal album, link shared album and user shared album

jrasm91 (approved):

PR #3074, opened 2023-07-02
fix(web): use native image decoder

### LINKED ISSUE #3073: thumbnails don't load sometimes
### The bug

There's no obvious cause, but it seems that thumbnails won't load occasionally. I found I can make it happen more often by using the scrollbar. Seems like a race condition somewhere?

image

N/A

### Version of Immich Server

v1.65.0

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
N/A
```

### Additional information

_No response_

uhthomas:
The decode function should be more reliable than event listeners and doesn't need additional logic for destruction. The semantics of decode are also different to the 'load' event listener. 'decode' will wait until the image has been fully decoded and can be rendered, unlike 'load' which will trigger before the image has actually been decoded.

Fixes: #3073


uhthomas:
This is probably still a nice thing to have, as the code is simpler and less likely to be prone to failures, but unfortunately does not resolve the linked issue.

image

jrasm91 (approved):

jrasm91 (approved):

martabal (approved):

PR #3106, opened 2023-07-04
feat(web): set asset as profile picture

faupau03:
Adds the ability to set the profile picture in web.
grafik

It even includes the zoom functionality from the asset viewer ;)


web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,75 @@
...
+  import Button from '../elements/buttons/button.svelte';
+  const dispatch = createEventDispatcher();
+  export let asset: AssetResponseDto;
+  export let publicSharedKey = '';
+  import domtoimage from 'dom-to-image';
+  import type { UserResponseDto } from '@api';
+  import { notificationController, NotificationType } from './notification/notification';
+  import { lastUpdatedProfilePicture } from '$lib/stores/preferences.store';
+  onMount(() => {
+    console.log(asset);
```
uhthomas, line 17: remove

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+  let user: UserResponseDto;
+
+  const handleSetProfilePicture = async () => {
+    const div: HTMLElement | null = document.getElementById('profile-picture')?.childNodes[0] as HTMLElement;
```
uhthomas, line 23: Use svelte binding instead?

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,75 @@
...
+    if (!div) {
+      return;
+    }
+    domtoimage.toBlob(div).then(function (blob: Blob) {
```
uhthomas, line 27: Prefer async code.

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,75 @@
...
+      const file: File = new File([blob], 'profile-picture.png', { type: 'image/png' });
+      try {
+        api.userApi.createProfileImage({ file }).then((res) => {
+          console.log(res);
```
uhthomas, line 31: Remove

web/src/lib/components/shared-components/user-avatar.svelte
```diff
@@ -15,6 +16,14 @@
   export let showTitle = true;
   export let autoColor = false;
   let showFallback = true;
+  let appendix = '';
+
+  lastUpdatedProfilePicture.subscribe((value) => {
+    console.log('isUpdateProfilePicture', value);
```
uhthomas, line 22: remove

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,75 @@
...
+    <span class="flex gap-2 place-items-center">
+      <p class="font-medium">Set profile picture</p>
+    </span>
+  </svelte:fragment>
+  <div class="flex justify-center place-items-center items-center">
+    <div
+      id="profile-picture"
+      class="w-1/2 aspect-square rounded-full overflow-hidden relative flex border-immich-primary border-4"
+    >
+      <PhotoViewer {publicSharedKey} {asset} on:close={() => console.log('close')} />
```
uhthomas, line 64: Remove the console log?

web/src/lib/stores/preferences.store.ts
```diff
@@ -46,3 +46,5 @@ export interface AlbumViewSettings {
 export const albumViewSettings = persisted<AlbumViewSettings>('album-view-settings', {
   sortBy: 'Most recent photo',
 });
+
+export const lastUpdatedProfilePicture = persisted<number>('last-updated-profile-picture', 0, {});
```
alextran1502, line 50: What is the purpose of this store?

jrasm91: I think to bypass the cache for the profile picture, although I'm sure there is a better way.

faupau03: The profile picture is loaded via an static url and by that gets cached by the browser. When ever the profile picture gets changed in the web, the store gets updated and accordingly the static url. Resulting in an updated profile picture.

uhthomas: Just add the 'no-cache' cache-control header to the response. The browser will still cache the asset data, but will revalidate the request against the etag. Saves some code and is what we intend to do for all the API endpoints at some point.

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -32,6 +33,7 @@
   let isShowDeleteConfirmation = false;
   let addToSharedAlbum = true;
   let shouldPlayMotionPhoto = false;
+  let isShowProfileImageCrop = false;
```
uhthomas, line 36: I wonder if there's a better way to handle this. A Boolean for each modal seems verbose and strange.

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,61 @@
...
+  export let publicSharedKey = '';
+  import domtoimage from 'dom-to-image';
+  import { notificationController, NotificationType } from './notification/notification';
+  import { lastUpdatedProfilePicture } from '$lib/stores/preferences.store';
+
+  let profilePicture: HTMLDivElement;
+
+  const handleSetProfilePicture = async () => {
+    const div = profilePicture.childNodes[0] as HTMLDivElement;
+    const blob: Blob = await domtoimage.toBlob(div);
```
uhthomas, line 18: are these type annotations really needed?

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,61 @@
...
+    <div
+      bind:this={profilePicture}
+      class="w-1/2 aspect-square rounded-full overflow-hidden relative flex border-immich-primary border-4"
+    >
+      <PhotoViewer {publicSharedKey} {asset} />
+    </div>
+  </div>
+  <span class="p-4 flex justify-end">
+    <Button on:click={handleSetProfilePicture}>
+      <p class="">Set as profile picture</p>
```
uhthomas, line 56: remove the class attribute?

web/src/lib/components/shared-components/user-avatar.svelte
```diff
@@ -15,6 +16,13 @@
   export let showTitle = true;
   export let autoColor = false;
   let showFallback = true;
+  let appendix = '';
```
uhthomas, line 19: this should really be named 'query', though given my earlier comment about cache control this code will probably be removed anyway

web/src/lib/components/asset-viewer/asset-viewer.svelte
faupau03: Well i think its out of scope of this pr to change that, maybe in a future pr.

server/src/domain/user/user.service.ts
```diff
@@ -120,6 +121,14 @@ export class UserService {
     return this.userCore.getUserProfileImage(user);
   }
 
+  async getUserProfileImageHash(userId: string): Promise<string> {
```
uhthomas, line 124: This shouldn't be necessary. I believe a weak etag is automatically generated.

uhthomas (changes requested):

server/src/domain/user/user.service.ts
faupau03: thanks, i realized that a moment ago too. I will revert the change and just set cache-control to nocache

web/package.json
```diff
@@ -58,10 +58,12 @@
 	},
 	"type": "module",
 	"dependencies": {
+		"@types/dom-to-image": "^2.6.4",
```
alextran1502, line 61: This should be in dev-dependency

server/src/infra/entities/user.entity.ts
```diff
@@ -42,6 +42,9 @@ export class UserEntity {
   @Column({ default: '' })
   profileImagePath!: string;
 
+  @Column({ default: '' })
```
alextran1502: You will have to generate a database migration for the newly added column

alextran1502: Please see https://immich.app/docs/developer/database-migrations

server/src/domain/user/user.service.ts
alextran1502: Can you add tests for this service method?

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,59 @@
...
+    try {
+      await api.userApi.createProfileImage({ file });
+      dispatch('close');
+      notificationController.show({
+        type: NotificationType.Info,
+        message: 'Profile picture set.',
+        timeout: 3000,
+      });
+    } catch (err) {
+      console.error('Error [profile-image-cropper]:', err);
```
jrasm91, line 28: Use handleError instead

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,59 @@
...
+      notificationController.show({
+        type: NotificationType.Error,
+        message: 'Error setting profile picture.',
+        timeout: 3000,
+      });
+    }
+  };
+</script>
+
+<BaseModal on:close={() => dispatch('close')}>
```
jrasm91, line 38: Can you use bind:close instead?

faupau03: I just copied from album-selection-model.svelte. How would bind:close be used?

server/src/infra/entities/user.entity.ts
faupau03: I reverted the changes regarding the user entity, as it is not needed. I think the database migration is not needed then.

server/src/domain/user/user.service.ts
faupau03: I reverted the changes regarding profileImageHash, as its not needed. I think a test is not needed then.

web/package.json
```diff
@@ -1,77 +1,79 @@
 {
```
jrasm91, line 1: Please revert this and only add your dependency as a change.

web/src/lib/components/shared-components/profile-image-cropper.svelte
jrasm91: Sorry, I was thinking of `on:click` which I believe is shorthand for this. If the implementation just emits the event with the same name again. Does that work here?

jrasm91 (approved):

web/package.json
```diff
@@ -74,4 +76,4 @@
 		"svelte-material-icons": "^3.0.4",
 		"thumbhash": "^0.1.1"
 	}
-}
```
uhthomas, line 77: nit: a new line here would be nice

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,55 @@
+<script lang="ts">
+  import { AssetResponseDto, api } from '@api';
+  import { createEventDispatcher } from 'svelte';
+  import PhotoViewer from '../asset-viewer/photo-viewer.svelte';
+  import BaseModal from './base-modal.svelte';
+  import Button from '../elements/buttons/button.svelte';
+  const dispatch = createEventDispatcher();
```
uhthomas, line 7: I don't like mixed imports and general statements. There should be a big block of imports and then the application code.

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,55 @@
...
+
+  let profilePicture: HTMLDivElement;
+
+  const handleSetProfilePicture = async () => {
+    const div = profilePicture.childNodes[0];
+    const blob = await domtoimage.toBlob(div);
+    const file = new File([blob], 'profile-picture.png', { type: 'image/png' });
+    try {
+      await api.userApi.createProfileImage({ file });
+      dispatch('close');
```
uhthomas, line 22: Should this also be dispatched when an error is thrown?

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,55 @@
...
+  export let asset: AssetResponseDto;
+  export let publicSharedKey = '';
+  import domtoimage from 'dom-to-image';
+  import { notificationController, NotificationType } from './notification/notification';
+  import { handleError } from '$lib/utils/handle-error';
+
+  let profilePicture: HTMLDivElement;
+
+  const handleSetProfilePicture = async () => {
+    const div = profilePicture.childNodes[0];
```
uhthomas, line 17: Is there a better way to target the node? Can it not be directly bound?

web/src/lib/components/shared-components/profile-image-cropper.svelte
faupau03: I think when `await api.userApi.createProfileImage({file})` throws an error, the execution will directly go into the catch block and not dispatch the close event. Is that not the case?

web/src/lib/components/shared-components/profile-image-cropper.svelte
faupau03: The problem with directly binding the PhotoViewer component is, that it is a Svelte Component and thus does not expose the DOM Node needed.

uhthomas: In which case, can you bind the element in the PhotoViewer and export it?

web/src/lib/components/shared-components/profile-image-cropper.svelte
uhthomas: I believe so, I was asking if it should be dispatched even in the face of an error. I suppose not?

alextran1502:
For some portrait photos, I saw a half circle blocking the profile picture positioning 
image

web/src/lib/components/shared-components/profile-image-cropper.svelte
faupau03: I tried to bind the imgElement, which already is a variable in photo-viewer.svelte. But now i have the problem, that the imgElement binding has to passed through the BaseModal component. Could you give some advice on how to achieve this?

faupau03:
> For some portrait photos, I saw a half circle blocking the profile picture positioning image

Hi thanks for the feedback. I did not encounter this behaivour before. Would you be able to provide the asset you used? Thx

alextran1502:
IMG_0601
Here you go

web/src/lib/components/shared-components/profile-image-cropper.svelte
uhthomas: Why would it need to be passed through the BaseModal component, sorry?

web/src/lib/components/asset-viewer/photo-viewer.svelte
```diff
@@ -9,13 +9,15 @@
...
 
   let assetData: string;
 
   let copyImageToClipboard: (src: string) => Promise<Blob>;
   let canCopyImagesToClipboard: () => boolean;
 
+  $: console.log(imgElement);
+
   onMount(async () => {
     // Import hack :( see https://github.com/vadimkorr/svelte-carousel/issues/27#issuecomment-851022295
```
uhthomas, line 22: This is funny and can probably be removed given the web uses ESM.

web/src/lib/components/shared-components/profile-image-cropper.svelte
faupau03: The imgElement has the correct value, when logged in photo-viewer.svelte. Logging the imgElement variable in profile-image-cropper.svelte results in an undefined object.
grafik

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,58 @@
...
+
+<BaseModal on:close>
+  <svelte:fragment slot="title">
+    <span class="flex gap-2 place-items-center">
+      <p class="font-medium">Set profile picture</p>
+    </span>
+  </svelte:fragment>
+  <div class="flex justify-center place-items-center items-center">
+    <div class="w-1/2 aspect-square rounded-full overflow-hidden relative flex border-immich-primary border-4">
+      <PhotoViewer {imgElement} {publicSharedKey} {asset} />
```
uhthomas, line 49: I think you have to `bind:imgElement={imgElememt}`.

faupau03: Thanks for help, got it working now.

web/src/lib/components/asset-viewer/photo-viewer.svelte
faupau03: removed ;)

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,51 @@
...
+        timeout: 3000,
+      });
+    } catch (err) {
+      handleError(err, 'Error setting profile picture.');
+    }
+    dispatch('close');
+  };
+</script>
+
+<BaseModal on:close>
```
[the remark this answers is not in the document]
faupau03: it forwards the close event to the BaseModal component

uhthomas (approved):
Assuming the issue @alextran1502 raised is fixed, this looks good. Thanks!

faupau03:
its not fixed yet. I think i have to dig further into the issue. For now, if the asset is not fittet into the profile picture by zooming in, it shows a blue background in the avatar.
grafik

faupau03:
fixed now. When the asset is not zoomed in enough or shows other transparent areas, it shows an error message.

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,85 @@
...
+  import Button from '../elements/buttons/button.svelte';
+
+  export let asset: AssetResponseDto;
+  export let publicSharedKey = '';
+
+  const dispatch = createEventDispatcher();
+  let imgElement: HTMLDivElement;
+
+  const hasTransparentPixels = async (blob: Blob) => {
+    const img = new Image();
```
uhthomas, line 18: use `await img.decode`.

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,85 @@
...
+    img.src = URL.createObjectURL(blob);
+    return new Promise((resolve) => {
+      img.onload = function () {
+        const canvas = document.createElement('canvas');
+        canvas.width = img.width;
+        canvas.height = img.height;
+        const ctx = canvas.getContext('2d');
+        ctx?.drawImage(img, 0, 0);
```
uhthomas, line 26:
```suggestion
        ctx.drawImage(img, 0, 0);
```

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,85 @@
...
+        const imageData = ctx?.getImageData(0, 0, canvas.width, canvas.height);
```
uhthomas, line 27:
```suggestion
        const imageData = ctx.getImageData(0, 0, canvas.width, canvas.height);
```

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,85 @@
...
+    });
+  };
+
+  const handleSetProfilePicture = async () => {
+    try {
+      const blob = await domtoimage.toBlob(imgElement);
+      if (await hasTransparentPixels(blob)) {
+        notificationController.show({
+          type: NotificationType.Error,
+          message: 'Profile pictures cannot have transparent pixels.',
```
uhthomas, line 49: why?

faupau03: I think transparent areas, caused by bad positioning, are not really pleasing in a profile picture. As i havent found an easy way to restrict the profile picture selection the the image area itself, thats what i came up with for now. But it might not be necessary.

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,91 @@
...
+
+  const dispatch = createEventDispatcher();
+  let imgElement: HTMLDivElement;
+
+  const hasTransparentPixels = async (blob: Blob) => {
+    const img = new Image();
+    img.src = URL.createObjectURL(blob);
+    await img.decode();
+    return new Promise((resolve) => {
+      img.onload = function () {
```
uhthomas, line 22: This promise and callback are not needed with `img.decode()`.

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,91 @@
...
+        const canvas = document.createElement('canvas');
+        canvas.width = img.width;
+        canvas.height = img.height;
+        const ctx = canvas.getContext('2d');
+        if (!ctx) {
```
uhthomas, line 27: is this ever possible? I guess maybe for SSR?

web/src/lib/components/shared-components/profile-image-cropper.svelte
faupau03: thanks, sry, fixed now

web/src/lib/components/shared-components/profile-image-cropper.svelte
faupau03: The svelte check throws an error, as it could be possibly null. I just added this as a fix.

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -0,0 +1,86 @@
...
+  <svelte:fragment slot="title">
+    <span class="flex gap-2 place-items-center">
+      <p class="font-medium">Set profile picture</p>
+    </span>
+  </svelte:fragment>
+  <div class="flex justify-center place-items-center items-center">
+    <div
+      class="w-1/2 aspect-square rounded-full overflow-hidden relative flex border-immich-primary border-4 bg-immich-"
+    >
+      <PhotoViewer bind:element={imgElement} {publicSharedKey} {asset} />
```
jrasm91, line 77: publicSharedKey can be omitted here. Setting a profile picture from a shared link isn't a valid scenario.

neozhu:
@faupau03 
Is it possible to add a feature where new users can upload their own photos as profile pictures, and then add a functionality to find all photos that are similar to this profile picture? I think this feature would be useful.

faupau03:
> @faupau03 Is it possible to add a feature where new users can upload their own photos as profile pictures, and then add a functionality to find all photos that are similar to this profile picture? I think this feature would be useful.

I think thats out of scope for this pr. Feel free to open a feature request.

PR #3107, opened 2023-07-04
fix(web): previous/next asset navigation

brighteyed:
This PR fixes the next/previous navigation in the `AssetViewer` including navigation across not loaded buckets

Example of the bug:
1. Open the main timeline (the first N bucket(s) are loaded now)
2. Click on the scrollbar somewhere near the bottom (some bucket(s) near that location are loaded. Note: `_assetGridState.assets` now contains assets from nonadjacent buckets)
3. Click on any asset to open `AssetViewer`
4. Click the Previous button multiple times. When we reach the end of the bucket(s) loaded in (2), we begin to show assets from buckets loaded in (1) thus missing a lot of assets in intermediate buckets.

Details:

I couldn't find a way to write `navigateAsset` function better. But I think the current implementation is at least clear and readable. It just finds the adjacent bucket when needed


web/src/lib/stores/asset-interaction.store.ts
```diff
@@ -67,31 +67,60 @@ function createAssetInteractionStore() {
...
-        }
-      }
+    const currentBucketIndex = _assetGridState.loadedAssets[_viewingAssetStoreState.id];
+    if (currentBucketIndex < 0 || currentBucketIndex >= _assetGridState.buckets.length) {
+      return;
+    }
+
+    const currentBucket = _assetGridState.buckets[currentBucketIndex];
+
+    const index = currentBucket.assets.findIndex((a) => a.id == _viewingAssetStoreState.id);
```
uhthomas, line 77:
```suggestion
    const index = currentBucket.assets.findIndex(({ id }) => id == _viewingAssetStoreState.id);
```

web/src/lib/stores/asset-interaction.store.ts
```diff
@@ -67,31 +67,60 @@ function createAssetInteractionStore() {
...
+    if (index == -1) {
```
uhthomas, line 78:
```suggestion
    if (index === -1) {
```

web/src/lib/stores/asset-interaction.store.ts
```diff
@@ -67,31 +67,60 @@ function createAssetInteractionStore() {
...
+      if (index + 1 < currentBucket.assets.length) {
+        const asset = currentBucket.assets[index + 1];
+        if (asset) {
+          setViewingAsset(asset);
+        }
+      } else {
+        const nextBucketIndex = currentBucketIndex + 1;
+        if (nextBucketIndex < _assetGridState.buckets.length) {
+          const nextBucket = _assetGridState.buckets[nextBucketIndex];
+          if (nextBucket.assets.length == 0) {
```
uhthomas, line 92:
```suggestion
          if (nextBucket.assets.length === 0) {
```

web/src/lib/stores/asset-interaction.store.ts
```diff
@@ -67,31 +67,60 @@ function createAssetInteractionStore() {
...
+      if (index > 0) {
+        const asset = currentBucket.assets[index - 1];
+        if (asset) {
+          setViewingAsset(asset);
+        }
+      } else {
+        const prevBucketIndex = currentBucketIndex - 1;
+        if (prevBucketIndex >= 0) {
+          const prevBucket = _assetGridState.buckets[prevBucketIndex];
+          if (prevBucket.assets.length == 0) {
```
uhthomas, line 113:
```suggestion
          if (prevBucket.assets.length === 0) {
```

web/src/lib/stores/asset-interaction.store.ts
```diff
@@ -67,31 +67,60 @@ function createAssetInteractionStore() {
...
 
-    const asset = _assetGridState.assets[index];
-    if (asset) {
-      setViewingAsset(asset);
+    if (direction === 'next') {
+      if (index + 1 < currentBucket.assets.length) {
+        const asset = currentBucket.assets[index + 1];
+        if (asset) {
+          setViewingAsset(asset);
+        }
```
jrasm91, line 87: There are a lot of nested if statements here and a lot of similar conditions. It would be more readable if it was flattened. You could just set asset to a variable and only do setViewingAsset once at the end of the file if it is not null. You could also do something like add an internal method for getNextAsset and getPreviousAsset which could return early from if statements, which would further cut down on the nesting (no more else statements in those implementations).

brighteyed: > There are a lot of nested if statements here and a lot of similar conditions. It would be more readable if it was flattened. You could just set asset to a variable and only do setViewingAsset once at the end of the file if it is not null. You could also do something like add an internal method for getNextAsset and getPreviousAsset which could return early from if statements, which would further cut down on the nesting (no more else statements in those implementations).

Thank you for the idea! I believe you meant something like I just pushed. This code is better, IMO

jrasm91 (approved):
Looks a lot better IMO! A few more suggestions, but looks good to me.

web/src/lib/stores/asset-interaction.store.ts
jrasm91: Yup, exactly. Looks a lot better. Thanks!

web/src/lib/stores/asset-interaction.store.ts
```diff
@@ -66,52 +66,72 @@ function createAssetInteractionStore() {
...
+      return null;
+    }
+
+    const nextBucket = _assetGridState.buckets[nextBucketIndex];
+    if (nextBucket.assets.length !== 0) {
+      return nextBucket.assets[0];
+    }
+
+    await assetStore.getAssetsByBucket(nextBucket.bucketDate, BucketPosition.Below);
+    navigateAsset('next');
```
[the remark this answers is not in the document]
brighteyed: Nice catch!

web/src/lib/stores/asset-interaction.store.ts
uhthomas: I was going to comment the same thing, thanks for mentioning this @jrasm91.

web/src/lib/stores/asset-interaction.store.ts
```diff
@@ -66,30 +66,70 @@ function createAssetInteractionStore() {
...
+    }
+
+    const nextBucket = _assetGridState.buckets[nextBucketIndex];
+    await assetStore.getAssetsByBucket(nextBucket.bucketDate, BucketPosition.Below);
+
+    if (nextBucket.assets.length !== 0) {
+      return nextBucket.assets[0];
+    }
+
+    return null;
```
uhthomas, line 92:
```suggestion
    return nextBucket.assets.values().next() ?? null;
```

alextran1502 (approved):

PR #3137, opened 2023-07-07
fix(server): thumbnail content type not being passed to stream handle

### LINKED ISSUE #3018: thumbnails should be served with the correct content type

mertalev:
## Description

This PR explicitly sets the content type for thumbnails to avoid having them sent as `application/octet-stream`. It also adds a warning in situations where the requested thumbnail format isn't available and a fallback is used.

Fixes #3018

## How Has This Been Tested?

This was tested by scrolling in the main gallery view to check that all images load as expected with the correct content type.

### DIFF
--- a/server/immich-openapi-specs.json
+++ b/server/immich-openapi-specs.json
@@ -1673,7 +1673,13 @@
         "responses": {
           "200": {
             "content": {
-              "application/octet-stream": {
+              "image/jpeg": {
+                "schema": {
+                  "type": "string",
+                  "format": "binary"
+                }
+              },
+              "image/webp": {
                 "schema": {
                   "type": "string",
                   "format": "binary"
@@ -2704,7 +2710,7 @@
         "responses": {
           "200": {
             "content": {
-              "application/octet-stream": {
+              "image/jpeg": {
                 "schema": {
                   "type": "string",
                   "format": "binary"
--- a/mobile/openapi/doc/AssetApi.md
+++ b/mobile/openapi/doc/AssetApi.md
@@ -822,7 +822,7 @@ Name | Type | Description  | Notes
 ### HTTP request headers
 
  - **Content-Type**: Not defined
- - **Accept**: application/octet-stream
+ - **Accept**: image/jpeg, image/webp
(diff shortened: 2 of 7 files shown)


jrasm91 (approved):

alextran1502:
So this is strange, if you run `make api` from the root of the project, you get new generated value for the SDK, like so

image

And when this happens, the web tests won't pass.

image

alextran1502:
I just ran the openapi generator script, and you can see new files get generated.

server/immich-openapi-specs.json
```diff
@@ -1672,20 +1664,6 @@
         ],
         "responses": {
           "200": {
-            "content": {
```
mertalev, line 1675: Is your `immich-openapi-specs.json` different than mine? The changes in the branch are getting undone here. I just ran `make api` locally with this branch up-to-date and it worked.

jrasm91:
The problem is the open API spec is changed in your commit. I'm not sure if those changes are correct. Obviously if you remove the response the type will become an issue. We need the spec to say binary still.

server/immich-openapi-specs.json
alextran1502: No I think this is the correct spec, because you remove the `@ApiOkResponse` so these values should get removed, right?

mertalev:
It should be good now lol.

alextran1502:
something is wrong with this branch, I can see that you push the changes with newly generated files but the diff shows otherwise

mertalev:
I added back the ApiOkResponse decorators, regenerated the files and force pushed to overwrite the bad SDK files. Is there something in particular that doesn't look right?

PR #3148, opened 2023-07-07
refactor(server): upload config

jrasm91:
Tested Scenarios
- Upload profile picture from mobile app
- Upload file via drag and drop
- Upload file via upload button
- Upload duplicate via drag and drop
- Upload duplicate via upload button

Not Tested:
- XMP
- Live Photos

The code/logic should be the same, it has just been restructured in such a way that the logic (and tests) now exists in `AssetService`.


alextran1502 (approved):
Looks good, tested with LivePhotos and it works fine

PR #3155, opened 2023-07-08
feat(web): add emptyplaceholder when no assets

faupau03:
Adds the emptyplaceholder to the timeline view.
grafik

### DIFF
--- a/web/src/routes/(user)/photos/+page.svelte
+++ b/web/src/routes/(user)/photos/+page.svelte
@@ -12,18 +12,31 @@
   import AssetSelectControlBar from '$lib/components/photos-page/asset-select-control-bar.svelte';
   import { assetInteractionStore, isMultiSelectStoreState, selectedAssets } from '$lib/stores/asset-interaction.store';
   import { assetStore } from '$lib/stores/assets.store';
-  import { onDestroy } from 'svelte';
+  import { onDestroy, onMount } from 'svelte';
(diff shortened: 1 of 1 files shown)


web/src/routes/(user)/photos/+page.svelte
```diff
@@ -12,18 +12,31 @@
...
+  import { onDestroy, onMount } from 'svelte';
   import DotsVertical from 'svelte-material-icons/DotsVertical.svelte';
   import Plus from 'svelte-material-icons/Plus.svelte';
   import type { PageData } from './$types';
+  import { api } from '@api';
+  import EmptyPlaceholder from '$lib/components/shared-components/empty-placeholder.svelte';
+  import { openFileUploadDialog } from '$lib/utils/file-uploader';
 
   export let data: PageData;
+  let assetCount: number = 1;
```
uhthomas, line 24: Type annotation shouldn't be needed here?

uhthomas (commented):
Looks cool, though I wonder if there's a more elegant way to do this? It might be possible to render this if there are no assets available in the store, which would avoid an API call. The only problem I envision there is the placeholder could show whilst the timeline is loading? I'm sure there's a way to avoid that though.

uhthomas:
also, could the placeholder be vertically centred?

jrasm91 (approved):

PR #3171, opened 2023-07-09
feat(server): transcoding hardware acceleration

### LINKED ISSUE #1225: Hardware acceleration transcoding option for videos
### Feature detail

Could you add an option to harness the integrated hardware transcoding capabilities in most modern CPUs (and GPUs) for video transcoding? This would significantly reduce the load on the server when transcoding video files.

An option could be added under the ffmpeg settings in the administration area to select the transcoding method (software vs. hardware)

You would need to pass through the transcoding device into the docker. For details or as an example you can see how PhotoPrism does that.  For details see:
[https://docs.photoprism.app/getting-started/advanced/transcoding/](https://docs.photoprism.app/getting-started/advanced/transcoding/) 

Best

Daniel

### Platform

Server

### LINKED ISSUE #3084: should support svg files
### The bug

Immich does not currently allow svg files.

N/A

### Version of Immich Server

v1.65.0

N/A

### Platform with the issue

- [X] Server
- [X] Web
- [X] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
N/A
```

### Additional information

_No response_

mertalev:
## Description

Support for NVENC, Quick Sync and VAAPI transcoding backends. This will make transcoding much faster at the cost of lower quality at the same bitrate compared to software encoding.

Additionally, this substantially changes the server/microservices Dockerfile:

- Uses a Debian-based image
  - Alpine is incompatible with many acceleration APIs
- Builds Imagemagick and libvips from source
  - Debian's Imagemagick package has issues with RAW images 

Fixes #1225, fixes #3084

Notes:
- This adds a new file `hwaccel.yml` that `immich-microservices` uses to set the appropriate settings for the docker container. Be sure to uncomment the right sections in this file to ensure hardware acceleration works.
- This uses a Debian-based image, so if running in a dev environment you should delete your current `immich-microservices` and `immich-server` containers and the `dist` and `node_modules` folders relating to them. Otherwise you might see errors stemming from it trying to load Alpine binaries.

## How Has This Been Tested?
Current testing in production build on three systems, one with an Nvidia GPU, another with an older Intel CPU, and a third with a 13th gen Intel CPU:

- NVENC
  - ✅ H.264
  - ✅ HEVC
  - ❌ VP9 does not work (not supported)
- VAAPI
  - ✅ H.264
  - ✅ HEVC
  - ✅ VP9 (on a supporting Intel CPU)
- QSV
  - ✅ H.264
  - ✅ HEVC
  - ✅ VP9 (on a supporting CPU)

NOTE: Only a limited variety of image formats were tested with this, primarily HEIC and JPEG. It would be good to check other formats as well to ensure Imagemagick and libvips work as expected.


martabal:
Do they support raw images ?

mertalev:
The image has to support image formats like RAW since it's responsible for thumbnail generation.

martabal:
Does it work ? With [https://github.com/imagegenius/docker-immich](this image) we have to build imagemagick & libvips to support raw images.

martabal:
And you can have a version mismatch between libvips & sharp

mertalev:
It seems Imagemagick can handle raw images, but sharp throws an error. I can try compiling libvips as well to see if that helps.

server/Dockerfile.accel
```diff
@@ -0,0 +1,54 @@
+FROM node:18-bookworm as builder
+
+WORKDIR /usr/src/app
+
+RUN --mount=type=cache,target=/var/cache/apt apt-get update && apt-get install -yqq ffmpeg libraw-dev libjxl-dev libvips-dev libheif-dev
+# debian build for imagemagick has broken RAW support, so build manually
+RUN wget https://imagemagick.org/archive/ImageMagick.tar.gz \
```
brighteyed, line 7: Do we always want to use the latest released version?

docker/docker-compose.yml
```diff
@@ -21,13 +21,30 @@ services:
...
+      # - /dev/dxg  # If using hwaccel in WSL2
+      # - /dev/dri/card0  # If using VAAPI for GPU hwaccel
+      # - /dev/dri/renderD129  # alternate device if `card0` doesn't work
+    # deploy:  # Uncomment this section if using NVIDIA GPU
+    #   resources:
+    #     reservations:
+    #       devices:
+    #         - driver: nvidia
+    #           count: 1
+    #           capabilities: [gpu]
```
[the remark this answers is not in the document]
mertalev: That's a lot cleaner, I like it.

server/src/domain/media/media.util.ts
```diff
@@ -189,3 +196,184 @@ export class VP9Config extends BaseConfig {
...
+  }
+}
+
+export class QSVConfig extends BaseConfig implements VideoCodecHWConfig {
+  getSupportedCodecs() {
+    return [VideoCodec.H264, VideoCodec.HEVC, VideoCodec.VP9];
+  }
+
+  getBaseInputOptions() {
+    return ['-init_hw_device qsv=accel:/dev/dri/renderD128', '-filter_hw_device accel'];
```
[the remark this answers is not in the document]
mertalev: VAAPI supports all kinds of devices so it needs more logic, whereas QSV generally wants `renderD128`. But it would be good to check if this device exists instead of just assuming it does.

server/Dockerfile.accel
mertalev: We don't particularly need to, but if we set it to a certain version then we need to keep track of updating it. The AIO image seems to have envs for targeting a version and assigns the latest version to them if not specified. I think something like that would be good.

mertalev:
I can confirm RAW images work now after compiling libvips. The image is also a decent bit smaller at 1.07gb instead of 1.22gb.

server/src/domain/media/media.util.ts
```diff
@@ -189,3 +196,184 @@ export class VP9Config extends BaseConfig {
...
+  }
+}
+
+export class VAAPIConfig extends BaseConfig implements VideoCodecHWConfig {
+  getSupportedCodecs() {
+    return [VideoCodec.H264, VideoCodec.HEVC, VideoCodec.VP9];
+  }
+
+  getBaseInputOptions() {
+    const devices = readdirSync('/dev/dri/');
```
jrasm91, line 326: I think we want to avoid these types of calls in the domain code. They can't be mocked in tests very easily. Ideally you should call a repository method, like `storageRepository.readDir('/dev/dri/')` or something

samip5:
I tried this with Quick Sync, and well, libva errors.

```
[Nest] 7  - 07/17/2023, 10:20:31 PM     LOG [MediaService] Start encoding video 5fceb8ef-581b-43ea-9610-951b7675f0ee {"inputOptions":["-init_hw_device qsv=accel:/dev/dri/renderD128","-filter_hw_device accel"],"outputOptions":["-vcodec h264_qsv","-acodec aac","-movflags faststart","-fps_mode passthrough","-vf format=nv12,hwupload=extra_hw_frames=64","-preset 7","-global_quality 23"],"twoPass":false}
[Nest] 7  - 07/17/2023, 10:20:31 PM   ERROR [MediaRepository] ffmpeg version 6.0 Copyright (c) 2000-2023 the FFmpeg developers
  built with gcc 12.2.1 (Alpine 12.2.1_git20220924-r10) 20220924
  configuration: --prefix=/usr --disable-librtmp --disable-lzma --disable-static --disable-stripping --enable-avfilter --enable-gnutls --enable-gpl --enable-libaom --enable-libass --enable-libbluray --enable-libdav1d --enable-libdrm --enable-libfontconfig --enable-libfreetype --enable-libfribidi --enable-libmp3lame --enable-libopenmpt --enable-libopus --enable-libplacebo --enable-libpulse --enable-librist --enable-libsoxr --enable-libsrt --enable-libssh --enable-libtheora --enable-libv4l2 --enable-libvidstab --enable-libvorbis --enable-libvpx --enable-libwebp --enable-libx264 --enable-libx265 --enable-libxcb --enable-libxml2 --enable-libxvid --enable-libzimg --enable-libzmq --enable-lto --enable-pic --enable-postproc --enable-pthreads --enable-shared --enable-vaapi --enable-vdpau --enable-vulkan --optflags=-O3 --enable-libjxl --enable-libsvtav1 --enable-libvpl
  libavutil      58.  2.100 / 58.  2.100
  libavcodec     60.  3.100 / 60.  3.100
  libavformat    60.  3.100 / 60.  3.100
  libavdevice    60.  1.100 / 60.  1.100
  libavfilter     9.  3.100 /  9.  3.100
  libswscale      7.  1.100 /  7.  1.100
  libswresample   5.  0.100 /  5.  0.100
  libpostproc    57.  1.100 / 57.  1.100
[AVHWDeviceContext @ 0x7fc4ffe2ef00] Failed to initialise VAAPI connection: -1 (unknown libva error).
Device creation failed: -5.
Failed to set value 'qsv=accel:/dev/dri/renderD128' for option 'init_hw_device': I/O error
Error parsing global options: I/O error

[Nest] 7  - 07/17/2023, 10:20:31 PM   ERROR [MediaService] Error: ffmpeg exited with code 1: Device creation failed: -5.
Failed to set value 'qsv=accel:/dev/dri/renderD128' for option 'init_hw_device': I/O error
Error parsing global options: I/O error
```

mertalev:
From that log it's using Alpine binaries. I think you need to set both immich-server and immich-microservices to use Dockerfile.accel and rebuild the containers.

(That should probably be the default for dev, though the bigger conversation to have is whether this image should replace the existing Alpine one or be an alternate version.)

Also double check that you have the hwaccel.yml and enabled the renderD128 device on it.

samip5:
> From that log it's using Alpine binaries. For the dev environment, I think you need to set both immich-server and immich-microservices to use Dockerfile.accel and rebuild the containers.
> 
> (That should probably be the default for dev, though the bigger conversation to have is whether this image should replace the existing Alpine one or be an alternate version.)
> 
> Also double check that you have the hwaccel.yml and enabled the renderD128 device on it.

I'm using the ghcr.io/immich-app/server:pr-3171 image tag though?

Can be seen [here](https://github.com/samip5/k8s-cluster/blob/main/cluster/apps/default/immich-dev/app/microservices/helm-release.yaml#L33)

jrasm91:
Does the PR build push hardware accelerated enabled images?

samip5:
> Does the PR build push hardware accelerated enabled images?

The option exists in settings, but it seems to be alpine images.

mertalev:
It seems GH was building with the Alpine-based Dockerfile before this. I moved `Dockerfile.accel` to `Dockerfile` for now to avoid the headache of managing both.

server/src/domain/media/media.util.ts
mertalev: I moved the call to `storageRepository.readdir` in the service.

server/Dockerfile.accel
mertalev: I ended up setting target versions for both since it feels more stable that way. Plus it means we can use sha256 checksums so it's safer.

jrasm91 (approved):
Code looks good!

server/src/domain/media/media.service.spec.ts
```diff
@@ -765,5 +765,434 @@ describe(MediaService.name, () => {
...
+            '-preset p1',
+            '-cq:v 23',
+          ],
+          twoPass: false,
+        },
+      );
+    });
+
+    it('should set options for qsv', async () => {
+      storageMock.readdir.mockImplementation(async () => ['renderD128']);
```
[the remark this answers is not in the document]
mertalev: I tried `storageMock.readdir.mockResolvedValue(new Promise(() => ['renderD128']))` and it just stalled. But I apparently didn't read the name `mockResolvedValue` closely enough to just pass `['renderD128']` 😛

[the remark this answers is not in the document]
jrasm91: Lol, nice.

brighteyed (approved):
I have not tested this PR, but LGTM!

mertalev:
As of the latest version of this PR, I can confirm all three APIs work when used on a supporting system and the `hwaccel.yml` is configured.

alextran1502:
I just run the first test uploading from iPhone and ran into this issue with websocket. 

image

mertalev:
Hmm that's odd, I didn't touch that part of the codebase at all.

alextran1502:
@mertalev Agree, might it be the changed in the node distro? Do you see the same thing when upload from the mobile app?

mertalev:
I can't seem to reproduce this issue on iOS. I can upload and view images without any errors.

alextran1502:
> I can't seem to reproduce this issue on iOS. I can upload and view images without any errors.

Hmm let me try again, I can confirm that the upload is success and I can view the image, just a bunch of those errors showing up

martabal:
Hey @mertalev can you store the libvips/imagemagick version in a txt/json file for easy parsing ?

mertalev:
I made a `build-lock.json` file for build versions and sha256.

martabal:
> I made a `build-lock.json` file for build versions and sha256.

This is perfect. Thanks !

docs/docs/features/hardware-transcoding.md
```diff
@@ -0,0 +1,60 @@
...
+3. In the `docker-compose.yml` under `immich-microservices`, uncomment the lines relating to the `hwaccel.yml` file.
+4. Redeploy the `immich-microservices` container with these updated settings.
+5. In the Admin page under `FFmpeg settings`, change the hardware acceleration setting to the appropriate option and save.
+
+## Tips
+
+- You may want to choose a slower preset than for software transcoding to maintain quality and efficiency
+- While you can use VAAPI with Nvidia GPUs and Intel CPUs, prefer the more specific APIs since they're more optimized for their respective devices
+
+[hw-file]: https://github.com/immich-app/immich/releases/latest/download/hwaccel.yml
```
jrasm91, line 57: Have you changed the github action to include this in the release?

server/Dockerfile
jrasm91, line 1: Lots of changes here.. A couple questions:
- How does the impact the size of the image
- Why are the build from source changes in this PR? Are they related to hardware acceleration or just because you were already making so many other build-related changes?

server/src/domain/media/media.service.ts
```diff
@@ -228,7 +236,15 @@ export class MediaService {
     }
   }
 
-  private getCodecConfig(config: SystemConfigFFmpegDto) {
+  async getCodecConfig(config: SystemConfigFFmpegDto) {
+    if (!config.accel || config.accel === TranscodeHWAccel.DISABLED) {
```
jrasm91, line 240: Did we decide this would always have a value?

server/src/domain/media/media.service.ts
```diff
@@ -228,7 +236,15 @@ export class MediaService {
...
+      return this.getSWCodecConfig(config);
+    } else {
```
jrasm91, line 242: Don't need an else if the previous line returns.

server/src/domain/media/media.util.ts
```diff
@@ -137,6 +144,34 @@ class BaseConfig implements VideoCodecSWConfig {
...
+      .filter((device) => device.startsWith('renderD') || device.startsWith('card'))
+      .sort((a, b) => {
+        // order GPU devices first
+        if (a.startsWith('card') && b.startsWith('renderD')) {
+          return -1;
+        } else if (a.startsWith('renderD') && b.startsWith('card')) {
+          return 1;
+        } else {
+          return -a.localeCompare(b);
+        }
```
jrasm91, line 170: I'd prefer if + return, if + return etc.

server/src/domain/system-config/system-config.core.ts
```diff
@@ -27,6 +28,7 @@ export const defaults = Object.freeze<SystemConfig>({
     maxBitrate: '0',
     twoPass: false,
     transcode: TranscodePolicy.REQUIRED,
+    accel: TranscodeHWAccel.DISABLED,
```
jrasm91, line 31: This should be the default, so it should never not have a value, just FYI.

server/src/infra/repositories/filesystem.provider.ts
```diff
@@ -92,4 +92,8 @@ export class FilesystemProvider implements IStorageRepository {
       total: stats.blocks * stats.bsize,
     };
   }
+
+  async readdir(folder: string): Promise<string[]> {
+    return fs.readdir(folder);
+  }
```
jrasm91, line 98: I think you can also do `readdir = readdir` if you do `import { readdir } from 'fs/promises'`

jrasm91 (approved):
Looks complicated AF! :smile:

server/src/domain/media/media.service.ts
mertalev: This is from back when I wasn't sure if it would have a value, but yeah it always should

server/Dockerfile
mertalev: The size of the image increases from a little over 700mb to 1.1gb, mostly because of the switch from Alpine to Debian. It's more or less inescapable unfortunately since Alpine doesn't play nice with these APIs.

The build changes aren't needed for hardware transcoding, but I wanted to avoid causing a regression in image format compatibility because Debian's package has an issue with RAW files. I can take that part out and make a separate PR if you prefer.

docs/docs/features/hardware-transcoding.md
mertalev: No 😅

server/Dockerfile
jrasm91: I think it's fine, but we should validate/test that, not just hw changes as well.

mertalev: Agreed

mertalev:
> Looks complicated AF! 😄

There was a brief moment where I thought I'd just need to add `-hwaccel auto` and it would automagically take care of everything... I wish.

server/Dockerfile
jrasm91: Can you update the PR description to mention these changes?

mertalev: Done

mertalev:
Tested QSV on a 13th gen CPU. For Quick Sync support for 12th gen and above, it seems FFmpeg needs to be built with `libvpl`. The current version of FFmpeg in Debian Bookworm is built with `libmfx` instead so it doesn't work on these CPUs. 

How do we want to handle this? We could install from a different repo or maybe even use [Jellyfin's own FFmpeg package](https://github.com/jellyfin/jellyfin-ffmpeg/releases), or we could just add it to the list of limitations in the docs for now.

mertalev:
I can confirm that switching to Jellyfin's FFmpeg build fixed the issue with Quick Sync (and shrunk the image down to under 1GiB). VP9 also works in Quick Sync. With that, the only outstanding issue seems to be the errors @alextran1502 got. If we can resolve that, it otherwise seems good. Maybe a rebase could fix it?

mertalev:
Rebased

PR #3198, opened 2023-07-10
Add typesense_host into example env file

CliffLin:
When I deployed typesense with another container name, I found the application is not able to connect to typesense.   
I think it's better to put the env name to example file to have a chance to lead user find it when they need to change it.


alextran1502:
It is already included in the environment variable section in the documentation page

image

In the future, we will simplify the `.env.example` and only include the required variables

PR #3202, opened 2023-07-10
feat(server): Use Testcontainers for e2e tests

etnoy:
This adds testcontainers for e2e testing (including future integration testing).

Testcontainers is an easy way to spin up dbs for testing. The essence of this PR are the following few lines:

image

and bam! you have a clean pg and redis instance for the tests to run. No need to start tedious db setups in your dev machine.

This cuts the time of e2e github actions in half!


uhthomas (approved):

server/e2e/load-testcontainers.ts
```diff
@@ -0,0 +1,24 @@
+import { GenericContainer, PostgreSqlContainer } from 'testcontainers';
```
jrasm91, line 1: Can we rename this setup.ts?

jrasm91 (approved):

server/e2e/load-testcontainers.ts
etnoy: Will do

etnoy: Done

PR #3205, opened 2023-07-11
Add TZ variable to example env file and installation documentation

catleeball:
I saw in the FAQ that setting the TZ environment variable will help immich display timestamp metadata correctly, so I added it to the example `.env` file and updated the docker-compose installation steps to mention it.

### DIFF
--- a/docker/example.env
+++ b/docker/example.env
@@ -114,3 +114,16 @@ IMMICH_MACHINE_LEARNING_URL=http://immich-machine-learning:3003
 ###################################################################################
 
 #IMMICH_VERSION=
+
+###################################################################################
+# Timezone
+#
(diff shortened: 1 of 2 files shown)


jrasm91 (commented):
Thank you for the contribution!

However, we're actually trying to move _away_ from the `.env` as "documentation", in favor of the website. We actually have an environment variables section that documents the use of `TZ` here:

It sounds like we need to add a link in the `example.env` that points to that page and remove some of the option sections of the file.

etnoy:
I added the TZ part to the faq. Honestly, it's a bit of a hack that night work worse than many think. The real fix is too make sure files have the correct metadata.

PR #3213, opened 2023-07-11
Document OAuth config env variables for docker

osresearch:
This patch adds documentation to Administration - OAuth for setting the docker-compose environment to include the OIDC provider information and also adds a sample configuration to the example.env file.

### DIFF
--- a/docker/example.env
+++ b/docker/example.env
@@ -114,3 +114,21 @@ IMMICH_MACHINE_LEARNING_URL=http://immich-machine-learning:3003
 ###################################################################################
 
 #IMMICH_VERSION=
+
+###################################################################################
+# OAuth configuration - Optional
+#
+# This allows users to login with external credentials from an OAuth / OIDC provider
+# like Authentik or Keycloak instead of having a local account.  This sort of SSO
+# is useful if immich is installed as part of a community or managed group.
+#
+# For more info see https://immich.app/docs/administration/oauth
+###################################################################################
+#OAUTH_ENABLED=true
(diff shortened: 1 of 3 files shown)


bo0tzz:
I don't think the env vars for oauth config are supported anymore?

osresearch:
Why were they removed?  `.env` vars are necessary for automated deployments and requiring a manual step via the website to set it up is problematic compared to being able to include them in the runtime configuration.

jrasm91:
> Why were they removed? `.env` vars are necessary for automated deployments and requiring a manual step via the website to set it up is problematic compared to being able to include them in the runtime configuration.

I don't disagree, but the trade-off to keep them was increased maintenance work. It is very uncommon to need to regularly run automated deployments of immich _and_ preconfigure OAuth, so we went with a simpler approach of only configuring these values through the UI.

osresearch:
I'm hoping to integrate immich into the https://v.st/ self-hosted system, which uses single-sign-on for all of the included tools (mastodon, matrix, hedgedoc, nextcloud, etc).  Being able to help folks setup a community with all of the tools pre-configured to work together using a single `docker-compose` command is much easier for groups that don't want to have to separately go through each tool to figure out how its SSO or OIDC vars are configured.

jrasm91:
In the short term, I don't think we are going to add support for auto configuring OAuth or any other settings via environment variables. While not ideal, I'm sure you can accomplish the same thing via a script. Either using the CLI to updates the values, inserting them directly into the database, or injecting a migration file into the container via a volume mount or extending the image.

PR #3238, opened 2023-07-13
fix(web): remove processing key events in photo viewer window

### LINKED ISSUE #3165: image metadata cannot be copied to clipboard
### The bug

I want to copy the filename, but can't because Immich is quite aggressively pushing the image to the clipboard instead.

image

N/A

### Version of Immich Server

v1.65.0

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
N/A
```

### Additional information

_No response_

tranquanghuy0801:
Resolves #3165 

Currently processing key events are always executed. Update this to allow only when range selection is not active

### DIFF
--- a/web/src/lib/components/asset-viewer/photo-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/photo-viewer.svelte
@@ -45,6 +45,9 @@
   };
 
   const handleKeypress = async ({ metaKey, ctrlKey, key }: KeyboardEvent) => {
+    if (window.getSelection()?.type === 'Range') {
+      return;
+    }
     if ((metaKey || ctrlKey) && key === 'c') {
       await doCopy();
     }


bo0tzz:
Does the image copying feature still work with this change?

tranquanghuy0801:
> Does the image copying feature still work with this change?

If you click on the `Copy` button it still works. Or is there another way to copy the image?

jrasm91 (commented):
We want to change it from "ctrl + c always copies the image" to "ctrl + c only copies the image when there is no text selected". ie selection type is not "Range".

https://developer.mozilla.org/en-US/docs/Web/API/Selection/type

tranquanghuy0801:
> We want to change it from "ctrl + c always copies the image" to "ctrl + c only copies the image when there is no text selected". ie selection type is not "Range".
> 
> https://developer.mozilla.org/en-US/docs/Web/API/Selection/type

Sorry for that. Also thanks for the guidance. Updated my work

bo0tzz (approved):

jrasm91 (approved):

PR #3286, opened 2023-07-15
fix(web): navigation buttons z-order

### LINKED ISSUE #2946: [BUG] Video player has greyed out next button despite arrow key working for next
### The bug

The web video player seems to have an right arrow that will often times go greyed out when watching videos in an album, despite arrow key on keyboard will still work despite it being greyed out.

Example of greyed out:
image

k3s

### Version of Immich Server

v1.63.2

n/a

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

```bash
1. Create an album with just videos
2. Begin watching any video
3. Notice that sometimes when wanting to move to the next one, you cannot unless keyboard is used.
```

### Additional information

_No response_

brighteyed:
## Description
This PR:
* Fixes #2946 
* Simplifies `AssetViewer` component by extracting `NavigationArea` component

## How Has This Been Tested?
Chrome/Edge/Firefox:
- Navigate by next/prev buttons
- Navigate by arrow keys
- Navigate by clicking on the navigation area
- Highlight by hovering over the next/prev button or navigation area
- Z-order (menu, notification, download panel)
- Close `AssetViewer` by `Escape` key


web/src/lib/components/asset-viewer/navigation-area.svelte
jrasm91, line 1: Instead of tracking hover in this class, can we just use this from tailwind?

https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state

brighteyed: Thanks! Also removed `NavigationButton` since we don't need it any more

jrasm91 (approved):

PR #3317, opened 2023-07-17
fix(server): add back missed extensions

jrasm91:
Missed in a recent refactor.


jrasm91:
Closing in favor of #3318

PR #3197, opened 2023-07-10
refactor(server): mime types

jrasm91:
In this PR:
- Remove mime type from asset property and database table
- Remove mime type from asset response
- Remove mime-types server dependency
- Infer mime-type from file extension (on upload and retrieval)

Tested Scenarios:
- Verify `image/webp` content type for image serve (thumbnail path)
- Verify `image/jpeg` content type for image serve (resize path)
- Verify `video/mp4` content type for video serve (mp4)
- Verify `video/webm` content type for video serve (webm)
- Verify `image/jpeg` content type for file download (jpeg)
- Verify `video/mp4` content type for file download (mp4)
- Verify uploaded extensions are accepted: `.heif`, `.tiff`, `.dng`, `.ari`, `.arw`, `.raf`, `.heic`, `.mov`, `.srw`, `.pef`, `.jpg`, `.jpg`, `.mp4`, `.webm`

Next Steps:
- Add endpoint for file extension list
- Remove extension mapping from web/mobile/cli
- Remove asset type from create dto

PR #3318, opened 2023-07-17
fix(server): add missing extensions and mime types

### LINKED ISSUE #3300: regression(web): mts files unsupported and uncaught error for unsupported file extensions
### The bug

Attempting to upload a `.mts` video via the web on v1.67.2 does nothing and shows errors in the console.

image

image

N/A

### Version of Immich Server

v1.67.2

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
N/A
```

### Additional information

_No response_

uhthomas:
Add extensions and mime types which were accidentally removed in #3197.

Fixes: #3300

### DIFF
--- a/server/src/domain/asset/asset.service.spec.ts
+++ b/server/src/domain/asset/asset.service.spec.ts
@@ -12,7 +12,6 @@ import {
 import { when } from 'jest-when';
 import { Readable } from 'stream';
 import { ICryptoRepository } from '../crypto';
-import { mimeTypes } from '../domain.constant';
 import { IStorageRepository } from '../storage';
 import { AssetStats, IAssetRepository } from './asset.repository';
 import { AssetService, UploadFieldName } from './asset.service';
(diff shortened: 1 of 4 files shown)


server/src/domain/domain.constant.ts
```diff
@@ -30,70 +30,70 @@ export function assertMachineLearningEnabled() {
...
+  '.xmp': ['application/xml', 'text/xml'],
 };
 
-const isType = (filename: string, lookup: Record<string, string>) => !!lookup[extname(filename).toLowerCase()];
-const getType = (filename: string, lookup: Record<string, string>) => lookup[extname(filename).toLowerCase()];
-const lookup = (filename: string) =>
-  getType(filename, { ...image, ...video, ...sidecar }) || 'application/octet-stream';
+const isType = (filename: string, r: Record<string, string | string[]>) => extname(filename).toLowerCase() in r;
+
+const lookup = (filename: string) => ({ ...image, ...video, ...sidecar }[extname(filename).toLowerCase()]).at(0) ?? 'application/octet-stream';
```
jrasm91, line 96: This looks wrong. You are indexing into `Record<string, string | string[]>`. This will return the first letter of the string or the first item in the array.

server/src/domain/asset/asset.service.spec.ts
```diff
@@ -132,6 +179,14 @@ describe(AssetService.name, () => {
             );
           });
         }
+
+        it('should be sorted (valid)', () => {
+          expect(valid).toEqual([...valid].sort());
+        });
+
+        it('should be sorted (invalid)', () => {
+          expect(valid).toEqual([...valid].sort());
```
jrasm91, line 188: These should be invalid actually

server/src/domain/domain.constant.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+    { mimetype: 'video/quicktime', extension: 'mov' },
+    { mimetype: 'video/vnd.avi', extension: 'avi' },
+    { mimetype: 'video/webm', extension: 'webm' },
+    { mimetype: 'video/x-flv', extension: 'flv' },
+    { mimetype: 'video/x-matroska', extension: 'mkv' },
+    { mimetype: 'video/x-ms-wmv', extension: 'wmv' },
+    { mimetype: 'video/x-msvideo', extension: 'avi' },
+  ]) {
+    it(`should map ${extension} to ${mimetype}`, async () => {
+      expect({ ...mimeTypes.image, ...mimeTypes.video }[`.${extension}`]).toContain(mimetype);
```
jrasm91, line 86: You are reconstructing this object every test. Just do it once outside of the for loop.

jrasm91: Can we keep extensions the same here? Or add the period in front of the extension in the list

server/src/domain/asset/asset.service.spec.ts
uhthomas: Thanks, I think I copied this from #3317 and missed it.

server/src/domain/domain.constant.spec.ts
uhthomas: >Can we keep extensions the same here?

What do you mean by this? I understand the second option, but not this one.

server/src/domain/domain.constant.ts
```diff
@@ -30,70 +30,73 @@ export function assertMachineLearningEnabled() {
...
+  '.xmp': ['application/xml', 'text/xml'],
 };
 
-const isType = (filename: string, lookup: Record<string, string>) => !!lookup[extname(filename).toLowerCase()];
-const getType = (filename: string, lookup: Record<string, string>) => lookup[extname(filename).toLowerCase()];
+const isType = (filename: string, r: Record<string, string | string[]>) => extname(filename).toLowerCase() in r;
+
 const lookup = (filename: string) =>
-  getType(filename, { ...image, ...video, ...sidecar }) || 'application/octet-stream';
+  [{ ...image, ...video, ...sidecar }[extname(filename).toLowerCase()]].flat().at(0) ?? 'application/octet-stream';
```
jrasm91, line 99: This looks pretty ugly to me. Can we instead change the objects above to Record<string, string[]> and move the array flattening stuff?

server/src/domain/domain.constant.spec.ts
jrasm91: Also, we are extension driven now, not mime-type driven, so I think it makes more sense to have extension be the first key and mime type be the last key, and then it would make sense for this to be sorted by extension name not mime type.

server/src/domain/domain.constant.ts
jrasm91: Can you also change `??` back to `||`. I prefer and use `||` over `??` when I don't consider `false` or `''` as valid values.

server/src/domain/domain.constant.spec.ts
uhthomas: > Also, we are extension driven now, not mime-type driven, so I think it makes more sense to have extension be the first key and mime type be the last key, and then it would make sense for this to be sorted by extension name not mime type.

This is tough - I think sorting by extensions makes sense when we have `extension: [mimeType]` but for `{ mimeType, extension }` it makes the most sense to sort by mimeType as to keep the grouping.

server/src/domain/domain.constant.ts
uhthomas: > Can you also change `??` back to `||`. I prefer and use `||` over `??` when I don't consider `false` or `''` as valid values.

Given the value can never be false or '', I think the intent behind `??` is clearer.

uhthomas: > This looks pretty ugly to me. Can we instead change the objects above to Record<string, string[]> and remove the array flattening stuff?

Yeah I suppose we can. I just thought it might be nice for flexibility. Though uniformity is probably better.

server/src/domain/domain.constant.spec.ts
uhthomas: > You are reconstructing this object every test. Just do it once outside of the for loop.

I do get this, though this is a test and is unlikely to have any impact on performance. Having to define some constant elsewhere just because feels weird.

jrasm91: Right now if you want to add a new file extension you need to:
- add it to the mapping with a mime type
- update the asset upload test to include the extension
- And now go here and also update this list, which isn't even sorted in the same way. 

I don't know why it needs to be so complicated. Why do we need to validate all variations of mime types, why do we need it sorted alphabetically in the first place? Why can't the test be validating the actual lookup function itself, which you still aren't even testing. I think it is confusing to have this be mime type driven when everything else is extension driven.

jrasm91 (approved):

PR #3332, opened 2023-07-19
fix(server): cancel error

jrasm91:


web/src/lib/utils/handle-error.ts
```diff
@@ -17,7 +17,7 @@ export async function getServerErrorMessage(error: unknown) {
 }
 
 export async function handleError(error: unknown, message: string) {
-  if (error instanceof CanceledError) {
+  if (error instanceof axios.CanceledError) {
```
uhthomas, line 20: Any reason not to use [isCancel](https://github.com/axios/axios/blob/21a5ad34c4a5956d81d338059ac0dd34a19ed094/lib/cancel/isCancel.js#L3)?

alextran1502 (approved):

web/src/lib/utils/handle-error.ts
uhthomas: https://github.com/immich-app/immich/pull/4227

PR #3339, opened 2023-07-19
feat(mobile): render login fields/buttons based on server configuration

ddshd:
Issue #3337, Discussion #2206 

Shows/hides email/password fields based on config on the server
Shows/hides oauth buttons based on config on the server
Shows error message same as web interface is both are disabled

Only Password
Both Enabled
Only OAuth
Nothing Enabled

### DIFF
--- a/mobile/lib/modules/login/ui/login_form.dart
+++ b/mobile/lib/modules/login/ui/login_form.dart
@@ -36,6 +36,7 @@ class LoginForm extends HookConsumerWidget {
     final isLoading = useState<bool>(false);
     final isLoadingServer = useState<bool>(false);
     final isOauthEnable = useState<bool>(false);
+    final isPasswordLoginEnable = useState<bool>(false);
     final oAuthButtonLabel = useState<String>('OAuth');
     final logoAnimationController = useAnimationController(
       duration: const Duration(seconds: 60),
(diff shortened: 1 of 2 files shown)


mobile/lib/modules/login/ui/login_form.dart
```diff
@@ -317,6 +326,10 @@ class LoginForm extends HookConsumerWidget {
                       ],
                     ],
                   ),
+            if (!isOauthEnable.value && !isPasswordLoginEnable.value)
+              const Center(
+                child: Text("Login has been disabled."),
```
ddshd, line 331: How to support translation? Just add a field to the English json?

jrasm91:
Can you attach some screenshots to the PR description that show the updated flow?

mobile/lib/modules/login/ui/login_form.dart
alextran1502: Correct, and the end of the Text it should have a `tr()` method. You can reference other files for example

alextran1502 (approved):

alextran1502:
@ddshd Can you allow the maintainer to push to your branch? 

Here is how you can get the key rendered correctly 
image

ddshd:
> @ddshd Can you allow the maintainer to push to your branch?
> 
> Here is how you can get the key rendered correctly image

Flutter threw an error when i did that for some reason but 

Enabled access for changes

PR #3350, opened 2023-07-20
fix(server): properly handle SIGTERM

### LINKED ISSUE #1568: [Bug]: Handle SIGTERM
The node processes started by Immich currently don't handle the `SIGTERM` signal. This is required to let Docker gracefully shutdown the containers.

Ideally this should close any open connections and also be logged.

bt90:
## Description

Fixes #1568

This adds tini for all node and python based containers as explained in https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#handling-kernel-signals and https://github.com/nodejs/docker-node/issues/1620#issuecomment-1239804844

see https://github.com/krallin/tini

## How Has This Been Tested?

Fired up a Github codespace and every container launched.

## Checklist:


uhthomas:
It looks like there's an alpine package for tini, might be better just to use that.

https://github.com/krallin/tini#alpine-linux-package

bt90:
Fair point.

docker/docker-compose.prod.yml
```diff
@@ -7,7 +7,8 @@ services:
     build:
       context: ../server
       dockerfile: Dockerfile
-    command: ["./start-server.sh"]
+    command: [ "start.sh", "immich" ]
```
uhthomas, line 10: I'm not opposed to making this change, though it is unrelated and is incomplete as the old "start-server.sh" file is hanging around still.

bt90: I was a bit reluctant to remove these files as it might break existing setups that have not yet been adapted to the changes introduced in https://github.com/immich-app/immich/pull/2689

bt90:
@bo0tzz @uhthomas added a proof of concept for using tini

bt90:
```
@bt90 ➜ /workspaces/immich (docker_init) $ docker top immich_web 
UID                 PID                 PPID                C                   STIME               TTY                 TIME                CMD
root                55170               55141               0                   15:53               ?                   00:00:00            tini -- node /usr/src/app/build/index.js
root                55274               55170               0                   15:53               ?                   00:00:00            node /usr/src/app/build/index.js
@bt90 ➜ /workspaces/immich (docker_init) $ docker top immich_server 
UID                 PID                 PPID                C                   STIME               TTY                 TIME                CMD
root                54884               54865               0                   15:53               ?                   00:00:00            tini -- node dist/main immich
root                55008               54884               7                   15:53               ?                   00:00:05            node dist/main immich
@bt90 ➜ /workspaces/immich (docker_init) $ docker top immich_microservices 
UID                 PID                 PPID                C                   STIME               TTY                 TIME                CMD
root                55053               55036               0                   15:53               ?                   00:00:00            tini -- node dist/main microservices
root                55121               55053               25                  15:53               ?                   00:00:18            node dist/main microservices
@bt90 ➜ /workspaces/immich (docker_init) $ docker top immich_machine_learning 
UID                 PID                 PPID                C                   STIME               TTY                 TIME                CMD
root                56214               56192               0                   15:54               ?                   00:00:00            tini -- python -m app.main
root                56265               56214               29                  15:54               ?                   00:00:08            python -m app.main
```

server/start.sh
```diff
@@ -25,4 +25,4 @@ if [ "$REDIS_PASSWORD_FILE" ]; then
 	unset REDIS_PASSWORD_FILE
 fi
 
-exec node dist/main $1
```
uhthomas, line 28: this should probably be in the `Dockerfile` instead. Something like

```Dockerfile
ENTRYPOINT ["tini", "--", "sh", "-c"]
```

bo0tzz:
Cool stuff! With `tini` in the Dockerfile, we don't need the changes to the docker-compose.yml anymore, right?

bt90:
Correct. I already reverted all compose related changes.

server/start.sh
bt90: Apart from the machine-learning container, the default entrypoint is `/bin/sh`. Changing this might affect cli or test execution via docker.

docker/docker-compose.prod.yml
jrasm91: To be fair, this isn't the docker compose file included in the release, but to build production like containers locally. There is no issue with changing this imo.

jrasm91: This change actually makes it consistent with the docker-compose.yml file.

machine-learning/Dockerfile
```diff
@@ -25,4 +27,4 @@ ENV NODE_ENV=production \
 
 COPY --from=builder /opt/venv /opt/venv
 COPY app .
-ENTRYPOINT ["python", "-m", "app.main"]
+ENTRYPOINT ["tini", "--", "python", "-m", "app.main"]
```
uhthomas, line 30: TBH idk why the python command is in the entry point. It should be in a CMD. Would it be possible to make it consistent with the other images?

server/start.sh
uhthomas: I think it will be fine, just feels weird to have the script use tini as it's only required for containers.

bt90: Ideally we should switch all these entrypoints to

```
ENTRYPOINT ["/usr/bin/tini", "--"]
```

The only downside would be that we'd have to tinker with the compose file and adjust the commands for server and machine-learning. e.g. `command: [ "./start.sh", "immich" ]`

uhthomas: I agree, but I don't see why any changes are required to the compose file.

bt90: The current commands in the compose file are missing the `./` prefix. You'd end up with this:

```
tini -- start.sh server
```

As `start.sh` is not on the path this will fail.

machine-learning/Dockerfile
bo0tzz: cc @mertalev ?

bt90: I agree. Separating the python bits into `CMD` is a lot cleaner.

```Dockerfile
ENTRYPOINT [ "tini", "--" ]
CMD [ "python", "-m", "app.main" ]
```

server/start.sh
bt90: Completely out of scope for this PR but why don't you create separate images for server and microservices? With the correct usage of `ARG` docker will reuse 99% of the image and we could drop the `command` from our compose file.

https://docs.docker.com/engine/reference/builder/#impact-on-build-caching

bt90:
The node and python based images are using the same entrypoint:

```
ENTRYPOINT ["tini", "--"]
```

I still had to change to compose file as explained in https://github.com/immich-app/immich/pull/3350#discussion_r1270233494

server/start.sh
uhthomas: Because it would be the exact same image. Would be wasteful and more complex for no real benefit.

server/Dockerfile
```diff
@@ -39,4 +39,4 @@ VOLUME /usr/src/app/upload
 
 EXPOSE 3001
 
-ENTRYPOINT ["/bin/sh"]
+ENTRYPOINT ["tini", "--"]
```
uhthomas, line 42: I think this can be changed to avoid modifying the docker compose file. The default entry point is "/bin/sh -c", so it should remain that way.


```suggestion
ENTRYPOINT ["tini", "--", "/bin/sh", "-c"]
```

bt90: The problem is how the CMD is passed:

```sh
docker inspect immich_server
```

```
[
    {
        "Id": "4978ed236c4578e552e72b276d7e6739e25e3a3c6d786f4bd699697206923d9e",
        "Created": "2023-07-21T13:04:46.1112611Z",
        "Path": "tini",
        "Args": [
            "--",
            "/bin/sh",
            "-c",
            "start.sh",
            "immich"
        ],
```

The result:
```sh
tini -- /bin/sh -c start.sh immich
```

```
immich: line 0: start.sh: not found
```

There are two problems:
1. the command passed to `-c` needs to be a single string like `'start.sh immich'`
2. `start.sh` is not on the path so this would still yield `/bin/sh: start.sh: not found`

bt90: I'll test without `-c`

bt90: Works :slightly_smiling_face:

bo0tzz (approved):

PR #3353, opened 2023-07-20
chore: Remove old startscripts

bt90:
## Description

as discussed in https://github.com/immich-app/immich/pull/3350#discussion_r1269514277

### DIFF
--- a/docker/docker-compose.prod.yml
+++ b/docker/docker-compose.prod.yml
@@ -7,7 +7,7 @@ services:
     build:
       context: ../server
       dockerfile: Dockerfile
-    command: ["./start-server.sh"]
+    command: [ "start.sh", "immich" ]
     volumes:
       - ${UPLOAD_LOCATION}:/usr/src/app/upload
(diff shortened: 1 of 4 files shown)


bt90:
@alextran1502 @jrasm91 this is a follow-up to #2689 which removes the old scripts. This should be part of the release changelog, as this will break installations that haven't yet adapted their configurations for the new start script.

jrasm91:
Do we need to make this breaking change now for any specific reason?

bt90:
Nope. Just a cleanup PR.

jrasm91:
Up to @alextran1502 if we want to do this now. The original idea was to group it in with the next release that required a docker compose change.

bt90:
#3350 *might* require changes to the compose file. Let's wait for the dust to settle.

jrasm91:
No breaking docker compose changes at the moment. This is easy enough to re-create in the future so I'm going to close it for now.

PR #3364, opened 2023-07-21
fix(mobile): Uses gray box placeholder for loading images by default and fixes odd spinner 

martyfuhry:
- Replaced this odd spinner image with a normal spinner that isn't thin and scaled poorly
image

- Uses gray box placeholder as default image loading (especially in memories)


martyfuhry:
I actually don't care for how this looks in the memories page. I'd prefer if it just faded in from nothing instead of grey.

martyfuhry:
OK, now there are two optional parameters in the `ImmichImage`:

```dart
      this.useGrayBoxPlaceholder = false,
      this.useProgressIndicator = false,
```

If neither is set, the image should just fade in (I think). If both are set, it will show a loading icon on top of a gray box placeholder. If one or the other is set, it will show one or the other. This is to give you more fine-grained control of what you want the widget to display while it loads.

alextran1502 (approved):

PR #3371, opened 2023-07-21
fix(cli): move runtime dependencies to correct section

oddlama:
Cleaned cli dependencies to ensure this will work after pruning developmen-only dependencies

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -2,9 +2,14 @@
   "name": "immich-cli",
   "dependencies": {
     "axios": "^1.4.0",
+    "byte-size": "^8.1.1",
+    "cli-progress": "^3.12.0",
+    "commander": "^11.0.0",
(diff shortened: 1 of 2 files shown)


etnoy (approved):

cli/package.json
```diff
@@ -2,9 +2,15 @@
   "name": "immich-cli",
   "dependencies": {
     "axios": "^1.4.0",
+    "byte-size": "^8.1.1",
+    "chai": "^4.3.7",
```
jrasm91, line 6: This is a testing library so it probably should not be here. If it doesn't compile there is probably another test file that needs to be excluded.

oddlama: Oh sorry, I missed that one, fixed! :D

jrasm91 (approved):

PR #3406, opened 2023-07-23
Update example.env to specify REDIS_PASSWORD_FILE

abhi-chakrab:
update to the example .env file to indicate docker secret support for REDIS_PASSWORD

### DIFF
--- a/docker/example.env
+++ b/docker/example.env
@@ -28,6 +28,7 @@ REDIS_HOSTNAME=immich_redis
 # Note: these parameters are not automatically passed to the Redis Container
 # to do so, please edit the docker-compose.yml file as well. Redis is not configured
 # via environment variables, only redis.conf or the command line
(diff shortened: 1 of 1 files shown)


jrasm91:
Can you add this to the environment documentation page in the docs/ instead? We're moving away from documenting in this example file.

abhi-chakrab:
Sure I can. But I couldn't figure out how to edit the documentation.

jrasm91:
In the repo, you can enter the docs/ folder, then run `npm i` and `npm start`. It'll serve the docs locally so you can edit any of the markdown pages and see what they look like.

Sounds like we need a docs page for how to contribute to the docs 😛

Have you used nodejs before?

jrasm91:
Actually it looks like this is doing the same thing.

https://github.com/immich-app/immich/pull/3402

abhi-chakrab:
> Actually it looks like this is doing the same thing.
> 
> #3402

Yesh looks like that. Should I close this PR?

jrasm91:
That's fine. Unless you have something additional to add.

abhi-chakrab:
Closing pr.

PR #3431, opened 2023-07-26
fix: people in shared assets

martabal:
Currently, when assets with people in it are shared in albums, users who don't own the asset fail to load people thumbnails. This PR removes the people information for those who don't own the asset.

Here an example with an account looking at a shared asset with a face in this asset : 
| Before | After |
| --- | --- |
| Screenshot from 2023-07-27 00-03-22 | Screenshot from 2023-07-27 00-03-42 |

### DIFF
--- a/server/src/immich/api-v1/asset/asset.service.ts
+++ b/server/src/immich/api-v1/asset/asset.service.ts
@@ -207,10 +207,10 @@ export class AssetService {
 
     const allowExif = this.getExifPermission(authUser);
     const asset = await this._assetRepository.getById(assetId);
-    let data = allowExif ? mapAsset(asset) : mapAssetWithoutExif(asset);
+    const data = allowExif ? mapAsset(asset) : mapAssetWithoutExif(asset);
 
     if (data.ownerId !== authUser.id) {
-      delete data.people;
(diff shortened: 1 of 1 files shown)


server/src/immich/api-v1/asset/asset.service.ts
```diff
@@ -207,12 +207,13 @@ export class AssetService {
...
     const allowExif = this.getExifPermission(authUser);
     const asset = await this._assetRepository.getById(assetId);
+    let data = allowExif ? mapAsset(asset) : mapAssetWithoutExif(asset);
 
-    if (allowExif) {
-      return mapAsset(asset);
-    } else {
-      return mapAssetWithoutExif(asset);
+    if (data.ownerId !== authUser.id) {
+      delete data.people;
```
jrasm91, line 213: Should we set it to an empty list instead or delete it?

jrasm91 (approved):
Nice

server/src/immich/api-v1/asset/asset.service.ts
martabal: `people` is optional so removing it is valid, but it's probably better to set it to an empty list

PR #3456, opened 2023-07-28
Clarify immich-admin command usage

mnahinkhan:
Some parts of the documentation state that administrative tasks like changing the admin password can be done via the `immich` function, but that is incorrect; instead, the `immich-admin` function should be used.


jrasm91 (approved):
We should update the screenshot or replace them with code blocks, but this is better than that we currently have!

alextran1502 (approved):

PR #3461, opened 2023-07-29
fix the grammar for albums to be backed up

Jacob-Roberts:
It looks like this is managed in a 3rd party service like lokalise or something. But this sounds better in English


gthb96:
I think #3682 changed it back to the wrong translation 🙃

PR #3486, opened 2023-07-31
fix(mobile): Improve vertical swipe detection in gallery viewer

mPyKen:
On my device, I have had hard times opening up the details section / closing the photo viewer via the swipe-up / -down gesture in the gallery viewer. Either my device reports smaller deltas more frequently or I am generally slow with my gestures^^ How is it on other devices?

This PR changes this behavior by looking at the direction of the gesture (ratio of y/x difference). I could have reduced the existing delta sensitivity, but I wasn't sure if that affects other devices negatively.


alextran1502 (approved):
Thank you

alextran1502 (approved):

PR #3513, opened 2023-08-02
refactor(web): asset grid state

jrasm91:
In this PR:
- Refactor asset grid state
- Only create an abort controller before starting an http request
- Track asset lookup map to bucket, bucket index, and asset index

Tested Scenarios:
- View an asset, hit next until the next time bucket loads (observed in the network tab)
- View an asset towards end of timeline, hit previous until a previous time bucket loads (observed in the network tab)
- View first asset in the timeline, observe clicking previous does nothing
- Scrub the timeline and observe relatively quick loading/rendering.
- Add an artificial delay to get asset by time bucket to verify abort controller is firing and cancelling requests when the bucket is no longer intersecting
- Delete photos, make sure they are removed from the grid
- Archive photos, make sure they are removed from the grid
- Select all photos, make sure the count matches the statistics for the user
- Select the first asset, jump a long way down, shift select an asset, then scroll up and observe everything is selected

### DIFF
--- a/web/src/lib/models/asset-grid-state.ts
+++ b/web/src/lib/models/asset-grid-state.ts
@@ -1,10 +1,24 @@
-import type { AssetResponseDto } from '@api';
+import { api, AssetCountByTimeBucket, AssetResponseDto } from '@api';
+import { writable } from 'svelte/store';
+import type { AssetStore } from '../stores/assets.store';
+import { handleError } from '../utils/handle-error';
 
 export enum BucketPosition {
   Above = 'above',
   Below = 'below',
   Visible = 'visible',
   Unknown = 'unknown',
+}
+
+export interface Viewport {
+  width: number;
+  height: number;
+}
+
+interface AssetLookup {
+  bucket: AssetBucket;
+  bucketIndex: number;
+  assetIndex: number;
 }
 
 export class AssetBucket {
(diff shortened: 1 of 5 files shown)


brighteyed (commented):
Brilliant! I'm a big fan of your refactoring PRs.

Looks like `handleSelectAssets` doesn't work as expected:
  - Open photos page and select one asset
  - Click on the scrollbar somewhere near the bottom and Shift-select another asset

Assets got selected, but `addGroupToMultiselectGroup` was not called for intermediate date groups

brighteyed:
> Brilliant! I'm a big fan of your refactoring PRs.
> 
> Looks like `handleSelectAssets` doesn't work as expected:
> 
> * Open photos page and select one asset
> * Click on the scrollbar somewhere near the bottom and Shift-select another asset
> 
> Assets got selected, but `addGroupToMultiselectGroup` was not called for intermediate date groups

`getBucketIndexByAssetId` returns `null` instead of `0` in my case:
image

brighteyed (approved):

web/src/lib/models/asset-grid-state.ts
```diff
@@ -15,44 +29,220 @@ export class AssetBucket {
...
+    if (bucketIndex === this.buckets.length - 1) {
+      return null;
+    }
+
+    const nextBucket = this.buckets[bucketIndex + 1];
+    await this.loadBucket(nextBucket.bucketDate, BucketPosition.Unknown);
+    return nextBucket.assets[0]?.id || null;
+  }
+
+  private emit(recalculate: boolean) {
```
alextran1502, line 231: Is this for updating the state?

web/src/lib/stores/assets.store.ts
```diff
@@ -1,268 +1,38 @@
...
   return {
-    setInitialState,
-    getAssetsByBucket,
-    removeAsset,
-    updateBucketHeight,
-    cancelBucketRequest,
-    getAdjacentAsset,
-    updateAsset,
-    subscribe,
+    init: store.init.bind(store),
```
alextran1502, line 28: Can you help explain what does the `bind` method do here?

web/src/lib/models/asset-grid-state.ts
jrasm91: Yes. It updates the svelte store.

web/src/lib/stores/assets.store.ts
jrasm91: The `this` in a method is "whatever object called the method". It's kind of complicated, but basically 

```
on:click={obj.foo}
on:click={() => obj.foo()}
```

Have different values for `this`. Only the second one would work.

jrasm91: Bind makes a method always use the same `this` value, which is passed into `bind`.

PR #3521, opened 2023-08-02
Local merged videos should now play locally instead of loading remote

### LINKED ISSUE #3523: Locally present video is downloaded from server when opened on mobile
### The bug

On the Android app, when clicking on a video that's already downloaded on the device (the cloud icon with a tick inside is shown), it seems to download the video from the remote server, potentially using mobile data. I first noticed this on the network usage indicator on the status bar, but checking on Immich's mobile data usage from Android Settings before & after opening the video did indeed confirm this. It only seems to affect videos, images don't use data & load fine without an internet connection.

There are settings in the mobile app for "Load original image" and "Prefer remote images", both of them are off.

Ubuntu 22.04

### Version of Immich Server

v1.71.0

v1.71.0

### Platform with the issue

- [ ] Server
- [ ] Web
- [X] Mobile

```YAML
Not relevant
```

### Your .env content

```Shell
Not relevant
```

### Reproduction steps

```bash
1. Upload any video to Immich
2. Make sure it's present locally
3. Open the video from the mobile app
```

### Additional information

_No response_

martyfuhry:
Fixes a bug where an `AssetState.merged` video asset would prefer to play the remote video rather than the local one.

Fixes #3523

### DIFF
--- a/mobile/lib/modules/asset_viewer/views/video_viewer_page.dart
+++ b/mobile/lib/modules/asset_viewer/views/video_viewer_page.dart
@@ -34,7 +34,7 @@ class VideoViewerPage extends HookConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    if (asset.storage == AssetState.local && asset.livePhotoVideoId == null) {
+    if (asset.isLocal && asset.livePhotoVideoId == null) {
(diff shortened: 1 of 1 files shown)


PandaDriver156:
Should fix #3523

alextran1502 (approved):

PR #3529, opened 2023-08-03
chore: remove without thumbs

jrasm91:
Both the mobile app and web app now handle assets without thumbnails. This PR removes the query param and changes the endpoint to always return all assets regardless of thumbnail status.

Tested Scenarios:
- View asset without thumbnail on the timeline
- View asset without thumbnail on the archive page
- View asset without thumbnail on the favorites page

Not tested - mobile app :-)


jrasm91:
This is a pretty trivial PR. Any chance you could take a look at it in the next few days @alextran1502 ?

PR #3532, opened 2023-08-03
fix(server): ffprobe not found

jrasm91:


mertalev:
Ah, the Dockerfile only adds ffmpeg to `PATH` in prod. This approach is also fine though, so so I think you can remove the `PATH` line if you stick with linking in the script.

mertalev (approved):

alextran1502 (approved):

PR #3536, opened 2023-08-03
fix(web): scrollbar

jrasm91:
In this PR:
- Fix issue where scroll top becomes hidden
- Fix issue where scroll line does not go all the way to the bottom
- Fix issue where dragging outside of the asset grid (and then coming back) stops scrolling the timeline
- Refactor click offsets (derived from the delta between window height and viewport height)
- Scrollbar component automatically hide's itself now based on the asset store timeline height
- Simplify interface between scrollbar and asset grid
- Make it more obvious how the virtual scrollbar position is translated back and forth to the virtual position

| Current | Previous |
| :---------------------- | :-------------------- |
|image| image
|image|image

### DIFF
--- a/web/src/lib/components/photos-page/asset-grid.svelte
+++ b/web/src/lib/components/photos-page/asset-grid.svelte
@@ -8,10 +8,7 @@
   import AssetViewer from '../asset-viewer/asset-viewer.svelte';
   import IntersectionObserver from '../asset-viewer/intersection-observer.svelte';
   import Portal from '../shared-components/portal/portal.svelte';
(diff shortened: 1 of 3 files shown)


alextran1502 (approved):

brighteyed:
The code looks good to me, but I have never debugged the scrollbar or improved it, so I'm not familiar with edge cases. 
While testing, I encountered an issue where dragging the scrollbar from the bottom up causes the mouse pointer to move to the top edge of my screen by itself. I managed to reproduce this issue on the main branch, but on this branch it is much easier

PR #3545, opened 2023-08-04
refactor(ml): model downloading

mertalev:
## Description

This PR distinguishes between downloading and loading models. Specifically, it adds a new `download` method that can be used independently from the `load` method for each model type. 

There are several useful benefits to this:
- Lazy loading of models without compromising on the ability to download models at startup
- Simplified startup logic
    - Model classes can be instantiated freely since they can handle eager/lazy loading internally
- Faster startup times
    - Insightface's logic for loading models is very inefficient: it loads every model in the folder and inspects their inputs and outputs, then unloads the ones that aren't relevant. 
    - The new implementation only loads relevant models, and in fact only extracts these models from the zip file on download.

## How Has This Been Tested?

- No errors on startup
- Deleting the model cache and restarting downloads each model *once*
- After restarting the container, it does not download any models
- Object tagging, CLIP and facial recognition jobs run as normal

### DIFF
--- a/machine-learning/app/config.py
+++ b/machine-learning/app/config.py
@@ -20,7 +20,7 @@ class Settings(BaseSettings):
     min_face_score: float = 0.7
     test_full: bool = False
 
-    class Config(BaseSettings.Config):
+    class Config:
         env_prefix = "MACHINE_LEARNING_"
         case_sensitive = False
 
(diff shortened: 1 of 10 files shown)


mertalev:
I should probably update the tests...

jrasm91 (approved):
Nice work!

alextran1502 (approved):

Issue #3154, opened 2023-07-08
[BUG] Files without exif offset_time attribute are assumed to contain utc date/time instead of local timezone


JayBigGuy10:
### The bug

If I upload a photo taken by a device which stores the local timezone time/date directly into the exif date taken fields, and doesn't store a offset_time, the server incorrectly applies an offset.

For example a photo that I took when it was 2022/08/02 16:18 in new Zealand, appears on the server as if it was taken 13 hours later because the server believes the 2022/08/02 16:18 datetime to be UTC(+0) which needs offsetting, not the NZDST(+13) that it actually is

I have looked into various photos exif dumps and this happens with the Samsung Galaxy S9 and my OnePlus Nord (Android 10) photos from my hiking group which store local timezone, but not with the Samsung S10+, Samsung S22+ and Olympus TG-6 camera which store UTC/offset.

I guess the server will have to somehow assume that for these cases that the photos were taken in the servers current TZ?

### The OS that Immich Server is running on

Docker Desktop (WSL Backend) For Windows 10

### Version of Immich Server

v1.66.1

### Version of Immich Mobile App

1.66.0 Build 89

### Platform with the issue

- [X] Server
- [X] Web
- [X] Mobile

### Your docker-compose.yml content

```YAML
version: "3.8"

services:
  immich-server:
    container_name: immich_server
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    command: [ "start.sh", "immich" ]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - C:\docker\immich\read-only:/mnt/media/read-only
    env_file:
      - .env
    depends_on:
      - redis
      - database
      - typesense
    restart: always

  immich-microservices:
    container_name: immich_microservices
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    command: [ "start.sh", "microservices" ]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - C:\docker\immich\read-only:/mnt/media/read-only
    env_file:
      - .env
    depends_on:
      - redis
      - database
      - typesense
    restart: always

  immich-machine-learning:
    container_name: immich_machine_learning
    image: ghcr.io/immich-app/immich-machine-learning:${IMMICH_VERSION:-release}
    volumes:
      - model-cache:/cache
    env_file:
      - .env
    restart: always

  immich-web:
    container_name: immich_web
[...]

Issue #2000, opened 2023-03-14
[BUG] Invalid timezone in DSLR photos uploaded via CLI


etnoy:
### The bug

Just spent a few hours digging into why my DSLR photos were one hour off in time compared to my mobile photos.

Interestingly enough, this issue has plagued me over at google photos, too.

In the end, it was due to the microservice container having its timezone set to UTC.

But wait, we are using exif-vendored so this shouldn't be an issue? Alex himself says so here: https://github.com/immich-app/immich/issues/1519#issuecomment-1414953736

But alas, timezones still matter, because exif-vendored has to rely on local timezone information in some special cases, just like the DSLR photos.

The issue was finally solved for me by setting a timezone environment variable in the microservices container corresponding to my local timezone.

The obvious problem is that setting timezones in docker containers is messy, so it might not be an ideal solution.

Another solution is to move more of the timezone logic to somewhere that has native timezone information, like the CLI. Right now, the CLI uses a different exif library but I published a PR that makes the logic more consistent with immich itself:  too: https://github.com/immich-app/CLI/pull/69 

Note that the computed timezone data from the CLI is not even considered in [the current logic](https://github.com/immich-app/immich/blob/2ca560ebf80aadf1f7904bd88d7447d9a2a59ed3/server/apps/microservices/src/processors/metadata-extraction.processor.ts#L170) in most cases.

### The OS that Immich Server is running on

Ubuntu 22.04 via docker

### Version of Immich Server

1.50.1

### Version of Immich Mobile App

1.50.0 73

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
version: "3.8"

services:
  immich-server:
    container_name: immich_server
    image: altran1502/immich-server:release
    entrypoint: [ "/bin/sh", "./start-server.sh" ]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
    env_file:
      - .env
    environment:
      - NODE_ENV=production
      - TZ=Europe/Stockholm
    depends_on:
      - redis
      - database
    restart: always

  immich-microservices:
    container_name: immich_microservices
    image: altran1502/immich-server:release
    entrypoint: [ "/bin/sh", "./start-microservices.sh" ]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
    env_file:
      - .env
    environment:
      - NODE_ENV=production
      - TZ=Europe/Stockholm
    depends_on:
[...]

PR #3555, opened 2023-08-04
Update example.env to show that TZ can be set to fix image offset issues

JayBigGuy10:
Add timezone referenced from the FAQ and Environment Variables doc pages

https://github.com/immich-app/immich/issues/3154

https://github.com/immich-app/immich/issues/2000


alextran1502:
Hello, we are no longer trying to update the `.env` example in an effort to keep it easy to navigate. 

The environment variable is already available in the corresponding page 

image

PR #3565, opened 2023-08-05
refactor(server, web)!: store latest immich version available on the server

martabal:
> **warning**
> Breaking Change - the websocket message format has changed, so some functions will not work properly unless the server and mobile client are on the same version.
<hr>

Currently, every time the admins hit refresh, they hit the GitHub API to retrieve the latest Immich version. If the version on GitHub differs from the server version, the web and mobile clients will notify the admins about the availability of a new Immich version.

However, excessive web refreshes ( by consequences on the GitHub API too ) can lead Github to ban your public IP address. To mitigate this risk and improve the users privacies, this PR move the logic from the clients to the server. Now, the server will fetch the latest Immich version from GitHub every 1 hour (not configurable for now).

WebUI changes : 

Screenshot from 2023-08-12 11-26-38


server/src/domain/system-config/system-config.service.ts
```diff
@@ -53,6 +79,70 @@ export class SystemConfigService {
...
+        minor: parseInt(minor, 10),
+        patch: parseInt(patch, 10),
+      };
+
+      const config = await this.core.getConfig();
+      if (
+        GithubVersion.major > serverVersion.major ||
+        GithubVersion.minor > serverVersion.minor ||
+        GithubVersion.patch > serverVersion.patch
+      ) {
```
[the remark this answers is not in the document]
martabal: Indeed ! Maybe if only one is different ?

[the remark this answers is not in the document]
bo0tzz: There should be libraries out there that implement semver comparison. The logic isn't particularly complex, but it might be cleaner to just pull one of those in?

jrasm91:
Can we just set up a cron to run every hour or so and then store the updated version info in memory instead?

martabal:
I chose an interval rather than a cron, because I'm afraid that every x hours all users make a request to GitHub at the same time which would stress GitHub servers

martabal:
And yeah it's probably better to just store it in memory

server/src/domain/system-config/system-config.service.ts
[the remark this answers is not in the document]
mertalev: You can change the `serverVersion` in `domain.constant.ts` to be a class like so:
```
export class ServerVersion implements IServerVersion {
  major: number;
  minor: number;
  patch: number;

  constructor(major: string | number, minor: string | number, patch: string | number) {
    this.major = Number(major);
    this.minor = Number(minor);
    this.patch = Number(patch);
  }

  toString() {
    return `${this.major}.${this.minor}.${this.patch}`;
  };

  valueOf() {
    return this.major * 100000 + this.minor * 1000 + this.patch;
  }
}
```

If you use this class for the GH version as well, then you can compare in two ways.

1. `githubVersion > serverVersion`
    - Very elegant, but it assumes the minor and patch versions will be less than 100)
2. `githubVersion.toString().localeCompare(serverVersion.toString(), undefined, {numeric: True}) === 1`
    - Less readable, but the `numeric` property will ensure this works correctly in all cases (see [here](https://stackoverflow.com/a/55466325))

jrasm91:
What do you think about storing the current release information on immich.app? We would not need to worry about rate limits afaik

server/src/domain/system-config/system-config.service.ts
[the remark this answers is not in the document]
jrasm91: Just use this and call it a day. Almost 300k weekly downloads. It's the implementation npm uses.

martabal:
> What do you think about getting the current release information from immich.app? We would not need to worry about rate limits afaik

I think we'll need to have rate limits for immich.app,if we use it to retrieve the latest immich version. Plus it adds an intermediary: when releasing a new version, the version stored on immich.app will always be a few minutes out of date. I don't mind using immich.app, but it will probably need extra work

jrasm91:
We have a workflow that runs on release and the docker images are not available until a few minutes later anyways, so that doesn't seem like much of a problem. Afaik there are no rate limits on Vercel, especially for serving static content.

martabal:
@jrasm91 Should the server make an API request right after the startup ?

server/src/domain/system-config/system-config.service.ts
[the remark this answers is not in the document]
martabal: If Immich keeps the current semantic versionning, it's probably better to avoid an extra dependency ? I like a lot the second option proposed by @mertalev

bo0tzz:
I think something's gone wrong with your merge from main, some of the things that have been added there recently are showing up as new in your diff.

server/src/domain/domain.config.ts
```diff
@@ -29,7 +29,8 @@ export const immichAppConfig: ConfigModuleOptions = {
     MACHINE_LEARNING_PORT: Joi.number().optional(),
     MICROSERVICES_PORT: Joi.number().optional(),
     SERVER_PORT: Joi.number().optional(),
-    DISABLE_CHECK_LATEST_VERSION: Joi.boolean().optional(),
+    CHECK_NEW_VERSION_INTERVAL: Joi.boolean().optional(),
+    DISABLE_CHECK_LATEST_VERSION: Joi.number().optional(),
```
bo0tzz, line 33: These could be combined into one by having `CHECK_NEW_VERSION_INTERVAL=0` mean that it is disabled.

server/src/domain/user/user.service.ts
```diff
@@ -86,8 +86,8 @@ export class UserService {
     }
 
     if (this.systemConfigService.availableVersion)
-      return { availableVersion: this.systemConfigService.availableVersion, available: true };
-    else return { available: false };
+      return { availableVersion: this.systemConfigService.availableVersion };
+    else return { availableVersion: null };
   }
```
bo0tzz, line 91: Do you need the if here? Can't this whole thing just be `return { availableVersion: this.systemConfigService.availableVersion };`?

server/src/domain/domain.config.ts
martabal: Great idea !

server/src/domain/job/job.constants.ts
```diff
@@ -37,6 +37,9 @@ export enum JobName {
   USER_DELETION = 'user-deletion',
   USER_DELETE_CHECK = 'user-delete-check',
 
+  // server version
+  LATEST_IMMICH_VERSION = 'latest-immich-version',
```
jrasm91, line 41: Is this used anywhere?

server/src/domain/job/job.repository.ts
```diff
@@ -70,6 +71,9 @@ export type JobItem =
   // Asset Deletion
   | { name: JobName.PERSON_CLEANUP; data?: IBaseJob }
 
+  // Immich Version
+  | { name: JobName.LATEST_IMMICH_VERSION; data?: GithubRelease }
```
jrasm91, line 75: data it not actually used here. You could probably just put `IBaseJob` or possibly `never`

server/src/domain/job/job.interface.ts
```diff
@@ -4,6 +4,10 @@ export interface IBaseJob {
   force?: boolean;
 }
 
+export type GithubRelease = {
```
jrasm91, line 7: This type belongs wherever the http call is made.

server/src/domain/system-config/system-config.service.ts
```diff
@@ -1,6 +1,8 @@
-import { Inject, Injectable } from '@nestjs/common';
+import { Inject, Injectable, Logger } from '@nestjs/common';
+import axios from 'axios';
```
jrasm91, line 2: We should not be importing `axios` directly inside of `domain/` this should be replaced with a repository and the implementation can exist in `infra/`

server/src/domain/system-config/system-config.service.ts
```diff
@@ -17,11 +19,24 @@ import { SystemConfigCore, SystemConfigValidator } from './system-config.core';
...
   constructor(
     @Inject(ISystemConfigRepository) repository: ISystemConfigRepository,
     @Inject(IJobRepository) private jobRepository: IJobRepository,
   ) {
+    this.availableVersion = null;
     this.core = new SystemConfigCore(repository);
+    this.proxyHost = process.env.PROXY_HOST || null;
+    this.proxyProtocol = process.env.PROXY_PROTOCOL || null;
+    this.proxyPort = Number(process.env.PROXY_PORT) || null;
+    this.disableCheckLatestVersion = process.env.CHECK_NEW_VERSION_INTERVAL === '0';
```
jrasm91, line 39: Can we make this configurable via the UI instead?

server/src/domain/system-config/system-config.service.ts
```diff
@@ -53,6 +73,48 @@ export class SystemConfigService {
...
+        : false;
+
+    try {
+      const { data } = await axios.get<GithubRelease>(
+        'https://api.github.com/repos/immich-app/immich/releases/latest',
+        {
+          headers: {
+            Accept: 'application/vnd.github.v3+json',
+          },
+          proxy,
```
jrasm91, line 96: Why is it necessary to expose `proxy` settings here? I think there might be out of the box environment variables to configure this automatically `http_proxy` and `https_proxy`.

server/src/domain/system-config/system-config.service.ts
```diff
@@ -53,6 +73,48 @@ export class SystemConfigService {
...
+      const [, major, minor, patch] = data.tag_name.match(/v(\d+)\.(\d+)\.(\d+)/) || [];
+
+      const temp = new ServerVersion(major, minor, patch);
+
+      if (temp.toString().localeCompare(serverVersion.toString(), undefined, { numeric: true }) === 1) {
+        this.logger.debug(`New version Immich version available : ${temp.toString()}`);
+        this.availableVersion = temp;
+      } else {
+        this.logger.debug('No new Immich version available detected');
+      }
```
jrasm91, line 109: This code probably belongs in some utility file, which can have some separate tests for it directly.

server/src/domain/system-config/system-config.service.ts
```diff
@@ -53,6 +73,48 @@ export class SystemConfigService {
...
+
+      return true;
+    } catch (error) {
+      console.error('Error occurred:', error);
```
jrasm91, line 113: We should use logger not console for logging.

web/src/lib/components/shared-components/version-announcement-box.svelte
```diff
@@ -15,22 +16,30 @@
...
     try {
-      githubVersion = await getGithubVersion();
+      const { data } = await api.userApi.getLatestImmichVersionAvailable();
+      if (data.availableVersion) {
+        githubVersion = semverToName({
+          major: data.availableVersion.major,
+          minor: data.availableVersion.minor,
+          patch: data.availableVersion.patch,
+        });
+      }
```
jrasm91, line 35: IMO the server should be the one that determines if a new version is available. I don't think it makes sense for each client to have to manually do the computation. We could just return something like:

```typescript
{
  isAvailable: boolean;
  releaseType: 'patch' | 'minor' | 'major' | 'none',
  currentVersion: IServerVersion;
  releaseVersion: IServerVersion;
  
}
```

web/src/lib/components/shared-components/version-announcement-box.svelte
```diff
@@ -15,22 +16,30 @@
...
+
       if (localStorage.getItem('appVersion') === githubVersion) {
         // Updated version has already been acknowledged.
         return;
       }
 
-      if (githubVersion !== serverVersionName) {
+      if (data.availableVersion && !disable_check && !import.meta.env.DEV) {
```
jrasm91, line 42: Does this work? Only environment variables that start with PUBLIC_* are available on the web afaik.

web/src/lib/components/shared-components/version-announcement-box.svelte
```diff
@@ -1,12 +1,13 @@
...
-  import { getGithubVersion } from '$lib/utils/get-github-version';
   import { onMount } from 'svelte';
   import FullScreenModal from './full-screen-modal.svelte';
-  import type { ServerVersionReponseDto } from '@api';
+  import { api, type ServerVersionReponseDto } from '@api';
   import Button from '../elements/buttons/button.svelte';
 
   export let serverVersion: ServerVersionReponseDto;
 
+  const disable_check: boolean = import.meta.env.CHECK_NEW_VERSION_INTERVAL === '0';
```
jrasm91, line 9: Does this work? I don't think non PUBLIC_* environment variables are available in this context. Also, can't the logic to skip the check happen in the server? Like it can ignore the update check and always return false. I don't think we need to have a check client side and a check server side for this feature.

web/src/lib/components/shared-components/version-announcement-box.svelte
martabal: It doesn't. I wanted to use it just to make sure, but it's useless

server/src/domain/job/job.constants.ts
martabal: No

server/src/domain/system-config/system-config.service.ts
martabal: Yes, it would be better !

martabal:
@jrasm91 Do you know why the mobile unit tests fail ? They didn't fail before the merge

jrasm91:
Numbers in the dots need to have the API decorator and type int otherwise the dart code is wrong. You can look at any other number field in a dto to see what I mean.

martabal:
> Numbers in the dots need to have the API decorator and type int otherwise the dart code is wrong. You can look at any other number field in a dto to see what I mean.

Thanks !😄

server/src/domain/system-config/system-config.service.ts
martabal: What's the best between configuring the proxy in the Immich system settings and with an environment variable ?

alextran1502 (commented):
Can you help explain what scenario proxy can be used in?

martabal:
It is to improve privacy, one of the main benefit is to hide your server ip address, but this is probably not essential

server/src/domain/system-config/system-config.service.ts
jrasm91: I'd say the UI whenever possible. If we're going to add a proxy thing, can we make it a single string/url?

web/src/lib/components/admin-page/settings/proxy/proxy-settings.svelte
```diff
@@ -0,0 +1,143 @@
...
+  async function saveSetting() {
+    try {
+      const { data: current } = await api.systemConfigApi.getConfig();
+
+      if (!current.oauth.enabled && current.proxy.enabled && !proxyConfig.enabled) {
+        const confirmed = await openConfirmModal();
+        if (!confirmed) {
+          return;
+        }
+      }
```
jrasm91, line 50: Why is there OAuth stuff here? Why do you need a confirm modal here?

martabal: It was my template for the proxy settings, I didn't finish at all this part

web/src/lib/components/shared-components/version-announcement-box.svelte
martabal: what is releaseType for?

martabal:
Removed the proxy implementation as it's not essential for this PR and needs more discussion

server/src/domain/domain.constant.ts
```diff
@@ -12,12 +12,23 @@ export interface IServerVersion {
...
+    this.minor = Number(minor);
+    this.patch = Number(patch);
+  }
+
+  toString() {
+    return `${this.major}.${this.minor}.${this.patch}`;
+  }
+}
+
+export const serverVersion: ServerVersion = new ServerVersion(major, minor, patch);
```
jrasm91, line 31: If we're going to use this class, can you replace the `SERVER_VERSION` constant below to just to `serverVersion.toString()` instead?

server/src/domain/system-config/dto/system-config.dto.ts
```diff
@@ -44,6 +45,11 @@ export class SystemConfigDto implements SystemConfig {
   @ValidateNested()
   @IsObject()
   thumbnail!: SystemConfigThumbnailDto;
+
+  @Type(() => SystemConfigCheckAvailableVersionDto)
+  @ValidateNested()
+  @IsObject()
+  checkAvailableVersion!: SystemConfigCheckAvailableVersionDto;
```
jrasm91, line 52: Can we just call this `versionCheck` or `newVersionCheck` instead?

server/src/domain/system-config/system-config.core.ts
```diff
@@ -128,7 +133,11 @@ export class SystemConfigCore {
 
   public config$ = singleton;
 
-  constructor(private repository: ISystemConfigRepository) {}
+  public schedulerRegistry: SchedulerRegistry;
+
+  constructor(private repository: ISystemConfigRepository) {
+    this.schedulerRegistry = new SchedulerRegistry();
+  }
```
jrasm91, line 140: A few things here. We should not create the registry class directly, we should just inject it into the constructor. Also, we try to only inject repositories that can easily be mocked and are technology independent. I'm not 100% sure we need a repository for a basic set interval though. I'd be fine using it directly. It's only setup and access in one place anyway, so using the scheduler seems overly complicated. I think this is the wrong class too. It should be implemented in server-info.service probably, since this class is all about storage and retrieval of configuration not what they're used for.

server/src/domain/system-config/system-config.repository.ts
```diff
@@ -7,4 +8,5 @@ export interface ISystemConfigRepository {
   readFile(filename: string): Promise<Buffer>;
   saveAll(items: SystemConfigEntity[]): Promise<SystemConfigEntity[]>;
   deleteKeys(keys: string[]): Promise<void>;
+  getLatestAvailableVersion(): Promise<GithubRelease>;
```
jrasm91, line 11: Getting the latest release information from github has nothing to do with system configuration.

server/src/domain/system-config/system-config.util.spec.ts
```diff
@@ -0,0 +1,82 @@
+import { ServerVersion } from '..';
```
jrasm91, line 1: What do you think about adding the `compareVersions` method to the `ServerVersion` class instead, extracting it into a separate file `domain/server-info/server-version.ts` and having a separate `server-version.spec.ts` where all these tests live?

server/src/domain/user/response-dto/user-available-version.dto.ts
```diff
@@ -0,0 +1,17 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export class AvailableVersionResponseDto {
```
jrasm91, line 3: This data doesn't really have anything to do with the user resource, so I find it a bit weird to return it here. We already have `GET /server-info/version`, so maybe something like `GET /server-info/available-version` would make more sense.

server/src/domain/user/user.service.spec.ts
```diff
@@ -9,6 +9,7 @@ import {
   newUserRepositoryMock,
 } from '@test';
 import { when } from 'jest-when';
+import { SystemConfigService } from '..';
```
jrasm91, line 12: Don't import from parent directory

server/src/domain/user/user.service.ts
```diff
@@ -76,6 +80,15 @@ export class UserService {
     return mapUser(createdUser);
   }
 
+  async latestImmichVersionAvailable(): Promise<AvailableVersionResponseDto> {
```
jrasm91, line 83: Move this to server-info service

server/src/immich/app.service.ts
```diff
@@ -22,9 +61,22 @@ export class AppService {
...
+      );
+      this.configCore.schedulerRegistry.addInterval('check-available-version', interval);
+    }
   }
 
   async destroy() {
     this.searchService.teardown();
+    if (this.configCore.schedulerRegistry.doesExist('interval', 'check-available-version')) {
+      this.configCore.schedulerRegistry.deleteInterval('check-available-version');
+    }
```
jrasm91, line 80: Search service uses an interval the same way you should use it. No logic should exist here, just call `service.init()` and `service.destroy()` and move the logic into `domain/`.

server/src/infra/repositories/system-config.repository.ts
```diff
@@ -20,6 +25,11 @@ export class SystemConfigRepository implements ISystemConfigRepository {
     return this.repository.save(items);
   }
 
+  async getLatestAvailableVersion(): Promise<GithubRelease> {
+    const { data } = await axios.get<GithubRelease>('https://api.github.com/repos/immich-app/immich/releases/latest');
+    return data;
+  }
+
```
jrasm91, line 32: This is perfect adding it to a repository method, but I don't think it should be in system config. I don't think we have an existing repository that is appropriate for this either. Can you add a new one, like `IServerInfoRepository` or `IGitHubRepository`?

martabal:
@jrasm91 Thanks for the review ! I will finish this PR next week

server/src/domain/system-config/system-config.service.ts
```diff
@@ -13,14 +14,21 @@ import {
...
+  public availableVersion: ServerVersion | null;
+  public dateCheckAvailbleVersion: number | null;
+  private logger = new Logger();
+
   constructor(
-    @Inject(ISystemConfigRepository) repository: ISystemConfigRepository,
+    @Inject(ISystemConfigRepository) private repository: ISystemConfigRepository,
     @Inject(IJobRepository) private jobRepository: IJobRepository,
   ) {
+    this.dateCheckAvailbleVersion = null;
```
danieldietzler, line 30: Typo: `dateCheckAvailableVersion` :)

server/src/domain/system-config/system-config.service.ts
```diff
@@ -53,6 +61,24 @@ export class SystemConfigService {
...
+        return true;
+      } else {
+        this.logger.debug('No new version detected');
+      }
+    } catch (error) {
+      this.logger.error('Error occurred:', error);
+      return false;
+    }
+  }
+
```
danieldietzler, line 81: You should return false after the try catch instead of in l. 78. Currently, the method will return undefined if no new version is detected.

server/src/domain/system-config/system-config.util.ts
```diff
@@ -0,0 +1,16 @@
+import { ServerVersion } from '../domain.constant';
+
+export function compareVersions(remoteVersionString: string, serverVersion: ServerVersion): boolean {
+  const remoteVersion = stringToVersion(remoteVersionString);
+  return remoteVersion.toString().localeCompare(serverVersion.toString(), undefined, { numeric: true }) === 1;
```
danieldietzler, line 5: Instead of checking `=== 1` you should do either `!!remoteVersion.toString().localeCompare(...)` or `Boolean(remoteVersion.toString().localeCompare(...))`

server/src/domain/system-config/system-config.util.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export function stringToVersion(version: string): ServerVersion {
+  const matchResult = version.match(/(?:v)?(\d+)\.(\d+)\.(\d+)/i);
```
danieldietzler, line 9: Probably just nitpicking but I personally prefer named capture groups: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Groups_and_backreferences#using_named_groups

server/src/domain/system-config/system-config.util.ts
martabal: I don't think it works, if the server version is bigger than the remote version `remoteVersion.toString().localeCompare(...) or Boolean(remoteVersion.toString().localeCompare(...))` returns -1 which is transforms to `true`

server/src/infra/repositories/communication.repository.ts
```diff
@@ -34,7 +41,7 @@ export class CommunicationRepository implements OnGatewayConnection, OnGatewayDi
   }
 
   send(event: CommunicationEvent, userId: string, data: any) {
-    this.server.to(userId).emit(event, JSON.stringify(data));
+    this.server.to(userId).emit(event, data);
```
jrasm91, line 44: @alextran1502 @shenlong-tanwen - I noticed that `broadcast` did not use `JSON.stringify`, while `send` did. I changed both to omit calling `JSON.stringify`, can one of you help me validate the mobile app events still work properly with this change? Thanks.

jrasm91 (approved):
@martabal I simplified things quite a bit, hopefully they are all OK.
- The check runs on start up
- The check runs on a cron job every hour and does not deal with `setInterval` or the scheduler registry anymore
- The user gets an event pushed to them when they connect via a web socket
- The user gets an event pushed to them when there is a new version available
- The check is disabled when `NODE_ENV=development`

bo0tzz (changes requested):
Like Martin said, we should be very sure that we don't have all the instances hammering github at the exact same moment.

jrasm91:
I don't think it would be an issue personally, but it was easy enough to changed it to an interval (which is still declarative) and make it dependent on server start-up time instead.

bo0tzz (approved):
Thanks!

PR #2834, opened 2023-06-18
Add support for many missing raw formats

elliotclee:
Now that libraw support is in (thanks @uhthomas), as per issue #2745 this PR adds uploading of most of the raw photo file formats supported by libraw. While not every last possible format was added (libraw does not have a definitive list of what it supports), I got all the ones mentioned by gnome-raw-thumbnailer and most of the ones mentioned by digikam. Also allows uploading of .avif files (#2803 ).

Feedback welcome!

PR #3647, opened 2023-08-11
Allows for correct production of thumbnails and metadata extraction from Canon CR2 raw files.

coissac:
At the end of the pull request #2834, I discussed the non-generation of thumbnails for CR2 files.
Here is a proposition for a patch affecting the generation of thumbnails and the metadata extraction.

The patch relies on two packages 
- `js-img-type` : for identifying CR2 files
- `dcraw` : for "pre-converting" the CR2 file into a true TIFF buffer.

I tested the patch on the dev environment installed on an orbstack docker environment running on a M1 mac.
It allows correct generation of thumbnails with CR2 file import without compromising import of other formats (tested with JPEG and CR3 raw format, which are OK).

I'm not a JavaScript/typescript developer, hence perhaps a more elegant code style is possible ;-).
But at least that gives you a direction.

All the best

Eric


coissac:
Perhaps #3648 is a more general way for doing the same, just by extending the list of unsupported formats with the mime type of CR2 file.

andreybavt:
I haven't tried the metadata extraction, but thumbnails worked fine after I simply added 

`'image/cr2'` to the `unsupportedSharpTypes` in https://github.com/immich-app/immich/pull/3648

image

alextran1502:
Replaced by #3648

PR #3652, opened 2023-08-12
New FAQ entry: how move assets,albums,persons from one account to another

bmaster001:
Added an entry to document what can be done to move photos, albums and persons from one account to another

### DIFF
--- a/docs/docs/FAQ.md
+++ b/docs/docs/FAQ.md
@@ -102,7 +102,7 @@ If you are using `My Photo Stream`, the Photos app temporarily creates duplicate
 
 ### How can I move all data (photos, persons, albums) from one user to another?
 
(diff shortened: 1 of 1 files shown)


bo0tzz:
Thanks for the PR! I have two notes:
1. If any of the duplicate assets (that do not have their owner changed) are in albums, that entry will end up broken in some way. I'm not sure if there's an easy way around that though.
2. It might be good to add a warning note to this. Maybe something like "This is an advanced operation. If you can't to do it with the steps described here, this is not for you."  ?

bmaster001:
About note 1: I guess it's indeed possible that you end up with an album owned by 'destination user', that contains assets that still belong to the 'source user'. Not sure how Immich will react to such a situation... ?  But if you start from the idea that all those left-overs will be removed afterwards (they are duplicates anyway), this is no big deal?

jrasm91 (approved):

PR #3658, opened 2023-08-13
feat(server): wide gamut thumbnails

### LINKED ISSUE #3136: AdobeRGB images are dull
### The bug

It seems like images within the AdobeRGB colour space have dull thumbnails.

See this side-by-side comparison of the same image rendered by Geeqie and Immich.

image

https://github.com/lovell/sharp/issues/1114 seems related, maybe it's as simple as asking sharp to use the inbuilt profile?

N/A

### Version of Immich Server

v1.65.0

N/A

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
N/A
```

### Additional information

_No response_

mertalev:
## Description

As requested in #3623, this PR allows for targeting Display P3 instead of sRGB for thumbnails, allowing for more vibrant images that better capture the colors of the original. Additionally, it adds a quality setting so users can customize thumbnail generation to their preference.

This only applies to image thumbnails for now, not video thumbnails. I tried several approaches for adding the same features for video thumbnails but wasn't satisfied with any of them:

* FFmpeg does not tag the resulting image with an ICC profile even with `-flags2 iccprofiles` (added with [this](https://trac.ffmpeg.org/ticket/9672) issue), so colors do not appear correctly

* JPEG quality in FFmpeg ranges 1-31 (lower is better) instead of 1-100, and even the highest quality setting is notably worse than what sharp or Imagemagick produces

* It is possible to extract a lossless thumbnail from FFmpeg and process it in sharp, but the resulting thumbnail naturally has different colors than the transcoded video

* It is possible to target `smpte432` for videos as well, but they don't seem to be processed correctly in Firefox (and applying a thumbnail setting to transcoding is confusing)

Fixes #3136

## How Has This Been Tested?

I ran thumbnail generation on a library of a few thousand images, some of which are HDR images. Inspecting the thumbnails produced for these images with exiftool shows that they have the expected ICC profile. Comparing them to older thumbnails likewise shows the expected difference in color.

Before:

before

After:

after

(Credits to @adambeck7 for the image)

### DIFF
--- a/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
@@ -97,7 +97,6 @@ const faceSearch = {
 describe(FacialRecognitionService.name, () => {
   let sut: FacialRecognitionService;
   let assetMock: jest.Mocked<IAssetRepository>;
-  let configMock: jest.Mocked<ISystemConfigRepository>;
   let faceMock: jest.Mocked<IFaceRepository>;
   let jobMock: jest.Mocked<IJobRepository>;
   let machineLearningMock: jest.Mocked<IMachineLearningRepository>;
@@ -109,7 +108,6 @@ describe(FacialRecognitionService.name, () => {
 
   beforeEach(async () => {
     assetMock = newAssetRepositoryMock();
-    configMock = newSystemConfigRepositoryMock();
     faceMock = newFaceRepositoryMock();
     jobMock = newJobRepositoryMock();
     machineLearningMock = newMachineLearningRepositoryMock();
@@ -123,7 +121,6 @@ describe(FacialRecognitionService.name, () => {
 
     sut = new FacialRecognitionService(
       assetMock,
-      configMock,
       faceMock,
       jobMock,
       machineLearningMock,
--- a/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
@@ -96,6 +96,7 @@ const faceSearch = {
 
 describe(FacialRecognitionService.name, () => {
   let sut: FacialRecognitionService;
+  let configMock: jest.Mocked<ISystemConfigRepository>;
   let assetMock: jest.Mocked<IAssetRepository>;
   let faceMock: jest.Mocked<IFaceRepository>;
   let jobMock: jest.Mocked<IJobRepository>;
@@ -104,10 +105,10 @@ describe(FacialRecognitionService.name, () => {
   let personMock: jest.Mocked<IPersonRepository>;
   let searchMock: jest.Mocked<ISearchRepository>;
   let storageMock: jest.Mocked<IStorageRepository>;
-  let configMock: jest.Mocked<ISystemConfigRepository>;
 
   beforeEach(async () => {
     assetMock = newAssetRepositoryMock();
+    configMock = newSystemConfigRepositoryMock();
     faceMock = newFaceRepositoryMock();
     jobMock = newJobRepositoryMock();
     machineLearningMock = newMachineLearningRepositoryMock();
@@ -115,12 +116,12 @@ describe(FacialRecognitionService.name, () => {
     personMock = newPersonRepositoryMock();
     searchMock = newSearchRepositoryMock();
     storageMock = newStorageRepositoryMock();
-    configMock = newSystemConfigRepositoryMock();
 
     mediaMock.crop.mockResolvedValue(croppedFace);
 
     sut = new FacialRecognitionService(
       assetMock,
+      configMock,
       faceMock,
       jobMock,
       machineLearningMock,
@@ -128,7 +129,6 @@ describe(FacialRecognitionService.name, () => {
       personMock,
       searchMock,
       storageMock,
-      configMock,
     );
   });
 
--- a/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
@@ -96,8 +96,8 @@ const faceSearch = {
 
 describe(FacialRecognitionService.name, () => {
   let sut: FacialRecognitionService;
-  let configMock: jest.Mocked<ISystemConfigRepository>;
   let assetMock: jest.Mocked<IAssetRepository>;
+  let configMock: jest.Mocked<ISystemConfigRepository>;
   let faceMock: jest.Mocked<IFaceRepository>;
   let jobMock: jest.Mocked<IJobRepository>;
   let machineLearningMock: jest.Mocked<IMachineLearningRepository>;
--- a/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
@@ -1,3 +1,4 @@
+import { Colorspace } from '@app/infra/entities';
 import {
   assetStub,
   faceStub,
@@ -22,7 +23,6 @@ import { IStorageRepository } from '../storage';
 import { ISystemConfigRepository } from '../system-config';
 import { IFaceRepository } from './face.repository';
 import { FacialRecognitionService } from './facial-recognition.services';
-import { Colorspace } from '@app/infra/entities';
 
 const croppedFace = Buffer.from('Cropped Face');
 
--- a/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
@@ -104,6 +104,7 @@ describe(FacialRecognitionService.name, () => {
   let personMock: jest.Mocked<IPersonRepository>;
   let searchMock: jest.Mocked<ISearchRepository>;
   let storageMock: jest.Mocked<IStorageRepository>;
+  let configMock: jest.Mocked<ISystemConfigRepository>;
 
   beforeEach(async () => {
     assetMock = newAssetRepositoryMock();
@@ -129,6 +130,7 @@ describe(FacialRecognitionService.name, () => {
       personMock,
       searchMock,
       storageMock,
+      configMock,
     );
   });
 
@@ -292,6 +294,8 @@ describe(FacialRecognitionService.name, () => {
       expect(mediaMock.resize).toHaveBeenCalledWith(croppedFace, 'upload/thumbs/user-id/person-1.jpeg', {
         format: 'jpeg',
         size: 250,
+        quality: 80,
+        wideGamut: true,
       });
       expect(personMock.update).toHaveBeenCalledWith({
         id: 'person-1',
@@ -313,6 +317,8 @@ describe(FacialRecognitionService.name, () => {
       expect(mediaMock.resize).toHaveBeenCalledWith(croppedFace, 'upload/thumbs/user-id/person-1.jpeg', {
         format: 'jpeg',
         size: 250,
+        quality: 80,
+        wideGamut: true,
       });
     });
 
@@ -330,6 +336,8 @@ describe(FacialRecognitionService.name, () => {
       expect(mediaMock.resize).toHaveBeenCalledWith(croppedFace, 'upload/thumbs/user-id/person-1.jpeg', {
         format: 'jpeg',
         size: 250,
+        quality: 80,
+        wideGamut: true,
       });
     });
   });
--- a/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
@@ -295,7 +295,7 @@ describe(FacialRecognitionService.name, () => {
         format: 'jpeg',
         size: 250,
         quality: 80,
-        wideGamut: true,
+        colorspace: 'p3',
       });
       expect(personMock.update).toHaveBeenCalledWith({
         id: 'person-1',
@@ -318,7 +318,7 @@ describe(FacialRecognitionService.name, () => {
         format: 'jpeg',
         size: 250,
         quality: 80,
-        wideGamut: true,
+        colorspace: 'p3',
       });
     });
 
@@ -337,7 +337,7 @@ describe(FacialRecognitionService.name, () => {
         format: 'jpeg',
         size: 250,
         quality: 80,
-        wideGamut: true,
+        colorspace: 'p3',
       });
     });
   });
--- a/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
@@ -295,7 +295,7 @@ describe(FacialRecognitionService.name, () => {
         format: 'jpeg',
         size: 250,
         quality: 80,
-        colorspace: 'p3',
+        colorspace: Colorspace.P3,
       });
       expect(personMock.update).toHaveBeenCalledWith({
         id: 'person-1',
@@ -318,7 +318,7 @@ describe(FacialRecognitionService.name, () => {
         format: 'jpeg',
         size: 250,
         quality: 80,
-        colorspace: 'p3',
+        colorspace: Colorspace.P3,
       });
     });
 
@@ -337,7 +337,7 @@ describe(FacialRecognitionService.name, () => {
         format: 'jpeg',
         size: 250,
         quality: 80,
-        colorspace: 'p3',
+        colorspace: Colorspace.P3,
       });
     });
   });
--- a/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.service.spec.ts
@@ -22,6 +22,7 @@ import { IStorageRepository } from '../storage';
 import { ISystemConfigRepository } from '../system-config';
 import { IFaceRepository } from './face.repository';
 import { FacialRecognitionService } from './facial-recognition.services';
+import { Colorspace } from '@app/infra/entities';
 
 const croppedFace = Buffer.from('Cropped Face');
 
--- a/server/src/domain/facial-recognition/facial-recognition.services.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.services.ts
@@ -18,6 +18,7 @@ export class FacialRecognitionService {
 
   constructor(
     @Inject(IAssetRepository) private assetRepository: IAssetRepository,
+    @Inject(ISystemConfigRepository) systemConfig: ISystemConfigRepository,
     @Inject(IFaceRepository) private faceRepository: IFaceRepository,
     @Inject(IJobRepository) private jobRepository: IJobRepository,
     @Inject(IMachineLearningRepository) private machineLearning: IMachineLearningRepository,
@@ -25,7 +26,6 @@ export class FacialRecognitionService {
     @Inject(IPersonRepository) private personRepository: IPersonRepository,
     @Inject(ISearchRepository) private searchRepository: ISearchRepository,
     @Inject(IStorageRepository) private storageRepository: IStorageRepository,
-    @Inject(ISystemConfigRepository) systemConfig: ISystemConfigRepository,
   ) {
     this.configCore = new SystemConfigCore(systemConfig);
   }
--- a/server/src/domain/facial-recognition/facial-recognition.services.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.services.ts
@@ -167,7 +167,7 @@ export class FacialRecognitionService {
     const thumbnailOptions = {
       format: 'jpeg',
       size: FACE_THUMBNAIL_SIZE,
-      wideGamut: thumbnail.wideGamut,
+      colorspace: thumbnail.colorspace,
       quality: thumbnail.quality,
     } as const;
     await this.mediaRepository.resize(croppedOutput, output, thumbnailOptions);
--- a/server/src/domain/facial-recognition/facial-recognition.services.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.services.ts
@@ -163,7 +163,8 @@ export class FacialRecognitionService {
     };
 
     const croppedOutput = await this.mediaRepository.crop(asset.resizePath, cropOptions);
-    await this.mediaRepository.resize(croppedOutput, output, { size: FACE_THUMBNAIL_SIZE, format: 'jpeg' });
+    await this.mediaRepository.resize(croppedOutput, { size: FACE_THUMBNAIL_SIZE });
+    await this.mediaRepository.saveThumbnail(croppedOutput, output, { format: 'jpeg' });
     await this.personRepository.update({ id: personId, thumbnailPath: output });
 
     return true;
--- a/server/src/domain/facial-recognition/facial-recognition.services.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.services.ts
@@ -18,7 +18,7 @@ export class FacialRecognitionService {
 
   constructor(
     @Inject(IAssetRepository) private assetRepository: IAssetRepository,
-    @Inject(ISystemConfigRepository) systemConfig: ISystemConfigRepository,
+    @Inject(ISystemConfigRepository) configRepository: ISystemConfigRepository,
     @Inject(IFaceRepository) private faceRepository: IFaceRepository,
     @Inject(IJobRepository) private jobRepository: IJobRepository,
     @Inject(IMachineLearningRepository) private machineLearning: IMachineLearningRepository,
@@ -27,7 +27,7 @@ export class FacialRecognitionService {
     @Inject(ISearchRepository) private searchRepository: ISearchRepository,
     @Inject(IStorageRepository) private storageRepository: IStorageRepository,
   ) {
-    this.configCore = new SystemConfigCore(systemConfig);
+    this.configCore = new SystemConfigCore(configRepository);
   }
 
   async handleQueueRecognizeFaces({ force }: IBaseJob) {
--- a/server/src/domain/facial-recognition/facial-recognition.services.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.services.ts
@@ -162,9 +162,15 @@ export class FacialRecognitionService {
       height: newHalfSize * 2,
     };
 
+    const { thumbnail } = await this.configCore.getConfig();
     const croppedOutput = await this.mediaRepository.crop(asset.resizePath, cropOptions);
-    await this.mediaRepository.resize(croppedOutput, { size: FACE_THUMBNAIL_SIZE });
-    await this.mediaRepository.saveThumbnail(croppedOutput, output, { format: 'jpeg' });
+    const thumbnailOptions = {
+      format: 'jpeg',
+      size: FACE_THUMBNAIL_SIZE,
+      wideGamut: thumbnail.wideGamut,
+      quality: thumbnail.quality,
+    } as const;
+    await this.mediaRepository.resize(croppedOutput, output, thumbnailOptions);
     await this.personRepository.update({ id: personId, thumbnailPath: output });
 
     return true;
--- a/server/src/domain/facial-recognition/facial-recognition.services.ts
+++ b/server/src/domain/facial-recognition/facial-recognition.services.ts
@@ -18,7 +18,6 @@ export class FacialRecognitionService {
 
   constructor(
     @Inject(IAssetRepository) private assetRepository: IAssetRepository,
-    @Inject(ISystemConfigRepository) configRepository: ISystemConfigRepository,
     @Inject(IFaceRepository) private faceRepository: IFaceRepository,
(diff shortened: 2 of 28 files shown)


uhthomas:
I don't know if this is really what we want. I already have an open PR which attempts to solve the same problem: https://github.com/immich-app/immich/pull/3164.

As I've explained in that PR, we should really just be converting between colour spaces to sRGB correctly rather than simply rendering thumbnails in the original colour space.

uhthomas:
It looks like the work you've done here may translate quite well to what I was working on though. I believe providing the correct ICC profile before creating the thumbnail will produce a correct looking sRGB image.

mertalev:
There's something strange I've noticed about how Sharp is handling colors. It ignores `toColorspace`; only `withMetadata` has an effect. Calling `withMetadata` without any arguments produces a more desaturated image than calling it with `srgb`. Even though these are supposed to be equivalent, the former doesn't add an ICC profile while the latter does. 

So far I tested with HDR mode disabled on my monitor, but enabling it (and enabling wide gamut support with it) makes the image appear oversaturated. This is the expected effect when you view an sRGB image as if it were P3. My thinking is that sharp doesn't actually convert the image to P3 and only adds the ICC profile - even if `toColorspace` is called. This means the P3 tagged image only looks correct when color managed into sRGB by the browser (similar to tagging it as sRGB).

Some comparison images:

`withMetadata()`:

IMG_1912_emptyWith

`withMetadata({icc: 'srgb'})`:

IMG_1912_srgbWith

`withMetadata({icc: 'p3'})`:

IMG_1912_p3With

`withMetadata({icc: 'p3'}).toColorspace('srgb')` (identical to above):

IMG_1912_p3Withtosrgb

`withMetadata({icc: 'p3'}).toColorspace('p3')` (identical to above):

IMG_1912_p3Withto

mertalev:
I wonder if this is because it's using ImageMagick to do the actual conversion. I'll test with a 16-bit PNG to see if it's any different when libvips itself converts it.

uhthomas:
The whole thing is a bit confusing, but I have successfully converted an AdobeRGB image to sRGB with imagemagick and without desaturating the colours (see latest comment on other PR). All it needs is the original ICC profile.

I don't know if sharp is able to do this properly. See https://github.com/lovell/sharp/issues/1323

mertalev:
You might be looking for `withMetadata({icc: 'srgb'})`.

uhthomas:
> You might be looking for `withMetadata({icc: 'srgb'})`.

From what I understand, you actually need to provide the original ICC profile explicitly and then convert to sRGB. See again my example of properly doing the conversion with imagemagick.

I'd appreciate it if you could try to replicate the behaviour I observed to make sure we're on the same page and figure out a path forward.

mertalev:
I can try your approach later today, but if I understand the sharp code correctly, it uses the embedded ICC profile of the original image to transform to the target profile. I don't think using the original profile for the target image is a good idea since the colorspace can be different.

uhthomas:
> I can try your approach later today, but if I understand the sharp code correctly, it uses the embedded ICC profile of the original image to transform to the target profile. I don't think using the original profile for the target image is a good idea since the colorspace can be different.

From what I understand this is actually not what is happening. Sharp is essentially loading the original image as if it were srgb, even though it isn't and is why the colours are dull. Explicitly providing the correct ICC profile allows imagemagick to load the image and convert to sRGB correctly.

Does what I've written make sense? I'm feeling like we're not quite aligned.

mertalev:
To my understanding, browsers will color manage every image into sRGB if they detect an sRGB display. If the image has an sRGB color profile, then it will just use that. If it's a different color profile, then it will use it to convert to sRGB. And if it doesn't have an ICC profile, it will be viewed as a standard uncalibrated sRGB image (even if it isn't actually sRGB). 

In your PR, you mentioned that using `withMetadata()` still led to dull images. I think this is because it doesn't actually add an ICC profile unless called with the `icc` profile set, meaning the browser doesn't have the context needed to do color management. Adding the original ICC profile allowed it to color manage to sRGB. 

The main difference between the PRs, as I understand it, is that this is doing that conversion up-front and embedding a web-compatible ICC profile, while using the original profile leaves it to the client to color manage. 

I'll have to do some more testing, but I think the latter approach seems more bug-prone. The original profile may be for a different colorspace (e.g. CMYK vs sRGB), and color management is very browser-dependent.

mertalev:
> > I can try your approach later today, but if I understand the sharp code correctly, it uses the embedded ICC profile of the original image to transform to the target profile. I don't think using the original profile for the target image is a good idea since the colorspace can be different.
> 
> From what I understand this is actually not what is happening. Sharp is essentially loading the original image as if it were srgb, even though it isn't and is why the colours are dull. Explicitly providing the correct ICC profile allows imagemagick to load the image and convert to sRGB correctly.
> 
> Does what I've written make sense? I'm feeling like we're not quite aligned. 

Just saw this reply. I don't think this is the case. Sharp processes 16-bit images as 16-bit unless `pipelineColorspace` is called to override it. When embedding an ICC to the output, it calls `icc_transform` which uses the ICC profile in the image.

mertalev:
> There's something strange I've noticed about how Sharp is handling colors. It ignores `toColorspace`; only `withMetadata` has an effect. Calling `withMetadata` without any arguments produces a more desaturated image than calling it with `srgb`. Even though these are supposed to be equivalent, the former doesn't add an ICC profile while the latter does.
> 
> So far I tested with HDR mode disabled on my monitor, but enabling it (and enabling wide gamut support with it) makes the image appear oversaturated. This is the expected effect when you view an sRGB image as if it were P3. My thinking is that sharp doesn't actually convert the image to P3 and only adds the ICC profile - even if `toColorspace` is called. This means the P3 tagged image only looks correct when color managed into sRGB by the browser (similar to tagging it as sRGB).
> 
> Some comparison images:
> 
> `withMetadata()`:
> 
> IMG_1912_emptyWith
> 
> `withMetadata({icc: 'srgb'})`:
> 
> IMG_1912_srgbWith
> 
> `withMetadata({icc: 'p3'})`:
> 
> IMG_1912_p3With
> 
> `withMetadata({icc: 'p3'}).toColorspace('srgb')` (identical to above):
> 
> IMG_1912_p3Withtosrgb
> 
> `withMetadata({icc: 'p3'}).toColorspace('p3')` (identical to above):
> 
> IMG_1912_p3Withto

This also seems to be a false alarm. I tried opening the 16-bit image in Photoshop and it looked even more saturated. So it's not that the P3 image was getting oversaturated compared to sRGB, just that sRGB is that much more desaturated.

AngelaDMerkel:
I made the original feature request in #3623 and would like to say my thanks to @mertalev for all his work. My feedback in this pull request is that even properly converted sRGB thumbs only approximate the original. Every device I own, as well as nearly every device others users own supports P3 natively and the difference in appearance is significant because wider colour spaces simply have more colours in them. sRGB will always be lesser with respect to fidelity.

In my view self hosting offers more control. All my users can view P3 and I have some reference monitors with near BT2020 coverage. I would appreciate being able to leverage these capabilities and see no harm in properly converted sRGB in addition to P3 options.

mertalev:
> 
> 
> 
> In my view self hosting offers more control. All my users can view P3 and I have some reference monitors with near BT2020 coverage. I would appreciate being able to leverage these capabilities and see no harm in properly converted sRGB in addition to P3 options. 

Thanks for the kind words! :)

Working on this PR, I definitely agree that P3 preserves colors in a way that makes sRGB look flat. I'm glad you mentioned it since I thought sRGB was the only real option for web usage.

uhthomas:
I think offering support for different ICC profiles in thumbnails might be a nice feature long term, though properly converting to srgb will make a huge difference to all immich users and devices regardless of their display technology. In a lot of cases, P3 probably won't actually look that different to a good srgb conversion.

My concern is that there are more ICC profiles than just P3, and browsers have awful ICC profile support which is why I'm keen to fix sRGB conversion first.

Is this fair?

AngelaDMerkel:
@uhthomas Thanks also for your time. It is my personal opinion that P3 and sRGB will always look very different depending on how they were mastered, but I also work in colouring and am probably not representative. Any approach you take is fair and I am not the only user. 

P3 seemed an exception in my mind as it's in common use on the web, whereas AdobeRGB and BT2020 are not at all. My preferred solution would be to have both proper sRGB conversion and perhaps an optional toggle for P3. Unfortunately I don't have the means to judge the difficulty, but that us my preferred solution. 

Apple, as ever, is producing HDR images in their iPhones now too, which cannot really be converted to sRGB and are instead designed to be displayed as they are and managed by the display. A P3 option might be a positive long term step toward AVIF/JXL in other colour spaces, but that seems a long way off.

uhthomas:
Thanks for your thoughts @AngelaDMerkel, I think we're pretty much aligned. Proper sRGB conversion would be the priority, and then a separate toggle to preserve the P3 ICC profile as a nice addition. We're looking into this right now and will hopefully figure something out.

jrasm91 (approved):
Lgtm

mertalev:
I don't think these are conflicting asks. After fixing a bug causing some images to come out desaturated, this PR improves existing sRGB conversion in both adding a web-compatible ICC profile and in making the color pipeline more accurate. This is in addition to P3 support, which is a toggle in the thumbnail settings.

mertalev:
Also thanks to @uhthomas for helping me debug this issue. It was surprisingly difficult to find the root problem.

AngelaDMerkel:
@adambeck7 To me it looks like Immich is setting the colour temperature to Daylight which is resulting is a pretty strong cast, but it's hard to say

server/src/domain/media/media.service.ts
```diff
@@ -59,51 +58,89 @@ export class MediaService {
...
   async handleGenerateWebpThumbnail({ id }: IEntityJob) {
     const [asset] = await this.assetRepository.getByIds([id]);
     if (!asset || !asset.resizePath) {
       return false;
     }
 
-    const webpPath = asset.resizePath.replace('jpeg', 'webp').replace('jpg', 'webp');
+    switch (asset.type) {
+      case AssetType.IMAGE:
+        try {
```
uhthomas, line 121: Why duplicate this try? I would wrap the switch in it, or omit it entirely. I think you can await something and use .catch for the log?

server/src/infra/repositories/media.repository.ts
```diff
@@ -23,10 +24,25 @@ export class MediaRepository implements IMediaRepository {
   }
 
   async resize(input: string | Buffer, output: string, options: ResizeOptions): Promise<void> {
-    await sharp(input, { failOn: 'none' })
+    let colorspace = options.wideGamut ? 'p3' : 'srgb';
```
uhthomas, line 27: I'm not sure I like the variable 'wideGamut'. There are more profiles than just P3.

uhthomas (changes requested):

mertalev:
> Not sure if this is related, but I wonder if this will fix something I just noticed today. Lots of RAW photos are really warm, almost to the point of being sepia. Just uploaded these on 1.73.0, but I've noticed it to some extent in previous versions too. If this doesn't belong here, let me know and I can open a separate issue. Is there a release of the `feat/server-wide-gamut` branch so I could test it? 
> 
> 
> 
> Some photos are more obvious than others:
> 
> Screenshot 2023-08-16 at 8 38 07 AM
> 
> 
> 
> 
> 
> I thought it'd be worse on a photo with more white, but to my eyes it doesn't seem as severe: 
> 
> Screenshot 2023-08-16 at 8 39 30 AM
> 
> 

Imagemagick seems to do this when it converts RAW images. Related: https://stackoverflow.com/questions/71983812/wrong-whitebalace-colorgrade-for-arw-files-in-imagemagick-updated 

I think it ignores camera white-balance by default, but this can be changed on a file-type basis with the `-define` flag.

uhthomas:
@adambeck7 @AngelaDMerkel Rendering of raw images is unrelated to this change. I don't believe raw files actually have an associated colour space at all.

I would open a separate issue - we may be able to ask libraw to render images with the embedded colour temperature.

server/src/infra/repositories/media.repository.ts
mertalev: What other web-compatible wide-gamut profiles are there?

server/src/domain/media/media.service.ts
jrasm91: You can't wrap the whole thing in a try because that would also catch `UnsupportedMediaTypeException`.

server/src/infra/repositories/media.repository.ts
mertalev: `preserveColorspace` is misleading because it doesn't actually keep the image's colorspace.

I originally had a drop-down with sRGB and P3, but a slider makes for better UX in this case since these will continue to be the only real options for the foreseeable future. I can change the internal variable to store the color profile instead of being a boolean to handle the case where another viable option shows up.

uhthomas: That is a fine solution to me.

server/src/domain/media/media.service.ts
uhthomas: There just be ways to clean it up though. Even just moving logging elsewhere or merging log lines would help?

So instead of a successful log in two places, it could be moved out of the switch statement entirely and the asset kind formatted into the string. These logs also don't really provide any useful info other than just logging, so maybe the log could be printed downstream?

server/src/domain/media/media.service.ts
```diff
@@ -59,51 +58,89 @@ export class MediaService {
...
+          return false;
+        }
+        break;
+      default:
+        throw new UnsupportedMediaTypeException(`Unsupported asset type for thumbnail generation: ${asset.type}`);
+    }
 
-    const { thumbnail } = await this.configCore.getConfig();
-    await this.mediaRepository.resize(asset.resizePath, webpPath, { size: thumbnail.webpSize, format: 'webp' });
+    const webpPath = this.getPath(asset, 'webp');
```
uhthomas, line 142: nit: it's pretty clear from context this will be webp so it could be 'path' or 'thumbnailPath' if you really want to be explicit.

server/src/domain/media/media.service.ts
```diff
@@ -292,4 +329,10 @@ export class MediaService {
 
     return handler;
   }
+
+  getPath(asset: AssetEntity, extension: string): string {
```
uhthomas, line 333: The name of this function is quite misleading. It's changing the file system, not just getting a path or something.

server/src/domain/media/media.util.ts
```diff
@@ -50,7 +50,7 @@ class BaseConfig implements VideoCodecSWConfig {
     if (this.shouldToneMap(stream)) {
       options.push(...this.getToneMapping());
     }
-    options.push('format=yuv420p');
+    options.push(`format=yuv420p`);
```
uhthomas, line 53: ?

server/src/domain/media/media.service.ts
jrasm91: There actually is already a try/catch upstream that runs and logs whenever a job throws an unhandled error. We can probably just remove these.

server/src/domain/media/media.service.ts
mertalev: How about `ensurePath`?

uhthomas: Is the mkdir really required? I don't see it used before.

mertalev: It was there before; I just refactored it into a separate method. I'm not sure if the folder is guaranteed to exist prior to the job.

server/src/domain/media/media.util.ts
mertalev: Ah, this is a holdover from my FFmpeg attempt.

server/src/domain/media/media.service.ts
jrasm91: We can definitely clean all those up, probably in another PR though. At the moment if the directory doesn't exist the write file will fail.

server/src/domain/media/media.service.ts
```diff
@@ -59,51 +58,65 @@ export class MediaService {
...
+  }
+
   async handleGenerateWebpThumbnail({ id }: IEntityJob) {
     const [asset] = await this.assetRepository.getByIds([id]);
     if (!asset || !asset.resizePath) {
       return false;
     }
 
-    const webpPath = asset.resizePath.replace('jpeg', 'webp').replace('jpg', 'webp');
+    switch (asset.type) {
```
uhthomas, line 107: Could we create another function like 'generateThumbnail(asset, kind)` which deduplicates the switch on asset type?

server/src/infra/entities/system-config.entity.ts
```diff
@@ -90,6 +92,11 @@ export enum ToneMapping {
   DISABLED = 'disabled',
 }
 
+export enum Colorspace {
```
uhthomas, line 95: Colour space is two words, not one.

```suggestion
export enum ColorSpace {
```

server/src/infra/repositories/media.repository.ts
```diff
@@ -23,10 +25,25 @@ export class MediaRepository implements IMediaRepository {
...
 
   async resize(input: string | Buffer, output: string, options: ResizeOptions): Promise<void> {
-    await sharp(input, { failOn: 'none' })
+    let colorspace = options.colorspace;
+    if (options.colorspace !== Colorspace.SRGB) {
+      try {
+        const { space } = await sharp(input).metadata();
+        // if the image is already in srgb, keep it that way
+        if (space && space === Colorspace.SRGB) {
+          colorspace = Colorspace.SRGB;
```
uhthomas, line 34: This is confusing me. This whole block is essentially a no-op?

uhthomas (dismissed):

server/src/infra/repositories/media.repository.ts
jrasm91: Perhaps renaming the variables to `targetColorSpace` and `actualColorSpace` or similar would make it more clear. I believe this will ignore the targetColorSpace argument in the case the file is srgb.

mertalev: I changed the local variable to `colorProfile`, since this is what it actually sets in the sharp pipeline.

server/src/domain/media/media.service.ts
mertalev: Done

server/src/infra/repositories/media.repository.ts
```diff
@@ -23,10 +25,25 @@ export class MediaRepository implements IMediaRepository {
   }
 
   async resize(input: string | Buffer, output: string, options: ResizeOptions): Promise<void> {
-    await sharp(input, { failOn: 'none' })
+    let colorProfile = options.colorspace;
+    if (options.colorspace !== Colorspace.SRGB) {
+      try {
+        const { space } = await sharp(input).metadata();
+        // if the image is already in srgb, keep it that way
+        if (space && space === Colorspace.SRGB) {
```
uhthomas, line 33: There's no need for `space &&` here. Even if space is null, it still won't equal the constant.

uhthomas, line 33: I also don't get what's special about sRGB? If the original space is AdobeRGB and the target is P3 for instance, then why is it okay to use P3 for that and not sRGB?

mertalev: It doesn't make sense to convert sRGB content to P3. The reason to use P3 is because of its wider color gamut, but content in sRGB doesn't stand to gain anything from this and can only lose due to precision loss during conversion. 

AdobeRGB isn't a colorspace in libvips. It would show up as generic rgb16 or similar. P3 is a better choice for this and other content because there's a higher chance of it supporting the colors in the source.

mertalev: Done

mertalev:
@adambeck7 I tried the fix in #3806 along with the settings in this PR. It looks a lot better now!

#3806 only:

IMG_1912_wb


sRGB (#3806 + this PR):

IMG_1912_wb_srgb


P3 (#3806 + this PR):

IMG_1912_wb_p3

alextran1502:
@mertalev Can you help with rebasing this PR?

@uhthomas any other feedback do you have on this PR?

mertalev:
Rebased

server/src/domain/facial-recognition/facial-recognition.service.spec.ts
```diff
@@ -96,38 +97,38 @@ const faceSearch = {
...
     assetMock = newAssetRepositoryMock();
-    configMock = newSystemConfigRepositoryMock();
     faceMock = newFaceRepositoryMock();
     jobMock = newJobRepositoryMock();
     machineLearningMock = newMachineLearningRepositoryMock();
     mediaMock = newMediaRepositoryMock();
     personMock = newPersonRepositoryMock();
     searchMock = newSearchRepositoryMock();
     storageMock = newStorageRepositoryMock();
+    configMock = newSystemConfigRepositoryMock();
```
jrasm91, line 118: And here

server/src/domain/facial-recognition/facial-recognition.service.spec.ts
```diff
@@ -96,38 +97,38 @@ const faceSearch = {
...
       assetMock,
-      configMock,
       faceMock,
       jobMock,
       machineLearningMock,
       mediaMock,
       personMock,
       searchMock,
       storageMock,
+      configMock,
```
jrasm91, line 131: And here

server/src/domain/facial-recognition/facial-recognition.services.ts
```diff
@@ -18,16 +18,16 @@ export class FacialRecognitionService {
...
     @Inject(IAssetRepository) private assetRepository: IAssetRepository,
-    @Inject(ISystemConfigRepository) configRepository: ISystemConfigRepository,
     @Inject(IFaceRepository) private faceRepository: IFaceRepository,
     @Inject(IJobRepository) private jobRepository: IJobRepository,
     @Inject(IMachineLearningRepository) private machineLearning: IMachineLearningRepository,
     @Inject(IMediaRepository) private mediaRepository: IMediaRepository,
     @Inject(IPersonRepository) private personRepository: IPersonRepository,
     @Inject(ISearchRepository) private searchRepository: ISearchRepository,
     @Inject(IStorageRepository) private storageRepository: IStorageRepository,
+    @Inject(ISystemConfigRepository) systemConfig: ISystemConfigRepository,
```
jrasm91, line 28: And here

server/src/infra/repositories/media.repository.ts
```diff
@@ -23,10 +25,25 @@ export class MediaRepository implements IMediaRepository {
   }
 
   async resize(input: string | Buffer, output: string, options: ResizeOptions): Promise<void> {
-    await sharp(input, { failOn: 'none' })
+    let colorProfile = options.colorspace;
+    if (options.colorspace !== Colorspace.SRGB) {
+      try {
+        const { space } = await sharp(input).metadata();
+        // if the image is already in srgb, keep it that way
+        if (space === Colorspace.SRGB) {
```
jrasm91, line 33: This is a little strange that we are comparing our internal enum value against the case sensitive value returned from ffprobe. It might be better to use a string instead of the enum. I don't think how we represent the enum needs to be couples with ffprobes value.

web/src/lib/components/admin-page/settings/setting-switch.svelte
```diff
@@ -7,6 +7,13 @@
   export let checked = false;
   export let disabled = false;
   export let isEdited = false;
+  export let onToggle: (checked: boolean) => void = () => {
+    return;
+  };
+  export const handler = (e: Event) => {
+    const target = e.target as HTMLInputElement;
+    onToggle(target.checked);
+  };
```
jrasm91, line 16: This is exactly what event dispatchers are for

mertalev: I knew there had to be a better way 😛

server/src/infra/repositories/media.repository.ts
mertalev: Yeah, this was a bit awkward since it's actually comparing against another enum that happens to have the same value.

server/src/domain/facial-recognition/facial-recognition.service.spec.ts
```diff
@@ -96,38 +97,38 @@ const faceSearch = {
...
   let assetMock: jest.Mocked<IAssetRepository>;
-  let configMock: jest.Mocked<ISystemConfigRepository>;
   let faceMock: jest.Mocked<IFaceRepository>;
   let jobMock: jest.Mocked<IJobRepository>;
   let machineLearningMock: jest.Mocked<IMachineLearningRepository>;
   let mediaMock: jest.Mocked<IMediaRepository>;
   let personMock: jest.Mocked<IPersonRepository>;
   let searchMock: jest.Mocked<ISearchRepository>;
   let storageMock: jest.Mocked<IStorageRepository>;
+  let configMock: jest.Mocked<ISystemConfigRepository>;
```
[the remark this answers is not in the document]
mertalev: Done

server/src/domain/facial-recognition/facial-recognition.service.spec.ts
```diff
@@ -95,8 +96,8 @@ const faceSearch = {
 
 describe(FacialRecognitionService.name, () => {
   let sut: FacialRecognitionService;
-  let assetMock: jest.Mocked<IAssetRepository>;
   let configMock: jest.Mocked<ISystemConfigRepository>;
+  let assetMock: jest.Mocked<IAssetRepository>;
```
jrasm91, line 100: So close. This should just be how it was before :stuck_out_tongue:

server/src/domain/facial-recognition/facial-recognition.services.ts
```diff
@@ -27,7 +27,7 @@ export class FacialRecognitionService {
     @Inject(ISearchRepository) private searchRepository: ISearchRepository,
     @Inject(IStorageRepository) private storageRepository: IStorageRepository,
   ) {
-    this.configCore = new SystemConfigCore(configRepository);
+    this.configCore = new SystemConfigCore(systemConfig);
```
jrasm91, line 30: SystemConfigCore is instantiated 9 times and 7 of the 9 use `configRepository`. I think we should standardize on that name. What do you think?

server/src/infra/repositories/media.repository.ts
```diff
@@ -1,8 +1,10 @@
 import { CropOptions, IMediaRepository, ResizeOptions, TranscodeOptions, VideoInfo } from '@app/domain';
+import { Colorspace } from '@app/infra/entities';
 import { Logger } from '@nestjs/common';
 import ffmpeg, { FfprobeData } from 'fluent-ffmpeg';
 import fs from 'fs/promises';
 import sharp from 'sharp';
+import { Writable } from 'typeorm/platform/PlatformTools.js';
```
jrasm91, line 7: This is probably the wrong import, no?

web/src/lib/components/admin-page/settings/setting-switch.svelte
```diff
@@ -1,12 +1,19 @@
...
   import { fly } from 'svelte/transition';
+  import { createEventDispatcher } from 'svelte';
 
   export let title: string;
   export let subtitle = '';
   export let checked = false;
   export let disabled = false;
   export let isEdited = false;
+
+  const dispatch = createEventDispatcher();
```
jrasm91, line 12:
```suggestion
  const dispatch = createEventDispatcher<{ toggle: boolean }>();
```

Best to keep these events typed for improved intellisense and type checking.

web/src/lib/components/admin-page/settings/setting-switch.svelte
```diff
@@ -1,12 +1,19 @@
...
+
+  function onToggle(event: Event) {
+    dispatch('toggle', (event.target as HTMLInputElement).checked);
+  }
```
jrasm91, line 16: I'd be fine inlining this. If you keep it here, we use the form `const onToggle = (event: Event) => { ... }` more often than this function syntax.

jrasm91 (approved):
Few small things to clean up, but LGTM.

server/src/domain/facial-recognition/facial-recognition.service.spec.ts
mertalev: lol I just put it on line 99 without looking at the diff

server/src/domain/facial-recognition/facial-recognition.services.ts
mertalev: Hmm, you've got a point there. I didn't notice the different variable names. I'll rename it to `configRepository`.

server/src/infra/repositories/media.repository.ts
mertalev: Yeah, not sure why it auto-imported this one

jrasm91 (approved):

waclaw66:
@mertalev Could you please fix colorspace handling according https://github.com/immich-app/immich/pull/4040#issuecomment-1716761519? Thanks in advance.

PR #3661, opened 2023-08-13
feat(mobile): map view

shenlong-tanwen:
#### Changes made in the PR:

- Initial implementation of Map view for the mobile app
- Dark mode, Favorite only and Relative date filters to filter the map markers
- Tapping the center of the heat map will display a asset marker closer to the tap point (The threshold for detecting taps has to be improved)
- Bottom sheet to display assets in the current map bounds.
- Scrolling the bottom sheet will display the marker for the current asset on the map
- Zoom to the current asset button in the bottom sheet to zoom into the location of the current asset
- Selection handling for the AssetGrid in the bottom sheet to share, favorite or archive assets straight from the map view
- Move to current location button to go to the current location of the user to display assets from there

### Screenshots
Search Page | Map view | With Selection | Light-mode
:-: | :-: | :-: | :-:
search_page | map_view | with_selection | light_mode

### Screencap

https://github.com/immich-app/immich/assets/139912620/42e6a09d-f2f2-489b-b28e-7cc6d5b6f5e1

---

#### Changes made in the PR:

- Initial implementation of Map view for the mobile app
- Dark mode, Favorite only and Relative date filters to filter the map markers
- Tap marker to open the image in viewer
- Tap a cluster with > 10 assets to list all the assets based on the zoom level (This is a different behavior than the one in the web app where the first asset in the cluster is opened)
    -  zoom level > 12 - Assets from same city
    -  zoom level > 9 - Assets from same state
    - zoom level > 4 - Assets from the same country
    - zoom level < 4 - Nothing happens
- Just as the web version, if a cluster contains less than 10 assets, tapping it will show all of them in a circle

#### How has this been tested?
All the below were performed using the demo instance of Immich
- Enable / Disabled dark mode and ensured the map theme is updated accordingly
- Enabled favorite only and ensured only favorite assets with location data are displayed in map
- Ensured the relative date filter for various ranges

#### Screenshots
Search Page | Map Settings | Dark Mode | Light Mode
:-: | :-: | :-: | :-:
Search page | Map settings | Dark mode | Light mode

#### Screencap
https://github.com/immich-app/immich/assets/139912620/3191250c-ddbd-4199-bed4-d900f3eccf35

#### Feedback required
- Doing nothing when a cluster is tapped if the zoom level is < 4 might seem like a bug to some? 
- Test the changes once in devices / emulators of different resolution in both iOS and Android.
- Performance of the map in low peforming devices?
- The [plugin](https://github.com/lpongetti/flutter_map_marker_cluster) is forked and changes are made to make the clustering work the same way as the web version. You can find the fork [here](https://github.com/shalong-tanwen/flutter_map_marker_cluster/tree/immich). The changes are too niche to be added to the original plugin and I doubt they'll accept the changes into their repo. Nevertheless, we can try opening a PR for the same. Meanwhile, should we continue using the fork with the changes? If so, I think it would be best to move it into `immich-app` org. 
⚠️ Note: `pubspec.yaml` and `pubspec.lock` are not updated with this cluster plugin yet. Once we decide on what to do with the fork, I'll add the proper git url as dependency
- Yet to test the behavior in a slower network, and also need to decide whether to show a loader while fetching the map markers and/or when an error occurs?


alextran1502:
Nicely done! Thank you for the PR.

What do you think about following GPhotos design of display "your map" component as the first item in the Places list?

shalong-tanwen:
> What do you think about following GPhotos's design of displaying "your map" component as the first item in the Places list?

Ah, I do not have backup turned on in GPhotos, so thought they were displaying only the map because of the lack of auto back. Yes, instead of displaying both the map as well as the places, we can follow their approach and hide the places under the View all button.

What do you think about making the "your map" thumbnail dynamic by displaying a single marker from the user's asset? We can query the local db for an asset with location and probably use that instead of making a request to the server and fallback to a constant co-ordinates when we cannot find one?

radh21301:
Hey. Just a few suggestions. I love using the following two options on map view on google photos.

1. Give google photos access to my current location. Then zoom into where I am and see what photos/videos I had taken around that region.

2. Have the map view update as I scroll or move across the map. The top half of the screen has the map view, and the bottom has something like a timeline view. That view updates as I move/scroll across the map. Something like the figure below.
image

alextran1502:
> What do you think about making the "your map" thumbnail dynamic by displaying a single marker from the user's asset? We can query the local db for an asset with location and probably use that instead of making a request to the server and fallback to a constant co-ordinates when we cannot find one?

This would be nice, if it is too complex to implement, we can push it to the subsequent enhancement PR

> Ah, I do not have backup turned on in GPhotos, so thought they were displaying only the map because of the lack of auto back. Yes, instead of displaying both the map as well as the places, we can follow their approach and hide the places under the View all button.

It is actually the first item on the Places list. See image below
image

shalong-tanwen:
> Hey. Just a few suggestions. I love using the following two options on map view on google photos.

I've tried to implement the features from photos in the map page. Here is the latest screen cap showing the current state of the map page in Immich

https://github.com/immich-app/immich/assets/139912620/b766a2e1-91b4-4c25-b50d-611897dc401a

The code needs a little refactoring and the following things are yet to be added as well:
- Bottom sheet grid asset selection and asset actions
- Search page map view widget look

shalong-tanwen:
The final version of the map view for mobile apps

### Screenshots
Search Page | Map view | With Selection | Light-mode
:-: | :-: | :-: | :-:
search_page | map_view | with_selection | light_mode

### Screencap

https://github.com/immich-app/immich/assets/139912620/42e6a09d-f2f2-489b-b28e-7cc6d5b6f5e1

jrasm91:
Can you update the description with the final version instead? Looks dope.

alextran1502:
Let me know when this PR is ready for review and I can help taking a look, and to test it

shalong-tanwen:
> Let me know when this PR is ready for review and I can help taking a look, and to test it

The PR is ready for review and testing. I guess the code could use a refactoring, but it can be done in future PRs.

alextran1502:
I ran into this issue when opening the app, so all the markers aren't loaded

image

shalong-tanwen:
> I ran into this issue when opening the app, so all the markers aren't loaded

Looks like the marker has an invalid lat, lon data. Latitude should be between -90 to +90 and longitude should be between -180 to +180. From your screenshot, looks like a latitude value of 106 has been provided, maybe the values are switched for an asset?

In any cases, I've added additional checks to handle them gracefully instead of throwing an exception

alextran1502 (approved):

ayykamp:
The mobile map should probably also respect the admin Tile URL setting, especially because https://github.com/immich-app/immich/discussions/3866

shalong-tanwen:
> The mobile map should probably also respect the admin Tile URL setting, especially because #3866

PR https://github.com/immich-app/immich/pull/4045 is awaiting merge

PR #3669, opened 2023-08-13
fix(server): Does not assign lat/lon if they are at 0,0 #2991

### LINKED ISSUE #2991: [BUG] Map shows some pictures on Null Island
### The bug

Some photos in my library have location data for [Null Island](https://en.wikipedia.org/wiki/Null_Island). It is the exact center point of the map.
image
Pictures can have location data of 0 longitude, 0 latitude when the data gets stripped out or something went wrong trying to get the location when taking the picture.

Alpine Linux, but it is dockerized, another image creator

### Version of Immich Server

v1.64.0

v1.64.0 build.87

### Platform with the issue

- [ ] Server
- [X] Web
- [X] Mobile

```YAML
version: "3"
services:
  immich:
    container_name: immich
    environment:
      - PUID=1000
      - PGID=1000
      - TZ=Europe/Berlin
      - DB_HOSTNAME=postgres
      - DB_USERNAME=postgres
      - DB_PASSWORD=postgres
      - DB_DATABASE_NAME=immich
      - REDIS_HOSTNAME=redis
      #- DISABLE_MACHINE_LEARNING=false
      #- DISABLE_TYPESENSE=false
      - DB_PORT=5432 #optional
      - REDIS_PORT=6379 #optional
      - REDIS_PASSWORD= #optional
    volumes:
      - ./config:/config
      - ./photos:/photos
      - ./ml:/config/machine-learning #optional
    ports:
      - 1034:8080
    restart: unless-stopped
  redis:
    image: redis
    ports:
      - 6379:6379
    container_name: redis
  postgres:
    image: postgres:14
    ports:
      - 5432:5432
    container_name: postgres14
    environment:
      POSTGRES_USER: postgres
      POSTGRES_PASSWORD: postgres
      POSTGRES_DB: immich
    volumes:
      - postgres:/var/lib/postgresql/data

volumes:
   postgres:
```

### Your .env content

```Shell
I put the environment vars in the docker compose file
```

### Reproduction steps

```bash
1. Obtain a photo with Null Island location info
2. View the "Map" tab in Immich
3. See the issue in the middle of the map
```

### Additional information

Immich should detect that 0,0 location data and just not show those photos on the Map.

sellnat77:
## Description
  - Allows null input to `parseLatitude` and `parseLongitude`
    - Returns null
  - If the latitude AND longitude are both 0, their exif coordinates are not set

Fixes #2991
Closes #3665 

## How Has This Been Tested?

- [x] Edit picture metadata to have lat,lon (0,0) -> open map with default settings -> verify null island is empty

## Checklist:

### DIFF
--- a/server/src/infra/migrations/1692057328660-fixGPSNullIsland.ts
+++ b/server/src/infra/migrations/1692057328660-fixGPSNullIsland.ts
@@ -1,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from "typeorm"
+
+export class FixGPSNullIsland1692057328660 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`UPDATE "exif" SET latitude = NULL, longitude = NULL WHERE latitude = 0 AND longitude = 0;`);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`UPDATE "exif" SET latitude = 0, longitude = 0 WHERE latitude = NULL AND longitude = NULL;`);
+    }
+
+}
--- a/server/src/infra/migrations/1692057328660-fixGPSNullIsland.ts
+++ b/server/src/infra/migrations/1692057328660-fixGPSNullIsland.ts
@@ -6,7 +6,8 @@ export class FixGPSNullIsland1692057328660 implements MigrationInterface {
         await queryRunner.query(`UPDATE "exif" SET latitude = NULL, longitude = NULL WHERE latitude = 0 AND longitude = 0;`);
     }
 
-    public async down(queryRunner: QueryRunner): Promise<void> {
+    public async down(): Promise<void> {
+        // Setting lat,lon to 0 not necessary
     }
 
 }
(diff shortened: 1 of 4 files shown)


bo0tzz (approved):

jrasm91 (commented):
Can we add a migration to fix existing data? Is that something you would be able to do? We can SQL update the data and set 0s to null on the next release.

sellnat77:
I can try! Any hints on where to look for a previous example?

jrasm91:
The src/infra/migrations folder

jrasm91:
Here's the docs for them as well:

https://immich.app/docs/developer/database-migrations

sellnat77:
Added migration file, not sure i did it right? specifically the `down` portion

When i transferred the migration to the dir the 0,0 entries were properly nulled out
if the `down` migration were to be ran, all the null entries would be placed on null island

jrasm91 (commented):
Looks great. We don't need a down for this migration 😁

(You can leave it empty)

jrasm91 (approved):

JackBailey:
For anyone with this issue with imported files, here's how you perform the migration manually for docker installs:

```bash
docker exec -it immich-postgres psql --password --username immich-user immich-database
```

Enter your password, then run the following:

```sql
UPDATE "exif" SET latitude = NULL, longitude = NULL WHERE latitude = 0 AND longitude = 0;
```

PR #3695, opened 2023-08-15
fix(web,server): disable partner's archive access

brighteyed:
## Description
Before 1.72.0 user didn't have access to partner's archived assets. This PR fixes the regression by introducing the new permission `ARCHIVE_READ`

## How Has This Been Tested?
 - Partner's timeline now doesn't include archived assets
 - `getTimeBuckets` and `getByTimeBucket` with `userId` specified and `isArchived` other than `false` throws an error now


jrasm91 (approved):
Thanks for fixing this. The permission doesn't quite apply everywhere (get asset by id could still return archived assets), but I think it is better than what we had before.

PR #3698, opened 2023-08-15
fix(web,server): album share performance

### LINKED ISSUE #3177: sharing a large album causes the web UI to hang
### The bug

I tried to share an album with ~7000 assets and found the web UI would become unresponsive and never recover. The album would be successfully shared however.

N/A

### Version of Immich Server

v1.66.1

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
N/A
```

### Additional information

_No response_

jrasm91:
Fixes #3177

Improve performance while sharing (adding and removing a user from) an album.

In this PR:
- Don't return assets on the update album endpoint response
- Don't return assets on the add album user endpoint response
- Don't load assets from the database when removing a user (mobile app needs changes before assets can be skipping when adding a new user)
- Don't ask for assets when refreshing the album info (web)

Before vs After Tests (on an album with 10k assets)

1.9 seconds to add a user (before) vs 800ms to add a user (after)
1.7 seconds to remove a user (before) vs 600ms to remove a user (after)

image

Future Improvements:
- Adding/removing assets to an album should be done without loading all the assets into memory
- Loading all the assets into memory is currently required to get the album asset count. This can be avoided, but may require using query builder or running two selects.

Tested Scenarios:
- Add a user to an album
- Remove a user from an album
- Leaving a shared album
- Creating a new album

### DIFF
--- a/server/src/domain/album/album.repository.ts
+++ b/server/src/domain/album/album.repository.ts
@@ -7,8 +7,12 @@ export interface AlbumAssetCount {
   assetCount: number;
 }
 
+export interface AlbumInfoOptions {
+  withAssets: boolean;
+}
+
 export interface IAlbumRepository {
-  getById(id: string): Promise<AlbumEntity | null>;
+  getById(id: string, options: AlbumInfoOptions): Promise<AlbumEntity | null>;
   getByIds(ids: string[]): Promise<AlbumEntity[]>;
   getByAssetId(ownerId: string, assetId: string): Promise<AlbumEntity[]>;
   hasAsset(id: string, assetId: string): Promise<boolean>;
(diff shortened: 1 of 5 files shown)


alextran1502 (approved):
Tested on mobile and it works fine.

Ah wait, I have to run this PR on mobile as well? right

jrasm91:
> Tested on mobile and it works fine.
> 
> Ah wait, I have to run this PR on mobile as well? right

There are no mobile changes in this code, so as long as the mobile app (on main/release) can create an album, add a user, remove a user, update the title/description, etc. we should be good.

PR #3721, opened 2023-08-16
feat: set person birth date (web only)

daniele-athome:
This PR implements #3348: setting the date of birth for a person and showing their age in a photo.

In a person page, there is a new context menu item (the same menu item is also in the people page in each face context menu):

context_menu

which will open a modal:

popup_dialog

Opening as draft since it somewhat works now but it lacks tests and some final polishing.

Things to address before marking as ready:

- [x] handle malformed input
- [x] handle resetting birth date (i.e. setting it to null)
- [ ] context menu doesn't close when the modal opens, not sure how to fix that yet
- [x] domain tests (server - person.service.spec.ts)
- [x] test fixtures (server - person.stub.ts)
- [x] update PR screenshots

Things to expect from review:

* using input type=date vs a proper (non-popup) date picker
* using Date vs string in DTOs, entities, etc.
* birthDate vs dateOfBirth (mainly naming differences between the database and the UI)
* e2e tests (nothing to start from, people features have no e2e tests)


jrasm91:
Can you use typed events? Look at the `edit-description.svelte` component for an example. It makes the handlers cleaner IMO.

```typescript
  const dispatch = createEventDispatcher<{
    close: void;
    updated: string;
  }>();
 
 dispatch('updated', description); 
 
// on:updated={({ detail: description }) => handleUpdateDescription(description)}
// const handleUpdateDescription = async (description: string) => {
```

daniele-athome:
I finished the development and tested it. Things to think about from a UX standpoint:

* people page: the birth date and person name modals could probably be merged into one (and that would really start to make Immich look like a contacts app :smile:)
* person page: the modal could be converted into an inline field close to the name field

I still need to make the context menu disappear from the person page after the menu item is clicked, I'll address it tonight. Nevertheless I'd like to submit this for review in the meantime.

server/src/domain/domain.util.ts
```diff
@@ -31,6 +31,13 @@ export const toBoolean = ({ value }: IValue) => {
   return value;
 };
 
+export const toDate = ({ value }: IValue) => {
+  if (value && value.length > 0) {
+    return new Date(value);
+  }
+  return value;
+};
+
```
jrasm91, line 40: This is not necessary.

server/src/domain/person/person.dto.ts
```diff
@@ -12,6 +12,15 @@ export class PersonUpdateDto {
   @IsString()
   name?: string;
 
+  /**
+   * Person date of birth.
+   */
+  @IsOptional()
+  @IsDate()
+  @Transform(toDate)
```
jrasm91, line 20: Can you replace this with `@Type(() => Date)`

server/src/domain/person/person.service.ts
```diff
@@ -63,8 +63,8 @@ export class PersonService {
   async update(authUser: AuthUserDto, id: string, dto: PersonUpdateDto): Promise<PersonResponseDto> {
     let person = await this.findOrFail(authUser, id);
 
-    if (dto.name != undefined || dto.isHidden !== undefined) {
-      person = await this.repository.update({ id, name: dto.name, isHidden: dto.isHidden });
+    if (dto.name != undefined || dto.birthDate !== undefined || dto.isHidden !== undefined) {
+      person = await this.repository.update({ id, name: dto.name, birthDate: dto.birthDate, isHidden: dto.isHidden });
```
jrasm91, line 67: This is only here, really, to determine when `SEARCH_INDEX_ASSET` needs to run. It is quite an expensive operation (loading every asset for a person and re-indexing it). It only needs to run when the name or visibility of the people changes. This is not the case for person birthdays.

Can we change it so the `await this.repository.update` always runs and the `SEARCH_INDEX_ASSET` part only runs if `name` or `isHidden` are updated?

server/src/infra/entities/person.entity.ts
```diff
@@ -30,6 +30,9 @@ export class PersonEntity {
   @Column({ default: '' })
   name!: string;
 
+  @Column({ type: 'date', nullable: true })
+  birthDate?: Date | null;
```
jrasm91, line 34:
```suggestion
  birthDate!: Date | null;
```

This is not optional, it will/should always be present.

web/src/lib/components/faces-page/people-card.svelte
```diff
@@ -54,6 +58,7 @@
       <ContextMenu on:outclick={() => (showContextMenu = false)}>
         <MenuOption on:click={() => onHideFaceClicked()} text="Hide face" />
         <MenuOption on:click={() => onChangeNameClicked()} text="Change name" />
+        <MenuOption on:click={() => onSetBirthDateClicked()} text="Set date of birth" />
```
jrasm91, line 61:
```suggestion
        <MenuOption on:click={() => dispatch('set-birth-date', person)} text="Set date of birth" />
```

web/src/lib/components/faces-page/people-card.svelte
```diff
@@ -17,6 +17,10 @@
     dispatch('change-name', person);
   };
 
+  const onSetBirthDateClicked = () => {
+    dispatch('set-birth-date', person);
```
jrasm91, line 21: I realize this is following exist code in this component, but in general if the component doesn't do anything besides emit an event and it only does it in one place I usually inline it. All of these events could be inlined. Also, we should use typed events here as well. Basically just adding

```typescript
let dispatch = createEventDispatcher<{
  'change-name': PersonResponseDto,
  'set-birth-date': PersonResponseDto,
  'merge-faces': PersonResponseDto,
}>();
```

In all reality, these could/should probably all be `event: void,`, since the dto isn't changed and the parent is passing it in already.

web/src/routes/(user)/people/+page.svelte
```diff
@@ -261,6 +268,37 @@
...
   };
 
+  const submitBirthDateChange = async (value: string) => {
+    showSetBirthDateModal = false;
+    if (!edittingPerson) {
+      return;
+    }
+    if (value === edittingPerson.name) {
+      return;
+    }
```
jrasm91, line 278: These can be merged into a single condition.

web/src/routes/(user)/people/+page.svelte
```diff
@@ -324,6 +362,7 @@
             <PeopleCard
               {person}
               on:change-name={handleChangeName}
+              on:set-birth-date={handleSetBirthDate}
```
jrasm91, line 365: You could also just do to avoid using the `CustomEvent<T>` interface in the handlers. It's cleaner IMO.

```suggestion
              on:set-birth-date={() => handleSetBirthDate(person)}
```

jrasm91 (commented):
The API specifically allows setting the birthday to `null` as far as I can tell. Is it possible to clear a birthday from the UI as well? Often we ignore/skip empty values, which is fine, but then we might want to add a "Remove Birthdate" type of button.

server/src/domain/person/person.dto.ts
daniele-athome: I couldn't handle null values if I used `@Type(() => Date`... I'll try another approach, also see my comment about API best practices below.

jrasm91:
Sounds good. I think throwing an error for an empty string in the API and mapping empty string to null in the client is a good approach.

I don't really care very much about where/how the "remove birthday" works. Setting it to an empty string and hitting save/confirm seems fine IMO. I've also seen dedicated menu items and/or an explicit remove button, but I don't think these are necessary here.

server/src/domain/person/person.dto.ts
jrasm91: You can add `ValidateIf((value) => value !== null)` instead.

web/src/lib/components/faces-page/people-card.svelte
daniele-athome: Sure about this? All the others are simple one-liner methods. Or should I inline the others as well?

web/src/lib/components/faces-page/people-card.svelte
daniele-athome: I changed all of them following your suggestion. However I left PersonResponseDto because it would need more refactorings.

web/src/routes/(user)/people/+page.svelte
daniele-athome: I also fixed the other one in submitNameChange (I was too fixated on following existing code eheh).

daniele-athome:
I'm trying to find a way to close the context menu after opening the modal, but the rest is done.

jrasm91 (approved):

DrSpaldo:
Once the birthday is added, how do you run a job to remove all the photos that were incorrectly tagged prior to the birthday?

daniele-athome:
There is no such job, but you could delete them directly from the database if you feel up to it (thumbnails won't be purged though).

PR #3724, opened 2023-08-16
chore(web): better explain what the thumbnails type are for

daniele-athome:
I also changed the title of the settings (for some users WEBP and JPG might make little sense). I don't know if it's correct to spell "small thumbnails resolution" with or without the trailing "S" in the word "thumbnails" - sorry, not a native speaker.

I rewrote the helper text several times trying to find the best way to explain thumbnails usage; this is the best I could come up with. Feel free to suggest anything else of course.


daniele-athome:
Maybe "used when viewing group of photos" is missing an "a":

> Used when viewing a group of photos"

jrasm91 (approved):

PR #3726, opened 2023-08-16
fix(web): label for attribute

daniele-athome:
Fixed also in a comment because why not :-)


jrasm91 (approved):
Is there anyway to automate this? It seems like there should be some tools or rules out there to enforce basic accessibility.

alextran1502 (approved):

daniele-athome:
I wouldn't know, sorry. My IDE was warning me about that so I had to fix it :-)

PR #3753, opened 2023-08-17
fix(web): escape shortcut

martabal:
This PR aims to enhance how the 'Escape' key shortcut is handled on the web app.

### DIFF
--- a/web/src/lib/components/album-page/edit-description-modal.svelte
+++ b/web/src/lib/components/album-page/edit-description-modal.svelte
@@ -6,18 +6,17 @@
 
   const dispatch = createEventDispatcher<{
     close: void;
-    updated: string;
+    save: string;
   }>();
   export let album: AlbumResponseDto;
 
   let description = album.description;
 
-  const handleSave = () => {
-    dispatch('updated', description);
-  };
+  const handleCancel = () => dispatch('close');
+  const handleSubmit = () => dispatch('save', description);
 </script>
 
-<FullScreenModal on:clickOutside={() => dispatch('close')}>
+<FullScreenModal on:clickOutside={handleCancel} on:escape={handleCancel}>
   <div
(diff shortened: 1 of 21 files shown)


danieldietzler:
Hey @martabal, a few minor things:

- Escape in the search bar closes the "smart search is enabled by default..." thing but the focus is still in the search bar. Is this intentional?
- Currently it is not possible to exit the map menu with escape. I think that should be added for consistency.
- Also for consistency: Some pages allow going back using escape. Create Album for instance brings you back to the albums view when pressing escape. Other pages however, don't currently support this. I've seen it not working with: Add to album and sharing -> shared links (create shared album works)
- User Management -> Create User/Edit User: Escape not working
- User Settings -> Partner Sharing -> Add Partner: Escape not working
- User Settings -> API Keys -> New API Key: Escape not working

martabal:
@danieldietzler thanks for the suggestions, I really appreciate that :smile: 

> * Escape in the search bar closes the "smart search is enabled by default..." thing but the focus is still in the search bar. Is this intentional?

It is. IMO, if the "big search bar" (the smart search is enabled by default..." thing) is showed, the first press should hide it, then the second press should bring the user to the previous route (like if you clicked on the left arrow).

jrasm91 (approved):
I mean, it works for the most part, but it is all a bit messy IMO. It would be great to clean up the interfaces in the future. I think, especially for modals, something like a way to specify the action that should happen when you click `escape` and the action that should happen when you click outside. It would be good for there to be an easy, standardized way to set this behavior. Maybe you can take a look at this in the future. Something like:

```
<FullScreenModal escape="none" backdrop="none" on:confirm/>
<FullScreenModal escape="cancel" backdrop="none" on:confirm on:cancel/>
<FullScreenModal escape="cancel" backdrop="cancel" on:confirm on:cancel/>
```

Essentially these events all map to either (1) ignore or (2) cancel. I don't think it makes sense to clutter everything up with so many listeners that are all propagated up the stack.

PR #3768, opened 2023-08-18
feat(web,server)!: configure machine learning via the UI

jrasm91:
# Breaking Change
The way we disabled machine learning has changed in this pull request.

To disable machine learning now, use `IMMICH_MACHINE_LEARNING_ENABLED=false` 

(previously `IMMICH_MACHINE_LEARNING_URL=false`)
<hr>

In this PR:

- Make machine learning url configurable via the admin settings (still can be set/disabled via `IMMICH_MACHINE_LEARNING_URL`)
- Add options to disable certain ML features in the UI (these skip corresponding jobs)
- Remove unused search config endpoint (replaced with the new features endpoint)
- Update jobs to become disabled with corresponding feature flags

Tested scenarios:
- Start up with ML disabled via env verify jobs are disabled and ML doesn't run on upload
- Start up with ML enabled and disabled via the UI and verify ML stops running
- Start up with ML enabled and disable individual features. Verify associated jobs stop running
- Start up with ML enabled and verify all jobs still function correctly

image


martabal:
Awesome !
One concern : Users who use the AIO image don't have to worry about configuring the url of the machine-learning (it's 127.0.0.1). Can we still rely on the `IMMICH_MACHINE_LEARNING_URL` env for the default ?

bo0tzz:
The ML URL is the sort of environment config that I think should really be left in an env var if we can. I do like all the feature flag options!

jrasm91:
We can certainly leave it as an environment variable. I personally switch mine between two machines all the time, so would love to be able to configure it dynamically. I guess I'll think about it some more and see if there is a clean way to do it.

jrasm91:
> The ML URL is the sort of environment config that I think should really be left in an env var if we can. I do like all the feature flag options!

What about if we make the "default" value configurable via an environment variable, but you can override it via the UI and a value from the database? So overrides get merged into defaults, but we can make the default value something like `process.env.MACHINE_LEARNING_URL || immich-machine-learning:3003`?

server/src/domain/system-config/system-config.core.ts
```diff
@@ -44,6 +44,13 @@ export const defaults = Object.freeze<SystemConfig>({
     [QueueName.THUMBNAIL_GENERATION]: { concurrency: 5 },
     [QueueName.VIDEO_CONVERSION]: { concurrency: 1 },
   },
+  machineLearning: {
+    enabled: process.env.IMMICH_MACHINE_LEARNING_URL !== 'false',
```
martabal, line 48: Currently, typesense has the environment variable `TYPESENSE_ENABLED`. Can we use `MACHINE_LEARNING_ENABLED` to be more consistent ?

jrasm91: Definitely. It'll just be a breaking change for those who have it set currently.

bo0tzz: I think this is a change we should make as well and this seems like a good moment to do it.

jrasm91:
There are some things that I would like to improve upon, but I think it might be better to do it in a separate pull request. It would be great to organize, standardize, and consolidate all the environment variables and configuration settings. There is an open PR to load config from a file and we also have some variables with `IMMICH_` prefix and others without. I would love to merge this and the config file PR, standardize the names, and update the documentation all in a single release.

alextran1502 (approved):

PR #3855, opened 2023-08-25
refactor(web): shared link key auth

jrasm91:
## In this PR:
- Set the shared link key on application start-up `api.setKey()`
- Remove all _props_ used to pass the key down the component tree
- Use `api.getKey()` on all api requests that might use it
- Add `api.isSharedLink` that can be used globally for conditional display/loading logic

Basically remove all the passed props in lieu of setting it once and then using it directly in the api request.

## Tested Scenarios
- View a public shared link for an album
- View a public shared link for individual assets
- View asset details for a shared link album asset
- View asset details for a shared link individual asset
- View asset details for an asset from a logged in user's timeline
- Download individual assets from a shared link

### DIFF
--- a/web/src/api/api.ts
+++ b/web/src/api/api.ts
@@ -39,6 +39,11 @@ export class ImmichApi {
   public userApi: UserApi;
 
   private config: Configuration;
+  private key?: string;
+
+  get isSharedLink() {
+    return !!this.key;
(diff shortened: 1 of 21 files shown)


alextran1502 (approved):
Neat! Thank you!

PR #3867, opened 2023-08-25
feat(web): skeleton on asset loading

JasBogans:


jrasm91 (approved):
Lgtm

jrasm91 (commented):
It would be nice to add this to other pages. Maybe there is a way to build it into the Asset grid component instead.

JasBogans:
> It would be nice to add this to other pages. Maybe there is a way to build it into the Asset grid component instead.

It can be used in the Favourites, Archived and also in the Search page. 
Off topic: one thing I noticed is that in the search there is not date, is that for a particular reason?
image

PR #3872, opened 2023-08-26
feat(mobile): upload image assets before videos

shenlong-tanwen:
### Changes made in this PR:
Asset backup flow is updated to always prefer image files before uploading video files. Mobile device uploads are sequential and so, uploading a huge video file might block the other uploads from getting processed faster.  Since most of the times, image assets are smaller than video assets, it makes sense to try to upload them so as to maximize the number of uploads. This is especially useful in iOS if a large video asset gets retried in a loop blocking other smaller image assets from getting uploaded
- #3341 

### How was this tested?
The backup service code is common for both background upload and manual asset upload. I've tested the change only with manual upload. Background upload should work properly as well

- Tried uploading a mix of image and video assets manually and ensured that images are getting uploaded before the video assets are processed.

### Feedback required
- I'm not sure if the current flow of uploading the assets as they come is useful. But if people prefer that, we can have this "image before video" behind a toggle in the backup page?

### DIFF
--- a/mobile/lib/modules/backup/services/backup.service.dart
+++ b/mobile/lib/modules/backup/services/backup.service.dart
@@ -218,7 +218,14 @@ class BackupService {
     bool anyErrors = false;
     final List<String> duplicatedAssetIds = [];
 
-    for (var entity in assetList) {
+    // Upload images before video assets
+    // AssetType::audio is greater than image / video
+    // But since we do not support them, it shouldn't be an issue
+    List<AssetEntity> sortedAssets = assetList.sorted(
+      (a, b) => a.typeInt - b.typeInt,
+    );
+
+    for (var entity in sortedAssets) {
       try {
         if (entity.type == AssetType.video) {
           file = await entity.originFile;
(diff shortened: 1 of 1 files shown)


mobile/lib/modules/backup/services/backup.service.dart
```diff
@@ -196,6 +196,16 @@ class BackupService {
         : candidates.whereNot((e) => existing.contains(e.id)).toList();
   }
 
+  // AssetType::audio is greater than image / video
+  // But since we do not support them, it shouldn't be an issue
+  int _sortByAssetType(AssetEntity a, AssetEntity b) {
+    return a.typeInt < b.typeInt
```
fyfrey, line 202: Simply `a.typeInt - b.typeInt` should work the same

shalong-tanwen: oh my, that's true. Let me update this real fast. thank you!

fyfrey:
Thank you for the PR - it's probably a good idea to prefer images over videos in most cases.
The default sorting is an unstable sorting, thus the images are no longer in any particular order.
Could you make the sorting depending on multiple values: asset type and then creation date (so that older assets are uploaded first - FIFO as so not to starve old assets)

shalong-tanwen:
> Could you make the sorting depending on multiple values: asset type and then creation date (so that older assets are uploaded first - FIFO as so not to starve old assets)

I've now updated the sorting to handle both the properties while backing up an asset. The asset gets sorted by type, and if they are equal then by creation date. Great suggestion, thank you!

fyfrey:
Thank you for the PR - it's probably a good idea to prefer images over videos in most cases.
The default sorting is an unstable sorting, thus the images are no longer in any particular order.
Could you make the sorting depending on multiple values: asset type and then creation date (so that older assets are uploaded first - FIFO as so not to starve old assets)

fyfrey (approved):
Excellent. Looks good to me now!

PR #3890, opened 2023-08-27
feat(web,server): Shared accounts / sharing pictures with family / pictures in timeline / "family space"

### LINKED ISSUE #1043: [Feature]: User Groups for sharing/permissions
### Feature detail

I'd like be able to use groups of users. Either created by the admin, users, imported from oauth, or some combination thereof. Ideally the end result would be able to use all of those options together. Admin created through local and if they control the provider oauth information, with users being able to create their own local groups.

Reasoning behind this request is that all sides of my family have requested me to create a space for them to share photos with the family that isn't tied to some specific platform. Immich is far and above my top choice for this compared to other projects I have seen out there. Downside is that with my parents being divorced and then my wifes side of the family that's essentially three separate units. My mother doesn't want my father and his side of the family to see pictures of her birthday party for example. I don't want to run three instances of Immich because then photos I'd like all three family's to see need to be duplicated. 

With user groups I could create a "Family A" group and share an event like "my mothers birthday party" album that was just pertinent to that side of the family without having to specify every member every time we create an album for an event. Same for "Family B" and "Family C". Then if there was some album I wanted to share with everyone, say my own birthday party I could share to all three groups or a group that contains those three nested.

The other advantage of groups vs individual sharing would be as new members of the family join the server and placed in their group they gain access to all albums shared to that family instead of having to go back and add their accounts to all existing albums.

This request in conjunction with nested albums requested in #1010 would create my perfect scenario for sharing and permissions.  If #402 and #34 were implemented, Immich would become the only real choice in the photo hosting/backup realm in my humble opinion.

### Platform

Server

korotovsky:
Hi Immich team,

Thanks for the great project,

As was discussed, I created this PR to collect further feedback and let it wait until libraries functionality gets merged.

**The info and background about this PR**: It works in a way like non interactive login and you can assign other accounts to be logged into this account programmatically once you have performed an interactive login with password/SAML. As as result multiple users ends up into "shared" account where all photos belongs to the same user aka one family group. Family group can also share its own albums to other users in Immich, this way we achieve two-hop sharing (see FAQ below for more info).

Audit log has been patched as well, you can also see you is actually using this meta account.

Screenshot 2023-08-25 at 00 01 50

Screenshot 2023-08-24 at 23 55 59

This might be not ideal, but amount of effort vs product outcome in my opinion is great:

**This implementation gives us**:
- One shared family space as an virtual account;
- Closes https://github.com/immich-app/immich/discussions/3886
- Closes https://github.com/immich-app/immich/discussions/3781
- Partially closes https://github.com/immich-app/immich/issues/1043
- Closes https://github.com/immich-app/immich/discussions/3843 if belongs to the same family
- Closes https://github.com/immich-app/immich/discussions/3418 if belongs to the same family;
- Display shared photos in main photo feed;
- Search capabilities;
- Facial data can be shared;
- Geo data can be shared;
- Edit/delete functionalities within "family" account;
- **Simple**: does not touch core immich queries for selecting photos, etc because implemented as part of authentication layer.

**FAQ**:
What is "_non-interactive login account_":
- Account has restricted abilities for interactive login: password/SAML is blocked and you can not login into this account directly from the login page.

How do you see managing the change of password with your solution?
- Password change and initial password-set forms are not affected by this functionality for accounts with interactive login capabilities.

What is two-hop sharing?
- Account "A" (with Google OpenID/password credentials) -> ---- does programmatic authentication ----- -> account C (non interactive account "Family Space") - > ----- shares album to other accounts in Immich -> Account E (with Google OpenID/password credentials) 
  - Account "A" - me
  - Account "B" - other person in same household
  - Account "C" - family space (non interactive login account)
  - Account "E" - parent

### DIFF
--- a/server/immich-openapi-specs.json
+++ b/server/immich-openapi-specs.json
@@ -5301,6 +5301,10 @@
           "id": {
             "type": "string"
           },
+          "parentAuthId": {
+            "nullable": true,
+            "type": "string"
+          },
           "updatedAt": {
             "type": "string"
           }
@@ -5311,7 +5315,8 @@
           "updatedAt",
           "current",
           "deviceType",
-          "deviceOS"
+          "deviceOS",
+          "parentAuthId"
         ],
         "type": "object"
       },
@@ -5568,6 +5573,9 @@
           },
           "firstName": {
             "type": "string"
+          },
+          "interactiveLoginEnabled": {
+            "type": "boolean"
           },
           "lastName": {
             "type": "string"
@@ -7288,6 +7296,9 @@
             "format": "uuid",
             "type": "string"
           },
+          "interactiveLoginEnabled": {
+            "type": "boolean"
+          },
           "isAdmin": {
             "type": "boolean"
           },
@@ -7298,6 +7309,9 @@
             "type": "boolean"
           },
           "password": {
+            "type": "string"
+          },
+          "sharedAccountId": {
             "type": "string"
           },
           "shouldChangePassword": {
@@ -7379,6 +7393,9 @@
           "id": {
             "type": "string"
           },
+          "interactiveLoginEnabled": {
+            "type": "boolean"
+          },
           "isAdmin": {
             "type": "boolean"
           },
@@ -7392,6 +7409,9 @@
             "type": "string"
           },
           "profileImagePath": {
+            "type": "string"
+          },
+          "sharedAccountId": {
             "type": "string"
           },
           "shouldChangePassword": {
--- a/server/src/domain/auth/auth.service.ts
+++ b/server/src/domain/auth/auth.service.ts
@@ -36,6 +36,7 @@ export interface LoginDetails {
   clientIp: string;
   deviceType: string;
   deviceOS: string;
+  authParent: string | null;
 }
 
 interface LoginResponse {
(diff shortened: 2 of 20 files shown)


AngelaDMerkel:
In my opinion this is PR moves further away from the solution I would prefer to this problem. Having one shared user rather than being able to seamlessly view shared images in the timeline only complicates ownership of images. In a case where multiple different users are on the same instance, maybe four friends, it doesn't make sense to have a "family" user for our shared events. A preferred behaviour would be to allow users to take take ownership of an image shared to them as in Google Drive's "Save to my Cloud" feature. Doing this retains a record of who owns the original and prevents it from being duplicated. Not to mention many people take uninteresting and ugly photos and you might want to only see good ones from your friends in your shared timeline.

daniele-athome:
This seems to me like a workaround to a problem addressable maybe with the concept of groups IMHO.
A user account is a first-class concept in an application (non-tech users won't understand the concept of "non-interactive login user account"), and using a "fake account" to work around the fact that groups are more difficult to implement seems like taking in techical debt voluntarily.

Also I agree with @AngelaDMerkel about keeping information about asset's ownership and origin.

jarneson33:
I would also like to add on to the current discussion and agree that this does not seem like the right direction and I personally would be disappointed if we considered all the issues linked in this PR as "closed" especially since if you have to add the conditional of "if belongs to the same family" then you aren't really closing it.

Reading the initial feature requests it looks like the users just want an easy way to add a bunch of users to a shared album. Seems like we should be able to give the user the ability to create a group and that group is an option when sharing. When you create a shared album today and add users you have the ability to add multiple at a time, which means the app already handles adding a collection of users to an album, so it seems very feasible that we can create a group object that is just a collection of users and have it as a selectable option when sharing.

jarneson33:
oh and I too Also I agree with @AngelaDMerkel about keeping information about asset's ownership and origin. That is one point of frustration with our families shared albums is not seeing who added what.

fivestones:
> it looks like the users just want an easy way to add a bunch of users to a shared album.

That would be a good feature, but for me personally (and probably many others) what I'm wanting is for my wife's pictures and my pictures to all be visible in my timeline, together, but to be able to easily filter the timeline to show only mine or only hers. I think this could broaden to be for any kind of shared photos too--just have filter settings for the timeline which let you turn on or off the inclusion of photos shared by different people.

This kind of filtering would optimally apply to lots of other things too like types of photos, camera that took the photos, screenshots, locations, etc. So I could quickly set a filter that shows me only photos shared by my wife, in California, in 2008, that were taken by a canon camera. And then I could run a CLIP search on those. And I could then make this filter set a saved set that could show up on the left side next to the other links, either static (basically an album made up of all the photos in the current search results) or like a live search which will use that same set of filters and in the future include new photos that also match, just by clicking the link for this filter set on the left.

Then I can have links on the left that say "My photos", "My wife's photos", "All the photos"--and clicking any of these shows a timeline with those respective sets of photos in it.

Somebody else may have links that say "My photos", "Housemates' photos", and "All the photos". Any one of these should be able to be set as the default that comes up in your timeline when you log in.

Maybe I just need to make this a new feature request.

bo0tzz:
I will close this PR as it doesn't seem like it will get merged. Thank you for getting some good discussion going!

PR #3905, opened 2023-08-29
fix (mobile): Fix slow album thumbnail generation for album picker

a-elkaim:
When generating the thumbnail of an album, all of the pictures of the album are retrieved but only the first picture of the album is used.

Retrieving all of the pictures of the album at once can cause huge performance issues on large albums.

Since only the first picture is used to generate the thumbnail, this commit uses the getAssetListPaged method instead of getAssetListRange, effectively retrieving only the first picture of the album.

Note that this issue has been observed on Android but may also happen on iOS devices.

### DIFF
--- a/mobile/lib/modules/backup/providers/backup.provider.dart
+++ b/mobile/lib/modules/backup/providers/backup.provider.dart
@@ -218,22 +218,18 @@ class BackupNotifier extends StateNotifier<BackUpState> {
       final assetCountInAlbum = await album.assetCountAsync;
       if (assetCountInAlbum > 0) {
         final assetList = await album.getAssetListPaged(page: 0, size: 1);
-
-        if (assetList.isNotEmpty) {
-          final thumbnailAsset = assetList.first;
-
-          try {
-            final thumbnailData = await thumbnailAsset
-                .thumbnailDataWithSize(const ThumbnailSize(512, 512));
-            availableAlbum =
-                availableAlbum.copyWith(thumbnailData: thumbnailData);
-          } catch (e, stack) {
-            log.severe(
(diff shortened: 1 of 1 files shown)


mobile/lib/modules/backup/providers/backup.provider.dart
```diff
@@ -217,8 +217,7 @@ class BackupNotifier extends StateNotifier<BackUpState> {
 
       final assetCountInAlbum = await album.assetCountAsync;
       if (assetCountInAlbum > 0) {
-        final assetList =
-            await album.getAssetListRange(start: 0, end: assetCountInAlbum);
+        final assetList = await album.getAssetListPaged(page: 0, size: 1);
 
         if (assetList.isNotEmpty) {
```
EnochPrime, line 222: If `assetList` could potentially be empty could the wrapping `assetCount > 0` be removed? Then it also shouldn't be necessary to query the asset count at all here.

azsde: I don't think so as we do not want to list empty albums, that being said we can probably remove `if (assetList.isNotEmpty)`

What do you think @alextran1502 ?

EnochPrime: Honestly I haven't looked at the implementation of each, but it seems logical to me that an asset count greater than 0 would be synonymous with an non empty list of assets from 0 to asset count.

I don't think an album with 0 assets would be able to return a non zero amount of assets.

azsde: Ok, i've uploaded another commit

alextran1502 (approved):

PR #3925, opened 2023-08-30
Create real anchors around admin sidebar buttons

daniele-athome:
This will make possible treating the admin sidebar buttons as links, thus enabling other gestures, e.g. Ctrl+Click.
The main (non-admin) sidebar already works like this (correctly).


alextran1502 (approved):
Thank you; TIL you have to make an anchor to be able to ctrl + click

PR #3926, opened 2023-08-30
Remove obsolete environment variables

daniele-athome:
Opening as draft since it hasn't been tested yet.

### DIFF
--- a/docs/docs/install/environment-variables.md
+++ b/docs/docs/install/environment-variables.md
@@ -51,7 +51,6 @@ These environment variables are used by the `docker-compose.yml` file and do **N
 | `PORT`                  | Web Port              | `3000`  | web              |
 | `SERVER_PORT`           | Server Port           | `3001`  | server           |
 | `MICROSERVICES_PORT`    | Microservices Port    | `3002`  | microservices    |
-| `MACHINE_LEARNING_PORT` | Machine Learning Port | `3003`  | machine learning |
 
 ## URLs
 
@@ -59,8 +58,6 @@ These environment variables are used by the `docker-compose.yml` file and do **N
 | :-------------------------------- | :--------------------------- | :-----------------------------------: | :-------------------- |
 | `IMMICH_WEB_URL`                  | Immich Web URL               |       `http://immich-web:3000`        | proxy                 |
 | `IMMICH_SERVER_URL`               | Immich Server URL            |      `http://immich-server:3001`      | web, proxy            |
-| `IMMICH_MACHINE_LEARNING_ENABLED` | Enabled machine learning     |                `true`                 | server, microservices |
-| `IMMICH_MACHINE_LEARNING_URL`     | Immich Machine Learning URL, | `http://immich-machine-learning:3003` | server, microservices |
 | `PUBLIC_IMMICH_SERVER_URL`        | Public Immich URL            |      `http://immich-server:3001`      | web                   |
 | `IMMICH_API_URL_EXTERNAL`         | Immich API URL External      |                `/api`                 | web                   |
 
@@ -180,16 +177,15 @@ Typesense URL example JSON before encoding:
 
 | Variable                                    | Description                    |        Default        | Services         |
 | :------------------------------------------ | :----------------------------- | :-------------------: | :--------------- |
-| `MACHINE_LEARNING_MIN_FACE_SCORE`           | Minimum Face Score             |         `0.7`         | machine learning |
 | `MACHINE_LEARNING_MODEL_TTL`                | Model TTL                      |         `300`         | machine learning |
-| `MACHINE_LEARNING_EAGER_STARTUP`            | Eager Startup                  |        `true`         | machine learning |
-| `MACHINE_LEARNING_MIN_TAG_SCORE`            | Minimum Tag Score              |         `0.9`         | machine learning |
-| `MACHINE_LEARNING_FACIAL_RECOGNITION_MODEL` | Facial Recognition Model       |      `buffalo_l`      | machine learning |
-| `MACHINE_LEARNING_CLIP_TEXT_MODEL`          | Clip Text Model                |    `clip-ViT-B-32`    | machine learning |
-| `MACHINE_LEARNING_CLIP_IMAGE_MODEL`         | Clip Image Model               |    `clip-ViT-B-32`    | machine learning |
-| `MACHINE_LEARNING_CLASSIFICATION_MODEL`     | Classification Model           | `microsoft/resnet-50` | machine learning |
 | `MACHINE_LEARNING_CACHE_FOLDER`             | ML Cache Location              |       `/cache`        | machine learning |
 | `TRANSFORMERS_CACHE`                        | ML Transformers Cache Location |       `/cache`        | machine learning |
+
+:::info
+
+Other machine learning parameters can be tuned from the admin UI.
+
+:::
 
 ## Docker Secrets
 
--- a/docs/docs/install/environment-variables.md
+++ b/docs/docs/install/environment-variables.md
@@ -47,7 +47,7 @@ These environment variables are used by the `docker-compose.yml` file and do **N
 ## Ports
 
 | Variable                | Description           |  Default  | Services         |
-|:------------------------|:----------------------|:---------:| :--------------- |
+| :---------------------- | :-------------------- | :-------: | :--------------- |
 | `PORT`                  | Web Port              |  `3000`   | web              |
 | `SERVER_PORT`           | Server Port           |  `3001`   | server           |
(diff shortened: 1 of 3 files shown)


docs/docs/install/environment-variables.md
```diff
@@ -51,16 +51,13 @@ These environment variables are used by the `docker-compose.yml` file and do **N
 | `PORT`                  | Web Port              | `3000`  | web              |
 | `SERVER_PORT`           | Server Port           | `3001`  | server           |
 | `MICROSERVICES_PORT`    | Microservices Port    | `3002`  | microservices    |
-| `MACHINE_LEARNING_PORT` | Machine Learning Port | `3003`  | machine learning |
```
mertalev, line 54: This one is still used by the ML service. The UI setting points requests at the given URL, while this is the port the ML service actually listens to.

docs/docs/install/environment-variables.md
```diff
@@ -51,16 +51,13 @@ These environment variables are used by the `docker-compose.yml` file and do **N
...
 
 ## URLs
 
 | Variable                          | Description                  |                Default                | Services              |
 | :-------------------------------- | :--------------------------- | :-----------------------------------: | :-------------------- |
 | `IMMICH_WEB_URL`                  | Immich Web URL               |       `http://immich-web:3000`        | proxy                 |
 | `IMMICH_SERVER_URL`               | Immich Server URL            |      `http://immich-server:3001`      | web, proxy            |
-| `IMMICH_MACHINE_LEARNING_ENABLED` | Enabled machine learning     |                `true`                 | server, microservices |
-| `IMMICH_MACHINE_LEARNING_URL`     | Immich Machine Learning URL, | `http://immich-machine-learning:3003` | server, microservices |
```
mertalev, line 63: There's a `MACHINE_LEARNING_HOST` env in the ML service used in addition to `MACHINE_LEARNING_PORT`, but it seems it's not listed on this page. It doesn't make sense to change it from the default `0.0.0.0` if running in a container, but for native deployments it could be useful.

docs/docs/install/environment-variables.md
```diff
@@ -180,17 +177,16 @@ Typesense URL example JSON before encoding:
...
 | `MACHINE_LEARNING_MODEL_TTL`                | Model TTL                      |         `300`         | machine learning |
-| `MACHINE_LEARNING_EAGER_STARTUP`            | Eager Startup                  |        `true`         | machine learning |
-| `MACHINE_LEARNING_MIN_TAG_SCORE`            | Minimum Tag Score              |         `0.9`         | machine learning |
-| `MACHINE_LEARNING_FACIAL_RECOGNITION_MODEL` | Facial Recognition Model       |      `buffalo_l`      | machine learning |
-| `MACHINE_LEARNING_CLIP_TEXT_MODEL`          | Clip Text Model                |    `clip-ViT-B-32`    | machine learning |
-| `MACHINE_LEARNING_CLIP_IMAGE_MODEL`         | Clip Image Model               |    `clip-ViT-B-32`    | machine learning |
-| `MACHINE_LEARNING_CLASSIFICATION_MODEL`     | Classification Model           | `microsoft/resnet-50` | machine learning |
 | `MACHINE_LEARNING_CACHE_FOLDER`             | ML Cache Location              |       `/cache`        | machine learning |
 | `TRANSFORMERS_CACHE`                        | ML Transformers Cache Location |       `/cache`        | machine learning |
```
mertalev, line 182: `TRANSFORMERS_CACHE` can be removed. It doesn't actually do anything since the cache is set explicitly in the code to the model's own cache folder.

mertalev: Also, there are a few new threading env variables that are missing here: `MACHINE_LEARNING_REQUEST_THREADS`, `MACHINE_LEARNING_MODEL_INTER_OP_THREADS`  and `MACHINE_LEARNING_MODEL_INTRA_OP_THREADS`. 

`MACHINE_LEARNING_REQUEST_THREADS`: sets the size of the request thread pool, defaults to cpu count

`MACHINE_LEARNING_MODEL_INTER_OP_THREADS`: sets the number of model ops to process in parallel, defaults to 1

`MACHINE_LEARNING_MODEL_INTRA_OP_THREADS`: sets the number of threads to handle each model op, defaults to 1
* Recommended to try changing this first rather than `MACHINE_LEARNING_MODEL_INTER_OP_THREADS` if making changes

mertalev (changes requested):
Thanks for cleaning this up!

docs/docs/install/environment-variables.md
daniele-athome: I added that too in the "Ports" section - it seemed the right place although the section is called "Ports"... maybe we should rename it to something like "Ports/Bindings" or something?

docs/docs/install/environment-variables.md
daniele-athome: I don't think I understand exactly the function of INTER_OP_THREADS vs INTRA_OP_THREADS - well, I understand the cardinality effects on the concurrency, but I don't understand their purpose so I might not have explained them well. Suggestions are very welcome :-)

docs/docs/install/environment-variables.md
mertalev: Yeah, that's be good

mertalev (approved):
LGTM

stavros-k:
I thinki the variable `IMMICH_MACHINE_LEARNING_URL` was removed accidentaly!

daniele-athome:
I believe that variable has been replaced by an admin setting.

stavros-k:
> I believe that variable has been replaced by an admin setting.

I see it used here:
https://github.com/immich-app/immich/blob/45f9c52e7f35a8e2fc6a4069569b87d2e9bf465f/server/src/domain/system-config/system-config.core.ts#L61

But if the intent is to remove that and move it in UI config setting, it will break both TrueNAS and TrueCharts Helm Packages.
As we won't be able to automatically define the URL of the machine learning Pod

daniele-athome:
@mertalev I'm going to restore `IMMICH_MACHINE_LEARNING_URL` but what about `IMMICH_MACHINE_LEARNING_ENABLED`? I removed that too but it seems to be used still, right?

stavros-k:
> @mertalev I'm going to restore `IMMICH_MACHINE_LEARNING_URL` but what about `IMMICH_MACHINE_LEARNING_ENABLED`? I removed that too but it seems to be used still, right?

Yes that need to be restored too! I missed that this was also removed

mertalev:
We allow settings to be set through a JSON config, and the direction has been to have a single source of truth for settings. You can set both `url` and `enabled` in this config, so having envs for them seems redundant.

daniele-athome:
So it's either:

1. remove any use of environment variables regarding ML enabled flag and URL
2. fix documentation until environment variables are removed for real

For option 2 you can take [my commit here](https://github.com/daniele-athome/immich/commit/d7a92ae3239cc8cd02a3677e8496a3dda7caed48).

stavros-k:
Keep in mind that people (eg TrueNAS and TrueCharts) that packaging this for easy deployment for other users in **kubernetes** will mount the json file with a configmap/secret. 
Which will be a read-only file, thus preventing users to modify it. 
With env's projects like the one mentioned, only have to specify what is needed to connect for example database, redis, ML, etc. And let the user define extra env's for fine tuning/optimizing

In the case of the config file. those projects will have to expose every option in the UI, resulting in a "bloated" UI.

jrasm91:
Obviously there is a balance between ease-of-automatic-deployment, ease-of-change, and ease-of-maintenance. Originally we were going to remove the two ML environment variables, but they got added back to avoid breaking existing deployments. This is not to say we won't remove them in the future though.

The TLDR is we're (at least right now) mainly interested/focused on ease-of-maintenance and ease-of-change (at perhaps the expense of ease-of-automatic-deployment).

The goal is definitely to keep the amount of configuration related code minimal while still making it is easy to change after the app is set up. As of now, you can use a configuration file OR the admin settings UI. Intermixing both leads to additional complexity and confusion.

stavros-k:
I agree, you can't have every possible option always, but defining things like host/port/passwords, should be part of the deployment. And in cases for other projects that pre-packages various projects, it helps "locking" such config options, to prevent users that do not have the knowledge of setting up containers/databases and linking them together, from shooting themselves on the foot.

On the other hand, tuning the machine learning tidbits is more suited for a "live" tuning and very dependent of the user/system hence UI makes more sense.

Anyway, this is of course my opinion. I can't force (nor I want to) what the choice will be.
Just dropped some use cases that are out there that might have slipped through the cracks.
For my personal use, as a power user, env or json doesn't make huge difference.
Thanks.

Issue #3166, opened 2023-07-08
web UI becomes unresponsive when uploading files


uhthomas:
### The bug

Uploading multiple files via the web brings the entire UI to a crawl and becomes unusable until the uploads complete.

### The OS that Immich Server is running on

N/A

### Version of Immich Server

73e82303e7db6f4192eb74ae0aa60ac7255c93e0

### Version of Immich Mobile App

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
N/A
```

lifeart:
We could have upload progress outside of modal window and call resolve before upload here: https://github.com/immich-app/immich/blob/main/web/src/lib/utils/file-uploader.ts#L72

uhthomas:
I believe this has something to do with how thumbnails of uploading assets are rendered? I can see they never fully load and I'm not sure how they're generated. If the entire image is used, then that would cause problems as these files are 6336 x 9504.

uhthomas:
Yeah... We should change this.

https://github.com/immich-app/immich/blob/c8f1a15f21e11d9eda0af652d7c5dca947b962f3/web/src/lib/components/shared-components/upload-asset-preview.svelte#L12C1-L12C1

bo0tzz:
Fixed in #4219

PR #3931, opened 2023-08-31
fix(web) UI becomes unresponsive when uploading files

valonsodev:
This is an approach to https://github.com/immich-app/immich/issues/3166. 
As commented there the biggest problem is how previews are simply the whole image as an ObjectURL so now asset previews use [createImageBitmap](https://caniuse.com/createimagebitmap) to asynchronously create a 70x70 pixels BitMap from each File that is then fed to a Canvas so we can extract a dataUrl. This is considerably more performant and can be seen as the browser doesn't stutter anymore. 
While the preview is created we display a loading spinner and we fall back to the Immich logo in case of error as we previously did.

Considering that Safari doesn't support options.resizeHeight nor options.resizeWidth (according to [Mozilla](https://developer.mozilla.org/en-US/docs/Web/API/createImageBitmap#browser_compatibility)) the following can be used instead, less clean tho.
```js
  async function createPreview() {
    const bmp = await createImageBitmap(uploadAsset.file);
    const canvas = document.createElement('canvas');
    const context = canvas.getContext('2d');
    canvas.height = 70;
    canvas.width = 70;
    context?.drawImage(bmp, 0, 0, canvas.width, canvas.height);
    return canvas.toDataURL();
}
```
But I don't really know how you guys handle these things (I've seen 1 pollyfill).
Let me know what you think.

### DIFF
--- a/web/src/lib/components/shared-components/upload-asset-preview.svelte
+++ b/web/src/lib/components/shared-components/upload-asset-preview.svelte
@@ -5,11 +5,16 @@
   import { fade } from 'svelte/transition';
   import ImmichLogo from './immich-logo.svelte';
   import { getFilenameExtension } from '../../utils/asset-utils';
+  import LoadingSpinner from '$lib/components/shared-components/loading-spinner.svelte';
 
   export let uploadAsset: UploadAsset;
-
-  let showFallbackImage = false;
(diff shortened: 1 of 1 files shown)


jrasm91:
Closing, as we've not removed image previews from the upload component.

PR #3939, opened 2023-09-01
fix(server): non-nullable `IsOptional`

mertalev:
## Description

`class-validator`'s `IsOptional` does not validate null inputs, which is problematic as `null` has different semantics and behavior in JavaScript compared to `undefined`. This PR adds a custom `IsOptional` decorator with a `nullable` option that defaults to `false`, and replaces existing usages with this.

## How Has This Been Tested?

I added `modelType: null` to the `classification` section of `immich-config.json` (aside: the value of this is not actually used, but is assignable in the config). This did not cause any errors on main, but was rejected in this PR.


server/src/domain/domain.util.ts
```diff
@@ -92,3 +100,22 @@ export async function* usePagination<T>(
...
+ */
+// https://stackoverflow.com/a/71353929
+export function IsOptional(nullable = false, validationOptions?: ValidationOptions) {
+  if (nullable) {
+    return IsOptionalValidator(validationOptions);
+  }
+
+  return ValidateIf((obj: any, v: any) => {
+    return v !== undefined;
+  }, validationOptions);
```
jrasm91, line 120:
```suggestion
  return ValidateIf((obj: any, v: any) => v !== undefined, validationOptions);
```

mertalev:
I can't tell if the e2e tests failing is good here. Do we want to allow `null` for birth dates?

jrasm91 (commented):
Thanks for looking into this. I'd love to add a few e2e tests that validate nulls are rejected now. We specifically have some e2e test for albums already so they should be easy to add. What do you think?

Also, I don't think we ever want is optional to allow null values. I'd also be fine with just this:

```typescript
export const IsOptional = ValidateIf((obj: any, value: any) => value !== undefined)
```

We may accidentally import `IsOptional` from the wrong package too, it is not easy to differentiate, so it also could be worth re-naming it so something else. Perhaps `Optional()`?

jrasm91:
I think we have:
- The property is optional and nulls are valid - `@ValidateIf(value => value !== null)`
- The property is optional and nulls are invalid `@ValidateIf(value => value !== undefined)`
I don't think we ever want want

> I can't tell if the e2e tests failing is good here. Do we want to allow `null` for birth dates?

This is a tricky one because it is impossible for the mobile app to send null values AFAIK. In order to get it to do partial updates (without sending `null` values for optional properties), we configured it to never send null values.

server/src/domain/person/person.dto.ts
```diff
@@ -1,47 +1,37 @@
...
-  @IsOptional()
+  @Optional()
   @IsString()
   name?: string;
 
   /**
    * Person date of birth.
    */
-  @IsOptional()
+  @Optional(true)
```
jrasm91: I'd love the interface to be `({ nullable: true })` as it is easier to know what the option is doing.

server/src/immich/api-v1/asset/dto/create-exif.dto.ts
```diff
@@ -1,45 +1,46 @@
-import { IsNotEmpty, IsOptional } from 'class-validator';
+import { Optional } from '@app/domain';
```
jrasm91, line 1: This file can be deleted.

jrasm91 (approved):

server/src/domain/person/person.dto.ts
mertalev: Yeah, it's unclear what `true` even means here

server/test/e2e/auth.e2e-spec.ts
```diff
@@ -64,27 +73,37 @@ describe(`${AuthController.name} (e2e)`, () => {
...
 
-      const { status, body } = await request(server).post('/auth/admin-sign-up').send(signupStub);
+      const { status, body } = await request(server).post('/auth/admin-sign-up').send(adminSignupStub);
 
       expect(status).toBe(400);
       expect(body).toEqual(errorStub.alreadyHasAdmin);
     });
+
+    it('should not allow null values', async () => {
+      for (const key of Object.keys(adminSignupStub)) {
```
jrasm91, line 99: If you swap these two lines you can have it log the key name in the test description.

jrasm91 (approved):
Thanks for adding tests

etnoy (approved):
Thanks for adding this. We're adding a bunch of e2e tests all over right now, and for libraries I ran into the exact same issue.

server/test/e2e/auth.e2e-spec.ts
mertalev: Oh, good point

PR #3967, opened 2023-09-04
fix(mobile): change password page not navigation properly after password change

ddshd:
- Fixes bug where the change password page would not navigate the user back after the password was changed
- Added translation supported notifications when the password was/wasn't successfully reset
- Added back button for user to be able to go back to the login page without changing password

How this was tested:
- Login with user with required password reset, update password, observe app properly sending user back to login page and showing success message
- Login with user with required password reset, disable internet, update password, observe app showing error message
- Login with user with required password reset, observe app showing back button and allowing user to go back

image


alextran1502 (approved):
Thank you for working on this.

PR #3969, opened 2023-09-04
[Localizely] Polish translations update

Marcinbar:


jrasm91:
Hello, we actually use [localizely](https://localizely.com/) to manage our translations and we export them to json from there. If you send me or alex your email we can add you as a translator. Our emails are in our github profiles or you can DM us on discord.

PR #3972, opened 2023-09-04
fix(web): skeleton loading

jrasm91:
Follow up to #3867.

Render the skeleton before the asset grid "element" is present. This will render it immediately, which has a better experience and avoids the sudden re-rendering that happened previously.


JasBogans:
> 

I tested it, now it work as expected :+1:

PR #3976, opened 2023-09-04
feat(web): Load full size images on the web viewer for web supported images

ddshd:
- A little workaround for the web viewer to load full-size images for the web viewer if the original image is one of the commonly supported web format
- Contains fallback to the web optimized format if the original files to load

**This may not fit the development vision of the web viewer so feel free to decline this PR.** Opened so if anyone is savvy and wants to add this work around to their local version they can find it easily.

How this was tested:
- Uploaded and opened a JPG and watched the network a\tab show difference in file size between the current implementation and this work around.
image
image

- Uploaded and opened a HEIC and watched the browser load the JPG version instead
image

- Tested PNG, GIF, WebP samples and they load properly still with this change

- Tested toggle button to switch between loading full size image and not

image
image

- Tested toggle does not show up on videos and file types unsupported on the web (will always load web optimized version)


mertalev:
Since images now have resolution, quality and wide gamut settings, I feel like this is a fairly niche feature. But I don't see a problem with it as long as there's a toggle (and the default maintains current behavior).

ddshd:
> Since images now have resolution, quality and wide gamut settings, I feel like this is a fairly niche feature. But I don't see a problem with it as long as there's a toggle (and the default maintains current behavior).

Where do you think this setting should reside? A button on the image to load full size image, or something within the settings?

mertalev:
Good question. I think this would be best kept in a local store, so somewhere in the image viewer UI? It should only show for images though, since this PR doesn't apply to videos.

ddshd:
> Good question. I think this would be best kept in a local store, so somewhere in the image viewer UI? It should only show for images though, since this PR doesn't apply to videos (or the name should make it clear it's for images).

Added a toggle in the menu

mertalev (commented):
We might need to think a little more on the security aspect of this since serving un-sanitized images has some inherent risks.. Some formats like SVG make it easy to embed code, PNGs can have decompression bombs, etc. We could limit it to JPEG and WebP, but that would still require some level of validation to know that we're actually serving these files and not a different file type with that extension.

ddshd:
> We might need to think a little more on the security aspect of this since serving un-sanitized images has some inherent risks.. Some formats like SVG make it easy to embed code, PNGs can have decompression bombs, etc. We could limit it to JPEG and WebP, but that would still require some level of validation to know that we're actually serving these files and not a different file type with that extension.

What about `fileTypeFromBlob` from https://www.npmjs.com/package/file-type?

web/src/lib/utils/asset-utils.ts
```diff
@@ -173,3 +173,15 @@ export function getAssetRatio(asset: AssetResponseDto) {
   }
   return { width, height };
 }
+
+/**
+ * Return if an asset is an image commonly supported by web browsers
+ * @param asset Asset to check
+ * @returns true if the asset is an image commonly supported by web browsers
+ */
+export function isImageCommonlySupportedByWeb(asset: AssetResponseDto): boolean {
```
mertalev, line 182: Nit: This is a bit wordy. Could you change it to `isWebCompatible`?

web/src/lib/components/asset-viewer/photo-viewer.svelte
```diff
@@ -23,10 +26,20 @@
...
+   * it's already in a format commonly supported by web browsers.
+   */
+  const loadAssetData = async (forceWebImage?: boolean) => {
     try {
       const { data } = await api.assetApi.serveFile(
-        { id: asset.id, isThumb: false, isWeb: true, key: api.getKey() },
+        {
+          id: asset.id,
+          isThumb: false,
+          isWeb: forceWebImage || !isImageCommonlySupportedByWeb(asset),
```
mertalev, line 40: `isWeb` is apparently just supposed to mean `Is request made from web` (rather than mobile), but both this and `isThumb` are legacy flags that don't seem all that necessary. Do we want to repurpose this flag to mean "serve web-compatible asset" like this? @jrasm91

web/src/lib/utils/asset-utils.ts
```diff
@@ -173,3 +173,15 @@ export function getAssetRatio(asset: AssetResponseDto) {
...
+  const commonWebSupportedExtensions = ['jpg', 'jpeg', 'png', 'gif', 'webp'];
+  const imgExtention = getFilenameExtension(asset.originalPath);
+
+  return commonWebSupportedExtensions.includes(imgExtention.toLowerCase());
```
mertalev, line 186: `getFilenameExtension` always returns extensions in lowercase, so no need to do `toLowerCase` here.

mertalev (commented):
Looking good! I tested it and it works. `isWeb` is in a weird place with this, though. I wonder if the file type check here should be done server-side so we don't have separate implementations for mobile and web. `isWeb` (or another flag like `preferOriginal`) could be used by both and the server would decide what to send.

web/src/lib/components/asset-viewer/photo-viewer.svelte
jrasm91: Idk. The whole file serving thing needs to be cleaned up a bit. I'm not even 100% sure what these flags even do tbh.

jrasm91:
> Looking good! I tested it and it works. `isWeb` is in a weird place with this, though. I wonder if the file type check here should be done server-side so we don't have separate implementations for mobile and web. `isWeb` (or another flag like `preferOriginal`) could be used by both and the server would decide what to send.

That is a good idea. This might be ok in the short term though.

mertalev:
Yeah, I think this can be followed up with another PR to refactor the server code a bit. This is fine for now.

mertalev:
Oops, didn't mean to push that.

jrasm91:
Closing in favor of #4040

PR #3977, opened 2023-09-04
feat(web): show original uploader in shared album photo details

mrijke:
Hey! Been using Immich for quite a bit, sharing photos within my family. While doing this we noticed that it might be nice to see who originally uploaded a photo in a shared album.
This commit adds a new section "Uploaded by" to the asset detail panel if the asset owner is not the current user.
It also refactors the "Appears in" section to not render anything if there are no albums this asset appears in.

It's my first time using Svelte for anything serious, open for feedback :)


jrasm91:
Thanks for the contribution, I agree seeing the owner would be useful, especially in a shared album with multiple participants. I changed it to send the owner details in the asset response and removed the client-side user lookup. I'm not 100% sold on the UI implementation though. Google photos has this same feature, but they add a "Shared by <name>" instead. I think I like it better. What do you think about changing it to match?

image

mrijke:
Agree, "Shared by" makes more sense. I'll change that now.

alextran1502 (approved):

PR #3978, opened 2023-09-04
feat(server): consider shared albums in getByAssetId

mrijke:
This commit changes the `album.getByAssetId` API to also consider albums that have been shared with the current user. This way when the user is browing their timeline and clicks to show the asset details they will see if the asset appears in not only their own albums but also albums shared with them.


jrasm91 (approved):

jrasm91:
Thanks for the PR, this is a nice improvement :clap:

PR #3979, opened 2023-09-04
feat(mobile): Admin Controls Web View

ddshd:
- Adds WebView to manage admin functions for admin uses

TODO:
- Android Testing
- More iOS Testing
- UI Cleanup
- oAuth Admin Account Testing

Current UI:

image
image

### DIFF
--- a/mobile/assets/i18n/en-US.json
+++ b/mobile/assets/i18n/en-US.json
@@ -222,6 +222,7 @@
   "profile_drawer_app_logs": "Logs",
   "profile_drawer_client_server_up_to_date": "Client and Server are up-to-date",
   "profile_drawer_settings": "Settings",
+  "profile_drawer_admin_page": "Admin Web View",
   "profile_drawer_sign_out": "Sign Out",
   "recently_added_page_title": "Recently Added",
   "search_bar_hint": "Search your photos",
--- a/mobile/lib/modules/admin/views/admin_web_view.dart
+++ b/mobile/lib/modules/admin/views/admin_web_view.dart
@@ -1,0 +1,141 @@
+import 'package:easy_localization/easy_localization.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_inappwebview/flutter_inappwebview.dart';
+import 'package:immich_mobile/shared/models/store.dart';
+
+class AdminWebView extends StatefulWidget {
+  const AdminWebView({super.key});
+
+  @override
+  _AdminWebViewState createState() => _AdminWebViewState();
+}
+
+class _AdminWebViewState extends State<AdminWebView> {
+  final GlobalKey webViewKey = GlobalKey();
+
+  InAppWebViewController? webViewController;
+  InAppWebViewGroupOptions options = InAppWebViewGroupOptions(
+    crossPlatform: InAppWebViewOptions(
+      useShouldOverrideUrlLoading: true,
+      mediaPlaybackRequiresUserGesture: false,
+    ),
+    android: AndroidInAppWebViewOptions(
+      useHybridComposition: true,
+    ),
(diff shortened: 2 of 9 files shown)


jrasm91:
Would it be better to just add a dedicated set of views instead of a web view for this functionality?

ddshd:
> Would it be better to just add a dedicated set of views instead of a web view for this functionality?

Dedicated views would be better for UX though because of the very small number of users who would be able to access these views (likely single digit admin users for each install) it may not be worth the effort. I also foresee the the mobile views diverging from the web views as more features/options are added to the web without the mobile views being updated.

I think a webview is a good stop gap solution to give easy access to the admin controls without too much effort to develop or more importantly maintain the separate views.

If that's not the path we want to take, I'm fully open to admin controls being revisited at a later time when there are resources to keep the dedicated views maintained.

daniele-athome:
Personally I'd advise against using a webview and instead redirect the user to a browser directly. Webviews are more of a band-aid and don't integrate well with apps IMHO.

ddshd:
> Personally I'd advise against using a webview and instead redirect the user to a browser directly. Webviews are more of a band-aid and don't integrate well with apps IMHO.

Agreed - I'm looking into if it's possible to call the native SafariViewController or Chrome Tabs with cookies pre-set. So far it doesn't look like it's possible (at least on iOS) which is why I decided to use WebView instead.

It's still worth considering ditching this all together for dedicated views at a later time

daniele-athome:
Sure. This is just my personal view anyway. You should probably wait for a comment by the dev team :-)

Using a web view has its use cases of course, but in this case it doesn't give a feeling of integration, but more of a container opening a webapp (so, equivalent to opening a full browser but without the features of a browser).
If the admin area would be deployed as a separate webapp (or, better the same webapp but without the navigation elements) specialized to serve only admin routes, that would probably be a candidate for a webview (waiting for the native implementation of course).

ddshd:
> Sure. This is just my personal view anyway. You should probably wait for a comment by the dev team :-)
> 
> Using a web view has its use cases of course, but in this case it doesn't give a feeling of integration, but more of a container opening a webapp (so, equivalent to opening a full browser but without the features of a browser). If the admin area would be deployed as a separate webapp (or, better the same webapp but without the navigation elements) specialized to serve only admin routes, that would probably be a candidate for a webview (waiting for the native implementation of course).

Thank you for the input though, it's very valid.. I was able to get it to work with mobile browser with a small change on the server side.. Lmk if you think this would look better or the same..

image

ddshd:
> Sure. This is just my personal view anyway. You should probably wait for a comment by the dev team :-)
> 
> Using a web view has its use cases of course, but in this case it doesn't give a feeling of integration, but more of a container opening a webapp (so, equivalent to opening a full browser but without the features of a browser). If the admin area would be deployed as a separate webapp (or, better the same webapp but without the navigation elements) specialized to serve only admin routes, that would probably be a candidate for a webview (waiting for the native implementation of course).

Thank you for the input though, it's very valid.. I was able to get it to work with mobile browser with a small change on the server side.. Lmk if you think this would look better or the same..

image

daniele-athome:
I would have just opened the browser, but as I said I'm just a contributor :-) you should ping the devs on Discord on #contributing channel.

P.S. out of curiosity, what did you modify on the server for this?

jrasm91:
I'd be fine with adding a link to the admin page in the sidebar (only visible for the admin of course).

ddshd:
> I'd be fine with adding a link to the admin page in the sidebar (only visible for the admin of course).

Should the link open in app (Screenshot above) or take the user out of the app?

jrasm91:
I think it can just take the user out of the app.

PR #4023, opened 2023-09-08
feat(server): don't clear console on nest startup

etnoy:
Currently, when nest starts it will clear the console. This is quite annoying since docker-compose prints the logs of all containers at once, and when microservices or server clears the console, everything is deleted. Also, since it happens for both micro and server the clear is executed twice.

This minor PR passes --preserveWatchOutput to nest so it does not clear the console, making it easier to debug and troubleshoot


alextran1502 (approved):

jrasm91 (commented):
Do you mind moving this to the tsconfig file instead?

etnoy:
> Do you mind moving this to the tsconfig file instead?

I had no idea this could be done in tsconfig. But I'm not sure how that is done

PR #4040, opened 2023-09-09
feat(web): load original image when zooming

lmichelin:
## Description

This PR allows to load the full image when zooming using the mouse wheel or the zoom button.

## How Has This Been Tested?

- upload a big image
- click on the image thumbnail to view it
- zoom using the mouse wheel
- the original image is being downloaded
- when download is complete, the original image is shown instead of the preview

Animation

## Checklist:

### DIFF
--- a/web/src/lib/components/asset-viewer/photo-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/photo-viewer.svelte
@@ -91,6 +91,7 @@
 
   $: if (imgElement) {
     createZoomImageWheel(imgElement, {
+      maxZoom: 10,
       wheelZoomRatio: 0.2,
     });
   }
--- a/web/src/lib/components/asset-viewer/photo-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/photo-viewer.svelte
@@ -6,12 +6,14 @@
   import { notificationController, NotificationType } from '../shared-components/notification/notification';
   import { useZoomImageWheel } from '@zoom-image/svelte';
   import { photoZoomState } from '$lib/stores/zoom-image.store';
+  import { isWebCompatibleImage } from '$lib/utils/asset-utils';
 
   export let asset: AssetResponseDto;
   export let element: HTMLDivElement | undefined = undefined;
 
   let imgElement: HTMLDivElement;
   let assetData: string;
+  let hasZoomed = false;
   let copyImageToClipboard: (src: string) => Promise<Blob>;
   let canCopyImagesToClipboard: () => boolean;
 
@@ -23,10 +25,10 @@
     canCopyImagesToClipboard = module.canCopyImagesToClipboard;
   });
 
-  const loadAssetData = async () => {
+  const loadAssetData = async ({ loadOriginal }: { loadOriginal: boolean }) => {
     try {
       const { data } = await api.assetApi.serveFile(
-        { id: asset.id, isThumb: false, isWeb: true, key: api.getKey() },
+        { id: asset.id, isThumb: false, isWeb: !loadOriginal, key: api.getKey() },
         {
           responseType: 'blob',
         },
@@ -37,7 +39,6 @@
       }
 
       assetData = URL.createObjectURL(data);
-      return assetData;
     } catch {
       // Do nothing
     }
@@ -87,6 +88,11 @@
 
   zoomImageWheelState.subscribe((state) => {
     photoZoomState.set(state);
+
+    if (state.currentZoom > 1 && isWebCompatibleImage(asset) && !hasZoomed) {
+      hasZoomed = true;
+      loadAssetData({ loadOriginal: true });
+    }
   });
 
   $: if (imgElement) {
@@ -104,9 +110,9 @@
   transition:fade={{ duration: 150 }}
   class="flex h-full select-none place-content-center place-items-center"
 >
-  {#await loadAssetData()}
+  {#await loadAssetData({ loadOriginal: false })}
     <LoadingSpinner />
-  {:then assetData}
+  {:then}
     <div bind:this={imgElement} class="h-full w-full">
       <img
         transition:fade={{ duration: 150 }}
--- a/web/src/lib/utils/asset-utils.ts
+++ b/web/src/lib/utils/asset-utils.ts
@@ -173,3 +173,15 @@ export function getAssetRatio(asset: AssetResponseDto) {
   }
   return { width, height };
 }
+
+// list of supported image extensions from https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types excluding svg
+const supportedImageExtensions = new Set(['apng', 'avif', 'gif', 'jpg', 'jpeg', 'jfif', 'pjpeg', 'pjp', 'png', 'webp']);
+
+/**
+ * Returns true if the asset is an image supported by web browsers, false otherwise
+ */
+export function isWebCompatibleImage(asset: AssetResponseDto): boolean {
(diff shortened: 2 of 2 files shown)


web/src/lib/utils/asset-utils.ts
```diff
@@ -173,3 +173,15 @@ export function getAssetRatio(asset: AssetResponseDto) {
   }
   return { width, height };
 }
+
+// list of supported image extensions from https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types
+const supportedImageExtensions = ['apng', 'avif', 'gif', 'jpg', 'jpeg', 'jfif', 'pjpeg', 'pjp', 'png', 'svg', 'webp'];
```
mertalev, line 178: Remove SVG here.

web/src/lib/components/asset-viewer/photo-viewer.svelte
```diff
@@ -87,10 +88,16 @@
 
   zoomImageWheelState.subscribe((state) => {
     photoZoomState.set(state);
+
+    if (state.currentZoom > 1 && isImageWebCompatible(asset) && !hasOriginalDataFetchingBeenTriggered) {
+      hasOriginalDataFetchingBeenTriggered = true;
+      loadAssetData({ loadOriginal: true });
```
mertalev, line 94: What happens if the original image fails to load? Ideally, it should continue to show the preview image.

web/src/lib/components/asset-viewer/photo-viewer.svelte
```diff
@@ -6,12 +6,14 @@
...
   import { useZoomImageWheel } from '@zoom-image/svelte';
   import { photoZoomState } from '$lib/stores/zoom-image.store';
+  import { isImageWebCompatible } from '$lib/utils/asset-utils';
 
   export let asset: AssetResponseDto;
   export let element: HTMLDivElement | undefined = undefined;
 
   let imgElement: HTMLDivElement;
   let assetData: string;
+  let hasOriginalDataFetchingBeenTriggered = false;
```
mertalev, line 16: This is a bit verbose. Could you replace it with `fetchedOriginal`?

mertalev (commented):
This is really clean. It has a lot of overlap with #3976, so how do you feel about which to merge first or if they should be combined?

web/src/lib/components/asset-viewer/photo-viewer.svelte
lmichelin: Yes, that's exactly what happens. In case of failed request, the `assetData` variable is not overriden and it continue to show the preview image.

web/src/lib/components/asset-viewer/photo-viewer.svelte
lmichelin: I think `fetchedOriginal` is a bit confusing as this variable becomes true just before starting to fetch the data. So it can be true even if the data has not been fetched, when fetching is in progress. The idea is to prevent the download from starting more than once. What do you think about `originalDataFetchingTriggered`?

web/src/lib/utils/asset-utils.ts
lmichelin: I also turned the array into a Set to improve performance while checking if an extension is supported

lmichelin:
@mertalev regarding the overlap with #3976, I think it could be combined, maybe we can imagine a setting with 3 options for the user :
- never load original image
- load original on zoom
- always load original image

The 3rd option could even be chosen automatically on desktop if a good bandwidth is detected?

Regarding the feature introduced with these 2 PRs, loading the original image on zoom is closer to Google Photo's behaviour, so I feel ok to merge it first.

web/src/lib/components/asset-viewer/photo-viewer.svelte
mertalev: Yeah, I can see how that could be confusing when debugging. 

I like `originalDataFetchingTriggered`: it's less wordy, but still conveys what it actually means.

web/src/lib/components/asset-viewer/photo-viewer.svelte
mertalev: Perfect!

mertalev:
Yeah, I think that's a great idea.  Asset serving is due for a refactor, so it's a good time to add this setting.

I'm also comfortable merging this first since it's more self-contained.

mertalev (approved):

web/src/lib/components/asset-viewer/photo-viewer.svelte
```diff
@@ -6,12 +6,14 @@
...
   import { useZoomImageWheel } from '@zoom-image/svelte';
   import { photoZoomState } from '$lib/stores/zoom-image.store';
+  import { isWebCompatibleImage } from '$lib/utils/asset-utils';
 
   export let asset: AssetResponseDto;
   export let element: HTMLDivElement | undefined = undefined;
 
   let imgElement: HTMLDivElement;
   let assetData: string;
+  let originalDataFetchingTriggered = false;
```
jrasm91, line 16: I also am not a fan of this name. What about `loadingOriginal`?

jrasm91 (approved):

web/src/lib/components/asset-viewer/photo-viewer.svelte
lmichelin: When this boolean is true, it means that the request is in progress or finished, that's why I tried to find a naming covering these 2 states. What about `isOriginalLoadingOrLoaded`?

jrasm91: What about `hasZoomed`? It starts as false and is set to true the first time you zoom in.

AngelaDMerkel:
> The 3rd option could even be chosen automatically on desktop if a good bandwidth is detected?

If this hasn't been merged, loading on zoom is better behaviour than bandwidth detection in scenarios where the server hosts multiple users. I have gigabit and loading several >50MB images in a row is noticeable for other services on the network or when multiple users access the same newly added photos simultaneously.

web/src/lib/components/asset-viewer/photo-viewer.svelte
lmichelin: Good idea!

waclaw66:
I've noticed that some preview images (not zoomed) have different color temperature (more redish) than the original. What could be the problem?

preview:
e3913a70-3724-42da-8179-d874524acc8d

original:

mertalev:
I noticed that too. I looked into it and it seems [this](https://github.com/immich-app/immich/blob/86a8a687151b799d60bee6ef9243b52698df46fd/server/src/infra/repositories/media.repository.ts#L43) line is the culprit. This was added for correct handling of 8-bit non-sRGB images, but it seems to have a side effect for 8-bit sRGB images. If the image is 8-bit sRGB, then the image processor seems to handle it differently when it's processed as 16-bit. 

Setting the pipeline colorspace based on the input image should fix it.

MaxLyt:
Was a setting for loading originals by default implemented? I cannot find it anywhere. If not, it would be great to have this setting because jpeg thumbnails are downgraded to 8-bit colour without gainmaps, which distorts images with higher colour depth (e.g., 10-bit) and/or HDR colourspace or gainmaps. So loading originals is the only way to see the correct colours. Zooming in every time to accomplish it is rather inconvenient.

GiorgosPap777:
Does anyone know if this exists?
>Zooming in every time to accomplish it is rather inconvenient.

i would be happy if at least some thing like this existed

mertalev:
There's an option in the user settings to prefer the original.

PR #4045, opened 2023-09-10
feat(mobile): use map settings from server-config

shenlong-tanwen:
### Changes made in this PR

- Read and store server features and config data in ServerInfoState
- Enable / Disable map view based on the value from server features
- Use the map tile url from server config

### How was this tested
- Disable map from web console and ensured that "Your places" thumbnail is removed from curated places
- Enabled map from web console and ensured that "Your places" gets added back
- Change map tile url in web and ensured that the new tile URL is used by the mobile app as well

##### Note:
- Different map tile URLs have different max zoom level and we might have to probably handle that as a map setting as well.
- Map page attribution is static for now. It has to be updated based on the tile url or not show it at all if the default tile is not used
- Server features and Server config data are only fetched on app opening and are reloaded only when the app is restarted. Once we start using websockets for pushing server config, we can update the store in mobile devices using that as well

### DIFF
--- a/mobile/lib/shared/providers/server_info.provider.dart
+++ b/mobile/lib/shared/providers/server_info.provider.dart
@@ -15,12 +15,35 @@ class ServerInfoNotifier extends StateNotifier<ServerInfoState> {
               patch_: 0,
               minor: 0,
             ),
+            serverFeatures: ServerFeaturesDto(
+              clipEncode: true,
+              configFile: false,
+              facialRecognition: true,
+              map: true,
+              oauth: false,
+              oauthAutoLaunch: false,
+              passwordLogin: true,
+              search: true,
+              sidecar: true,
+              tagImage: true,
+            ),
+            serverConfig: ServerConfigDto(
+              loginPageMessage: "",
+              mapTileUrl: "https://tile.openstreetmap.org/{z}/{x}/{y}.png",
+              oauthButtonText: "",
+            ),
             isVersionMismatch: false,
             versionMismatchErrorMessage: "",
           ),
         );
 
   final ServerInfoService _serverInfoService;
+
+  getServerInfo() async {
+    await getServerVersion();
+    await getServerFeatures();
+    await getServerConfig();
+  }
 
   getServerVersion() async {
     ServerVersionResponseDto? serverVersion =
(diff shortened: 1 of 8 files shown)


magicedy:
I've tested it, and everything works as expected. 
I'm having trouble using tile.openstreetmap.org and I'm really looking forward to this feature. 
Can we merge this PR before the next release @alextran1502

mobile/lib/shared/providers/server_info.provider.dart
```diff
@@ -15,13 +15,36 @@ class ServerInfoNotifier extends StateNotifier<ServerInfoState> {
...
+            ),
             isVersionMismatch: false,
             versionMismatchErrorMessage: "",
           ),
         );
 
   final ServerInfoService _serverInfoService;
 
+  getServerInfo() async {
+    await getServerVersion();
```
fyfrey, line 43: no need for async or any await here - the calls can happen asynchronously in parallel.

mobile/lib/shared/providers/server_info.provider.dart
```diff
@@ -15,13 +15,36 @@ class ServerInfoNotifier extends StateNotifier<ServerInfoState> {
...
+              oauth: false,
+              oauthAutoLaunch: false,
+              passwordLogin: true,
+              search: true,
+              sidecar: true,
+              tagImage: true,
+            ),
+            serverConfig: ServerConfigDto(
+              loginPageMessage: "",
+              mapTileUrl: "https://tile.openstreetmap.org/{z}/{x}/{y}.png",
```
fyfrey, line 32: the hardcoded URL before was `"https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"`. The new one is missing `{s}` is this correct?

mobile/lib/shared/providers/server_info.provider.dart
shalong-tanwen: Updated!

mobile/lib/shared/providers/server_info.provider.dart
shalong-tanwen: Yes. The template `{s}` was previously used to handle subdomains for openstreetmap. But since we are accepting tile urls as user inputs, we can no longer determine the subdomains on which they are served. Users can now configure the url more directly as `https://a.tile.openstreemap.org/{z}/{x}/{y}.png` or other similar subdomains instead

fyfrey (approved):

alextran1502 (approved):

PR #4053, opened 2023-09-11
fix(web): disable shortcut when writting text

martabal:
## Changes made in this PR

This PR fixes #4049 and #3954

## Screenshots

| Before | After |
| --- | --- |
| <video src="https://github.com/immich-app/immich/assets/74269598/55e08172-d3f2-4788-a287-f094b8ae3ae9"> | <video src="https://github.com/immich-app/immich/assets/74269598/f76c3c97-29f7-4edd-8ed6-d87cc2410940"> |

PR #4057, opened 2023-09-11
refactor(web): disable shortcut when writting

martabal:
Following https://github.com/immich-app/immich/pull/4053#discussion_r1321408726

### DIFF
--- a/web/src/lib/components/album-page/album-viewer.svelte
+++ b/web/src/lib/components/album-page/album-viewer.svelte
@@ -20,6 +20,7 @@
   import ControlAppBar from '../shared-components/control-app-bar.svelte';
   import ImmichLogo from '../shared-components/immich-logo.svelte';
   import ThemeButton from '../shared-components/theme-button.svelte';
+  import { shouldIgnoreShortcut } from '$lib/utils/shortcut';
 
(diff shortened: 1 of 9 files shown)


jrasm91 (commented):
Looks a lot better. I think we have had variations of this issue on different pages, and some of them use different solutions as well. Can you find all places where we use `document.addEventListener` to handle keyboard shortcuts and make sure this new if(disableKeyboard(event)) { return }` is used on all of them?

To me, `shouldDisableShortcut` sounds like the keyboard shortcuts should become "disabled", which makes me think they need to be re-enabled in the future. In reality this is just checking the source of the event and determining if it applies to shortcuts or not (whether it should be handled or skipped). Something like `isKeyboardShortcut(event)` or `shouldIgnoreShortcut(event)` makes more sense to me.

martabal:
Yes `shouldIgnoreShortcut()` is a lot better !

jrasm91 (approved):
Nice! Looks good to me. Did you test all the shortcuts still work and all the text boxes input fields also still work?

martabal:
Yes, everything works as expected.

alextran1502 (approved):

PR #4086, opened 2023-09-13
fix(server): Error when loading album with deleted owner

### LINKED ISSUE #1988: [BUG] Cannot read properties of null (reading 'id')
### The bug

I notice this problem when an user that is not the administrator tries to access the "Sharing" tab in the main page (shared albums). From the browser the effect is that the the URL changes to `xxxxx.com/auth/login` (i.e. it logs out) and the page freezes until you refresh. Form the app the effect is that the newly shared albums don't show. (only old ones, this started to happen after 1.50 I think, but don't quote me on that). In the logs I see errors, I think the cause is due to the `Cannot read properties of null (reading 'id')` error.

Unraid (debian based)

### Version of Immich Server

v1.50.1

v1.50.0

### Platform with the issue

- [X] Server
- [X] Web
- [X] Mobile

```YAML
version: "3.8"

services:
  immich-server:
    volumes:
    env_file:
      - .env
    environment:
      - NODE_ENV=production
    depends_on:
      - redis
      - database
    restart: always

  immich-microservices:
    volumes:
    env_file:
      - .env
    environment:
      - NODE_ENV=production
    depends_on:
      - redis
      - database
    restart: always

  immich-machine-learning:
    image: altran1502/immich-machine-learning:release
    entrypoint: ["/bin/sh", "./entrypoint.sh"]
    volumes:
    env_file:
      - .env
    environment:
      - NODE_ENV=production
    depends_on:
      - database
    restart: always

  immich-web:
    image: altran1502/immich-web:release
    entrypoint: ["/bin/sh", "./entrypoint.sh"]
    env_file:
      - .env
    environment:
      # Rename these values for svelte public interface
      - PUBLIC_IMMICH_SERVER_URL=${IMMICH_SERVER_URL}
    restart: always

  redis:
    container_name: immich_redis
    image: redis:6.2
    restart: always

  database:
    image: postgres:14
    env_file:
      - .env
    environment:
      POSTGRES_USER: ${DB_USERNAME}
    volumes:
    restart: always

  immich-proxy:
    container_name: immich_proxy
    image: altran1502/immich-proxy:release
    environment:
      - IMMICH_SERVER_URL
      - IMMICH_WEB_URL
    ports:
      - 2283:8080
    logging:
      driver: none
    depends_on:
      - immich-server
    restart: always

volumes:
  pgdata:
```
```


### Your .env content

```Shell
# Database

DB_HOSTNAME=immich_postgres
DB_USERNAME=postgres
DB_PASSWORD=postgres
DB_DATABASE_NAME=immich

# Optional Database settings:
# DB_PORT=5432

# Redis

REDIS_HOSTNAME=immich_redis

# Optional Redis settings:
# REDIS_PORT=6379
# REDIS_DBINDEX=0
# REDIS_PASSWORD=
# REDIS_SOCKET=

# Upload File Location
#

UPLOAD_LOCATION=/mnt/user/Media/immich

# Log message level - [simple|verbose]

LOG_LEVEL=verbose

# JWT SECRET
#
# This JWT_SECRET is used to sign the authentication keys for user login
# You should set it to a long randomly generated value
# You can use this command to generate one: openssl rand -base64 128

JWT_SECRET=xxx

# Reverse Geocoding
#

# WEB - Optional
#
# For example:

PUBLIC_LOGIN_PAGE_MESSAGE=Helou

#

#

IMMICH_API_URL_EXTERNAL=https://xxx
```
```


### Reproduction steps

```bash
1. Create an shared album from administrator account
2. Share to someone else
3. The non-admin user tries to load the "Sharing" tab
4. Error in the logs occurs
```

### Additional information

The first error (ERR_STREAM_PREMATURE_CLOSE) occurs multiple times but doesn't seem to cause any issue, its the second one that occurs when you click the "Sharing" tab
```
[Nest] 1  - 03/13/2023, 12:19:02 PM   ERROR [ExpressAdapter] Premature close
Error [ERR_STREAM_PREMATURE_CLOSE]: Premature close
    at new NodeError (node:internal/errors:372:5)
    at ServerResponse.onclose (node:internal/streams/end-of-stream:142:30)
    at ServerResponse.emit (node:events:539:35)
    at Socket.onServerResponseClose (node:_http_server:236:23)
    at Socket.emit (node:events:539:35)
    at TCP.<anonymous> (node:net:709:12)
[Nest] 1  - 03/13/2023, 12:19:20 PM   ERROR [ExceptionsHandler] Cannot read properties of null (reading 'id')
TypeError: Cannot read properties of null (reading 'id')
    at mapUser (/usr/src/app/dist/apps/immich/libs/domain/src/user/response-dto/user-response.dto.js:13:20)
    at mapAlbumExcludeAssetInfo (/usr/src/app/dist/apps/immich/libs/domain/src/album/response-dto/album-response.dto.js:63:35)
    at /usr/src/app/dist/apps/immich/apps/immich/src/api-v1/album/album.service.js:69:76
    at Array.map (<anonymous>)
    at AlbumService.getAllAlbums (/usr/src/app/dist/apps/immich/apps/immich/src/api-v1/album/album.service.js:69:23)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async /usr/src/app/node_modules/@nestjs/core/router/router-execution-context.js:46:28
    at async /usr/src/app/node_modules/@nestjs/core/router/router-proxy.js:9:17
```

danieldietzler:
This PR soft deletes albums when their user gets soft deleted.
This ensures that they won't be loaded.

Fixes #1988

Using e2e testing it has been ensured that soft deleted albums won't get returned if shared via a person or using a shared link. Additionally, there is a test that verifies that soft deleted albums still get deleted when the user gets deleted.


server/src/domain/user/user.service.ts
```diff
@@ -91,6 +91,7 @@ export class UserService {
     if (!user) {
       throw new BadRequestException('User not found');
     }
+    this.albumRepository.softDeleteAll(userId);
```
jrasm91, line 94: This should be awaited

server/src/infra/entities/album.entity.ts
```diff
@@ -52,4 +53,7 @@ export class AlbumEntity {
 
   @OneToMany(() => SharedLinkEntity, (link) => link.album)
   sharedLinks!: SharedLinkEntity[];
+
+  @DeleteDateColumn()
+  deletedAt?: Date;
```
jrasm91, line 58: deletedAt!: Date | null

jrasm91 (commented):
What happens when the user is restored?

Can you put in the description how this has been tested?

Does deleting a user after 7 days still delete these albums? Do you need to have withDeleted in the delete all query?

server/src/infra/entities/album.entity.ts
danieldietzler: Shouldn't the | null part be implicit because of the '?:' ?

jrasm91: If you try to set this to null it will throw a type error, so no, it is not implied.

jrasm91 (approved):
Looks good! Thanks for working on this.

PR #4098, opened 2023-09-14
feat(web): hide face from detail page

### LINKED ISSUE #4099: [Feature] Ability to hide face from the details screen
### Discussed in https://github.com/immich-app/immich/discussions/4096

<sup>Originally posted by **jakob42** September 14, 2023</sup>
### The feature

Hi,

when viewing an image I sometimes see somebody is tagged in the bg I dont know. I would love to be able to easily hide that face since I now am sure I dont know them. I can get a detail view clicking on the face (where I can give it a name, merge it, set birthday), but I cannot hide it from there. But finding the face-thumb in the long list of persons is nearly impossible and not worth it.

Thanks for considering!

### Platform

- [ ] Server
- [X] Web
- [ ] Mobile</div>

martabal:
## Changes made in this PR

This PR adds the ability to hide a face from the detail page. Once the face is hidden, the user is redirected to the Explore page.

Closes #4099

## Screenshots

Screenshot from 2023-09-14 16-42-38

### DIFF
--- a/web/src/routes/(user)/people/[personId]/+page.svelte
+++ b/web/src/routes/(user)/people/[personId]/+page.svelte
@@ -76,6 +76,24 @@
     }
   });
 
+  const hideFace = async () => {
+    try {
+      await api.personApi.updatePerson({
+        id: data.person.id,
+        personUpdateDto: { isHidden: true },
+      });
+
+      notificationController.show({
+        message: 'Changed visibility succesfully',
+        type: NotificationType.Info,
+      });
+
+      goto(AppRoute.EXPLORE, { replaceState: true });
+    } catch (error) {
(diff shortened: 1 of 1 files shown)


jrasm91 (approved):
This looks fine. In general we have this pattern in the web app where we often have the same action available a few different places. For people, the exact same action seems to exist in the all people page > menu option and now person page > menu option. I don't think we need two different implementations of the same action.

For assets, we consolidated some of this code by creating the `photos-page/*-action.svelte` components. They encapsulate the action, call the api endpoint, and show the notification. They have callbacks afterwards so the parent component can do redirects or other state management if needed. It would be great to do the same for these people actions.

PR #4101, opened 2023-09-15
fix(server): use srgb pipeline for srgb images

mertalev:
## Description

There's a bug with sharp regarding sRGB images where the combination of `pipelineColorspace('rgb16')` and `withMetadata` causes the output image to become more saturated. There's a similar issue for this in sharp that will be fixed in the next release, but from local testing it doesn't fix this particular issue. I believe it stems from libvips treating all 8-bit RGB as sRGB and sharp treating all 16-bit RGB as P3. This PR handles this by looking at exif metadata to decide whether an image is sRGB. It mostly maintains current behavior until metadata extraction is run for images, at which point it processes them correctly.

Fixes the issue described [here](https://github.com/immich-app/immich/pull/4040#issuecomment-1716196839)

## How Has This Been Tested?

I ran thumbnail generation without any metadata extraction and observed that while non-sRGB images were processed correctly, sRGB thumbnails were more saturated. After running metadata extraction and re-running thumbnail generation, sRGB images have more accurate colors while non-sRGB images are the same.


jrasm91:
@mertalev Can you rebase this?

server/src/domain/media/media.service.ts
```diff
@@ -334,4 +335,13 @@ export class MediaService {
     this.storageRepository.mkdirSync(folderPath);
     return join(folderPath, `${asset.id}.${extension}`);
   }
+
+  isSRGB(asset: AssetEntity): boolean {
+    const { colorspace, profileDescription, bitsPerSample } = asset.exifInfo ?? {};
+    if (colorspace || profileDescription) {
+      return [colorspace, profileDescription].includes('sRGB');
```
jrasm91, line 342: Is this case sensitive and does it need to be?

jrasm91 (commented):
This is changing a lot of stuff.

If exif tool fails to parse a number it will return a string, which is why validate checked for it specifically.

Can we keep it that way? I think just adding your fields as numeric would be sufficient. Did you find some wrong with the previous implementation?

We also are getting file stats from exif now, which is probably fine, but I think I would prefer that as a separate PR that would be easy to test, verify, and merge independently. I see number parsing/validation code is shared, but that seems less important than the benefits of a focused PR

mertalev:
> This is changing a lot of stuff.
> 
> If exif tool fails to parse a number it will return a string, which is why validate checked for it specifically.
> 
> Can we keep it that way? I think just adding your fields as numeric would be sufficient. Did you find some wrong with the previous implementation?
> 
> We also are getting file stats from exif now, which is probably fine, but I think I would prefer that as a separate PR that would be easy to test, verify, and merge independently. I see number parsing/validation code is shared, but that seems less important than the benefits of a focused PR

TBH I found `validate` to be really confusing. I initially thought it was referring to a dependency, then spent a minute re-reading it to make sure I understood what the function was doing, then wondered why its idea of validation is returning null for strings, then noticed it was only being used for number-like fields, then saw the `numericTags` field and checked the documentation for exiftool to understand its behavior and defaults. The function only made sense after all of that while something more explicit would have been immediately clear.

server/src/domain/media/media.service.ts
mertalev: The value for `ColorSpace` is numeric and parsed into `sRGB` based on the below EXIF spec, so it should always have that casing.

```
0x1 = sRGB
0x2 = Adobe RGB
0xfffd = Wide Gamut RGB
0xfffe = ICC Profile
0xffff = Uncalibrated
```

But the profile description can be anything, so this should actually just search for `srgb` in the strings uncased.

mertalev:
I reverted the file size change. I think the stats call is redundant since exiftool does it too, but you're right that it's out of scope for this PR.

jrasm91:
Agreed, it is definitely not very intuitive.

I'm fine updating it to be more transparent and clear too. I just have a hard time figuring out what changes you made for colorspace on my phone with that change in there. Maybe we can update it in a second pull request and add a comment about the way exit parsing works.

jrasm91:
> I reverted the file size change. I think the stats call is redundant since exiftool does it too, but you're right that it's out of scope for this PR.

I think we should potentially implement the change though, in another PR. Just need to make sure it works correctly and it is always a number, etc.

jrasm91 (approved):

server/src/domain/media/media.service.ts
jrasm91: Great, looks good.

PR #4106, opened 2023-09-16
Libraries: move assets automatically to the proper library

daniele-athome:
Read-only assets will be moved to an external library, the others will be moved to an upload library.

The external library will be created only if a user has read-only assets.

### Tests

- [x] user without read-only assets has assets moved only to the default upload library
- [x] user with both read-only and uploaded assets has them moved to the right library (upload and external)
- [x] user with only read-only assets has all of them moved to the (newly created) external library, but a default library is created anyway

### DIFF
--- a/server/src/infra/migrations/1688392120838-AddLibraryTable.ts
+++ b/server/src/infra/migrations/1688392120838-AddLibraryTable.ts
@@ -23,17 +23,40 @@ export class AddLibraries1688392120838 implements MigrationInterface {
       `ALTER TABLE "assets" ADD CONSTRAINT "FK_9977c3c1de01c3d848039a6b90c" FOREIGN KEY ("libraryId") REFERENCES "libraries"("id") ON DELETE CASCADE ON UPDATE CASCADE`,
     );
 
-    // Create default library for each user and assign all assets to it
-    const userIds: string[] = (await queryRunner.query(`SELECT id FROM "users"`)).map((user: any) => user.id);
+    // Create default library for each user and assign all assets to it.
+    // Read-only assets will be moved to an external library instead
+    const users: [{id: string, externalPath: string | null, numReadonlyAssets: number}] =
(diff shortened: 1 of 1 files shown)


daniele-athome:
Converting to draft because I want to add a dialog notifying the user about what happened after the migration with some warnings (e.g. "you need to set ignored paths etc.").

daniele-athome:
I realized there is no infrastructure to show a one-time message to users - the fact that it's been showed would have to be stored somewhere, and maybe something more generic would be better (something that other new features can use in the future to notice users).

However, maybe warning the user about something that something so technical happened seems like a UX fail. I don't want to give up this feature either though. Any suggestions?

P.S. given that a notice dialog would need a bigger development (see above), I'm turning this PR back into ready state so it can be reviewed.

jrasm91:
Can we change the scan/rescan stuff to not do anything if the import paths are empty?

There is no guarantee the storage path for a user is a correct/valid/intended import path. It might be sometimes, but not always.

Otherwise I don't think it is a good idea to automatically migrate these assets.

Worst case the user can set up an external library from the start, which they might do anyway.

daniele-athome:
> Can we change the scan/rescan stuff to not do anything if the import paths are empty?

I though it already did nothing if they are empty. It doesn't? What paths the job is going to iterate over otherwise?

> There is no guarantee the storage path for a user is a correct/valid/intended import path. It might be sometimes, but not always.

The migration will skip users that did not set users.externalPath. It might not be what some users want, but generally if a user comes from an externalPath scenario, they will want to use that (probably exact?) same path for an external library.

Anyway the main issue here is losing asset information after migrating to the new library feature - because if I don't migrate my existing read-only assets, I have to start everything from scratch. If we can have a way to improve this transition I believe it will benefit many users.

daniele-athome:
Reporting the rationale for closure by @etnoy on Discord:

> I've been thinking about it quite a bit honestly. First, thanks to the team we were able to get the library pr merged just now. Secondly, I think we'll keep the imported read-only assets as-is for now and allow them to migrated later on as things stabilize a bit

PR #4112, opened 2023-09-16
chore(server): Store generated files (thumbnails, encoded video) in subdirectories

### LINKED ISSUE #3314: [BUG] Thumbs dir size becomes too big
### The bug

Since all thumbnails are stored in a single directory (`./uploads/thumbs/$user_uuid`), this directory accrues many files. I uploaded my whole collection of photo's and albums, and now this folder contains 185K files, and is about 10MB in size (I'm talking the _inode size of the directory_, not the total size of all files in the directory!)

Thumbnails should also be stored in a nested folder structure, either based on the timestamp, or the leading characters of the file name. Eg. `b561e421-23e1-42c2-a416-3aeeb0f9c19f.jpeg` could be `b5/61/b561e421-23e1-42c2-a416-3aeeb0f9c19f.jpeg`.

Docker on Ubuntu 22.04.2 LTS

### Version of Immich Server

v1.67.2

n/a

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
n/a
```

### Your .env content

```Shell
n/a
```

### Reproduction steps

1. Install Immich as usual
2. Add 100K photo's, video's and albums
3. Watch the size of the thumbs folder grow:

```bash
$ ls -lh ./uploads/thumbs/                         
totaal 13M                                                                   
drwxr-xr-x 2 root root 4,0K jul 17 12:56 6f718f4e-089b-4c3f-9b07-04ff9dcc1221
drwxr-xr-x 2 root root 8,0K jul 17 15:51 b0acc1ad-06a2-4daa-940c-26241ac88de1
drwxr-xr-x 2 root root  11M jul 17 16:07 c9f4b1b9-0cc9-483d-824f-3b07f8ea1839
drwxr-xr-x 2 root root    6 jul 16 02:06 f843c72a-0483-49e5-a481-1fdc44f3c62a
```

### Additional information

Filesystem is XFS

danieldietzler:
This PR fixes #3314.

It helps improving the performance by storing generated files in subdirectories.
This yields smaller directory sizes (in terms of files per directory), decreasing the size of the inode entry.

- Migration job to migrate all files to new folder structure (and possibly more migration stuff in the future)
- Clean up directories if assets/people get deleted
- New asset and face thumbnails will be stored at i.e. `89/ab/`, if their id starts with `89abxxxx-xxxx...`

There are already pretty good tests in place to check thumbnail creation (+ path), so I've not written further tests and updated those. Furthermore, I tried different edge cases like mix of old & new folder structure, does the migration sort them correctly, live photos (?), do empty folders get deleted if files get deleted and the (nested) folders would be empty afterwards?

### The new migration job
Screenshot 2023-09-17 at 21 38 26


server/src/domain/media/media.service.ts
```diff
@@ -330,7 +378,20 @@ export class MediaService {
...
+    return this.ensurePath(asset, StorageFolder.ENCODED_VIDEO, extension);
+  }
+
+  private ensurePath(
+    asset: AssetEntity,
+    folder: StorageFolder.ENCODED_VIDEO | StorageFolder.UPLOAD | StorageFolder.PROFILE | StorageFolder.THUMBNAILS,
+    extension: string,
+  ): string {
+    let folderPath = this.storageCore.getFolderLocation(folder, asset.ownerId);
+    folderPath = join(folderPath, asset.id.substring(0, 2), asset.id.substring(2, 4), asset.id.substring(4, 6));
```
bo0tzz, line 394: Do we need this much nesting? Most other times I've seen a system like this, it just nested one level with `substring(0, 2)`.

danieldietzler: I thought it wouldn't hurt and I don't know how many assets some people might have in their library ;)
The original idea in the issue was a depth of 2 and I don't see any disadvantage in a little deeper nesting?

bo0tzz: Since it's the thumbs folder this isn't particularly important, but if someone wanted to poke around these files deep nesting might make that a bit of a pain. I don't really have any strong points for/against though :)

danieldietzler: Tbh I can't really estimate how much a performance impact large directories actually make and how many assets would end up in one folder if nested 1, 2 or 3 deep. xD
Let's just wait for others and see what they think about :)

danieldietzler: Fixing all the tests with a nesting of 3 is a pain though haha
`upload/encoded-video/user-id/as/se/t-/asset-id.mp4` nice path lol

mertalev: I wrote a script to test with 150k random UUIDs.

depth=1 
585.9375 assets per leaf folder

depth=2
2.5508 assets per leaf folder

depth=3
1.0045 assets per leaf folder

So anything more than 1 layer deep puts almost every asset in its own folder.

mertalev: Which makes sense since the characters are hexadecimal so each two letter sequence has 16*16=256 combinations, so the number of assets in the leaf folders will be roughly `(num_assets) / 256^depth`.

danieldietzler: That makes sense. So even with ~1M assets there would still be like <15 (15 * 2^16 = something greater than 1M) assets per folder. Alright, so I guess it is perfectly fine if I just go with a depth of 2 instead.
Thanks @mertalev for testing!

server/src/domain/facial-recognition/facial-recognition.services.ts
```diff
@@ -175,4 +198,12 @@ export class FacialRecognitionService {
 
     return true;
   }
+
+  ensureFaceThumbnailPath(ownerId: string, personId: string, extension: string): string {
+    let folderPath = this.storageCore.getFolderLocation(StorageFolder.THUMBNAILS, ownerId);
+    folderPath = join(folderPath, personId.substring(0, 2), personId.substring(2, 4));
+    this.storageRepository.mkdirSync(folderPath);
```
jrasm91, line 205: This same logic exists in each service. Can we move it to the storage core instead? Also I think we can just pass `filename` in and substring that. In this case you can just pass in `${personId}.jpeg}`, since generating a filename is more dependent on the calling implementation than it is on the common code to generate directories.

server/src/domain/job/job.constants.ts
```diff
@@ -44,6 +45,11 @@ export enum JobName {
   STORAGE_TEMPLATE_MIGRATION_SINGLE = 'storage-template-migration-single',
   SYSTEM_CONFIG_CHANGE = 'system-config-change',
 
+  // migration
+  QUEUE_MIGRATION = 'queue-migration',
+  ASSET_MIGRATION = 'asset-migration',
```
jrasm91, line 50:
```suggestion
  MIGRATE_ASSET = 'migrate-asset',
```

server/src/domain/job/job.constants.ts
```diff
@@ -44,6 +45,11 @@ export enum JobName {
...
+  FACE_THUMBNAIL_MIGRATION = 'face-thumbnail-migration',
```
jrasm91, line 51:
```suggestion
  MIGRATE_PERSON = 'migrate-person',
```

We're passing the entity of a person. Who knows, maybe in the future they will have videos too. I think keep it general "migrate this person" seems better, and matches migrating an asset, which is made up of multiple files that are looked at.

server/src/domain/job/job.interface.ts
```diff
@@ -17,9 +17,15 @@ export interface IFaceThumbnailJob extends IAssetFaceJob {
   personId: string;
 }
 
+export interface IPersonJob extends IBaseJob {
+  personId: string;
+  ownerId: string;
```
jrasm91, line 22: I don't think either of these are needed. You can just use `IEntityJob` and use id to pass the id of the entity (a person here)

server/src/domain/storage/storage.service.ts
```diff
@@ -26,6 +26,10 @@ export class StorageService {
 
       try {
         await this.storageRepository.unlink(file);
+        const found = file.match(/(?<path>.*\/.*)\/.*\/.*\..*/);
+        if (found?.groups) {
+          await this.storageRepository.removeEmptyDirs(found.groups.path, true);
+        }
```
jrasm91, line 32: Does this remove the directory if it is the last file in the folder?

jrasm91 (commented):
Looks like a great start. Did you try updating the migration concurrency and verify it is persisted correctly? I think you need to edit system config entity, to add the key for the override.

server/src/domain/job/job.interface.ts
```diff
@@ -17,9 +17,15 @@ export interface IFaceThumbnailJob extends IAssetFaceJob {
...
+}
+
 export interface IEntityJob extends IBaseJob {
   id: string;
   source?: 'upload';
+  deleteOld?: boolean;
```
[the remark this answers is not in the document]
danieldietzler: Nope. I had it at some point and then forgot to delete it. Thanks!

web/src/lib/components/admin-page/jobs/jobs-panel.svelte
```diff
@@ -102,6 +102,12 @@
       allowForceCommand: false,
       component: StorageMigrationDescription,
     },
+    [JobName.Migration]: {
+      icon: FolderMove,
+      title: api.getJobName(JobName.Migration),
+      subtitle: 'Migrate assets and face thumbnails',
```
jrasm91, line 108: We should maybe make this more descriptive, like

> Migrate thumbnails for assets and faces to the latest folder structure.

alextran1502:
Thank you for working on this. I have a request regarding the file name that goes into the `upload` directory. Can we add the original filename after the random UUID just for some traceability? so `UUID_originalName.ext`

jrasm91:
I think there was just some confusion - this PR doesn't change how anything works in the `/usr/src/app/upload/upload/` folder. We can make a separate PR to append original name to the file when it is uploaded originally though.

server/src/domain/media/media.service.ts
```diff
@@ -81,6 +79,63 @@ export class MediaService {
...
+    const people = await this.personRepository.getAll();
+
+    for (const person of people) {
+      await this.jobRepository.queue({
+        name: JobName.MIGRATE_PERSON,
+        data: { id: person.id },
+      });
+    }
+
+    while ((await this.jobRepository.getJobCounts(QueueName.MIGRATION)).waiting > 1) {
```
alextran1502, line 102: Can you help explain this while block?

mertalev: I think the idea is to poll for when all the other jobs complete so the queue job is the only one left. When this is the case, it means migration is complete and it can remove the other directories. 

But I'm not sure if this check actually has this behavior. A job could be active, paused, etc. but not waiting, so it seems like it could break out of this while loop before the other jobs are finished. And this job would be active so if there's only one waiting job, it would be a job that should complete before this code moves on.

danieldietzler: @mertalev is right.
The idea was to avoid deleting directories that have just been created by another job and not yet filled with new files.
And I feel like the paused jobs actually don't make a difference, because if paused, it is also fine to delete all empty folders. In that case I won't interfere with a running case, deleting directories that have just been created.

About the active jobs however, I think you're right and I will need to wait until there are no active jobs as well.

Nevertheless, it might very well be that I'm a) completely overlooking something or that there is b), a much better way to do this. In both cases feedback would be much appreciated!

alextran1502: What do you think about having a cronjob to run at a particular time of the day to delete empty directories when there are no jobs active?

danieldietzler: Hm, I'm not sure... It feels like unexpected behavior. I believe we should keep the folder structure untouched as long as not explicitly told to migrate/cleanup something. Maybe others would like to comment on this?

jrasm91: Honestly probably not that important for thumbnail directories. I am fine running it at the beginning of the job though.

danieldietzler:
Why is the random asset e2e test failing although it worked perfectly in the original PR? lol

web/src/lib/components/admin-page/jobs/jobs-panel.svelte
```diff
@@ -110,6 +110,12 @@
       allowForceCommand: false,
       component: StorageMigrationDescription,
     },
+    [JobName.Migration]: {
```
alextran1502, line 113: I assume this job will be a temporary job, correct? We might be able to remove it after a few releases?

danieldietzler: The idea was that the Migration Job (and Queue respectively) will replace the storage template migration job at some point. So that Migration will be the more general form for everything file-related.

danieldietzler: Jason said that storage template migration could need a refactoring anyway, so they would be merged then

server/src/domain/storage/storage.service.ts
```diff
@@ -26,6 +26,10 @@ export class StorageService {
 
       try {
         await this.storageRepository.unlink(file);
+        const found = file.match(/(?<path>.*\/.*)\/.*\/.*\..*/);
```
alextran1502, line 29: Can you help write a few words on the purpose of this RegEx as a comment above this line?

danieldietzler: Yeah sure, that is a good idea!

danieldietzler:
``` 
// extracts the file's parent folder
// upload/thumbs/89/ab/89abcd.jpeg -> upload/thumbs/89
```
would you say that is understandable?

alextran1502: yes this looks good, thank you

jrasm91 (approved):
Looks good. I tested it with about 3k photos and it seems to work fine, especially the migration process.

waclaw66:
Some assets were not migrated. I've checked file permissions, all correct.

obrazek

jrasm91:
Have you verified that those files are currently referenced in the database? (ie not unused/orphaned)

waclaw66:
> Have you verified that those files are currently referenced in the database? (ie not unused/orphaned)

You're right, there are no refs in db, probably orphaned. Thanks for a hint.

jrasm91:
Thanks for testing. Other than that, does everything work as expected?

waclaw66:
> Thanks for testing. Other than that, does everything work as expected?

Yes, everything seems to be working fine. 👍

waclaw66:
I've noticed that there has been created empty folder structure within `encoded-video` folder, however I have no encoded videos at all. According the timestamp, it seems to have been created while migration.
obrazek

danieldietzler:
> I've noticed that there has been created empty folder structure within `encoded-video` folder, however I have no encoded videos at all. According the timestamp, it seems to have been created while migration.
> 
> obrazek
> 

So all those directories are empty? If so, could you try re-running the job? That should cause the deletion of empty directories

waclaw66:
Yes, they are empty. When I deleted content of `encoded-video` folder, the folder structure is recreated after migration job restart.

danieldietzler:
Ok, I think I fixed the bug in #4251 @waclaw66.
Thanks for pointing this out.
Apologies for the inconvenience!

waclaw66:
Works like a charm. Thanks!

eygraber:
Does the migration need to be run constantly, or once it's run the first time then it doesn't need to be run again? If the latter, could that be removed from the settings once it's been run?

alextran1502:
@eygraber Once it run the first time, it doesn't need to run again, new incoming files will follow this rule from now on

eygraber:
So is it possible to remove the migration from the UI once that's done? Should I file an issue for that?

briandipalma:
> Have you verified that those files are currently referenced in the database? (ie not unused/orphaned)

Is there anything that could be done to highlight these assets? Some sort of job that can be run occasionally to list discrepancies between the database and the library on the filesystem? Recover files that have been deleted in the filesystem or delete files from the database?

PR #4071, opened 2023-09-12
fix(web): Timeline scrollbar shown over menu

danieldietzler:
This PR fixes #3735.

### Before
Screenshot 2023-09-12 at 16 26 08

### After
Mouse on the button
Screenshot 2023-09-12 at 16 27 30

Mose below the button, timeline text not overflowing
Screenshot 2023-09-12 at 16 27 38

PR #4127, opened 2023-09-17
fix(web): Context menu underflowing

danieldietzler:
Fixes an issue that has been introduced in #4071 as described by https://github.com/immich-app/immich/pull/4071#issuecomment-1722430395.

I've verified that the fix in #4071 still applies.


jrasm91 (approved):

waclaw66:
That bug applies also for faces (described https://github.com/immich-app/immich/pull/4071#issuecomment-1722430395), that hasn't been fixed.

PR #4128, opened 2023-09-17
feat: add setting for minimum face count for face detection

GenericGuy:
Adds the minimum face count setting to the web interface to circumvent detection of strangers and random background people if desired by the user. Originally discussed in the PR for face recognition (https://github.com/immich-app/immich/pull/2180#discussion_r1196146598). The default behavior is unchanged (i.e. every match is shown).

I tested it manually by adding multiple pictures of a person and changing the setting around, which worked flawlessly.

image

### DIFF
--- a/server/src/domain/smart-info/dto/model-config.dto.ts
+++ b/server/src/domain/smart-info/dto/model-config.dto.ts
@@ -51,7 +51,6 @@ export class RecognitionConfig extends ModelConfig {
 
   @IsNumber()
   @Min(1)
-  @Max(5)
   @Type(() => Number)
   @ApiProperty({ type: 'integer' })
   minFaces!: number;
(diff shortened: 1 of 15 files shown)


server/src/domain/smart-info/dto/model-config.dto.ts
```diff
@@ -48,4 +48,11 @@ export class RecognitionConfig extends ModelConfig {
   @Type(() => Number)
   @ApiProperty({ type: 'integer' })
   maxDistance!: number;
+
+  @IsNumber()
+  @Min(1)
+  @Max(5)
```
jrasm91, line 54: I don't know if we need a max check here.

jrasm91 (approved):
Looks good!

alextran1502 (approved):

thariq-shanavas:
Hi @GenericGuy, thank you so much for this PR!

As of v1.81, even if there are fewer than the set number of pictures containing person A, I can still open a picture containing person A and name them from the sidebar. Would it be possible to make them appear in the People section if they have a name assigned, *even if there are fewer pictures of them than the set value*?

I can open a discussion if you so prefer. Thanks again!

image

PR #4119, opened 2023-09-17
feat: make settings accordion click area larger

Ploonet:
# Description

Changes the setting-accordion Svelte component to make the click area larger.

Currently the only clickable area to open the accordion is the arrow to the right side.

This PR extends that area to contain the title, description and the space between the text and the arrow.

## Why

I believe this makes clicking the items easier, which helps accessibility on large high res displays and on mobile in my opinion.

## Other info

- This is my first open source PR ever, i hope i'm doing things right
- I've had issues running the dev version locally (couldn't get past the admin registration), so i was not able to test directly on the Settings pages. I built this by creating a SettingAccordion in the registration page.

PR #4129, opened 2023-09-17
Fix/clickable accordion layout

Ploonet:
After adding a feature in this #4119 , i've made a few mistakes that caused a regression in the UI ( [see this comment](https://github.com/immich-app/immich/pull/4119#discussion_r1328108722) )

This PR aims to fix said issue and remove a useless pointer class.


jrasm91 (approved):
LGTM

PR #4138, opened 2023-09-18
chore(server): Use access core for person permissions

danieldietzler:
Migrate all person service methods to use access core instead of making their own queries on the person table

### DIFF
--- a/server/src/domain/person/person.service.ts
+++ b/server/src/domain/person/person.service.ts
@@ -149,13 +149,12 @@ export class PersonService {
 
   async mergePerson(authUser: AuthUserDto, id: string, dto: MergePersonDto): Promise<BulkIdResponseDto[]> {
     const mergeIds = dto.ids;
-    await this.access.requirePermission(authUser, Permission.PERSON_READ, id);
+    await this.access.requirePermission(authUser, Permission.PERSON_WRITE, id);
     const primaryPerson = await this.findOrFail(id);
     const primaryName = primaryPerson.name || primaryPerson.id;
 
     const results: BulkIdResponseDto[] = [];
 
-    await this.access.requirePermission(authUser, Permission.PERSON_WRITE, primaryPerson.id);
     for (const mergeId of mergeIds) {
       const hasPermission = await this.access.hasPermission(authUser, Permission.PERSON_MERGE, mergeId);
 
--- a/server/src/domain/person/person.service.ts
+++ b/server/src/domain/person/person.service.ts
@@ -1,4 +1,5 @@
 import { BadRequestException, Inject, Injectable, Logger, NotFoundException } from '@nestjs/common';
+import { AccessCore, IAccessRepository, Permission } from '..';
 import { AssetResponseDto, BulkIdErrorReason, BulkIdResponseDto, mapAsset } from '../asset';
 import { AuthUserDto } from '../auth';
 import { mimeTypes } from '../domain.constant';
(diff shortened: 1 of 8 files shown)


server/src/domain/person/person.service.ts
```diff
@@ -48,12 +52,13 @@ export class PersonService {
     };
   }
 
-  getById(authUser: AuthUserDto, id: string): Promise<PersonResponseDto> {
+  async getById(authUser: AuthUserDto, id: string): Promise<PersonResponseDto> {
```
jrasm91, line 55: async isn't needed since you are not awaiting anything

server/src/domain/person/person.service.ts
```diff
@@ -62,17 +67,19 @@ export class PersonService {
...
+    await this.access.requirePermission(authUser, Permission.PERSON_READ, id);
+    const assets = await this.repository.getAssets(id);
     return assets.map(mapAsset);
   }
 
   async update(authUser: AuthUserDto, id: string, dto: PersonUpdateDto): Promise<PersonResponseDto> {
     let person = await this.findOrFail(authUser, id);
 
     if (dto.name !== undefined || dto.birthDate !== undefined || dto.isHidden !== undefined) {
+      await this.access.requirePermission(authUser, Permission.PERSON_WRITE, id);
```
jrasm91, line 79: Doing anything in this method should require PERSON_WRITE, including the thumbnail change. Can you move this to the top, before the find or fail?

server/src/domain/person/person.service.ts
```diff
@@ -189,7 +198,8 @@ export class PersonService {
   }
 
   private async findOrFail(authUser: AuthUserDto, id: string) {
-    const person = await this.repository.getById(authUser.id, id);
+    await this.access.requirePermission(authUser, Permission.PERSON_READ, id);
```
jrasm91, line 201: I would prefer _not_ requiring this check here and _only_ having access checks at the beginning of the service methods. We don't need to check for person.read and then later (here) for person.write.

server/src/infra/repositories/access.repository.ts
```diff
@@ -1,14 +1,15 @@
...
-import { AlbumEntity, AssetEntity, PartnerEntity, SharedLinkEntity } from '../entities';
+import { AlbumEntity, AssetEntity, PartnerEntity, PersonEntity, SharedLinkEntity } from '../entities';
 
 export class AccessRepository implements IAccessRepository {
   constructor(
     @InjectRepository(AssetEntity) private assetRepository: Repository<AssetEntity>,
     @InjectRepository(AlbumEntity) private albumRepository: Repository<AlbumEntity>,
     @InjectRepository(PartnerEntity) private partnerRepository: Repository<PartnerEntity>,
     @InjectRepository(SharedLinkEntity) private sharedLinkRepository: Repository<SharedLinkEntity>,
+    @InjectRepository(PersonEntity) private personRepository: Repository<PersonEntity>,
```
jrasm91, line 12: Can you move this after `partnerRepository` and keep if alphabetical?

server/src/infra/repositories/access.repository.ts
```diff
@@ -156,4 +157,15 @@ export class AccessRepository implements IAccessRepository {
       });
     },
   };
+
+  person = {
+    hasOwnerAccess: (userId: string, personId: string): Promise<boolean> => {
+      return this.personRepository.exist({
```
jrasm91, line 163: I don't do this other places, but I think there is a variation you can use `existBy` the is basically just the `where` object.

jrasm91 (approved):
Looks good! Maybe we can add some face sharing with partners in the future.

martabal:
> Looks good! Maybe we can add some face sharing with partners in the future.

It's coming :eyes:

martabal (approved):

PR #4145, opened 2023-09-19
feat(web): cancel pending requests on image change

lmichelin:
## Description

This PR cancels useless pending requests to save bandwidth in 2 cases:
- The user navigates quickly in the photo viewer using the previous or the next button.
- The user zooms a picture and then navigates to other picture before the original image finished to load. In this case the original image download is cancelled.

## How Has This Been Tested?

- [x] Test A
	- Turn on network throttling in the developer tools of the browser
	- Open a picture and press the right arrow key several times quickly
	- We can see the canceled requests in the network panel
- [x] Test B
	- Turn on network throttling in the developer tools of the browser
	- Open a picture
	- Zoom in using the mouse wheel
	- Navigate to the next picture before the original image finished to load
	- The original picture download is cancelled

## Checklist:

### DIFF
--- a/web/src/lib/components/asset-viewer/photo-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/photo-viewer.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import { fade } from 'svelte/transition';
-  import { onMount } from 'svelte';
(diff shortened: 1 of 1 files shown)


jrasm91 (approved):
Nice!

PR #4146, opened 2023-09-19
refactor(web): album listing page

martabal:
## Changes made in this PR

This PR adds the following improvements:
- Sort based on the number of the assets
- Albums can be sorted in both ascending and descending orders

Edit : 

- Quick actions added on the List mode. The album name and description can be quickly edited and albums can be quicly deleted.
- Fixed a background issue with the navbar
- Switching to List or Cover mode can be done with a single click without needing a dropdown menu 

### Sorting options

| Before | After |
| --- | --- |
| <video src="https://github.com/immich-app/immich/assets/74269598/6f2fa9b1-9422-42b1-ac24-9f2d8817b6d9"> |<video src="https://github.com/immich-app/immich/assets/74269598/906ca1f4-becd-442d-be97-54fada0f16aa"> |

### Quick edit an album

image

### Responsive design

https://github.com/immich-app/immich/assets/74269598/ac5e3ec8-182e-431d-884b-8445bf54ee29

### Fixed small issue with the navbar

| Before | After |
| --- | --- |
| <video src="https://github.com/immich-app/immich/assets/74269598/03663cf3-4bcc-4eac-9837-cfd463ba6562"> |<video src="https://github.com/immich-app/immich/assets/74269598/fdef4918-8493-4453-a594-4b965f2ad915"> |

### DIFF
--- a/web/src/lib/components/elements/dropdown.svelte
+++ b/web/src/lib/components/elements/dropdown.svelte
@@ -40,7 +40,7 @@
       {#if icon}
         <svelte:component this={icon} size="18" />
       {/if}
-      {value}
+      <p class="hidden sm:block">{value}</p>
     </div>
   </LinkButton>
 
--- a/web/src/lib/components/elements/dropdown.svelte
+++ b/web/src/lib/components/elements/dropdown.svelte
@@ -4,13 +4,12 @@
   import { clickOutside } from '$lib/utils/click-outside';
   import { fly } from 'svelte/transition';
   import type Icon from 'svelte-material-icons/DotsVertical.svelte';
+  import { createEventDispatcher } from 'svelte';
 
-  interface DropdownOption {
-    value: string;
-    icon?: Icon;
-  }
-
-  export let options: DropdownOption[] | string[] = [];
+  const dispatch = createEventDispatcher<{
+    changeSort: string;
+  }>();
+  export let options: string[];
   export let value = options[0];
   export let icons: (typeof Icon)[] | undefined = undefined;
 
@@ -21,7 +20,12 @@
   };
 
   const handleSelectOption = (index: number) => {
-    value = options[index];
+    if (options[index] === value) {
+      dispatch('changeSort', value);
+    } else {
+      value = options[index];
+    }
+
     showMenu = false;
   };
 
--- a/web/src/lib/components/elements/dropdown.svelte
+++ b/web/src/lib/components/elements/dropdown.svelte
@@ -7,7 +7,7 @@
   import { createEventDispatcher } from 'svelte';
 
   const dispatch = createEventDispatcher<{
-    changeSort: string;
+    select: string;
   }>();
   export let options: string[];
   export let value = options[0];
@@ -21,7 +21,7 @@
 
   const handleSelectOption = (index: number) => {
     if (options[index] === value) {
-      dispatch('changeSort', value);
+      dispatch('select', value);
     } else {
       value = options[index];
     }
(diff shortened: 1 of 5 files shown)


web/src/lib/components/elements/dropdown.svelte
```diff
@@ -4,13 +4,12 @@
...
 
-  interface DropdownOption {
-    value: string;
-    icon?: Icon;
-  }
-
-  export let options: DropdownOption[] | string[] = [];
+  const dispatch = createEventDispatcher<{
+    changeSort: string;
+  }>();
```
jrasm91, line 11: Clicking the same option multiple times might not always mean change the sort order. This is a general dropdown component. What about just emitting an `on:select` event. In the parent, if it matches the current value, then you can do something.

web/src/routes/(user)/albums/+page.svelte
```diff
@@ -1,11 +1,20 @@
+<script lang="ts" context="module">
+  // table is the text printed in the table and order is the text printed in the dropDow menu
+  export interface Sort {
+    table: string;
+    dropDow: string;
```
jrasm91, line 5: spelling, should be `dropDown` probably. But `label` or `title` might be more appropriate. What about `displayTitle` and `sortTitle`? Maybe also `widthClass`

web/src/routes/(user)/albums/+page.svelte
```diff
@@ -86,16 +104,34 @@
 
   $: {
     const { sortBy } = $albumViewSettings;
-    if (sortBy === 'Most recent photo') {
+    if (sortBy === sortByOptions.mostRecent.dropDow) {
```
jrasm91, line 107: This looks like a nightname lol. Is there any way to clean this up?

web/src/routes/(user)/albums/+page.svelte
martabal: This is much better. Thanks :pray:

web/src/routes/(user)/albums/+page.svelte
martabal: We can hardcode the value like before, but it may be complicated if we add labels or other ways of sorting. :man_shrugging: 

If you have any other idea, I'm in!

web/src/routes/(user)/albums/+page.svelte
```diff
@@ -1,11 +1,20 @@
+<script lang="ts" context="module">
+  // table is the text printed in the table and order is the text printed in the dropDow menu
+  export interface Sort {
```
jrasm91, line 3: Maybe this interface can `sortFn: ` that is a method that takes in a list of albums and returns them sorted in descending order. Then you can just call that and optionally reverse it.

Another option is to add a function that you call on a single album that returns a string value that you can sort on. Then the main sort implementation is like `albums.sort((a, b) =>mapFn(a).localCompare(mapFn(b))`

martabal: > Maybe this interface can `sortFn: ` that is a method that takes in a list of albums and returns them sorted in descending order. Then you can just call that and optionally reverse it.

I don't fully understand; What is the list sorted by ?

jrasm91: Each option in the `sortByOptions` can have its own implementation to sort a list of albums. So it has a title, label, direction, and a sort function.

martabal: Yes, makes sense. Thanks !

jrasm91: Looks a lot better. I think we still are using lodash-es. Might be possible to use that order by method too for more concise code

waclaw66:
Please fix misalignment of headers and items in cover view as it was before.

klejejs:
This looks nice!
Something I recently noticed is that list and grid are the only options in the dropdown, so do you think it would make sense to change view with a single click instead of having a dropdown?
Would it make sense to change that in this PR, or should there be a separate PR?

alextran1502 (approved):

PR #4153, opened 2023-09-20
feat(server, web): unmerge person's assets

martabal:
## Changes made in this PR

This PR introduces the capability to detach an asset from an person. Upon unmerging, the asset is assigned with either a new or an existing person.

This PR adds a few improvements too:

- Currently, the Explore page displays a fixed count of 12 faces. This PR introduces a dynamic approach, adjusting the number of faces displayed based on the page width, presenting them in a single line.
- On the merge face selector modal, use the `CircleIconButton` to provide a clearer indication that the two-arrow button is interactive.

To do : 
- [x] Rework the flow
- [x] Give to the user the option to either link the asset to an existing person or link the asset to a new person.
- [x] Handle the case if the asset unmerged is the feature photo
- [ ] Unassign person's assets
- [ ] Add server tests

### From the asset-viewer

https://github.com/immich-app/immich/assets/74269598/d324ee46-233b-4a3a-a4ee-a20801413cf4

### From the person page

https://github.com/immich-app/immich/assets/74269598/8273941b-5317-4916-b9d5-cb87c55f4c50

### Other improvements

| Explore page | Merge face selector | Person page |
| :---: | :---: | :---: |
| <video src="https://github.com/immich-app/immich/assets/74269598/e28ec9d5-6151-4b50-beeb-4ff64f6ef016"> | <video src="https://github.com/immich-app/immich/assets/74269598/00fd3d18-1b29-41c2-93e2-09561d9bf68b"> | Screenshot from 2023-10-30 11-43-55-2 |


server/src/domain/person/person.service.ts
```diff
@@ -147,6 +148,56 @@ export class PersonService {
...
+
+    const oldPerson = await this.findOrFail(dto.personId);
+
+    const newPerson = await this.repository.create({ ownerId: authUser.id });
+    let result: BulkIdResponseDto;
+    try {
+      const mergePerson = await this.repository.getById(oldPerson.id);
+      if (!mergePerson) {
+        result = { id: oldPerson.id, success: false, error: BulkIdErrorReason.NOT_FOUND };
+      }
```
jrasm91, line 162: Why are you looking up old person a second time?

server/src/domain/person/person.service.ts
```diff
@@ -147,6 +148,56 @@ export class PersonService {
...
+      this.logger.log(`un-merging ${dto.assetId} from ${oldPerson.id}`);
+      await this.repository.removeFaceFromPerson(oldPerson.id, newPerson.id, dto.assetId);
+
+      await this.repository.update({
+        id: newPerson.id,
+        faceAssetId: dto.assetId,
+      });
+
+      const assetId = dto.assetId;
+      const face = await this.repository.getFaceById({ personId: newPerson.id, assetId });
```
jrasm91, line 173: You should look up the face right after the person and throw an error before doing anything if it is invalid.

server/src/domain/person/person.service.ts
```diff
@@ -147,6 +148,56 @@ export class PersonService {
...
+      if (!face) {
+        throw new BadRequestException('Invalid assetId for feature face');
+      }
+      await this.jobRepository.queue({
+        name: JobName.GENERATE_FACE_THUMBNAIL,
```
jrasm91, line 178: We should change this job to only pass asset Id and person id, since the bounding box is stored on the face now, so we can look it up in the job implementation instead.

server/src/domain/person/person.service.ts
```diff
@@ -147,6 +148,56 @@ export class PersonService {
...
+          imageWidth: face.imageWidth,
+        },
+      });
+
+      result = { id: oldPerson.id, success: true };
+    } catch (error: Error | any) {
+      this.logger.error(`Unable to un-merge asset ${dto.assetId} from ${oldPerson.id}`, error?.stack);
+      result = { id: oldPerson.id, success: false, error: BulkIdErrorReason.UNKNOWN };
+    }
+    return result;
```
jrasm91, line 198: No need to use the bulk stuff for a single request.

server/src/immich/controllers/person.controller.ts
```diff
@@ -28,6 +29,11 @@ function asStreamableFile({ stream, type, length }: ImmichReadStream) {
 export class PersonController {
   constructor(private service: PersonService) {}
 
+  @Post('/unmerge')
```
jrasm91, line 32: What about delete /person/:id/asset/:assetId instead? And have it return a person response dto

server/src/infra/repositories/person.repository.ts
```diff
@@ -40,6 +40,18 @@ export class PersonRepository implements IPersonRepository {
     return result.affected ?? 0;
   }
 
+  async removeFaceFromPerson(oldPersonId: string, newPersonId: string, assetId: string): Promise<number> {
+    const result = await this.assetFaceRepository
+      .createQueryBuilder()
```
jrasm91, line 45: Can you rewrite this without the query builder as a normal update?

Can we call this reassign face instead?

martabal:
When the user decides to create a new person, the client generates the feature photo based on the boundingbox sent back by the server. Currently, the feature photo dezoom is flawed

martabal:
I'm not satisfied with the current flow on the edit panel. Currently you can reset the changes you made, but with the addition of unassigned faces, it gets quickly messy (each assigned faces can have a blue minus button, a red minus button and a blue reset button and each unassigned faces can have a blue reset button and a blue plus button). I will rework the flow : With assigned faces, you will only be able to unassign them (red minus button) and with unassigned faces you will only be able to assign them (blue plus minus button).

martabal:
Little update : I started to work on re-assigning faces feature, I hope to open a new PR as stated in #4717 next week

oziee:
amazing work..  can't wait to see how future alterations make this the best home photo manager around.. but with this first release after a little playing around I did find some items confusing

clicking "+" `Create new person` does not add a name, and it does not seem too intuitive as to what the program is changing. 1. Feels like the people list just spins around..
2. it shows a blue "-" button which feels like clicking this would be `remove` or `delete`
3. when changing the person the `select face` lists all people, hidden too.. and they are hidden for a reason usually as they want to be ignored/exempt for being ever shown

martabal:
Thanks for the feeback ! FYI, this PR was the first implementation which was superseeded by #5511 and #4949

> clicking "+" `Create new person` does not add a name, and it does not seem too intuitive as to what the program is changing.

Yes that's the goal, you're creating a new person for the selected face, and by default a new person does not have a name

>  it shows a blue "-" button which feels like clicking this would be remove or delete

I completely agree the UX/UI should be changed,  #5511 will complexify it again :sweat_smile: . I'm not an expert and I will be happy to change it for a better UX/UI design

> when changing the person the select face lists all people, hidden too.. and they are hidden for a reason usually as they want to be ignored/exempt for being ever shown

I disagree, I think it's a good thing to be able to reassign a face with a hidden person. Else, you have to close the modal, go to the explore page, un-hide the person, come back to the asset, re-assign the face, come back to the explore page and hide again the person.

BDeus:
Hi @martabal many thanks for you great works. I follow every PR you done. Thanks again
I'm wondering if it's on waiting list to be able to remove a full person with all these assets to be able to reassign it.
I have my 2 kids merge into a single person and i really don't want reindex everything

martabal:
Hello !
Unfortunately, the only way to "unmerge" 2 people recognized as the same person in the same asset is to go through each assets and re-assign them individually. 

In the future, we hope to improve the mechanism to cluster faces which could solve your issue. And at some point, a tool to detect similar faces so you can easily help the server to correctly assign the faces. You can see [here](https://github.com/immich-app/immich/discussions/5532#discussioncomment-7807128) the discussion we had about it.

hectorzin:
It is a problem when in the same photo, there are two persons tagged as the same one, it is impossoble to untag only on of them.
and when looking at the photo there are two persons as the same one, there are two rectangles focusing the faces at the photo but only one at the right side, so it is impossible to name separately one from the other

PR #4156, opened 2023-09-21
fix(server): search for terms separated by hyphens and/or underscores in asset search

ochen1:


jrasm91 (commented):
You will have to change line 3 to be version 9.

ochen1:
Done! @jrasm91

daniele-athome:
Can underscore really be a token separator in natural language? If I understand the purpose of this change correctly.

ochen1:
This PR originates from Discord and attempts to fix searching by file names, which are typically not considered natural language. For example, I should be able to search MyName and get the result "123_MyName_Last-1.png"

Similar naming conventions are pretty standard for photography companies. Space is a character that should be avoided in filenames.

I don't believe this will cause any problems with other use cases?

daniele-athome:
Thank you, I wasn't thinking about filenames.

alextran1502 (approved):

PR #4177, opened 2023-09-22
Correct admin commands

erichelgeson:
Also use text for examples so they can be easily copied and updated.


jrasm91 (commented):
Can either (1) add `/usr/src/app #` at the beginning of all the commands or (2) remove the `#` and just have the command by itself on the first line in each example?

jrasm91 (approved):

PR #4184, opened 2023-09-22
feat(server): random assets API

daniele-athome:
Adding a new API for assets:

```
/api/asset/random?count=5
```

"count" parameter is optional, defaulting to 1.

It will return `count` random assets owned by the authenticated user.

Possible things for the review:

* automated tests: I didn't know what to test here, no special logic in the service, just doing a SQL query
* selection query: I had to use a custom SQL query - see comments inside
* I did some research and tests and I picked the fastest and most reliable methods among the ones I found

## How was this tested?

* called API with no parameters -> defaults to 1 asset
* called API with invalid parameter -> HTTP 400
* called API with valid parameter -> returns N assets

### DIFF
--- a/server/src/domain/asset/dto/asset.dto.ts
+++ b/server/src/domain/asset/dto/asset.dto.ts
@@ -1,4 +1,5 @@
-import { IsBoolean, IsString } from 'class-validator';
+import { Type } from 'class-transformer';
+import { IsBoolean, IsNumber, IsPositive, IsString } from 'class-validator';
 import { Optional } from '../../domain.util';
 import { BulkIdsDto } from '../response-dto';
 
@@ -24,4 +25,12 @@ export class UpdateAssetDto {
   @Optional()
   @IsString()
   description?: string;
+}
+
+export class RandomAssetsDto {
+  @Optional()
+  @IsNumber()
+  @IsPositive()
+  @Type(() => Number)
+  count?: number | null;
 }
--- a/server/src/domain/asset/dto/asset.dto.ts
+++ b/server/src/domain/asset/dto/asset.dto.ts
@@ -1,5 +1,5 @@
 import { Type } from 'class-transformer';
-import { IsBoolean, IsNumber, IsPositive, IsString } from 'class-validator';
+import { IsBoolean, IsInt, IsPositive, IsString } from 'class-validator';
 import { Optional } from '../../domain.util';
 import { BulkIdsDto } from '../response-dto';
 
@@ -29,8 +29,8 @@ export class UpdateAssetDto {
 
 export class RandomAssetsDto {
   @Optional()
-  @IsNumber()
+  @IsInt()
   @IsPositive()
   @Type(() => Number)
-  count?: number | null;
+  count?: number;
 }
--- a/server/src/infra/repositories/asset.repository.ts
+++ b/server/src/infra/repositories/asset.repository.ts
@@ -429,6 +429,17 @@ export class AssetRepository implements IAssetRepository {
     return result;
   }
 
+  getRandom(ownerId: string, count: number): Promise<AssetEntity[]> {
+    // can't use queryBuilder because of custom OFFSET clause
+    return this.repository.query(
+      `SELECT *
+                                  FROM assets
+                                  WHERE "ownerId" = $1
+                                  OFFSET FLOOR(RANDOM() * (SELECT COUNT(*) FROM ASSETS)) LIMIT $2`,
+      [ownerId, count],
+    );
+  }
+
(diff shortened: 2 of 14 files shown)


server/src/domain/asset/dto/asset.dto.ts
```diff
@@ -25,3 +26,11 @@ export class UpdateAssetDto {
   @IsString()
   description?: string;
 }
+
+export class RandomAssetsDto {
+  @Optional()
+  @IsNumber()
```
jrasm91, line 32: Is int

server/src/domain/asset/dto/asset.dto.ts
```diff
@@ -25,3 +26,11 @@ export class UpdateAssetDto {
...
+  @IsPositive()
+  @Type(() => Number)
+  count?: number | null;
```
jrasm91, line 35: Null is not valid here, so you can remove that type.

server/src/infra/repositories/asset.repository.ts
```diff
@@ -429,6 +429,17 @@ export class AssetRepository implements IAssetRepository {
     return result;
   }
 
+  getRandom(ownerId: string, count: number): Promise<AssetEntity[]> {
+    // can't use queryBuilder because of custom OFFSET clause
+    return this.repository.query(
+      `SELECT *
+                                  FROM assets
+                                  WHERE "ownerId" = $1
+                                  OFFSET FLOOR(RANDOM() * (SELECT COUNT(*) FROM ASSETS)) LIMIT $2`,
```
jrasm91, line 438: Looks like this could return less than the limit in a few situations if the offset was close to the number of assets. Probably not a big deal for most situations though.

jrasm91:
Can you add a few e2e tests with the situations you described? Should be easy enough to add.

server/src/infra/repositories/asset.repository.ts
daniele-athome: You're right, I didn't think about boundary values. I think I fixed it in a more or less proper way.

daniele-athome:
I implemented tests for all use cases, although I didn't find a way to properly `expect` a random asset in a "jesty" way - there seems to be extensions for jest that could implement those, but I didn't think it was worth to add a dependecy just for an OR check.

I think I also found a typo in another test - I isolated the fix in a separate commit.

server/test/e2e/asset.e2e-spec.ts
```diff
@@ -378,6 +385,58 @@ describe(`${AssetController.name} (e2e)`, () => {
...
+        .get('/asset/random')
+        .set('Authorization', `Bearer ${user1.accessToken}`);
+
+      expect(status).toBe(200);
+
+      const assets: AssetResponseDto[] = body;
+      expect(assets.length).toBe(1);
+      expect(assets[0].ownerId).toBe(user1.userId);
+      // assets owned by user1
+      expect(assets[0].id == asset1.id || assets[0].id == asset2.id || assets[0].id == asset3.id).toBeTruthy();
```
danieldietzler, line 407: I personally would prefer `expect([asset1.id, asset2.id, asset3.id]).toContain(assets[0].id)`

server/test/e2e/asset.e2e-spec.ts
```diff
@@ -378,6 +385,58 @@ describe(`${AssetController.name} (e2e)`, () => {
...
+      // assets owned by user2
+      expect(assets[0].id == asset4.id).toBeFalsy();
```
danieldietzler, line 409: You can do `expect(asset[0].id).not.toBe(asset4.id)` instead

server/test/e2e/asset.e2e-spec.ts
```diff
@@ -322,7 +329,7 @@ describe(`${AssetController.name} (e2e)`, () => {
     });
 
     it('should require authentication', async () => {
-      const { status, body } = await request(server).get('/album/statistics');
+      const { status, body } = await request(server).get('/asset/statistics');
```
danieldietzler, line 332: Thanks for the good catch! :) That was my bad.

server/test/e2e/asset.e2e-spec.ts
daniele-athome: I don't know why I didn't think of that, thank you :-)

danieldietzler: Well, it is somewhat counterintuitive :D

danieldietzler (approved):
LGTM! I would like to have Jason take another look at it though.

alextran1502 (approved):

PR #4204, opened 2023-09-24
Add filesize to backup asset info and improve allignment.

bee395:
A few improvements related to the current backup asset info box in the mobile app

1. Only show this tile when an backup is running
2. Align the icon of this tile to the other items in the list
3. Add the file size to the backup asset info
4. I also changed the upload percentage in this tile. Every digit uses the same amount of space, so the percentage bar changes size less often

Edit: added point 4

### DIFF
--- a/mobile/assets/i18n/en-US.json
+++ b/mobile/assets/i18n/en-US.json
@@ -74,6 +74,7 @@
   "backup_controller_page_failed": "Failed ({})",
   "backup_controller_page_filename": "File name: {} [{}]",
   "backup_controller_page_id": "ID: {}",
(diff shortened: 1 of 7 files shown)


alextran1502:
Hello, can you help with including the screenshot?

bee395:
Hey Alex,

Thanks for your reply
See 1 screenshot for the idle state
And 1 where the backup is running

Screenshot_20230924_174009
Screenshot_20230924_174227

alextran1502:
Outdated, as we are planning to refactor the whole backup process. This might not applied anymore

PR #4222, opened 2023-09-25
feat(server, web)!: Move reverse geocoding settings to the UI

danieldietzler:
> **Warning**
> **Breaking Changes**:
> - `DISABLE_REVERSE_GEOCODING` has been replaced with an admin setting (UI or config file).
> - `REVERSE_GEOCODING_PRECISION` has been replaced with an admin setting (UI or config file)

In this PR:
- Move reverse geocoding settings to the Admin Settings page, inside of "Map Settings"
- Enables updating reverse geocoding precision on the fly

Tested Scenarios:
- Disable reverse geocoding and verified no location info appeared in the exif info panel
- Enabled reverse geocoding and verified uploaded assets were correctly processed
- Changed precision and verified new csv files were downloaded and correct settings were applied when manually refreshing a single asset.

Future Work:
- Possibly separate reverse geocoding jobs from metadata extraction. (Extraction saves gps data, reverse geocoding runs ons gps data directly)
- Possibly look into supporting google maps as a reverse geocoder (https://github.com/immich-app/immich/discussions/2684(

image


jrasm91 (approved):

alextran1502:
Hello, thank you for working on this. I've just tested the PR and saw one issue

image

Upon opening up the setting panel without changing anything, I see the "Reset to default" which shouldn't be there. I think the logic is incorrect in this case

jrasm91:
Good catch @alextran1502  - fixed.

PR #4227, opened 2023-09-25
chore(web): use axios isCancel function

uhthomas:
Use of this function should be more idiomatic and reliable - it's much less likely for changes to Axios to cause regressions.


jrasm91 (approved):
Is this the only place it is used/checked?

uhthomas:
> Is this the only place it is used/checked?

I believe so - I don't see any other references.

https://github.com/search?q=repo%3Aimmich-app%2Fimmich%20CanceledError&type=code

image

PR #4237, opened 2023-09-26
refactor(server): merge facial-recognition and person

danieldietzler:
Manually testing it everything seems to be fine.
Since the tests all pass as well and they should be decently good in that area it should be good


server/src/domain/search/search.service.spec.ts
```diff
@@ -7,18 +7,18 @@ import {
...
-  newFaceRepositoryMock,
   newJobRepositoryMock,
   newMachineLearningRepositoryMock,
+  newPersonRepositoryMock,
   newSearchRepositoryMock,
   newSystemConfigRepositoryMock,
   searchStub,
 } from '@test';
 import { plainToInstance } from 'class-transformer';
+import { IPersonRepository } from '..';
```
jrasm91, line 18: can you fix this import?

jrasm91 (approved):
Nice!

jrasm91 (approved):

PR #4239, opened 2023-09-26
fix(server): normalize external path

etnoy:
This normalizes the user's external path. This should fix many bugs we've found with external paths and no assets being found.


jrasm91:
Can you give an example of a situation that this would fix?

etnoy:
> Can you give an example of a situation that this would fix?

After some consideration, this might fix fewer bugs than anticipated. However, whenever you have /../ in the external path, this makes things easier. Note that this will always be the case when we do e2e.

jrasm91:
In that situation can we normalize the path before it is saved instead?

etnoy:
> In that situation can we normalize the path before it is saved instead?

done!

alextran1502 (approved):

PR #4248, opened 2023-09-27
feat(server): improve library scanning

jrasm91:
- Better scanning algorithm (avoid nested for loops)
- Remove (inline) MARK_OFFLINE job


jrasm91:
Mostly re-implemented in subsequent PRs.

PR #4251, opened 2023-09-27
fix(server): Thumbnail migration creating unnecessary directories

danieldietzler:
As pointed out in https://github.com/immich-app/immich/pull/4112#issuecomment-1736784668, the migration job created all directories for each asset (so `jpeg`, `webp` and `encoded video`) before checking if those folders are actually needed.

### DIFF
--- a/server/src/domain/media/media.service.ts
+++ b/server/src/domain/media/media.service.ts
@@ -98,26 +98,29 @@ export class MediaService {
     if (!asset) {
       return false;
     }
-    const resizePath = this.ensureThumbnailPath(asset, 'jpeg', false);
-    const webpPath = this.ensureThumbnailPath(asset, 'webp', false);
-    const encodedVideoPath = this.ensureEncodedVideoPath(asset, 'mp4', false);
 
-    if (asset.resizePath && asset.resizePath !== resizePath) {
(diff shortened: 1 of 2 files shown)


server/src/domain/media/media.service.ts
```diff
@@ -98,21 +98,24 @@ export class MediaService {
...
       return false;
     }
-    const resizePath = this.ensureThumbnailPath(asset, 'jpeg');
-    const webpPath = this.ensureThumbnailPath(asset, 'webp');
-    const encodedVideoPath = this.ensureEncodedVideoPath(asset, 'mp4');
+    const resizePath = this.ensureThumbnailPath(asset, 'jpeg', false);
+    const webpPath = this.ensureThumbnailPath(asset, 'webp', false);
+    const encodedVideoPath = this.ensureEncodedVideoPath(asset, 'mp4', false);
 
     if (asset.resizePath && asset.resizePath !== resizePath) {
```
jrasm91, line 105: Maybe we should leave ensure path like you had it before, but then only call it if the asset has the file to begin with, etc. So move the ensure path calls inside of the if statements. And have one for if the asset has the path and a second nested is for if the path is different.

jrasm91 (approved):

PR #4262, opened 2023-09-28
feat(web) Option to see Archived Assets on the Map

badbreze:
With this PR i'm trying to add an option to see Archived Assets on my map (normally i archive photos when i put them inside some album)

Hope this is useful.

### DIFF
--- a/server/src/domain/asset/asset.repository.ts
+++ b/server/src/domain/asset/asset.repository.ts
@@ -23,7 +23,7 @@ export interface LivePhotoSearchOptions {
 
 export interface MapMarkerSearchOptions {
   isFavorite?: boolean;
-  includeArchived?: boolean;
+  isArchived?: boolean;
   fileCreatedBefore?: Date;
   fileCreatedAfter?: Date;
 }
--- a/server/src/domain/asset/asset.repository.ts
+++ b/server/src/domain/asset/asset.repository.ts
@@ -23,6 +23,7 @@ export interface LivePhotoSearchOptions {
 
 export interface MapMarkerSearchOptions {
   isFavorite?: boolean;
+  includeArchived?: boolean;
   fileCreatedBefore?: Date;
   fileCreatedAfter?: Date;
 }
--- a/web/src/lib/stores/preferences.store.ts
+++ b/web/src/lib/stores/preferences.store.ts
@@ -22,6 +22,7 @@ export const locale = persisted<string | undefined>('locale', undefined, {
 export interface MapSettings {
   allowDarkMode: boolean;
   onlyFavorites: boolean;
+  seeArchived: boolean;
   relativeDate: string;
   dateAfter: string;
   dateBefore: string;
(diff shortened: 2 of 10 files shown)


server/src/domain/asset/asset.repository.ts
```diff
@@ -23,6 +23,7 @@ export interface LivePhotoSearchOptions {
 
 export interface MapMarkerSearchOptions {
   isFavorite?: boolean;
+  includeArchived?: boolean;
```
jrasm91, line 26: The main timeline has this same situation. In that case the endpoint defaults to no value. The clients have to send `false` to exclude them. Can we do the same here? So:
- Rename this to `isArchived`
- Change web/mobile to send `isArchived: false` by default, now configurable in the web.

What do you think about using `isArchived` instead? We can then use it

web/src/lib/stores/preferences.store.ts
```diff
@@ -22,6 +22,7 @@ export const locale = persisted<string | undefined>('locale', undefined, {
 export interface MapSettings {
   allowDarkMode: boolean;
   onlyFavorites: boolean;
+  seeArchived: boolean;
```
jrasm91, line 25: What do you think about `withArchived` instead?

badbreze: can be an alternative! i try when come home!!!

server/src/domain/asset/asset.repository.ts
badbreze: isArchived is already used to toggle between Archived and Not Archived, not the objective of this option!

jrasm91: I know, but if you omit it, that can return all assets, and that provide three options/states, correct?:
- isArchived (not set) = all photos (including archived assets)
- isArchived: true = only archived
- isArchived: false = only unarchived

jrasm91: I believe in the web you want it to only return assets that are not archived (so send `isArchived: false` by default) and an option to include them (so don't send `isArchived` at all, to include all assets)

badbreze: now it's using the already known isArchived, but imo, this limits future opportunity to "Toggle Between", and makes some confusion on the scope of the property, but if you think thats better i'm ok!

jrasm91: Can you help me understand how this limits future toggling exactly? I am not following. What example specifically would not work?

badbreze: for example, if you wanted to give the option to see "only archived", it's a remote possibility, but it would be impossible to manage with a single variable.

jrasm91: Why would sending `isArchived: true` not work for that case?

badbreze: nevermind man, drop the PR.....

badbreze:
Closing the PR because i can't follow the discussion. thanks anyone

PR #4263, opened 2023-09-28
feat(server): Add week numbers for templating

danieldietzler:
Add the option to use week numbers ('W') and padded week numbers ('WW') for storage templating.

Screenshot 2023-09-28 at 18 59 17

Screenshot 2023-09-28 at 18 58 20

Closes #4260.


bo0tzz:
The week number that the example date falls in is a bit unfortunate as it doesn't show off the padding. Very minor, but should we change that?

jrasm91 (approved):

PR #4265, opened 2023-09-28
fix(mobile): reduce server version api calls

shenlong-tanwen:
#### Changes made in this PR:

- The server version number is only used in the app server info box and home page app bar. However, the `getServerVersion` was getting called on each bottom tab switch resulting in unnecessary calls to the server each time a tab was switched. This PR reduces this to only fetch the server version when the user navigates to the home page / timeline. 

> [!NOTE]  
> We can also entirely remove this call from tab navigation. The server info calls are also sent on app start, app resumes and during backups which should be more than sufficient for the actual usage value it has.

- The sharing page's top button did not have any top padding set.  Updated them to use the same padding as the one used in the library page to make them consistent

Before | After
:-: | :-:
Before | After

### DIFF
--- a/mobile/lib/modules/album/views/sharing_page.dart
+++ b/mobile/lib/modules/album/views/sharing_page.dart
@@ -117,6 +117,7 @@ class SharingPage extends HookConsumerWidget {
         padding: const EdgeInsets.only(
           left: 12.0,
           right: 12.0,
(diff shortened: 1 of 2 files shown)


jrasm91 (approved):
Nice!

alextran1502 (approved):

PR #4277, opened 2023-09-29
feat(web): shuffle slideshow order

jrasm91:
Add an option to randomize (shuffle) the order of the pictures in the slideshow:

image
image

Add an option to launch the slideshow from the album options:

image

I would love to have the shuffle button be different colors to indicate if it is active or not, but the circle button doesn't support anything like that right now.

### DIFF
--- a/web/src/lib/components/asset-viewer/asset-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/asset-viewer.svelte
@@ -32,7 +32,7 @@
   import CircleIconButton from '../elements/buttons/circle-icon-button.svelte';
   import ProgressBar, { ProgressBarStatus } from '../shared-components/progress-bar/progress-bar.svelte';
   import { shouldIgnoreShortcut } from '$lib/utils/shortcut';
-  import { assetViewingStore } from '$lib/stores/asset-viewing.store';
+  import { assetViewingStore, SlideshowState } from '$lib/stores/asset-viewing.store';
   import { featureFlags } from '$lib/stores/server-config.store';
   import {
     mdiChevronLeft,
@@ -65,7 +65,7 @@
 
   let reactions: ActivityResponseDto[] = [];
 
-  const { slideshow, slideshowShuffle, setAssetId } = assetViewingStore;
+  const { slideshowShuffle, slideshowState, setAssetId } = assetViewingStore;
 
   const dispatch = createEventDispatcher<{
     archived: AssetResponseDto;
@@ -192,7 +192,16 @@
   });
 
   $: asset.id && !sharedLink && getAllAlbums(); // Update the album information when the asset ID changes
-  $: $slideshow ? handlePlaySlideshow() : handleStopSlideshow();
+
+  $: $slideshowState,
+    (() => {
+      if ($slideshowState === SlideshowState.PlaySlideshow) {
+        handlePlaySlideshow();
+      } else if ($slideshowState === SlideshowState.StopSlideshow) {
+        handleStopSlideshow();
+      }
+    })();
+
   $: $slideshowShuffle && preload();
 
   const getAllAlbums = async () => {
@@ -284,11 +293,11 @@
   };
 
   const navigateAssetForward = async (e?: Event) => {
-    if ($slideshow && $slideshowShuffle) {
+    if ($slideshowState === SlideshowState.PlaySlideshow && $slideshowShuffle) {
       return navigateAssetRandom();
     }
 
-    if ($slideshow && assetStore && progressBar) {
+    if ($slideshowState === SlideshowState.PlaySlideshow && assetStore && progressBar) {
       const hasNext = await assetStore.getNextAssetId(asset.id);
       if (hasNext) {
         progressBar.restart(true);
@@ -302,11 +311,11 @@
   };
 
   const navigateAssetBackward = (e?: Event) => {
-    if ($slideshow && $slideshowShuffle) {
+    if ($slideshowState === SlideshowState.PlaySlideshow && $slideshowShuffle) {
       return navigateAssetRandom();
     }
 
-    if ($slideshow && progressBar) {
+    if ($slideshowState === SlideshowState.PlaySlideshow && progressBar) {
       progressBar.restart(true);
     }
 
@@ -460,13 +469,13 @@
   let progressBarStatus: ProgressBarStatus;
 
   const handleVideoStarted = () => {
-    if ($slideshow) {
+    if ($slideshowState === SlideshowState.PlaySlideshow) {
       progressBar.restart(false);
     }
   };
 
   const handleVideoEnded = async () => {
-    if ($slideshow) {
+    if ($slideshowState === SlideshowState.PlaySlideshow) {
       await navigateAssetForward();
     }
   };
@@ -481,30 +490,22 @@
   };
 
   const handlePlaySlideshow = async () => {
-    if ($slideshow) {
-      return;
-    }
-
     try {
       await assetViewerHtmlElement.requestFullscreen();
     } catch (error) {
       console.error('Error entering fullscreen', error);
-      $slideshow = false;
+      $slideshowState = SlideshowState.StopSlideshow;
     }
   };
 
   const handleStopSlideshow = async () => {
-    if (!$slideshow) {
-      return;
-    }
-
     try {
       await document.exitFullscreen();
     } catch (error) {
       console.error('Error exiting fullscreen', error);
     } finally {
-      $slideshow = false;
       progressBar.restart(false);
+      $slideshowState = SlideshowState.None;
     }
   };
 
@@ -544,11 +545,15 @@
   bind:this={assetViewerHtmlElement}
 >
   <div class="z-[1000] col-span-4 col-start-1 row-span-1 row-start-1 transition-transform">
-    {#if $slideshow}
+    {#if $slideshowState != SlideshowState.None}
       <!-- SlideShowController -->
       <div class="flex">
         <div class="m-4 flex gap-2">
-          <CircleIconButton icon={mdiClose} on:click={() => ($slideshow = false)} title="Exit Slideshow" />
+          <CircleIconButton
+            icon={mdiClose}
+            on:click={() => ($slideshowState = SlideshowState.StopSlideshow)}
+            title="Exit Slideshow"
+          />
           {#if $slideshowShuffle}
             <CircleIconButton icon={mdiShuffle} on:click={() => ($slideshowShuffle = false)} title="Shuffle" />
           {:else}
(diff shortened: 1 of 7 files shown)


brighteyed:
Maybe instead of preloading all the buckets we could randomly select one bucket and then randomly select an asset from that bucket each time  `getRandomAsset` is called
Also we could keep track of the N previous assets shown and use that list in the `navigateAssetBackward` call. This idea lets us cover the "hey, I'd like to see that again" use case

brighteyed:
I removed the preloading of all the buckets and implemented a slideshow asset history. I think the code is ready for review now

web/src/lib/utils/slideshow-history.ts
```diff
@@ -0,0 +1,40 @@
...
+  private slideshowHistoryIndex = 0;
+
+  constructor(private onNavigate: (assetId: string) => void) {}
+
+  reset = (assetId: string) => {
+    this.slideshowHistory = [assetId];
+    this.slideshowHistoryIndex = 0;
+  };
+
+  append = (assetId: string) => {
```
jrasm91, line 12: Can you make these _not_ arrow functions?

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -76,6 +91,21 @@
   onMount(async () => {
     document.addEventListener('keydown', onKeyboardPress);
 
+    slideshowStateUnsubscribe = slideshowState.subscribe((value) => {
```
jrasm91, line 94: So you can do it this way yes, but stores can also be automatically subscribed and unsubscribed by svelte (in a svelte file). You could try doing something like this instead, which I have found works nicely. `$: $slideshowState && onSlideshowStateChange()`, basically use reactivity with the store to trigger code changes instead of manually subscribing and unsubscribing

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -459,19 +506,56 @@
...
       />
-    {/if}
-  </div>
+    </div>
+  {/if}
 
-  {#if !isSlideshowMode && showNavigation}
-    <div class="column-span-1 z-[999] col-start-1 row-span-1 row-start-2 mb-[60px] justify-self-start">
+  {#if $slideshowState === SlideshowState.None && showNavigation}
+    <div class="z-[1001] column-span-1 col-start-1 row-span-1 row-start-2 mb-[60px] justify-self-start">
```
jrasm91, line 516: z-indexes have been a mess, especially on this component Why exactly does this one need to be changed to 1001? Have you verified that all context menus, modals, buttons, etc. continue to work as expected after changing the z-index?\

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -459,19 +506,56 @@
...
       <NavigationArea on:click={navigateAssetBackward}><Icon path={mdiChevronLeft} size="36" /></NavigationArea>
     </div>
   {/if}
+
   <!-- Asset Viewer -->
-  <div class="col-span-4 col-start-1 row-span-full row-start-1">
+  <div class="z-[1000] col-start-1 col-span-4 row-start-1 row-span-full" bind:this={assetViewerHtmlElement}>
```
jrasm91, line 522: Again with the z-index.

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -459,19 +506,56 @@
...
+    {#if $slideshowState != SlideshowState.None}
+      <!-- Slideshow actions bar -->
+      <div class="z-[1000] absolute w-full flex">
```
jrasm91, line 525: And here.

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -459,19 +506,56 @@
...
+          <CircleIconButton icon={mdiChevronRight} on:click={navigateAssetForward} title="Next" />
+        </div>
+        <ProgressBar
+          autoplay
+          bind:this={progressBar}
+          bind:status={progressBarStatus}
+          on:done={navigateAssetForward}
+          duration={5000}
+        />
+      </div>
```
jrasm91, line 556: I know the asset-viewer component is already _huge_. What do you think about moving all these elements into a separate slideshow-bar component (or similar)? Almost everything would work fine with just reading/writing to the slideshow store. The only thing that seems to be also needed is triggering the navigate asset forward and backwards. What do you think about emitting those events from the new component instead?

web/src/lib/utils/slideshow-history.ts
```diff
@@ -0,0 +1,40 @@
+export class SlideshowHistory {
+  private slideshowHistory: string[] = [];
+  private slideshowHistoryIndex = 0;
```
jrasm91, line 3: these are inside a class called `Slideshow*`, so you could probably omit `slideshow*` from the variable names.

web/src/lib/utils/slideshow-history.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  constructor(private onNavigate: (assetId: string) => void) {}
+
+  reset(assetId: string) {
+    this.slideshowHistory = [assetId];
+    this.slideshowHistoryIndex = 0;
+  }
+
+  append(assetId: string) {
```
jrasm91, line 12: I think `queue` might make more sense. I know in a playlist you can queue items to play after the next one finishes. This seems like a similar concept, basically a playlist.

web/src/lib/utils/slideshow-history.ts
```diff
@@ -0,0 +1,40 @@
+export class SlideshowHistory {
+  private slideshowHistory: string[] = [];
+  private slideshowHistoryIndex = 0;
+
+  constructor(private onNavigate: (assetId: string) => void) {}
+
+  reset(assetId: string) {
```
jrasm91, line 7: This seems odd to me. Would it be better to use `reset()` followed by `append(assetId)` instead?

web/src/lib/utils/slideshow-history.ts
```diff
@@ -0,0 +1,40 @@
...
+  append(assetId: string) {
+    this.slideshowHistory.push(assetId);
+
+    // If we were at the end of the slideshow history, move the index to the new end
+    if (this.slideshowHistoryIndex === this.slideshowHistory.length - 2) {
+      this.slideshowHistoryIndex++;
+    }
+  }
+
+  navigateForward(): boolean {
```
jrasm91, line 21: Going along the lines of the playlist concept, I think it would be more appropriate to name this next and previous. What do you think?

jrasm91 (commented):
Thanks for working on this. Left a few comments. Please ignore the z-index stuff if I did that :-) I can't remember anymore.

web/src/lib/utils/slideshow-history.ts
brighteyed: I like this analogy with the playlist

web/src/lib/components/asset-viewer/asset-viewer.svelte
brighteyed: Initially, an automatic subscription/unsubscription seemed a good idea. However, I encountered two issues that I managed to solve only through manual subscription/unsubscription in `onMount`

### Issue 1:
When we call the `Slideshow` command from an album's page (`setAssetId`, followed by `$slideshowState = SlideshowState.PlaySlideshow`) an automatic subscription fires too early because in `handlePlaySlideshow` we need an `assetViewerHtmlElement` to be loaded. As a result we can't enter fullscreen mode. The manual subscription to the `slideshowState` in the `onMount` guarantees that the `assetViewerHtmlElement` is loaded before the subscription is fired.

### Issue 2:
Unexpectedly (at least for me), if I add `$: $slideshowShuffle, handleShuffleChange` then the callback is called on each `setAssetId` call. As a result, the history gets reset on each asset change. This issue was also resolved by using a manual subscription to `slideshowShuffle` in `onMount`.

Maybe there are other ways to solve these issues in svelte but I couldn't find them.

jrasm91: OK, sounds good.

web/src/lib/utils/slideshow-history.ts
jrasm91: I like the slideshow changes, maybe it would make sense to re-name this file/class to `Slideshow` which itself has a history, so you have `slideshow.history` instead of `slideshowHistory.history`.

brighteyed:
> Thanks for working on this. Left a few comments. Please ignore the z-index stuff if I did that :-) I can't remember anymore.

Thank you for the detailed review! I believe z-index manipulations are mostly needed for the `VideoViewer` component. For some reason, without them, the player prevents users from interacting with buttons on top of it. The buttons are visible, but hovering and clicking do not work (for example, see #2946).
I double checked everything z-index related and tested it thoroughly and everything seems fine

web/src/lib/components/asset-viewer/asset-viewer.svelte
brighteyed: Yes, it is indeed huge. In this PR, I've decided to group the viewer area and navigation controls (next/prev navigation buttons, slideshow bar) together. I have a plan to try to extract the "Viewer with basic navigation" component  in future PRs, leaving the top action navigation bar and details panel in the main component. 
I agree, it would be good to extract the "Slideshow bar" component in this PR. I'll try to do it later today

jrasm91:
Everything looks good to me, but I can't approve it because it is my PR. Maybe you can approve it instead @brighteyed ?

brighteyed:
> Everything looks good to me, but I can't approve it because it is my PR. Maybe you can approve it instead @brighteyed ?

I've just pushed the new `SlideshowBar` component. If you like the implementation, please ping me again and I'll approve the PR then

jrasm91:
Looks good to me.

brighteyed (approved):

natedawg:
I just tried this on the latest update and I'm getting a lot of repeat images. I have about 3000 photos in immich and the same 20-30 photos keep showing up on the slideshow. Even if I restart the slideshow from a new image the same photos keep appearing. Sometime the same photo comes up two times out of three.

brighteyed:
> I just tried this on the latest update and I'm getting a lot of repeat images. I have about 3000 photos in immich and the same 20-30 photos keep showing up on the slideshow. Even if I restart the slideshow from a new image the same photos keep appearing. Sometime the same photo comes up two times out of three.

The code randomly selects a month (bucket) and then randomly selects an asset from that month (bucket). This approach is not truly uniform because it doesn't take into account the number of assets in each month. Do you have a lot of months with only a few assets?

natedawg:
Ahh that is the issue.  Below is what my timeline looks like. The shuffle shows hardly anything from 2023 and 2022. Is there no way to pull a list of assets, shuffle and then lopp through them? Would be good to not show any duplicates until all assets have are viewed. 
image

brighteyed:
> Ahh that is the issue. Below is what my timeline looks like. The shuffle shows hardly anything from 2023 and 2022. Is there no way to pull a list of assets, shuffle and then lopp through them? Would be good to not show any duplicates until all assets are viewed.

There is a way to implement a uniform randomization without preloading all the assets. We might store an additional `count` field for each bucket and use it to implement a fair uniform randomization without performance issues. I'll try to implement it.

jrasm91:
I believe the slideshow component only runs on pages that use the asset grid. Maybe we can do a weighted sampling of assets based on the distribution in the time buckets. We should have available that "stats" (pictures per bucket) to take into account.

natedawg:
Maybe use a deterministic algorithm instead of `random` for the shuffle. This way duplicates could be avoided. The algorithm could generate the order and when given the current sequence number would know the previous and next sequence numbers too. The time the slideshow started could be used as the seed.

PR #4284, opened 2023-09-30
feat(web): focus search bar input when using explore shortcut "/"

daniele-athome:
This PR will focus the search bar when using the explore (actually search) shortcut as in Google Photos.

The code needs a serious review because I have little to no experience with Svelte - I started exploring it with Immich - and I don't know if this is the right approach.
Mainly it's about communicating from asset-grid (which receives the keydown event) to search-bar which involves traversing the component tree in a weird way. Maybe there should be some refactoring, i.e. "global" shortcuts like this one should be handled in a proper parent (very high in the tree) component?
If someone would point me to the right direction I'll be glad to implement this in a different way.

### How was this tested?

- [x] clicking "Explore" link normally does not focus the search bar
- [x] typing "/" from an asset grid will focus the search bar after navigating into the explore page

### DIFF
--- a/web/src/lib/components/layouts/user-page-layout.svelte
+++ b/web/src/lib/components/layouts/user-page-layout.svelte
@@ -8,13 +8,25 @@
   export let showUploadButton = false;
   export let title: string | undefined = undefined;
   export let scrollbar = true;
+  export function focusSearchInput() {
+    if (navigationBar) {
+      navigationBar.focusSearchInput();
+    }
+  }
+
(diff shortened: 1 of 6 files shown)


jrasm91 (commented):
There is probably a way to emit a custom event, then listen for it on the window in the explore page.

Events are probably the best way to broadcast events between components or pages that don't have a direct parent/child relationship.

danieldietzler:
#8630

PR #4293, opened 2023-10-01
feat(web, mobile): Options to show archived assets in map

danieldietzler:
Finishing up #4262.

Screenshot 2023-10-01 at 15 39 09


jrasm91 (approved):
Thanks for finishing this!

Maybe you can write an e2e test for archived map assets and this option?

jrasm91:
@alextran1502 would you mind taking a look at the mobile part and validating it?

PR #4294, opened 2023-10-01
feat: Maplibre

danieldietzler:
Don't mind this just yet. @jrasm91 and I are testing out maplibre.

Screenshot 2023-10-18 at 22 55 28


server/src/domain/system-config/system-config.core.ts
```diff
@@ -80,7 +80,10 @@ export const defaults = Object.freeze<SystemConfig>({
   },
   map: {
     enabled: true,
-    tileUrl: 'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
+    styles: [
+      { name: 'Immich Light', url: 'http://localhost:3001/api/system-config/map-light', default: 'light' },
+      { name: 'Immich Dark', url: 'http://localhost:3001/api/system-config/map-dark', default: 'dark' },
```
bo0tzz, line 85: I think using a relative URL here (`/api/...`) should work.

web/src/routes/(user)/map/+page.svelte
```diff
@@ -43,6 +40,39 @@
...
+      case 'light':
+        style = await getLightStyle();
+        break;
+    }
+  }
+
+  async function getLightStyle() {
+    const url = $serverConfig.mapStyles.find((style) => style.default === 'light')?.url;
+    if (url) {
+      const { data } = await axios.get<StyleSpecification>(url);
```
bo0tzz, line 57: Do we need to fetch the data or can we just pass maplibre the style URL?

danieldietzler: Oh I think you're right, totally missed that. Thanks!

web/src/lib/components/shared-components/map/map.svelte
```diff
@@ -0,0 +1,107 @@
...
+  function getLightStyle() {
+    return $serverConfig.mapStyles.find((style) => style.theme === 'light')?.url || styleUrl;
+  }
+
+  function getDarkStyle() {
+    return $serverConfig.mapStyles.find((style) => style.theme === 'dark')?.url || styleUrl;
+  }
+</script>
+
+<!-- Re-visit this line later. There is a potential PR that may fix this and allow updates without re-rendering the whole component -->
```
bo0tzz, line 49: Please add a link to the referenced PR here

danieldietzler: That was more of like a personal note. Actually, that should be fixed in the next days, because there is already a release (https://github.com/dimfeld/svelte-maplibre/releases/tag/v0.5.0). Or I just upgrade manually... 🤔

danieldietzler: Done ;)

server/src/domain/map/map-styles.dto.ts
```diff
@@ -0,0 +1,20 @@
...
+  @IsString()
+  name!: string;
+
+  @IsString()
+  url!: string;
+
+  @IsEnum(MapTheme)
+  @IsOptional()
+  @ApiProperty({ type: String, enum: MapTheme, enumName: 'MapTheme', required: false })
+  theme!: 'light' | 'dark' | undefined;
```
jrasm91, line 19: This is not required? Then it should be `theme?: MapTheme`

server/src/domain/server-info/server-info.service.spec.ts
```diff
@@ -169,7 +169,10 @@ describe(ServerInfoService.name, () => {
         loginPageMessage: '',
         oauthButtonText: 'Login with OAuth',
         trashDays: 30,
-        mapTileUrl: 'https://tile.openstreetmap.org/{z}/{x}/{y}.png',
+        mapStyles: [
+          { name: 'Immich Light', theme: 'light', url: '/api/system-config/map-light' },
```
jrasm91, line 173: I think this url should not include `/api` and that part should come from the client.

server/src/domain/system-config/system-config.service.ts
```diff
@@ -44,6 +44,7 @@ export class SystemConfigService {
     const config = await this.core.updateConfig(dto);
     await this.jobRepository.queue({ name: JobName.SYSTEM_CONFIG_CHANGE });
     this.communicationRepository.broadcast(CommunicationEvent.CONFIG_UPDATE, {});
+    console.log(mapConfig(config).map.styles);
```
jrasm91, line 47: console.log

jrasm91 (commented):
I feel like there is a decent amount of complexity added to this by supporting a variable array of style urls, and I kind of doubt somebody is going to use more than one anyway. Thoughts on just having `map.darkStyleUrl` and `map.lightStyleUrl`. They are always there and always required. No more optional chaining, filtering, etc. I think a lot of code becomes trivial with those assumptions. I also don't think we lose hardly anything by not allowing additional styles.

server/src/domain/map/map-styles.dto.ts
danieldietzler: I thought since this is also a request dto, it would need to be `theme!: ...`?

server/src/domain/system-config/system-config.service.ts
danieldietzler: Oops, missed that in testing. Thanks

server/src/domain/server-info/server-info.service.spec.ts
danieldietzler: What do you mean by "that part should come from the client"?

server/Dockerfile
```diff
@@ -50,6 +50,7 @@ RUN ldconfig /usr/local/lib
 COPY --from=prod /usr/src/app/node_modules ./node_modules
 COPY --from=prod /usr/src/app/dist ./dist
 COPY --from=prod /usr/src/app/bin ./bin
+COPY --from=prod /usr/src/app/assets ./assets
```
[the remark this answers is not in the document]
danieldietzler: Just replace the existing line with that?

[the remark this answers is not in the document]
jrasm91: I believe so.

jrasm91 (commented):
Have you mistakenly reverted the config array enhancements?

server/src/domain/system-config/system-config.core.ts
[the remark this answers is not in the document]
danieldietzler: Nope, I just screwed up the rebase. Good catch!

jrasm91 (approved):

server/assets/style-dark.json
```diff
@@ -1,6 +1,7 @@
 {
   "version": 8,
   "name": "Immich Map",
+  "metadata": { "maputnik:renderer": "mbgljs" },
```
bo0tzz, line 4: Do we need this line in the prod style?

jrasm91:
There is a bug about it not working for non-admin users. Maybe you are running into that issue.

waclaw66:
> There is a bug about it not working for non-admin users. Maybe you are running into that issue.

I checked it under my admin account.

danieldietzler:
@waclaw66 Is it possible that it just takes a long time to load and eventually it will show up? We're running into latency issues with the tiles server currently

waclaw66:
@danieldietzler it works now suddenly.

danieldietzler:
> @danieldietzler it works now suddenly.

👍🏼 Yeah then that it is likely to be the same latency issue. We're working on that, sorry!

waclaw66:
Latitude and logitude coordinates are swapped on the `Open in OpenStreetMap` link.

waclaw66:
Map style settings shown as `undefined` and modified when settings is opened.

obrazek

danieldietzler:
> Latitude and logitude coordinates are swapped on the `Open in OpenStreetMap` link.

Thanks, good catch!


> Map style settings shown as `undefined` and modified when settings is opened.

I'm not seeing this behavior. Did you change anything?
Screenshot 2023-11-10 at 09 49 12

waclaw66:
Didn't change anything, even when `map-settings` is deleted from browser settings.

PR #4310, opened 2023-10-02
fix(server): delete face thumbnails when merging people

danieldietzler:
There are a few places a person is deleted:
- Nightly job that removes people without any faces (deleted assets leaves a person without faces)
- Merge one person into another (deletes one person and re-assigns all faces to another)
- Facial recognition job > All (deletes all people and starts over)

The code would delete the thumbnail on the nightly job, but not the other two. This code extracts the person delete and thumbnail delete into a single job, which now deletes both.


jrasm91 (approved):
Nice!

PR #4313, opened 2023-10-02
feat(server): use host system timezone

etnoy:
Adding two volumes to the server and microservices containers allow them to use the host system timezone.

This is very useful for library imports, since any file without exif timezone or gps will get their timezone set to the local timezone by exiftool-vendored. This PR makes sure that the local timezone isn't just UTC, but the timezone of the host system. This follows the principle of least surprise for the user.


jrasm91 (approved):
LGTM

daniele-athome:
Wasn't it necessary to just add the `TZ` environment variable?

etnoy:
> Wasn't it necessary to just add the `TZ` environment variable?

it's better to infer the TZ from the host rather than having to set it manually in env

PR #4329, opened 2023-10-03
Fix incorrect hash value

magnww:
`hashSha1` is just a hash of the path string.
`hashFile` should be used to generate the correct hash value.


bo0tzz:
IIRC this is deliberate to allow the same file to exist in different libraries

etnoy:
Hashing of path is intentional

Issue #4228, opened 2023-09-26
[BUG] Immich on android mobile fails to upload large file


mattL485:
### The bug

Single video file on android mobile fails to upload. Android app running v1.79.1, but has been stuck on uploading this file for months. 

The video file is 2.01 GiB and is 3840x2160, 60fps, 67809kbps.

File uploads fine after transferring to pc then using cli or pc webui.

When uploading from the mobile app, the upload reaches 100% and stalls indefintely. When trying to upload from chrome on mobile, I get "AxiosError: Request failed with status code 413" in the small upload gui. Popup red error box says message below:
![Screenshot_20230925_200219_Chrome](https://github.com/immich-app/immich/assets/46937622/3e089cd3-5470-4b9b-a84c-1865a28eb49b)

### The OS that Immich Server is running on

Unraid 6.9.2

### Version of Immich Server

v1.79.1

### Version of Immich Mobile App

v1.79.1

### Platform with the issue

- [ ] Server
- [X] Web
- [X] Mobile

### Your docker-compose.yml content

```YAML
version: "3.8"

services:
  immich-server:
    container_name: immich_server
    image: ghcr.io/immich-app/immich-server:release
    entrypoint: ["/bin/sh", "./start-server.sh"]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
    env_file:
      - .env
    depends_on:
      - redis
      - database
      - typesense
    restart: always

  immich-microservices:
    container_name: immich_microservices
    image: ghcr.io/immich-app/immich-server:release
    entrypoint: ["/bin/sh", "./start-microservices.sh"]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
    env_file:
      - .env
    depends_on:
      - redis
      - database
      - typesense
    restart: always

  immich-machine-learning:
    container_name: immich_machine_learning
    image: ghcr.io/immich-app/immich-machine-learning:release
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - model-cache:/cache
    env_file:
      - .env
    restart: always

  immich-web:
    container_name: immich_web
    image: ghcr.io/immich-app/immich-web:release
    entrypoint: ["/bin/sh", "./entrypoint.sh"]
    env_file:
      - .env
    restart: always

  typesense:
    container_name: immich_typesense
    image: typesense/typesense:0.24.0
    environment:
[...]

PR #4346, opened 2023-10-04
Update nginx config docs

wipfli:
`client_max_body_size 50000M;` needs to be put in the main nginx file too, update the docs for that...


wipfli:
#4228

etnoy (approved):

etnoy:
LGTM

JeremyTBradshaw:
Thanks for this fix.  I find it odd that this need to include it in nginx.conf http block additionally to including it in the site's server block is not documented by nginx, but clearly is true.  There are lots of threads out there where people talk about it being settable in http, server or location blocks.

Thanks again.

PR #4358, opened 2023-10-05
fix: time buckets

### LINKED ISSUE #3616: [BUG] time buckets show image in different day
### The bug

I have an image taken at 18:09 on Friday (Pá) (in local time), but it's shown in Saturdays (So) bucket.
I assume the problem comes from different timezone perspective. My current timezone is +02:00. The photo has exif only, no TimeZoneOffset field or XMP is present, just time and GPS position. Timezone GMT-04:00 shown in details is probably calculated from GPS position.
When I changed the timezone on my computer to -04:00 and reloaded the page, affected image has moved correctly to Fridays bucket as expected. Local timezone shouldn't affect day buckets.

obrazek
obrazek
obrazek
obrazek

Fedora 38

### Version of Immich Server

v1.72.2

v1.72.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

```bash
see above
```

### Additional information

_No response_

jrasm91:
Fixes #3616

Basically will work the same way as before, but now, after running a migration and metadata extraction can use the "local time" instead. Changed the implementation to use a date time with time zone and changed _both_ time bucket queries to use `localDateTime`. This resulted in need to tell the UI to not adjust the time to the local timezone, but instead keep it as is.


alextran1502 (approved):

etnoy (approved):

etnoy:
My testing shows this to work nicely. Also the original user who reported the issue says it's working

etnoy:
The latest commit also contains a fix for the case when the postgres database runs in non-utc timezone, typically seen outside of a docker-compose deployment

jrasm91:
I believe we now have it working for the following situations:
- Database running in any timezone
- Server running in any timezone
- Client running in any timezone

The assets should be grouped by date taken, based on local time.

PR #4359, opened 2023-10-05
fix: migration efficiency

jrasm91:
The typeorm migration commands used to compile the source code prior to being able to connect to the database, etc. This took a long time. This PR changes that to use the already pre-compiled code in the `dist/` folder instead. This does require a successful compilation beforehand, so it should work fine as long as the server is up and running. It is _much_ faster though.


alextran1502 (approved):

etnoy:
perhaps it's also time to rebase older migrations soon?

jrasm91:
The only issue with that is everybody has to upgrade to a specific version. I think it makes sense to do it in the 1.0 release though. Or if/when we migrate to prisma :-)

PR #4366, opened 2023-10-06
refactor: library asset delete

jrasm91:
First of several PRs cleaning up the library service. 

- Don't read the whole asset, when only id is necessary to queue the deletion.
- Queue inside of the pagination


danieldietzler (approved):
Looks good!

PR #4387, opened 2023-10-07
feat(server,web): improve performances in person page (1)

martabal:
## Changes made in this PR

Currently all people are loaded on each person page. They are all loaded for the merge suggestion modal and to suggest people when typing the person name. This PR introduces the ability to search for people's names using the API, eliminating the need to load all data on the person's page. An API request is triggered whenever the person's name is modified, and it returns a maximum of 20 people. If it returns fewer results, the web client does not make new requests when the person's name changes.

## Screenshots

https://github.com/immich-app/immich/assets/74269598/0a722390-c98a-44cc-925e-ab69111f55c3


jrasm91 (approved):
I haven't tested it yet, but the code looks good to me.

PR #4392, opened 2023-10-08
fix(server): make system config core singleton

danieldietzler:
Currently, config validators wouldn't work because system config core wasn't a singleton and therefore there existed different instances of it with different validators. The instance the service held had no validators at all


jrasm91 (approved):
Thanks for working on this!

PR #4394, opened 2023-10-09
chore(server): use force instead of forceRefresh

jrasm91:
There is already a job property on the base job for `force`, so updated the library jobs to use that one.


alextran1502 (approved):
LGTM

PR #4395, opened 2023-10-09
refactor(server): filesystem crawl

jrasm91:
Consolidate tests and crawl implementation. Crawl also now always returns absolute paths..


etnoy:
LGTM

etnoy (approved):

PR #4398, opened 2023-10-09
chore(compose): Remove unnecessary env values from proxy container

bo0tzz:


jrasm91 (commented):
I think these are still used.

https://github.com/immich-app/immich/blob/main/nginx/15-set-env-variables.envsh

bo0tzz:
They are still supported, but not used or needed in the default setup, and having them there confuses users who think they need to add them to their .env file.

jrasm91:
We can remove them, but then also that may break some installs since they will no longer be passed from the `.env` file. Is that the intended behavior?

bo0tzz:
I would expect that anybody using these env vars has an unusual setup and is aware that these vars need to be passed.

jrasm91:
Sounds good

jrasm91 (approved):

martabal:
It should probably be mentioned in the next release

PR #4405, opened 2023-10-09
feat: Run postgres container as non-root

TrAyZeN:
postgres images support running as non-root user as mentionned in [1]. However, the user needs to exist. An existing user is "postgres" (uid 70), as defined in [2].

[1]: https://github.com/docker-library/docs/blob/master/postgres/README.md#arbitrary---user-notes
[2]: https://github.com/docker-library/postgres/blob/2bff0ce33b76f03f4e6986497e0ab084dad10e9d/Dockerfile-alpine.template#L4-L11

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -114,6 +114,7 @@ services:
   database:
     container_name: immich_postgres
     image: postgres:14-alpine@sha256:28407a9961e76f2d285dc6991e8e48893503cc3836a4755bbc2d40bcc272a441
+    user: 70:70
     env_file:
       - .env
     environment:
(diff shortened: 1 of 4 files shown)


bo0tzz:
Is this something that needs to be in the default Immich setup? Is it something that will break the database if people apply this change to their existing instances?

etnoy:
> Is this something that needs to be in the default Immich setup? Is it something that will break the database if people apply this change to their existing instances?

I think this makes a lot of sense, no point in running things as root. In the future we might be able to run the other containers as non-root as well

jrasm91:
I agree it should run as non root. I don't know if this is the best way to implement it.

PR #4410, opened 2023-10-09
fix(server): add original path and library id index to asset

etnoy:
This adds an index to the original asset path. Since we often do lookups on original path this should speed things up for large libraries

### DIFF
--- a/server/immich-openapi-specs.json
+++ b/server/immich-openapi-specs.json
@@ -4357,6 +4357,16 @@
       "put": {
         "operationId": "updateConfig",
         "parameters": [],
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/SystemConfigDto"
+              }
+            }
+          },
+          "required": true
+        },
         "responses": {
           "200": {
             "content": {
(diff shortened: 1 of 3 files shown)


jrasm91 (commented):
Do you still have your big instance? I assumed the index should be on library id and source path, since that is what the query uses. Were you able to validate the query before/after the index?

etnoy:
> Do you still have your big instance? I assumed the index should be on library id and source path, since that is what the query uses. Were you able to validate the query before/after the index?

250k assets, all timings made by psql

Create originalPath index: 1957ms

Select asset where originalPath='xxx'
Without index: ~133ms
With index: ~1-2ms

Create originalpath, libraryId index: 2066 ms

Select asset where originalPath='xxx' and libraryId='yyy'
Without index: ~120-140ms
With index: ~1.1-2.0 ms

So this made an incredible difference, at least when doing the raw sql call.

Profiling the library refresh is a bit harder, though.

jrasm91 (approved):
Can you revert the open API spec change (I need to fix that elsewhere), otherwise looks great!

etnoy:
> Can you revert the open API spec change (I need to fix that elsewhere), otherwise looks great!

done! Hope for a fix for that soon :)

jrasm91 (approved):

PR #4414, opened 2023-10-10
fix(server): always show people with name, ignore count

GenericGuy:
As noted in https://github.com/immich-app/immich/discussions/4407, people should always be shown when a name has been set for them. In my opinion that's true, but feel free to discuss on the linked page.


martabal:
Isn't it already the case ? Named people have the priority as shown here #3713

martabal:
Oh my bad you just want to bypass the count.

alextran1502 (approved):

PR #4418, opened 2023-10-10
fix(server): improve library scan queuing performance

etnoy:
Here, we mark library assets as offline in a single db query rather than queueing individual jobs. The performance improvement is immense. We also removed all queue preparation steps that had quadratic complexity, greatly improving performance for very large libraries.

### DIFF
--- a/server/src/domain/library/library.service.ts
+++ b/server/src/domain/library/library.service.ts
@@ -380,7 +380,7 @@ export class LibraryService {
         const existingPaths = await this.repository.getOnlineAssetPaths(job.id);
         this.logger.debug(`Found ${existingPaths.length} existing asset(s) in library ${job.id}`);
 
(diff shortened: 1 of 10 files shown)


server/src/domain/library/library.service.ts
```diff
@@ -384,17 +377,12 @@ export class LibraryService {
 
     this.logger.debug(`Found ${crawledAssetPaths.length} assets when crawling import paths ${library.importPaths}`);
     const assetsInLibrary = await this.assetRepository.getByLibraryId([job.id]);
-    const offlineAssets = assetsInLibrary.filter((asset) => !crawledAssetPaths.includes(asset.originalPath));
-    this.logger.debug(`${offlineAssets.length} assets in library are not present on disk and will be marked offline`);
+    const offlineAssetIds = assetsInLibrary
```
jrasm91, line 380: Everything looks great, except this line which can be made into O(n) instead of O(n^2).

jrasm91 (approved):

PR #4433, opened 2023-10-11
fix(server,cli): don't float promises

etnoy:
Should this be done on the web, too?

### DIFF
--- a/server/src/domain/asset/asset.service.ts
+++ b/server/src/domain/asset/asset.service.ts
@@ -272,7 +272,7 @@ export class AssetService {
       zip.addFile(originalPath, filename);
     }
 
-    zip.finalize();
+    await zip.finalize();
 
     return { stream: zip.stream };
   }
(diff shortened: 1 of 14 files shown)


server/src/domain/asset/asset.service.ts
```diff
@@ -272,7 +272,7 @@ export class AssetService {
       zip.addFile(originalPath, filename);
     }
 
-    zip.finalize();
+    await zip.finalize();
```
jrasm91, line 275: This one breaks downloading files I believe. It needs to not be awaited, so I think we need to add a catch

jrasm91 (approved):
Minus the archive one this looks good, thank you.

server/src/domain/asset/asset.service.ts
jrasm91: https://github.com/archiverjs/node-archiver/issues/476

jrasm91 (approved):

etnoy:
> Minus the archive one this looks good, thank you.

I can't even remember if I fixed this one, could you check for me?

PR #4434, opened 2023-10-11
feat(cli): refactor and add album support

etnoy:
* Deploy CLI as npm package @immich/cli
* Remove read-only asset support
* Add support for automatic album creation
* Pull supported file extensions from server
* Improved documentation
* Use SDK instead of manual calls
* Make server-info command richer

How this has been tested:
* Uploading various folders to immich
* Verifying created albums are as expected


cli/README.md
```diff
@@ -2,21 +2,27 @@ A command-line interface for interfacing with Immich
...
 
-    $ ts-node cli/src login-key https://your-immich-instance/api your-api-key
+    $ immich login-key https://your-immich-instance/api your-api-key
 
 NOTE: This will store your api key under ~/.config/immich/auth.yml
 
 Next, you can run commands:
 
-    $ ts-node cli/src server-info
+    $ immich cli/src server-info
```
jrasm91, line 21: No need to pass src/* anymore

etnoy: Can't believe I missed that one!

alextran1502:
Hello @etnoy, thank you for your continuous work on this. I just tested the PR and saw an issue about album creation.

Here is what the file systems look like

image

In `Cute` and `With Da Space` there are more assets.

Here is the command I used

image

When they get uploaded to the server, here is the result in the Albums tab

image

And here is one of the album's name

image

So I guess the album was created based on the absolute path of the file.

I expected there will be two album that get created namely `Cute` and `With Da Space` while the rest that are not in any folder would not be put into any album

etnoy:
> Hello @etnoy, thank you for your continuous work on this. I just tested the PR and saw an issue about album creation.
> 
> Here is what the file systems look like
> 
> image
> 
> In `Cute` and `With Da Space` there are more assets.
> 
> Here is the command I used
> 
> image
> 
> When they get uploaded to the server, here is the result in the Albums tab
> 
> image
> 
> And here is one of the album's name
> 
> image
> 
> So I guess the album was created based on the absolute path of the file.
> 
> I expected there will be two album that get created namely `Cute` and `With Da Space` while the rest that are not in any folder would not be put into any album

Thanks for testing. This might be a Windows issue, it works on Linux. Also, I copied the logic from the legacy CLI so the same issue is likely they're too. Will look into this later today

etnoy:
> Hello @etnoy, thank you for your continuous work on this. I just tested the PR and saw an issue about album creation.
> 
> Here is what the file systems look like
> 
> image
> 
> In `Cute` and `With Da Space` there are more assets.
> 
> Here is the command I used
> 
> image
> 
> When they get uploaded to the server, here is the result in the Albums tab
> 
> image
> 
> And here is one of the album's name
> 
> image
> 
> So I guess the album was created based on the absolute path of the file.
> 
> I expected there will be two album that get created namely `Cute` and `With Da Space` while the rest that are not in any folder would not be put into any album

I've added a check for windows, try again :)

alextran1502:
Looks good, can you help fix the documentation build error?

image

Then feel free to merge

alextran1502 (approved):

PR #4443, opened 2023-10-11
style(mobile): video indicator

shenlong-tanwen:
### Changes made in the PR

- Video file indicator has been updated in the mobile app to be more prominent
- Duration indicator is also updated to be in line with how it looks in _the app that must not be named_

#### Screenshot

Before | After
:-: | :-:
Before | After


alextran1502 (approved):

cfitzw:
Probably worthy of another PR, but something that stands out to me is the cloud state not being visible when there is a white background — see bottom left picture from the example screenshots.

PR #4456, opened 2023-10-12
chore(ci): use custom base for server image build

bo0tzz:
This should speed up all the build times significantly :)

The build files for the base image live in https://github.com/immich-app/base-images


bo0tzz:
Converted to draft as this is currently still using the `:main` tag for the base image.

jrasm91 (commented):
Still need to install tini I think

jrasm91 (approved):
I think this is good to go now. The e2e tests pass and I was able to verify (on a prod build) that an heic file thumbnail is generated correctly.

The scheduled tag won't be added until Friday. I would be OK merging this as is, testing it a bit, and then changing it to a schedule-based tag in the future.

Follow up items probably include:
- Deleting pushed PR tags for the base image
- Revisiting image size to see if there are further optimizations we could make
- Adding java as a dev-dependency to see if we can run the open api generator from inside the server container

bo0tzz:
The base image repo itself isn't building successfully atm, let's hold off on merging this until that's fixed.

PR #4464, opened 2023-10-13
fix(server,web): correctly remove metadata from shared links

### LINKED ISSUE #4382: [BUG] Shared file security bug 1.81.1

etnoy:
* Add sanitized asset response return type
* Send only minimal metadata in time bucket when not showing metadata
* e2e tests to make sure we don't get any metadata

Combine with https://github.com/immich-app/immich/pull/4438 and we'll resolve #4382


server/src/domain/asset/response-dto/asset-response.dto.ts
```diff
@@ -71,15 +74,23 @@ function _map(entity: AssetEntity, withExif: boolean): AssetResponseDto {
     isExternal: entity.isExternal,
     isOffline: entity.isOffline,
     isReadOnly: entity.isReadOnly,
+    hasMetadata: true,
   };
 }
 
-export function mapAsset(entity: AssetEntity): AssetResponseDto {
```
jrasm91, line 77: These used to be separate and it was a pain because every new property or change had to be made in two places and sometimes there were differences (on accident). Can you instead keep it as a single method with a boolean flag. The implementation can change to something like this though:

```typescript
const sanitized = { 
   ...
 };

const unsanitized = { 
  ...sanitized
  ...
 };

return withMetadata? unsanitized : sanitized;
```

server/test/e2e/shared-link.e2e-spec.ts
```diff
@@ -1,11 +1,18 @@
...
 import { db } from '@test/db';
 import { errorStub, uuidStub } from '@test/fixtures';
-import { createTestApp } from '@test/test-utils';
+import {
+  IMMICH_TEST_ASSET_PATH,
+  IMMICH_TEST_ASSET_TEMP_PATH,
+  createTestApp,
+  restoreTempFolder,
+} from '@test/test-utils';
+import * as fs from 'fs';
```
jrasm91, line 14: If you are going to us fs, can you at least import it from `fs/promises`, and ideally just import the method directly, so like `import { cp } from 'fs/promises';

server/test/e2e/shared-link.e2e-spec.ts
```diff
@@ -248,4 +258,81 @@ describe(`${PartnerController.name} (e2e)`, () => {
       expect(status).toBe(200);
     });
   });
+
+  describe('Shared link metadata', () => {
```
jrasm91, line 262: I don't know how I feel about having a dependency on library asset scanning for this e2e test. I personally would prefer to keep e2e dependency on libraries to a minimum.

jrasm91 (approved):
Looks really good!

server/test/e2e/shared-link.e2e-spec.ts
etnoy: To be honest, I've just copy-pasted the code from my library e2e tests since that was the easiest. Let me change this in another PR after this one, I'll refactor it to use an upload endpoint instead. (The format e2e tests have the same thing)

server/test/e2e/shared-link.e2e-spec.ts
etnoy: Done

server/src/domain/asset/response-dto/asset-response.dto.ts
etnoy: Done!

jrasm91: Thank you!

server/test/e2e/shared-link.e2e-spec.ts
jrasm91: Yeah that's fair. We can improve it in the future

PR #4470, opened 2023-10-13
fix(web): empty placeholders

jrasm91:
- Consolidate empty placeholder implementation to reduce duplicated code
- Use the empty placeholder component on the sharing page
- Remove margin and padding for the asset grid when the grid is empty (better alignment for the empty placeholder)

| Before | After |
| - | - |
| image |image |


danieldietzler (approved):
LGTM

PR #4474, opened 2023-10-14
feat(server): allow unassigned asset-faces

martabal:
With this PR, people can be un-assigned, it's needed by #4153 and it requires to modify the database schema.


server/src/domain/asset/response-dto/asset-response.dto.ts
```diff
@@ -39,6 +39,7 @@ export class AssetResponseDto extends SanitizedAssetResponseDto {
   exifInfo?: ExifResponseDto;
   smartInfo?: SmartInfoResponseDto;
   tags?: TagResponseDto[];
+  unassigedFaces?: AssetFaceBoxDto[];
```
danieldietzler, line 42: Typo: `unassignedFaces`

martabal: Thanks ! :smile:

jrasm91 (commented):
This looks really good so far! So we will have faces vs people in the asset dto? I wonder if they should all be sent in a separate request. It seems the asset response is getting quite large.

martabal:
Yes but just for unassigned faces. The idea is to show "x faces available" in the people section.

Then there's two possibilities:
- The client can generate thumbnails with the bounding box information for all unassigned faces. These thumbnails are used on the person-side panel, where users can edit, reassign, create new people, and unassign faces. I don't know how intensive it can be with assets with many faces on it. Disadvantage: the more faces in the asset, the more time/resources it may take
-  When a face is unassigned, the server can automatically create a thumbnail for that face. This way, when users edit faces on the asset using the person-side panel, they only need to load all thumbnails directly from the server. Disadvantage: The more unassigned faces there are, the more storage it will take on the server.

martabal:
I removed the changes I made on the asset response. I don't think a new field is necessary, everything can be put in the people field but it will have to change a little bit.

server/src/domain/person/person.service.ts
```diff
@@ -345,7 +345,9 @@ export class PersonService {
     } as const;
 
     await this.mediaRepository.resize(croppedOutput, thumbnailPath, thumbnailOptions);
-    await this.repository.update({ id: personId, thumbnailPath });
+    if (personId) {
```
jrasm91, line 348: How can person id ever not be defined/available? 

This is called with the _id_ of a person.
The job returns if the person doesn't have a face

Under what situations will personId be null and should not be returned? If this is just a type checking issue, maybe you should update it to use `person.id` instead.

jrasm91 (commented):
Doesn't look too bad actually. I thought it was going to be more complicated than this :smile:

server/src/domain/person/person.service.ts
martabal: > Under what situations will personId be null and should not be returned?

Indeed, that situation does not exist.

jrasm91 (approved):
Looks good! I did a clean install and also an upgrade and face stuff in general seems to be working fine. I merged a few people, etc. and no issues.

PR #4490, opened 2023-10-15
feat(mobile): shared-links

shenlong-tanwen:
Closes #1863 

#### Changes made in the PR

- Adds support for displaying Shared links created by the user in the Shares tab
- Actions to delete, edit and copy the shares just as the web client
- Create share links from the mobile app
- The actions list in control bottom bar is made into a scroll-able list. This will make it easier to add other buttons in the future without cluttering the UI

#### Screenshots:

Sharing page | Shared links | Landscape
:-: | :-: | :-:
sharing-page | shared-links | image

Sharing from homepage | Share in album view | Create / Edit share page
:-: | :-: | :-:
Share in main | Share in album | Create share page

>[!Note]
> The home page bottom bar will now display two share buttons instead of one. The one which says "Share" is used to create the share link and the one which says "Share To" will share the selected assets to an external application


alextran1502:
Thank you for the PR, I've done the first route of testing, and I noticed 2 things.

1. After creating the link, we should show the user the link to let them copy it, or show the user the link and also copy it to the user's clipboard for good UX.
2. Let's say I have entered my server URL when signing in as `http://10.1.15.216:2283/api`, then the share link will be copied as `http://10.1.15.216:2283/api/share/4S_cN_lJbtElYU_MgDJPPj8phifXRKNfgOJFRM12yAfErvXHlCMzmwH4Fe-gsp4Krqw` which will not render from the web because there is the `/api` after it. I think we should check if the server URL has `/api` in it to exclude it when building the link

mobile/lib/modules/shared_link/providers/shared_link.provider.dart
```diff
@@ -0,0 +1,89 @@
...
+
+class SharedLinksNotifier
+    extends StateNotifier<AsyncValue<List<SharedLinkResponseDto>>> {
+  final SharedLinkService _sharedLinkService;
+
+  SharedLinksNotifier(this._sharedLinkService) : super(const AsyncLoading()) {
+    fetchLinks();
+  }
+
+  fetchLinks() async {
```
fyfrey, line 13: could you add return types to all these functions in the file? should be `Future<void>`

mobile/lib/modules/shared_link/providers/shared_link.provider.dart
```diff
@@ -0,0 +1,89 @@
+import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:immich_mobile/modules/shared_link/services/shared_link.service.dart';
+import 'package:openapi/api.dart';
```
fyfrey, line 3: could you rework this (and the service class) so that openapi (and thus the DTOs) are only used in the service - but not in the provider? This makes the app code more independent  of the DTOs (which might change any time without someone looking at the mobile code)

mobile/lib/modules/shared_link/ui/shared_link_item.dart
```diff
@@ -0,0 +1,332 @@
...
+import 'package:fluttertoast/fluttertoast.dart';
+import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:immich_mobile/modules/search/ui/thumbnail_with_info.dart';
+import 'package:immich_mobile/modules/shared_link/providers/shared_link.provider.dart';
+import 'package:immich_mobile/routing/router.dart';
+import 'package:immich_mobile/shared/models/store.dart';
+import 'package:immich_mobile/shared/ui/confirm_dialog.dart';
+import 'package:immich_mobile/shared/ui/immich_toast.dart';
+import 'package:immich_mobile/utils/image_url_builder.dart';
+import 'package:openapi/api.dart';
```
fyfrey, line 14: please try to use custom classes instead of the openapi DTOs here as well

mobile/lib/modules/shared_link/views/shared_link_edit_page.dart
```diff
@@ -0,0 +1,337 @@
+import 'package:auto_route/auto_route.dart';
+import 'package:easy_localization/easy_localization.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:immich_mobile/modules/shared_link/providers/shared_link.provider.dart';
+import 'package:openapi/api.dart';
```
fyfrey, line 7: again, please try to avoid DTOs all across the app

mobile/lib/modules/shared_link/views/shared_link_page.dart
```diff
@@ -0,0 +1,106 @@
+import 'package:easy_localization/easy_localization.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:immich_mobile/modules/shared_link/providers/shared_link.provider.dart';
+import 'package:immich_mobile/modules/shared_link/ui/shared_link_item.dart';
+import 'package:immich_mobile/shared/ui/immich_loading_indicator.dart';
+import 'package:openapi/api.dart';
```
fyfrey, line 8: also here

mobile/lib/routing/router.dart
```diff
@@ -51,6 +53,7 @@ import 'package:immich_mobile/shared/views/app_log_detail_page.dart';
 import 'package:immich_mobile/shared/views/app_log_page.dart';
 import 'package:immich_mobile/shared/views/splash_screen.dart';
 import 'package:immich_mobile/shared/views/tab_controller_page.dart';
+import 'package:openapi/api.dart';
```
fyfrey, line 56: remove :-)

fyfrey (commented):
thank you for finally adding this feature to the app!

I'm mainly concerned with adding many for DTO usages across the codebase which make it brittle in case someone makes changes on the server side. Let's make custom POD classes instead (like `Asset`, `User` etc.)

shenlong-tanwen:
1. The create page is now updated to display the shared link on successful creation and show a toast message on failure. The inputs are disabled after successful creation of the link.

Before Creation | After Creation
:-: | :-:
Before-Creation | After-Creation

2. This has been addressed by parsing the server url from strong into a uri and constructing the url manually.

mobile/lib/modules/shared_link/providers/shared_link.provider.dart
shenlong-tanwen: Updated.

mobile/lib/modules/shared_link/providers/shared_link.provider.dart
shenlong-tanwen: I've updated the provider and service as requested.

mobile/lib/modules/shared_link/ui/shared_link_item.dart
shenlong-tanwen: Handled.

mobile/lib/modules/shared_link/views/shared_link_edit_page.dart
shenlong-tanwen: Updated.

mobile/lib/routing/router.dart
shenlong-tanwen: Removed 😅

mobile/lib/modules/shared_link/views/shared_link_page.dart
shenlong-tanwen: Handled.

shenlong-tanwen:
> thank you for finally adding this feature to the app!
> 
> I'm mainly concerned with adding many for DTO usages across the codebase which make it brittle in case someone makes changes on the server side. Let's make custom POD classes instead (like `Asset`, `User` etc.)

Thank you for the suggestion. The code certainly feels cleaner now. There are few instances where DTOs are still used like the ServerInfo handling. I'll migrate them to use data classes in a separate PR as well

fyfrey:
> Thank you for the suggestion. The code certainly feels cleaner now. There are few instances where DTOs are still used like the ServerInfo handling. I'll migrate them to use data classes in a separate PR as well

Awesome! Thank you very much for making the changes.


Migrating/removing additional DTO usage in another PR is super welcome! At some point, we could maybe add automatic checks that openapi is only used in a few allowed places to enforce that.

alextran1502:
Can you help me rebase and fix the conflict? Thanks!

shenlong-tanwen:
> Can you help me rebase and fix the conflict? Thanks!

Conflicts are resolved now and the PR is equivalent to main now

alextran1502 (approved):

Pheggas:
Thank you very much for this feature!

lukehmcc:
Awesome work on this! Beat me to it!

mobile/lib/modules/home/ui/control_bottom_app_bar.dart
```diff
@@ -51,73 +49,73 @@ class ControlBottomAppBar extends ConsumerWidget {
...
+        if (!hasLocal)
+          ControlBoxButton(
+            iconData: Icons.filter_none_rounded,
+            label: "control_bottom_app_bar_stack".tr(),
+            onPressed: enabled ? onStack : null,
+          ),
+        if (!hasRemote)
+          ControlBoxButton(
+            iconData: Icons.backup_outlined,
+            label: "Upload",
```
waclaw66, line 106: @shenlong-tanwen Could you please add translation of this button as well?

shenlong-tanwen: Sure, I'll add it in any one of my upcoming PRs :D

PR #4503, opened 2023-10-16
feat: remove assets from memories

martabal:
## Changes made in this PR

Add an option on the memory page to remove the assets from memories. It can be re-added from the asset viewer.

## Screenshots

https://github.com/immich-app/immich/assets/74269598/92432f18-e805-493e-b371-dc2c165fbbef


jrasm91:
Is this related to https://github.com/immich-app/immich/discussions/4496? If so, I think the original request was just "asset delete" from the memory viewer. I'm not sure we want to add the ability to toggle assets showing up in memories or not _per asset_.

martabal:
No it's unrelated, I want to mimic the hide memory from google photo. Indeed hiding per asset is not ideal

jrasm91:
Hmm OK. I think they have a "remove from this memory" option. That implies the memory is generated and saved in a dedicated table/format not computed "on-the-fly" all the time. I would be OK with moving to this format myself. This would allow sharing memories, removing an asset from a memory, saving a memory, deleting a memory, etc.

Also, I would be on board with adding some more memory settings (per user), which in google look something like this

image


What do you think about either of those options instead of a per asset flag?

martabal:
Indeed it would be much better.

PR #4518, opened 2023-10-17
fix(web): scrollbar offset

jrasm91:


alextran1502:
I added in some more logic and rearranged the photo page to be layout similar to other pages but without the title. Let me know how it goes with your testing

alextran1502 (approved):

waclaw66:
This causes misalignment of Photos section. Other sections are aligned with search bar.
|before|after|
|--|--|
|obrazek|obrazek|

PR #4556, opened 2023-10-20
feat(server): Two updates to dev environment

ItalyPaleAle:
1. In the `docker-compose.dev.yml` file, increased ulimits for the containers that use TS code. This was one of the reasons for failures in my Podman (on macOS/arm64) environment. It wasn't the only failure with Podman, and didn't investigate further (I switched to Docker on Linux/amd64 after), but it can still help others.

2. Added a `make dev-down` to perform a `docker-compose down` on the dev environment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,9 @@ dev:
 dev-new:
 	docker compose -f ./docker/docker-compose.dev.yml up --remove-orphans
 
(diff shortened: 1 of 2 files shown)


jrasm91 (approved):
Seems fine

PR #4568, opened 2023-10-20
fix(mobile): asset deletion state management

shenlong-tanwen:
#### Changed made in the PR

- The previous handling removed local assets from DB even if the user denied the System prompt which is undesirable. This resulted in few different scenarios where the actual state of the asset is different from the one maintained in the DB.

  Few such cases:
  - User denied system prompt, but the asset is still removed from DB since `remoteDeleted` will not be empty
  - User accepts system prompt, but the deletion is soft. If we go to trash, the indicator would still display cloud with check mark whereas it should display cloud only indicator

  These are handled in the updated handling. 

- During sync, remote deleted asset details were not removed from `ExifInfo` table which is properly cleaned now.
- Emptying trash removed local assets without asking a prompt. The state is now properly updated to mark them as not backed up instead.

### DIFF
--- a/mobile/lib/modules/trash/providers/trashed_asset.provider.dart
+++ b/mobile/lib/modules/trash/providers/trashed_asset.provider.dart
@@ -39,7 +39,7 @@ class TrashNotifier extends StateNotifier<bool> {
 
       _ref
           .read(syncServiceProvider)
-          .handleRemoteAssetRemoval(idsToRemove.map((e) => e!).toList());
+          .handleRemoteAssetRemoval(idsToRemove.cast<String>().toList());
     } catch (error, stack) {
       _log.severe("Cannot empty trash ${error.toString()}", error, stack);
     }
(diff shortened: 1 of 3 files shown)


mobile/lib/shared/providers/asset.provider.dart
```diff
@@ -93,33 +93,79 @@ class AssetNotifier extends StateNotifier<bool> {
...
+            deleteAssets
+                .where((a) => a.storage == AssetState.local)
+                .map((e) => e.id),
+          );
+
+          if (remoteDeleted.any((e) => e.isLocal)) {
+            // Force delete: Add all local assets including merged assets
+            if (force) {
+              dbIds.addAll(
+                remoteDeleted.map((e) => e.id).toList(),
```
fyfrey, line 120: toList() is unnecessary

fyfrey (commented):
Thanks, this certainly improves the delete handling + robustness. I've found two small code optimization opportunities

mobile/lib/modules/trash/providers/trashed_asset.provider.dart
```diff
@@ -28,19 +28,18 @@ class TrashNotifier extends StateNotifier<bool> {
...
+          .remoteIdProperty()
           .findAll();
 
-      await _db.writeTxn(() async {
-        await _db.exifInfos.deleteAll(dbIds);
-        await _db.assets.deleteAll(dbIds);
-      });
+      _ref
+          .read(syncServiceProvider)
+          .handleRemoteAssetRemoval(idsToRemove.map((e) => e!).toList());
```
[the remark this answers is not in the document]
shenlong-tanwen: Thanks for the tip! Updated it to cast now 😄

mobile/lib/shared/providers/asset.provider.dart
shenlong-tanwen: Removed

fyfrey (approved):

PR #4575, opened 2023-10-21
fix(web): fix #4574 link button + improves accessibility 

### LINKED ISSUE #4574: [BUG] Navigation item Administration, in header is a link that wrapps a button
### The bug

Seems like we only need a link and we can drop the button in the HTML.

Instead we need to show text "Administration" or SVG icon on sm screens.

Code that needs to be fixed: https://github.com/immich-app/immich/blob/e0fc6b753cea9195de8079ecc9df312418a282d9/web/src/lib/components/shared-components/navigation-bar/navigation-bar.svelte#L79

Any

### Version of Immich Server

Latest

Latest

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
Not relevant
```

### Your .env content

```Shell
Not relevant
```

### Reproduction steps

```bash
1. open demo on https://demo.immich.app/photos
2. open browsers developer tools 
3. check code in the header navigation that is responsible for "Administration" item (on full size desktop with no zoom)
4. check same code with narrower view (simulate mobile device or zoom in)
```

### Additional information

_No response_

BogdanCerovac:
Button in a link fixed,
horizontal rule is decorative here 
and `aria-current="page"` informs screen-readers that user is on that page when focusing on the link.

fixes #4574

### DIFF
--- a/web/src/lib/components/shared-components/navigation-bar/navigation-bar.svelte
+++ b/web/src/lib/components/shared-components/navigation-bar/navigation-bar.svelte
@@ -100,10 +100,9 @@
                       ? 'text-immich-primary dark:text-immich-dark-primary'
                       : ''}"
                   />
-                <hr
-                  role="decoration"
+                <div
                   class={$page.url.pathname.includes('/admin')
-                    ? 'border-1 mx-auto block w-2/3 border-immich-primary dark:border-immich-dark-primary'
+                    ? 'border-t-1 mx-auto block w-2/3 border-immich-primary dark:border-immich-dark-primary'
                     : 'hidden'}
                 />
               </div>
--- a/web/src/lib/components/shared-components/navigation-bar/navigation-bar.svelte
+++ b/web/src/lib/components/shared-components/navigation-bar/navigation-bar.svelte
@@ -76,37 +76,38 @@
         {/if}
 
         {#if user.isAdmin}
-          <a data-sveltekit-preload-data="hover" href={AppRoute.ADMIN_USER_MANAGEMENT}
+          <a
+            data-sveltekit-preload-data="hover"
+            href={AppRoute.ADMIN_USER_MANAGEMENT}
             aria-label="Administration"
-            aria-current={$page.url.pathname.includes('/admin')
-            ? 'page'
-            : null
-            }
+            aria-current={$page.url.pathname.includes('/admin') ? 'page' : null}
+          >
+            <div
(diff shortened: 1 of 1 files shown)


jrasm91:
Can you add to the description before/after screenshots if applicable?

BogdanCerovac:
@jrasm91 - sure, please check them below. I also had to make a temporary fix for accessibility name of the link with `aria-label` because it seems that svelte-material-icons don't have a prop to set proper role to it (should be `role="img"` (will make a PR there before we can have a permanent fix here)).


Screenshots before:

Desktop:
Desktop - button nested inside a link

Mobile:
Mobile - button nested inside a link, only icon shown

Screenshot after:

Desktop:
Desktop - link without button

Mobile:
Mobile - link without button, only icon shown

BogdanCerovac:
Opened issue for SVG missing role in https://github.com/ramiroaisen/svelte-material-icons/issues/20

web/src/lib/components/shared-components/navigation-bar/navigation-bar.svelte
```diff
@@ -76,33 +76,38 @@
...
+              </div>
+              <div class="block sm:hidden" aria-hidden="true">
+                  <Cog
+                    size="1.5em"
+                    class="dark:text-immich-dark-fg {$page.url.pathname.includes('/admin')
+                      ? 'text-immich-primary dark:text-immich-dark-primary'
+                      : ''}"
+                  />
+                <hr
+                  role="decoration"
```
jrasm91, line 104: The linting checks are complaining that a hr cannot have a role. Can you fix/address this please?

BogdanCerovac: Yes, I have a better solution on it's way.

jrasm91 (approved):

jrasm91:
I justed wanted to follow up on this - we don't use svelte-material-icons anymore and our own `Icon` component should support the `role` attribute now, which defaults to `role="img"`. Does that solve resolve the comment [here](https://github.com/immich-app/immich/pull/4575#issuecomment-1773902034)

BogdanCerovac:
@jrasm91 - yes, thank you. `role="img"` is perfect for SVGs used as images.

Now we must make sure that icons that bear information or are used instead of text in interactive elements also have their accessibility name set (as discussed in https://github.com/ramiroaisen/svelte-material-icons/issues/20#issuecomment-1779221649).

Will check and open an issue or PR...

PR #4582, opened 2023-10-21
Increase search count from 100 to 1000

cfitzw:
This is to increase the number of search results shown from 100 to 1000.

### DIFF
--- a/server/src/infra/repositories/typesense.repository.ts
+++ b/server/src/infra/repositories/typesense.repository.ts
@@ -313,8 +313,8 @@ export class TypesenseRepository implements ISearchRepository {
         {
           collection: assetSchema.name,
           q: '*',
-          vector_query: `smartInfo.clipEmbedding:([${input.join(',')}], k:100)`,
-          per_page: 100,
+          vector_query: `smartInfo.clipEmbedding:([${input.join(',')}], k:1000)`,
+          per_page: 1000,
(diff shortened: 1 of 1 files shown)


mertalev:
The problem with this is that it will always return 1000 images even if only the first few are relevant. The current result count is a balance of returning enough results without pages of random images showing up for every search.

cfitzw:
@mertalev The problem for me is my 100 results fill with all really good results. I mainly created this PR so that I could have a quick-and-dirty docker build image to use temporarily. 

It does seem like this may need more attention though, if that's the case :/

server/src/infra/repositories/typesense.repository.ts
```diff
@@ -313,8 +313,8 @@ export class TypesenseRepository implements ISearchRepository {
         {
           collection: assetSchema.name,
           q: '*',
-          vector_query: `smartInfo.clipEmbedding:([${input.join(',')}], k:100)`,
-          per_page: 100,
+          vector_query: `smartInfo.clipEmbedding:([${input.join(',')}], k:1000)`,
+          per_page: 1000,
```
jrasm91, line 317: https://typesense.org/docs/0.25.1/api/search.html#pagination-parameters max is 250

cfitzw:
Closing - moved to #4584

PR #4589, opened 2023-10-22
fix(mobile): handle non-standard exposure time string

### LINKED ISSUE #4376: [BUG] Unable to sync assets when exposureTime has non-numeric chars
### The bug

images from mobile client are successfully uploaded however the client does not update the status leaving the images marked as "not upload". this means you can't sort to albums or otherwise manage it as expected.

After increasing logging to maximum and enabling troubleshooting the following related error can be seen

```
Error while getting remote assets: FormatException: Invalid double
24 sec.

#0      double.parse (dart:core-patch/double_patch.dart:112)
#1      DurationExtension.toDouble (package:immich_mobile/utils/builtin_extensions.dart:18)
#2      _exposureTimeToSeconds (package:immich_mobile/shared/models/exif_info.dart:168)
#3      new ExifInfo.fromDto (package:immich_mobile/shared/models/exif_info.dart:57)
#4      new Asset.remote (package:immich_mobile/shared/models/asset.dart:31)
#5      MemoryService.getMemoryLane.<anonymous closure> (package:immich_mobile/modules/memories/services/memory.service.dart:39)
#6      MappedListIterable.elementAt (dart:_internal/iterable.dart:415)
#7      ListIterator.moveNext (dart:_internal/iterable.dart:344)
#8      new _GrowableList._ofEfficientLengthIterable (dart:core-patch/growable_array.dart:189)
#9      new _GrowableList.of (dart:core-patch/growable_array.dart:150)
#10     new List.of (dart:core-patch/array_patch.dart:47)
#11     SetBase.toList (dart:collection/set.dart:119)
#12     AssetService._getRemoteAssets (package:immich_mobile/shared/services/asset.service.dart:76)
<asynchronous suspension>
#13     SyncService._syncRemoteAssetsFull (package:immich_mobile/shared/services/sync.service.dart:193)
<asynchronous suspension>
#14     SyncService.syncRemoteAssetsToDb.<anonymous closure> (package:immich_mobile/shared/services/sync.service.dart:81)
<asynchronous suspension>
#15     AssetService.refreshRemoteAssets (package:immich_mobile/shared/services/asset.service.dart:42)
<asynchronous suspension>
#16     AssetNotifier.getAllAsset (package:immich_mobile/shared/providers/asset.provider.dart:53)
<asynchronous suspension>
```

arch/docker

### Version of Immich Server

1.81.1

1.80.0

### Platform with the issue

- [ ] Server
- [ ] Web
- [X] Mobile

```YAML
n/a
```

### Your .env content

```Shell
n/a
```

### Reproduction steps

```bash
1.open app
2.new images load
3 remain marked as not uploaded even after uploading.
```

### Additional information
it was partially discussed on the Discord and suggested that it is an issue with parsing numbers. numbers need to be converted from wholes to doubles aka 24 to 24.0

shenlong-tanwen:
Fixes #4376 

#### Changes made in the PR:

- Non standard `exposureTime` values used to throw exceptions during parsing resulting in sync failures. All the double conversions are made optional and exposureTime is calculated only when we receive a valid string else set to empty to prevent sync failures

### DIFF
--- a/mobile/lib/shared/models/exif_info.dart
+++ b/mobile/lib/shared/models/exif_info.dart
@@ -1,6 +1,5 @@
 import 'package:isar/isar.dart';
 import 'package:openapi/api.dart';
-import 'package:immich_mobile/utils/builtin_extensions.dart';
 
 part 'exif_info.g.dart';
 
@@ -165,7 +164,11 @@ double? _exposureTimeToSeconds(String? s) {
   }
   final parts = s.split("/");
   if (parts.length == 2) {
-    return parts[0].toDouble() / parts[1].toDouble();
+    final numerator = double.tryParse(parts[0]);
+    final denominator = double.tryParse(parts[1]);
+    if (numerator != null && denominator != null) {
+      return numerator / denominator;
+    }
   }
   return null;
 }


fyfrey:
@jrasm91 maybe we should fix this on the server side by normalizing the values there and storing as float? This would also enable meaningful search/filter options, e.g. show only images with long exposure (>1sec)

jrasm91:
That seems like a good idea to me. I'm not 100% sure if all the metadata values for this field will be able to be converted to a float. I guess we can just drop/ignore data that does not conform. So, something like (1) migration from string to float for the column, (2) add better parsing/validation in the extraction process, (3) update mobile / web app to always treat the value as a double? That is probably better than implementing this same validation in each client implementation.

shenlong-tanwen:
@fyfrey / @jrasm91  I've updated the handling to address non-standard values now so as to prevent the sync failures. Storing it as a string / seconds in the server can be handled as a different PR.

fyfrey (approved):
good solution for the time being!

PR #4590, opened 2023-10-22
feat(ml): dynamic batching

mertalev:
### Description

Models are optimized for processing inputs in batches, but until now the ML service has processed each input individually without any batching. This is particularly consequential when using acceleration devices; the overhead of moving data repeatedly to and from a device can undermine its performance. On the other hand, batching adds complexity to server code, particularly in the case of failures.

This PR aggregates request payloads within a certain timeframe into a single model input, improving performance for acceleration devices while being self-contained in the ML service. This does come at a cost: dynamic batching is less efficient than fixed batching, and in the case of CPUs, the overhead it adds outweighs the performance benefit. Because of this, it is disabled by default and can be enabled by setting `MACHINE_LEARNING_MAX_BATCH_SIZE` greater than 1.

### DIFF
--- a/machine-learning/app/config.py
+++ b/machine-learning/app/config.py
@@ -21,7 +21,7 @@ class Settings(BaseSettings):
     request_threads: int = os.cpu_count() or 4
     model_inter_op_threads: int = 1
     model_intra_op_threads: int = 2
-    max_batch_size: int = 8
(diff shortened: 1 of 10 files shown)


mertalev:
Closing as this branch is very stale. It's better to make a new PR when this feature is revisited.

PR #4591, opened 2023-10-22
refactor(mobile): server info to use data classes instead of dtos

shenlong-tanwen:
#### Changes made in the PR

- Server Info handling using DTOs from openapi are migrated to use data classes to make it easier to handle updates to the DTOs in future


fyfrey (approved):
Excellent!

alextran1502 (approved):

PR #4593, opened 2023-10-22
docs: milestones

jrasm91:
A list of accomplishments for the project.


shenlong-tanwen:
Looks great. Can we also make the release tags point to the actual release URL for each milestone?

danieldietzler:
Do we want to have dates or at least month/year of the release?

docs/src/pages/milestones.tsx
```diff
@@ -0,0 +1,407 @@
...
+import Timeline from '../components/timeline';
+
+interface Item {
+  icon: string;
+  title: string;
+  description?: string;
+  release: string;
+}
+
+const items: Item[] = [
```
danieldietzler, line 55: I know this makes it a lot more complex, but what do you think of a "Milestone" tag on PRs? Then, this list could be dynamic. (Maybe fetch milestones once per hour or something) However, getting the release would be a bit of a pain this way.

jrasm91: Maybe. There are downsides with that approach as well. Potentially not a static build anymore. I'm not sure what it would take to pre-render it, not a react expert here. I also like the icons and PRs don't have that right now. I think the little bit of manual effort to maintain this is worth the simplicity.

martabal:
Amazing ! I added the dates and the release urls

Screenshot from 2023-10-22 17-15-26

alextran1502 (approved):

danieldietzler:
The date of Immich' 1st birthday doesn't match, right? First commit: 3/3/2022, Immich turns 1: 1/27/2023

martabal:
> The date of Immich' 1st birthday doesn't match, right? First commit: 3/3/2022, Immich turns 1: 1/27/2023

#4606 should fix it

PR #4606, opened 2023-10-23
fix(docs): dates and responsive design in milestone page

martabal:
## Changes made in this PR

- Fixed dates for the first commit and Immich first birthday
- Removed date in release name if tag is not set
- Removed margins on cards for small screen


docs/src/components/timeline.tsx
```diff
@@ -10,6 +10,12 @@ export interface Item {
   release: string;
   tag?: string;
   date: Date;
+  dateType: DateType;
+}
+
+export enum DateType {
+  release = 'Release Date',
+  date = 'Date',
```
jrasm91, line 18: Can you make the enums all uppercase?

martabal: of course

jrasm91 (approved):

PR #4613, opened 2023-10-23
docs: fix shell comnad for restoring the database

eliavandini:
Hi, i was restoring a database of an earlier immich instance on a brand new server, but the command provided by the documentation kept failing with 
`Error response from daemon: network immich-app_default not found
Error: failed to start containers: immich_postgres`. 
Manually starting the network before starting the database seemed to fix the issue. I was able to reproduce the issue on my second laptop and creating the network before starting the database also fixed that.


bo0tzz:
This is a bit risky as the network name depends on the name that compose assigns to the stack, which is usually taken from the folder name. Is there a command in docker compose to create the networks without starting the stack?

jrasm91:
Would `docker compose start database` work instead?

PR #4633, opened 2023-10-25
fix(web): fix Theme Custom CSS endpoint requiring the user to be logged in as the server admin

### LINKED ISSUE #4631: [BUG] Custom CSS only works for logged-in admins
### The bug

The new cusom css feature only works for admins, who are logged in, so it doesn't impact any unauthenticated routes (login page) or for any secondary users.

N/A

### Version of Immich Server

N/A

N/A

### Platform with the issue

- [X] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1. Create custom css
2. Go to the login page or Login with a non-admin user
3. Observe in immich-web log
...
```

### Additional information

_No response_

Wingysam:
My original implementation of Custom CSS has this bug: #4631. This PR resolves that issue.

fixes #4631

### DIFF
--- a/cli/src/api/open-api/api.ts
+++ b/cli/src/api/open-api/api.ts
@@ -2732,6 +2732,12 @@ export interface SearchResponseDto {
 export interface ServerConfigDto {
     /**
      * 
+     * @type {string}
+     * @memberof ServerConfigDto
+     */
+    'customCss': string;
+    /**
+     * 
      * @type {boolean}
      * @memberof ServerConfigDto
      */
--- a/cli/src/api/open-api/api.ts
+++ b/cli/src/api/open-api/api.ts
@@ -2732,12 +2732,6 @@ export interface SearchResponseDto {
 export interface ServerConfigDto {
     /**
      * 
-     * @type {string}
-     * @memberof ServerConfigDto
-     */
-    'customCss': string;
-    /**
-     * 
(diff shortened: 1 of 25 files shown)


jrasm91 (commented):
This is _OK_, but if also means that the css is sent with all the rest of the config changes, wherever and whenever they are sent, which includes:
- To the web anytime a value changes
- To the mobile app anytime a value changes
- On each client load

I think the idea is similar, which is make it publicly available, but I think it might be better to make a separate endpoint called `GET /server-config/custom.css`. What do you think?

Wingysam:
It already has a custom endpoint for the CSS at /custom.css. This could be moved to clean up the organization or if it's necessary to access the relevant API. I'm just not sure how to get the CSS config option on the server without auth.

Wingysam:
That should be the right way to do this.

jrasm91 (approved):
This looks good. Do you want to try writing an e2e test?

Wingysam:
I did write a server test but I hadn't written an e2e test. I may be able to write one tomorrow.

jrasm91:
(specifically to test/assert that the custom theme can be requested without authentication)

Wingysam:
That should be correct?

PR #4645, opened 2023-10-25
feat(server): hardware video acceleration for Rockchip SOCs via RKMPP

fyfrey:
This PR adds a new type of hardware acceleration for Rockchip SOCs.
The approach differs from the existing options as it requires custom/patched ffmpeg binaries+libraries from the host OS (cannot put them into an image due to licensing issues).
To enable a user must select the new `hwaccel-rkmpp` in docker-compose and choose RKMPP as hardware acceleration in the settings UI.

**Some performance tests with an RK3588 SoC (4 "fast" ARM cores, 4 "slow" ARM cores, encoding acceleration for H265, HEVC, decoding acceleration for many formats):**

- 10s HEVC 4K -> HEVC FHD: 8s with hwaccel, 60s in software
- 10s H264 FHD -> HEVC FHD: 4s with hwaccel, 50s in software
- hwaccel is between 7 and 12 times faster than software transcoding


docker/hwaccel-rkmpp.yml
```diff
@@ -0,0 +1,149 @@
...
+      - video
+    devices:
+      - /dev/rga:/dev/rga
+      - /dev/dri:/dev/dri
+      - /dev/dma_heap:/dev/dma_heap
+      - /dev/mpp_service:/dev/mpp_service
+    volumes:
+      - /lib/firmware/mali_csffw.bin:/lib/firmware/mali_csffw.bin:ro
+      - /usr/bin/ffmpeg:/usr/bin/ffmpeg_mpp:ro
+      - /lib/aarch64-linux-gnu/libavdevice.so.60:/lib/aarch64-linux-gnu/libavdevice.so.60:ro
```
jrasm91, line 20: do these really need to be enumerated? Could you use a volume of `/lib/aarch64-linux-gnu:/lib/aarch64-linux-gnu` instead?

server/src/domain/media/media.util.ts
```diff
@@ -555,3 +562,58 @@ export class VAAPIConfig extends BaseHWConfig {
     return this.config.cqMode !== CQMode.ICQ || this.config.targetVideoCodec === VideoCodec.VP9;
   }
 }
+
+export class RKMPPConfig extends BaseHWConfig {
+  getOptions(videoStream: VideoStreamInfo, audioStream?: AudioStreamInfo | undefined): TranscodeOptions {
+    const options = super.getOptions(videoStream, audioStream);
+    options.ffmpegPath = 'ffmpeg_mpp';
```
jrasm91, line 569: why the need for a different path? Can you instead re-link  `/usr/bin/ffmpeg` via a volume mount instead or something like that?

jrasm91 (commented):
Are you planning do add any documentation for this setting?

docker/hwaccel-rkmpp.yml
fyfrey: yeah, I tried that first. But it segfaults :-/

mertalev (commented):
Is there a particular reason you added this instead of using `getScaling`?

docker/hwaccel-rkmpp.yml
jrasm91: Oh dang, lol. Even without the read only?

server/src/domain/media/media.util.ts
fyfrey: it allows to use the software transcoding fallback using the normal ffmpeg binary from the docker image.
the patched rockship ffmpeg uses hardware even if you just specify `-c:v hevc` without asking for hardware transcoding :-)
also it might have bugs in the hardware decoder for some videos; thus it needs the fallback to the properly working ffmpeg binary

docker/hwaccel-rkmpp.yml
fyfrey: yes :-D
I also tried to link the libraries into another folder (so it does never overwrite anything else, e.g. libc in the container) and use `LD_LIBRARY_PATH` - segfault.

server/src/domain/media/media.util.ts
jrasm91: Ah, I see. Does the user have to manually do anything to make this binary available?

fyfrey:
> Is there a particular reason you added this instead of using `getScaling`?

`getScaling` does not return the actual width and height. only 1 component and the other is -1 or -2 which is not understood by the RKMPP interface (needs explicit pixel values for resizing)

server/src/domain/media/media.util.ts
fyfrey: install ffmpeg for the rockchip SBC from a specific ubuntu ppa (mentioned in `hwaccel-rkmpp.yml`)

docker/hwaccel-rkmpp.yml
fyfrey: I just found a way that might work, using multiple `LD_LIBRARY_PATH`.
I'll try to put this into env.. and update the PR if it works

fyfrey: found a way - setting `LD_LIBRARY_PATH` dynamically in code is a bit cumbersome as fluent ffmpeg does not support setting environment variables. it works, though.

server/src/infra/repositories/media.repository.ts
```diff
@@ -70,16 +70,28 @@ export class MediaRepository implements IMediaRepository {
...
-          .on('error', (err, stdout, stderr) => {
-            this.logger.error(stderr);
-            reject(err);
-          })
-          .on('end', resolve)
-          .run();
+        const oldLdLibraryPath = process.env['LD_LIBRARY_PATH'];
+        if (options.ldLibraryPath) {
+          // fluent ffmpeg does not allow to set environment variables, so we do it manually
+          process.env['LD_LIBRARY_PATH'] = (oldLdLibraryPath || '') + options.ldLibraryPath;
```
mertalev, line 76: Is there no better way to do this 😭

Also thinking about it, this seems like it has potential for concurrency issues.

server/src/domain/media/media.util.ts
```diff
@@ -555,3 +562,59 @@ export class VAAPIConfig extends BaseHWConfig {
...
+    options.ldLibraryPath = '/lib/aarch64-linux-gnu:/lib/ffmpeg-mpp';
+    options.outputOptions.push(...this.getSizeOptions(videoStream));
+    return options;
```
mertalev, line 572: Up to preference, but I think this syntax is more idiomatic:

```
{
   ...options,
   ffmpegPath: 'ffmpeg_mpp',
   ldLibraryPath = '/lib/aarch64-linux-gnu:/lib/ffmpeg-mpp'
}
```

mertalev (commented):
I think I'd prefer to keep the library stuff outside of the server code if possible. But better would be to have a separate image for it. Is there a link for the licensing issue you mentioned? [MPP](https://github.com/rockchip-linux/mpp) is Apache 2.0 at least.

docker/hwaccel-rkmpp.yml
mertalev: Would it maybe be possible to just add a script that installs this package, without including those packages in the image? So a user could run that script in the container and have it install the necessary packages locally.

fyfrey: Installing only works on Ubuntu Jammy. Our base image is debian bookworm.
We could make a script that builds a new local image into which the binaries/libraries are copied. This would need to be run by the user after every docker pull (and is this forgotten)

fyfrey:
> I think I'd prefer to keep the library stuff outside of the server code if possible. But better would be to have a separate image for it. Is there a link for the licensing issue you mentioned? [MPP](https://github.com/rockchip-linux/mpp) is Apache 2.0 at least.

ffmpeg is LGPL and I think the patched version violates that. Found a comment from the dev team mentioning a license issue: 
https://forum.radxa.com/t/ffmpeg-rkmppenc-hardware-encoding-with-ffmpeg/13488/2

I'd rather not distribute any image with this included.

server/src/infra/repositories/media.repository.ts
fyfrey: I know, it's awful. I could raise an issue at fluent-ffmpeg. But not when it even if they add support to set environment variables.

I've read the documentation of the process env in node. I guess there is no concurrency issue because a) the env is reset synchronously without waiting for ffmpeg to run, b) each node worker has its own process env

mertalev: My thought was that we have child processes from exiftool etc. that aren't tied to Node.js, so theoretically when a child process is spawned by it, that process could see either `LD_LIBRARY_PATH` depending on timing. But I'm not sure if this is an issue in practice.

mertalev:
Ohhh, it's a copyleft thing. That does make it infeasible to ship it ourselves.

docker/hwaccel-rkmpp.yml
mertalev: If it's at all possible to keep this scoped in the container, I feel like that's the way to go. We could publish a Jammy version of the server (or maybe switch to Jammy outright?) and gate this package behind a build arg. Then a user could set that flag in their docker-compose to build the image with this package. This would be easy for users to set up, it would avoid mixing host/container dependencies, and we wouldn't be distributing problematic packages ourselves.

mertalev: Or alternatively, we could add a secondary start.sh script that makes sure this package is installed before calling Node.js. That would avoid the need to build the image locally.

server/src/infra/repositories/media.repository.ts
fyfrey: well, that child process inherits the env that exiftool had when it was started - and that env is unmodified if the node documentation is correct (see above)

docker/hwaccel-rkmpp.yml
fyfrey: properly installing the patched ffmpeg has the issue that the normal ffmpeg also uses some of the modified libraries (very undesirable!) unless we do the `LD_LIBRARY_PATH` swapping that is currently implemented. so, there is less benefit in building a custom image.

fyfrey:
> Are you planning do add any documentation for this setting?

Yes. Once we settled on the final implementation / configuration (possibly in a follow-up PR)

server/src/domain/media/media.util.ts
fyfrey: tried but it's way longer/complex with with outputOptions. so I left it as it is.

server/src/domain/media/media.util.ts
```diff
@@ -555,3 +562,59 @@ export class VAAPIConfig extends BaseHWConfig {
...
+  getSizeOptions(videoStream: VideoStreamInfo) {
+    if (this.shouldScale(videoStream)) {
+      const { width, height } = this.getSize(videoStream);
+      return [`-width ${width}`, `-height ${height}`];
+    }
+    return [];
+  }
+
+  getPresetOptions() {
+    return ['-level 153'];
```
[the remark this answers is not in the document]
fyfrey: only in the help output from the patched ffmpeg binary. it's level 5.1 (added this info to the code)

server/src/domain/media/media.util.ts
```diff
@@ -555,3 +562,59 @@ export class VAAPIConfig extends BaseHWConfig {
...
+  }
+
+  getBitrateOptions() {
+    const bitrate = this.getMaxBitrateValue();
+    if (bitrate > 0) {
+      return ['-rc_mode 3', '-quality_min 0', '-quality_max 100', `-b:v ${bitrate}${this.getBitrateUnit()}`];
+    } else {
+      // convert CQP from 51-10 to 0-100
```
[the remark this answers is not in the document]
fyfrey: good point. fixed both

fyfrey:
Thank you @mertalev and  @jrasm91 for the thorough review, feedback and suggestions! I'm now happy enough with the current state given the circumstances  (copyleft issue, missing fluent-ffmpeg env variable support)

server/src/infra/repositories/media.repository.ts
mertalev: Gotcha, it should be fine in that case. 

I'd suggest refactoring this so the ffmpeg code is in its own method and the try finally wraps around a call to it.

mertalev (approved):
Thanks for adding this! That speed-up is really impressive. I still hope there's a cleaner solution for this, but I think this implementation is okay given the limitations.

server/src/infra/repositories/media.repository.ts
mertalev: Actually never mind, it's better if this is handled in the promise so the env gets moved back sooner instead of waiting for the promise to resolve.

fyfrey: It would look like this:

```ts
    if (!options.twoPass) {
      return new Promise((resolve, reject) => {
        const oldLdLibraryPath = process.env['LD_LIBRARY_PATH'];
        if (options.ldLibraryPath) {
          // fluent ffmpeg does not allow to set environment variables, so we do it manually
          process.env['LD_LIBRARY_PATH'] = (oldLdLibraryPath || '') + options.ldLibraryPath;
        }
        try {
          this.runFfmpeg(input, output, options, resolve, reject);
        } finally {
          if (options.ldLibraryPath) {
            process.env['LD_LIBRARY_PATH'] = oldLdLibraryPath;
          }
        }
      });
    }
```

and

```ts
  runFfmpeg(
    input: string,
    output: string | Writable,
    options: TranscodeOptions,
    resolve: (value: void | PromiseLike<void>) => void,
    reject: (reason?: any) => void,
  ) {
    ffmpeg(input, { niceness: 10 })
      .setFfmpegPath(options.ffmpegPath || 'ffmpeg')
      .inputOptions(options.inputOptions)
      .outputOptions(options.outputOptions)
      .output(output)
      .on('error', (err, stdout, stderr) => {
        this.logger.error(stderr);
        reject(err);
      })
      .on('end', resolve)
      .run();
  }
```

`runFfmpeg` can't easily be reused for the other two calls, so not sure if this is actually better than the current state. maybe there's something better?

fyfrey: found something re-usable

mertalev: Oh nice, I like it

server/src/domain/media/media.util.ts
```diff
@@ -564,7 +564,10 @@ export class VAAPIConfig extends BaseHWConfig {
 }
 
 export class RKMPPConfig extends BaseHWConfig {
-  getOptions(videoStream: VideoStreamInfo, audioStream?: AudioStreamInfo | undefined): TranscodeOptions {
+  getOptions(videoStream: VideoStreamInfo, audioStream?: AudioStreamInfo): TranscodeOptions {
+    if (this.config.twoPass) {
```
mertalev, line 568: `getOptions` calls `eligibleForTwoPass`, so configs that don't support two-pass override it to return false. You should do that here too for consistency.

server/src/infra/repositories/media.repository.ts
```diff
@@ -70,16 +70,18 @@ export class MediaRepository implements IMediaRepository {
...
-          .output(output)
-          .on('error', (err, stdout, stderr) => {
-            this.logger.error(stderr);
-            reject(err);
-          })
-          .on('end', resolve)
-          .run();
+        const oldLdLibraryPath = process.env['LD_LIBRARY_PATH'];
+        if (options.ldLibraryPath) {
+          // fluent ffmpeg does not allow to set environment variables, so we do it manually
```
jrasm91, line 75: Do we need to add a colon or something when it is already set?

jrasm91 (approved):

Dark-Existed:
@fyfrey  hi, when I want to enable hwaccel-rkmpp.yml for rk3588, I hava this problem, any idea to solve this? thx!
I have installed ffmpeg from https://ppa.launchpadcontent.net/liujianfeng1994/rockchip-multimedia/ubuntu.
```
Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/lib/aarch64-linux-gnu/pulseaudio/libpulsecommon-15.99.so" to rootfs at "/lib/ffmpeg-mpp/libpulsecommon-15.99.so": open /var/lib/docker/overlay2/fa7c9bd3839088343d0aa48f6bba182c04b5ada6863f668689034e309bb7d660/merged/usr/lib/ffmpeg-mpp/libpulsecommon-15.99.so: read-only file system: unknown 
```

```
ls -l /lib/aarch64-linux-gnu/pulseaudio
total 1204
-rw-r--r-- 1 root root 497776 Feb 16  2023 libpulsecommon-15.99.so
-rw-r--r-- 1 root root 686768 Feb 16  2023 libpulsecore-15.99.so
-rw-r--r-- 1 root root  42776 Feb 16  2023 libpulsedsp.so
```

fyfrey:
@Dark-Existed I have not encountered this error during my tests. It seems like a docker issue.

Dark-Existed:
@fyfrey 
```
version: "3.8"

# Hardware acceleration for transcoding using RKMPP for Rockchip SOCs
# This is only needed if you want to use hardware acceleration for transcoding.
# Supported host OS is Ubuntu Jammy 22.04 with custom ffmpeg from ppa:liujianfeng1994/rockchip-multimedia

services:
  hwaccel:
    security_opt: # enables full access to /sys and /proc, still far better than privileged: true
      - systempaths=unconfined
      - apparmor=unconfined
    group_add:
      - video
    devices:
      - /dev/rga:/dev/rga
      - /dev/dri:/dev/dri
      - /dev/dma_heap:/dev/dma_heap
      - /dev/mpp_service:/dev/mpp_service
    volumes:
      - /usr/bin/ffmpeg:/usr/bin/ffmpeg_mpp:ro
      - /lib/aarch64-linux-gnu:/lib/ffmpeg-mpp:ro
      - /lib/aarch64-linux-gnu/libblas.so.3:/lib/ffmpeg-mpp/libblas.so.3:ro # symlink is resolved by mounting
      - /lib/aarch64-linux-gnu/liblapack.so.3:/lib/ffmpeg-mpp/liblapack.so.3:ro # symlink is resolved by mounting
      - /lib/aarch64-linux-gnu/pulseaudio/libpulsecommon-15.99.so:/lib/ffmpeg-mpp/libpulsecommon-15.99.so:ro
```

This is the lastest hwaccel-rkmpp.yml, may be this two line cause it mount twice.
```
- /lib/aarch64-linux-gnu:/lib/ffmpeg-mpp:ro
```
```
- /lib/aarch64-linux-gnu/pulseaudio/libpulsecommon-15.99.so:/lib/ffmpeg-mpp/libpulsecommon-15.99.so:ro
```
The last line should be remove? 
I delete it, the immich-microserver container still hava lib/ffmpeg-mpp/libpulsecommon-15.99.so, may work well ?

0x0013:
Hi,

I'm trying to get this working on RK3588S nodes in a kubernetes cluster, but so far not successful.

Needed to diverge from the steps in dockerfile:

1. Had a similar error as [above](https://github.com/immich-app/immich/pull/4645#issuecomment-1862623496) when starting the pod. Maybe my containerd runtime doesn't like that `/lib/ffmpeg-mpp` is mounted read-only and can't place another mount in there. I removed the `libpulsecommon-15.99.so` mount and created the corresponding symlink on hosts.
2. For testing I set the container to run in privileged mode. AFAIK Kubernetes doesn't have an option to have unconfined `sysfs` access, and only unmasked `procfs` didn't work. So far I was unable to find a more restrictive security setting which would work.
3. Relevant devices exposed with [generic-device-plugin](https://github.com/squat/generic-device-plugin).

With that, I am able to `exec` into the container and successfully convert manually using a command similar to that generated by immich:
```
LD_LIBRARY_PATH=/lib/aarch64-linux-gnu:/lib/ffmpeg-mpp ffmpeg_mpp -i /mnt/ext/video/testvideo.mp4 -c:v hevc_rkmpp_encoder -c:a libopus -movflags faststart -fps_mode passthrough -map 0:1 -map 0:0 -g 256 -v verbose -level 153 -rc_mode 2 -quality_min 73 -quality_max 73 test.mkv
```

This succeeds. However, when the process is initiated by immich, it still fails:
```
immich-microservices-79c67f8f6c-npmz7 immich-microservices [Nest] 8  - 01/01/2024, 8:24:21 PMLOG [MediaService] Start encoding video cc39b8db-f66c-46e2-bb80-4127c8513e6c {"inputOptions":[],"outputOptions":["-c:v hevc_rkmpp_encoder","-c:a libopus","-movflags faststart","-fps_mode passthrough","-map 0:0","-map 0:1","-g 256","-v verbose","-level 153","-rc_mode 2","-quality_min 73","-quality_max 73"],"twoPass":false,"ffmpegPath":"ffmpeg_mpp","ldLibraryPath":"/lib/aarch64-linux-gnu:/lib/ffmpeg-mpp"}
immich-microservices-79c67f8f6c-npmz7 immich-microservices [Nest] 8  - 01/01/2024, 8:24:21 PM   ERROR [MediaRepository] Error: ffmpeg exited with code 127
immich-microservices-79c67f8f6c-npmz7 immich-microservices [Nest] 8  - 01/01/2024, 8:24:21 PM   ERROR [MediaService] Error: ffmpeg exited with code 127
immich-microservices-79c67f8f6c-npmz7 immich-microservices [Nest] 8  - 01/01/2024, 8:24:21 PM   ERROR [MediaService] Error occurred during transcoding. Retrying with RKMPP acceleration disabled.
```
The fallback software encoding then succeeds (after a long time).

Code 127 should mean "command not found", but I'm puzzled as to why it fails if it succeeds when calling it manually from shell. `ffmpeg_mpp` is mounted from host in `/usr/bin/` and even manually calling it, I do not use the full path, so it's apparently in `$PATH` as it should be.
@fyfrey any ideas what could be the issue?

fyfrey:
@0x0013 this is a tough one since it runs fine manually in the container...

can you try in docker (and not kubernetes)?

0x0013:
@fyfrey I partially managed to get it to work. It seems the issue somehow stems from `LD_LIBRARY_PATH`. 

I was trying to compare the environment of my manual shell command `ffmpeg_mpp`, and what is run by immich. But due to immediate error, I could only see the standard (integrated) `ffmpeg` command. I noticed this in environment variables - `LD_LIBRARY_PATH=undefined`.

This prompted me to try adding the variable when running the container:
```
    env:
    - name: LD_LIBRARY_PATH
      value: /lib/aarch64-linux-gnu:/lib/ffmpeg-mpp
```
and with this, it is encoding correctly with RKMPP, though I see the command is being run with double paths (second set added by immich) - `LD_LIBRARY_PATH=/lib/aarch64-linux-gnu:/lib/ffmpeg-mpp:/lib/aarch64-linux-gnu:/lib/ffmpeg-mpp`..

I wonder if the library paths constructed end up as something like `undefined:/lib/aarch64-linux-gnu:/lib/ffmpeg-mpp` and maybe that's why it would fail?

I'm not sure if adding this path manually for the whole container is a good idea - could this have bad impact elsewhere?

mertalev:
Setting that env globally means the libraries in it take precedence over libraries elsewhere, such as /lib and /usr/lib. Libraries from those folders will still get loaded if something can't be found, but it can theoretically cause issues if it finds a library with a version or build that isn't actually compatible.

0x0013:
@fyfrey Thanks for your help. I did some more testing:

I already hadn't defined `LD_LIBRARY_PATH` variable before. I tested again without setting it, and the first (rkmpp accelerated) process fails with the same error. Then I check the environment for the running fallback `ffmpeg` command, and see `LD_LIBRARY_PATH=undefined`. Is it possible that, when reading existing env variable
```
const oldLdLibraryPath = process.env.LD_LIBRARY_PATH;
```
code would mistakenly interpret unset value as undefined, which would then later be set as string when restoring `oldLdLibraryPath`? (sorry, I don't have experience with TypeScript, this is just a guess)

Then, while testing with the workaround I know is working (globally set `LD_LIBRARY_PATH`), I noticed something interesting - first time `ffmpeg_mpp` runs, I see just the library path I set globally - `LD_LIBRARY_PATH=/lib/aarch64-linux-gnu:/lib/ffmpeg-mpp`. On the second and following transcodings, I see double path - `LD_LIBRARY_PATH=/lib/aarch64-linux-gnu:/lib/ffmpeg-mpp:/lib/aarch64-linux-gnu:/lib/ffmpeg-mpp`.

This made me think - maybe the path doesn't get added on first run? I never tested this, because software encoding runs so slow, I would change settings and try something else, before it had time to retry a second time. So this time, I set the path globally to a random empty directory `LD_LIBRARY_PATH=/home/node`, just for testing, and reran the container. First run, `ffmpeg_mpp` fails just as before, and fallbacks to sofware encoding. I let it run and after a while when the first fallback process is done, the new `ffmpeg_mpp` processes run successfully:

```
immich-microservices-74685f4bf9-p8hrv immich-microservices [Nest] 7  - 01/03/2024, 2:42:37 PMLOG [MediaService] Start encoding video 78025895-9811-4f06-b121-0b1db66c4e7e {"inputOptions":[],"outputOptions":["-c:v hevc_rkmpp_encoder","-c:a libopus","-movflags faststart","-fps_mode passthrough","-map 0:0","-map 0:1","-g 256","-v verbose","-level 153","-rc_mode 2","-quality_min 73","-quality_max 73"],"twoPass":false,"ffmpegPath":"ffmpeg_mpp","ldLibraryPath":"/lib/aarch64-linux-gnu:/lib/ffmpeg-mpp"}
immich-microservices-74685f4bf9-p8hrv immich-microservices [Nest] 7  - 01/03/2024, 2:43:02 PMLOG [MediaService] Encoding success 78025895-9811-4f06-b121-0b1db66c4e7e
```

I didn't test it without setting anything for `LD_LIBRARY_PATH` and then waiting for one fallback, this was just with a random (non-library) path.

mertalev:
Your guess is right. The issue is that reading an unset env from `process.env` returns undefined, but assigning undefined to `process.env` casts to a string.

PR #4655, opened 2023-10-26
feat(server,web,mobile): Add optional password option for share links.

jarvis2f:
## Introduction

This Pull Request aims to introduce password protection for shared links. In this Pull Request, I have implemented the following functionalities:

- An option to use a password when creating shared links in both web and mobile applications.
- A prompt for users to enter the password when attempting to access a password-protected shared link through the web application.

## Testing

I have added new test cases to ensure the correct functioning of these changes. I have manually tested these changes in both locally environments, everything is running well.

## Screenshots

image
image
image


jrasm91 (commented):
This only requires the password for the initial shared link metadata. Is that the intended behavior?

jarvis2f:
> This only requires the password for the initial shared link metadata. Is that the intended behavior?

If you enter a password when creating a shared link, others will be asked to enter the password when accessing the link.

martabal:
The passwords are stored in plaintext ?

jarvis2f:
> The passwords are stored in plaintext ?

I think you may need to recheck the password you set when using it, so the password is not encrypted and stored. And these passwords are usually set relatively simple.

jrasm91:
I think passwords in plaintext in this situation is fine. The security requirements for them is not the same as say a user's login.

I think this looks like a good implementation. It adds passwords as an additional check, in addition to the shared key which is how it works today.

server/src/domain/shared-link/shared-link.service.ts
```diff
@@ -32,6 +32,10 @@ export class SharedLinkService {
 
     const sharedLink = await this.findOrFail(authUser, id);
 
+    if (sharedLink.password && sharedLink.password !== dto.password) {
+      throw new BadRequestException('Invalid password');
```
jrasm91, line 36: Can you change this to an unauthorized exception? That seems more appropriate.

web/src/lib/components/shared-components/create-share-link-modal/create-shared-link-modal.svelte
```diff
@@ -40,7 +41,10 @@
       if (editingLink.description) {
         description = editingLink.description;
       }
-      allowUpload = editingLink.allowUpload;
+      if (editingLink.password) {
+        password = editingLink.password;
+      }
+      if (editingLink) allowUpload = editingLink.allowUpload;
```
jrasm91, line 47: Can you add braces here? I need to update the linting rules for this.

web/src/routes/(user)/share/[key]/+page.svelte
```diff
@@ -1,20 +1,74 @@
...
+  let { sharedLink, passwordRequired, sharedLinkKey: key } = data;
 
-  let isOwned = data.user ? data.user.id === sharedLink.userId : false;
+  let isOwned = !passwordRequired && data.user ? data.user.id === sharedLink?.userId : false;
+  let password = '';
+
+  const handlePasswordSubmit = async () => {
+    try {
+      const result = await api.sharedLinkApi.getMySharedLink({ password, key });
+      if (result.data) {
```
Comment (jrasm91) [inline web/src/routes/(user)/share/[key]/+page.svelte:21]: I think this will always have data if the request doesn't throw an error. We don't check this anywhere else.

jrasm91 (commented):
Looks good so far. Thanks for working on this. Would you be able to add some e2e tests for shared links with password?

bo0tzz:
Point of clarification: Is the password only checked when loading the shared link metadata (ie the list of assets), or also when loading the assets themselves?

jrasm91:
Only when loading the initial metadata.

shenlong-tanwen:
You don't need to add translation text in all the language files for the mobile app. Adding them to en-US should be enough. You can remove the keys for now from all the other language files.

jarvis2f:
Thank you for your replies and suggestions. I have modified and submitted it. Thank you for your help.

jrasm91 (approved):
Code looks good from my side. Do you have to type in the password every time you refresh the page? Do we want to do anything about that?

jarvis2f:
> Code looks good from my side. Do you have to type in the password every time you refresh the page? Do we want to do anything about that?

Should we store the successfully verified password in local storage, take it out from local storage when refreshing the page, and then verify it again.

shenlong-tanwen:
Apologies for the nit, but can you also remove the additional new line that got added to the language files? This would bring down the number of files updated in the change-set to half 😅

jarvis2f:
> Apologies for the nit, but can you also remove the additional new line that got added to the language files? This would bring down the number of files updated in the change-set to half 😅

What should I do? I didn't add additional new line,I just deleted them.

jrasm91:
It looks like a whitespace issue. I would probably recommend rebasing, switching the changes into a single commit, undoing the edit in the first place and then force pushing.

jarvis2f:
Thanks, I have deleted unnecessary submissions and forced push. I optimized the problem of requiring password re-entering when the page is refreshed. After the password verification is passed, I encrypt the (id-password) combination using sha256 and then store it in a cookie so that the verified status can be maintained when the page is refreshed.

web/src/routes/(user)/share/[key]/+page.svelte
```diff
@@ -1,20 +1,77 @@
...
+  const handlePasswordSubmit = async () => {
+    try {
+      const result = await api.sharedLinkApi.getMySharedLink({ password, key });
+      passwordRequired = false;
+      sharedLink = result.data;
+      isOwned = data.user ? data.user.id === sharedLink.userId : false;
+      document.title = (sharedLink.album ? sharedLink.album.albumName : 'Public Share') + ' - Immich';
+      document
+        .querySelector('meta[name="description"]')
+        ?.setAttribute('content', sharedLink.description || `${sharedLink.assets.length} shared photos & videos.`);
```
Comment (jrasm91) [inline web/src/routes/(user)/share/[key]/+page.svelte:27]: This isn't the svelte way to do this.

web/src/routes/(user)/share/[key]/+page.svelte
```diff
@@ -1,20 +1,77 @@
...
+      document.cookie = `immich_shared_link_token=${sharedLink.token}; path=/; max-age=86400; samesite=Lax`;
```
Comment (jrasm91) [inline web/src/routes/(user)/share/[key]/+page.svelte:28]: This could be an http only cookie set by the server instead.

Comment (jarvis2f) [reply web/src/routes/(user)/share/[key]/+page.svelte:28]: Should it be set on the server？

Comment (jrasm91) [reply web/src/routes/(user)/share/[key]/+page.svelte:28]: Yes.

jrasm91:
Thanks for working on this @jarvis2f!

PR #4661, opened 2023-10-26
fix(server): config update queue

jrasm91:
Storage config is still processed by the migration queue, but more appropriate is the background queue. Additionally, it now gets a priority to make sure it is picked up ahead and any other jobs that might be in the queue.


danieldietzler (approved):
Finally, nice!

PR #4673, opened 2023-10-28
fix(web): table headers when there's no album

martabal:
## Changes made in this PR

When a user has no album and has selected the List mode, the table headers are displayed, but the table itself remains empty which looks weird.

## Screenshots

| Before | After |
| :---: | :---: |
| Screenshot from 2023-10-28 03-16-58 | Screenshot from 2023-10-28 03-15-08 |

### DIFF
--- a/web/src/routes/(user)/albums/+page.svelte
+++ b/web/src/routes/(user)/albums/+page.svelte
@@ -248,71 +248,75 @@
       </div>
     </LinkButton>
   </div>
-
-  <!-- Album Card -->
-  {#if $albumViewSettings.view === AlbumViewMode.Cover}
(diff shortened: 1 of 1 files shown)


alextran1502:
Hello, what is the different in the before-after view for one-album scenario? I cannot spot the differences

martabal:
> Hello, what is the different in the before-after view for one-album scenario? I cannot spot the differences

The goal of this PR is to remove this when users have no albums.

Screenshot from 2023-10-28 03-16-58-2

danieldietzler (approved):
Looks to me like an obvious thing that should have been done earlier :D

PR #4678, opened 2023-10-28
fix(server/oauth): Handle errors from OAuth Discovery

### LINKED ISSUE #4675: [BUG] oauthAutoLaunch can sometimes result in an AggregateError if oauth provider is not reacheable
### The bug

It seems that when the oAuth provider is either unreacheable or it's DNS doesn't solve, it throws multiple errors to the stack, which do get somewhat handled by exceptionsHandler but as it's an AggregateError, it doesn't show anything really useful in logs.

Kubernetes

### Version of Immich Server

v1.82.1

v1.82.0

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
not relevant
```

### Your .env content

```Shell
not really relevant
```

### Reproduction steps

```bash
1. Setup oauth with autoredirect
2. Somehow mess with DNS so it's not available anymore or turn your oauth provider off
3. See the issue
```

### Additional information

_No response_

samip5:
Also remove non-needed await.

Closes #4675

### DIFF
--- a/server/src/domain/auth/auth.service.ts
+++ b/server/src/domain/auth/auth.service.ts
@@ -221,7 +221,7 @@ export class AuthService {
     }
 
     const client = await this.getOAuthClient(config);
-    const url = await client.authorizationUrl({
+    const url = client.authorizationUrl({
       redirect_uri: this.normalize(config, dto.redirectUri),
       scope: config.oauth.scope,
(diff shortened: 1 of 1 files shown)


jrasm91 (approved):
Lgtm

PR #4682, opened 2023-10-28
feat(web,server): activity

martabal:
## Changes made in this PR

This PR introduces Activity. 

On shared albums, you can add comments and likes to assets. They can be found in the Activity tab, by clicking on the comment icon on the bottom right.

Comments can be deleted by the authors, the owner of the album.

There's a lot related to this new feature which will come in future PRs such as notifications to be alerted on new reactions (likes & comments), a global activity tab for each shared album, nested comments...

Related FR: #1660

## Screenshots

https://github.com/immich-app/immich/assets/74269598/9b5dafb5-aac8-4584-b5f5-fbddf3147623


jrasm91 (commented):
Can you make the activity controller more restful please? You should never have two named parameters immediately adjacent to each other. Ideally just resource/:id

martabal:
> Can you make the activity controller more restful please? You should never have two named parameters immediately adjacent to each other. Ideally just resource/:id

Should I set the assetId or the albumId in the Body ?

jrasm91:
When you create a new item via post, yes those should be in the body.

When you query for them they should be query params.

If album id is required on every single route, you should consider changing the endpoints to be something like

/album/:id/activity/:activityId and call it AlbumActivity instead.

martabal:
> When you create a new item via post, yes those should be in the body.
> 
> When you query for them they should be query params.
> 
> If album id is required on every single route, you should consider changing the endpoints to be something like
> 
> /album/:id/activity/:activityId and call it AlbumActivity instead.

Thanks for the clarification, I appreciate it

server/src/domain/activity/activity.service.ts
```diff
@@ -0,0 +1,91 @@
...
+  async updateLikeStatus(authUser: AuthUserDto, dto: ActivityLikeDto): Promise<ActivityLikeStatusResponseDto> {
+    await this.access.requirePermission(authUser, Permission.ACTIVITY_CREATE, dto.albumId);
+
+    const options = {
+      userId: authUser.id,
+      albumId: dto.albumId,
+      assetId: dto.assetId,
+    };
+
+    const [reaction] = await this.repository.search(options);
```
jrasm91, line 60: I think this needs to be reactions, and then reaction is find of one that isn't a comment. Or, include isLiked in the query. Missed this cleaning up earlier.

server/src/domain/activity/activity.service.ts
```diff
@@ -0,0 +1,91 @@
...
+        assetId: dto.assetId,
+        userId: authUser.id,
+        albumId: dto.albumId,
+        comment: dto.comment,
+      })
+      .then(mapActivity);
+  }
+
+  async deleteComment(authUser: AuthUserDto, id: string): Promise<void> {
+    await this.access.requirePermission(authUser, Permission.ACTIVITY_DELETE, id);
```
jrasm91, line 88: Delete uses activity id.

server/src/immich/controllers/activity.controller.ts
```diff
@@ -0,0 +1,54 @@
...
+@UseValidation()
+export class ActivityController {
+  constructor(private service: ActivityService) {}
+
+  @Get('statistics')
+  getActivityStatistics(@AuthUser() authUser: AuthUserDto, @Query() dto: ActivityDto): Promise<StatsResponseDto> {
+    return this.service.getStatistics(authUser, dto);
+  }
+
+  @Get('')
```
jrasm91, line 30: Get()

server/src/infra/entities/activity.entity.ts
```diff
@@ -0,0 +1,51 @@
...
+  Check,
+  Column,
+  CreateDateColumn,
+  Entity,
+  Index,
+  ManyToOne,
+  PrimaryGeneratedColumn,
+  UpdateDateColumn,
+} from 'typeorm';
+import { AlbumEntity } from '.';
```
jrasm91, line 11: Import from album.entity

server/src/infra/entities/asset.entity.ts
```diff
@@ -158,6 +159,9 @@ export class AssetEntity {
 
   @OneToMany(() => AssetEntity, (asset) => asset.stackParent)
   stack?: AssetEntity[];
+
+  @OneToMany(() => ActivityEntity, (activity) => activity.asset)
+  activity!: ActivityEntity[];
```
jrasm91, line 164: Is this relationship needed?

server/src/infra/entities/user.entity.ts
```diff
@@ -62,4 +63,7 @@ export class UserEntity {
 
   @OneToMany(() => AssetEntity, (asset) => asset.owner)
   assets!: AssetEntity[];
+
+  @OneToMany(() => ActivityEntity, (assetFace) => assetFace.user)
```
jrasm91, line 67: Surely this side isn't needed.

server/src/infra/repositories/activity.repository.ts
```diff
@@ -0,0 +1,79 @@
+import { IActivityRepository } from '@app/domain';
+import { Injectable } from '@nestjs/common';
+import { InjectRepository } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+import { ActivityEntity } from '../entities/activity.entity';
+
+export interface ActivitySearch {
+  id?: string;
```
jrasm91, line 8: Remove this

server/test/fixtures/activity.stub.ts
```diff
@@ -0,0 +1,34 @@
...
+    userId: authStub.admin.id,
+    user: userStub.admin,
+    assetId: assetStub.image.id,
+    asset: assetStub.image,
+    albumId: albumStub.oneAsset.id,
+    album: albumStub.oneAsset,
+    createdAt: new Date(),
+    updatedAt: new Date(),
+  }),
+  favorite: Object.freeze<ActivityEntity>({
```
jrasm91, line 21: Liked

jrasm91 (commented):
Did not get too much time tonight, so left myself a few notes.

jrasm91 (approved):

alextran1502 (approved):
Amazing work! Thank you all!

waclaw66:
I think activity icons shouldn't be visible while slideshow.

PR #4695, opened 2023-10-29
fix(server): better fix for the OAuth Discovery errors

samip5:


jrasm91:
Probably yes with your tsconfig change

samip5:
I have fixed it since @jrasm91

server/tsconfig.json
```diff
@@ -26,7 +26,7 @@
       "@app/infra/*": ["src/infra/*"],
       "@app/domain": ["src/domain"],
       "@app/domain/*": ["src/domain/*"]
-    }
+    },
```
jrasm91, line 29:
```suggestion
    }
```

server/tsconfig.json
```diff
@@ -8,7 +8,7 @@
     "experimentalDecorators": true,
     "allowSyntheticDefaultImports": true,
     "resolveJsonModule": true,
-    "target": "es2017",
+    "target": "es2021",
```
jrasm91, line 11: Why did you change this? Does the type not exist in es2017 or something?

server/src/domain/auth/auth.service.ts
```diff
@@ -339,8 +339,12 @@ export class AuthService {
...
 
       return new issuer.Client(metadata);
-    } catch (error: Error | any) {
-      this.logger.error(`Error in OAuth discovery: ${error}`, error?.stack);
+    } catch (error: any) {
+      if (error instanceof AggregateError) {
+        this.logger.error(`Error in OAuth discovery: ${error}`, error?.stack, error?.errors);
+      } else if (error instanceof Error) {
+        this.logger.error(`Error in OAuth discovery: ${error}`, error?.stack);
+      }
```
jrasm91, line 347: I believe you can just change line 342 to be `error: any | AggregateError` and then use a single logger.error with `error?.errors`, instead of if/else with copy/pasted log lines.

server/tsconfig.json
samip5: Correct, it doesn't exist until 2021.

server/src/domain/auth/auth.service.ts
samip5: Original problem was that it wouldn't even find the AggregateError, but now it does, so fixed it to be like that instead.

jrasm91 (approved):

PR #4720, opened 2023-10-30
refactor(server): auth delete device

jrasm91:
Migrate delete user token to delete by id, instead of user id and device id (standardizing). This implies using access core to check for device access and throwing an error for a not found / not owned device id (instead of silently returning a 200 success)


danieldietzler (approved):
LGTM!

PR #4779, opened 2023-11-01
feat(web,server): user avatar color

martabal:
## Changes made in this PR

This PR adds the ability to a user to change its avatar color. Once he has set the avatar color, if he had a profile image, this profile image is deleted. Now, when a user is created, its avatar color is picked randomly from 10 colors.

Now when the user set a new profile image, the previous profile image is deleted from the file system.

Resolves #4767.

## Screenshots

https://github.com/immich-app/immich/assets/74269598/3556968c-3f65-4e33-962c-3a7f98f75808

### Responsive design

image


server/immich-openapi-specs.json
```diff
@@ -5440,6 +5440,29 @@
       }
     },
     "/user/profile-image": {
+      "delete": {
+        "operationId": "deleteProfileImage",
+        "parameters": [],
+        "responses": {
+          "200": {
```
jrasm91, line 5447: This should be 204 no content

server/src/domain/user/dto/update-user.dto.ts
```diff
@@ -1,6 +1,7 @@
 import { ApiProperty } from '@nestjs/swagger';
 import { Transform } from 'class-transformer';
 import { IsBoolean, IsEmail, IsNotEmpty, IsString, IsUUID } from 'class-validator';
+import { UserAvatarColor } from '..';
```
jrasm91, line 4: Can you fix this import?

server/src/infra/entities/user.entity.ts
```diff
@@ -1,3 +1,4 @@
+import { UserAvatarColor } from '@app/domain/user/response-dto/user-response.dto';
```
jrasm91, line 1: Importing from app/domain into user entities can cause circular imports. Can you move the enum to this file instead?

web/src/lib/components/user-settings-page/avatar-selector.svelte
```diff
@@ -0,0 +1,50 @@
...
+  import CircleIconButton from '../elements/buttons/circle-icon-button.svelte';
+  import { createEventDispatcher } from 'svelte';
+  import FullScreenModal from '../shared-components/full-screen-modal.svelte';
+  import UserAvatar from '../shared-components/user-avatar.svelte';
+  import { UpdateUserDtoAvatarColorEnum, UserResponseDto } from '@api';
+
+  export let user: UserResponseDto;
+
+  const dispatch = createEventDispatcher();
+  const colors: Array<UpdateUserDtoAvatarColorEnum> = [
```
jrasm91, line 12: Can this just be Object.values(UpdateUserDtoAvatarColorEnum)? Otherwise you have to remember to update the list whenever it changes.

server/src/domain/user/dto/update-user.dto.ts
```diff
@@ -49,4 +50,13 @@ export class UpdateUserDto {
   @Optional()
   @IsBoolean()
   memoriesEnabled?: boolean;
+
+  @Optional()
+  @IsString()
+  @IsNotEmpty()
+  avatarColor?: UserAvatarColor;
```
jrasm91, line 57: This should not be `IsString()` but `IsEnum(UseAvatarColor)`, right? Also can you add the `@ApiProperty()` stuff for named enums, like we do everywhere else?

server/src/domain/user/user.service.ts
```diff
@@ -93,10 +93,25 @@ export class UserService {
...
     const updatedUser = await this.userRepository.update(authUser.id, { profileImagePath: fileInfo.path });
+    if (user.profileImagePath !== '') {
+      const files = [user.profileImagePath];
+      await this.jobRepository.queue({ name: JobName.DELETE_FILES, data: { files } });
+    }
     return mapCreateProfileImageResponse(updatedUser.id, updatedUser.profileImagePath);
   }
 
+  async deleteProfileImage(authUser: AuthUserDto): Promise<void> {
+    const user = await this.findOrFail(authUser.id, { withDeleted: true });
```
jrasm91, line 106: You can delete a deleted user's profile pictures? Seems a bit strange.

server/src/domain/user/user.service.ts
```diff
@@ -93,10 +93,25 @@ export class UserService {
     authUser: AuthUserDto,
     fileInfo: Express.Multer.File,
   ): Promise<CreateProfileImageResponseDto> {
+    const user = await this.findOrFail(authUser.id, { withDeleted: true });
     const updatedUser = await this.userRepository.update(authUser.id, { profileImagePath: fileInfo.path });
+    if (user.profileImagePath !== '') {
```
jrasm91, line 98: How or why would this ever be an empty string when using "createProfileImage"?

server/src/domain/user/dto/update-user.dto.ts
```diff
@@ -49,4 +50,13 @@ export class UpdateUserDto {
...
+
+  @Optional()
+  @IsString()
+  profileImagePath?: string;
```
jrasm91, line 61: Why is this added here?

web/src/lib/components/shared-components/user-avatar.svelte
```diff
@@ -39,12 +45,14 @@
     md: 'w-10 h-10',
     lg: 'w-12 h-12',
     xl: 'w-20 h-20',
+    xxl: 'w-24 h-24',
+    xxxl: 'w-24 h-24',
```
jrasm91, line 49: are these supposed to be the same size?

server/src/domain/user/user.service.ts
martabal: If a user has never set is profile image, the default value is an empty string

jrasm91: Oh, I see. What do you think about having it be a static path instead of a random uuid and then having to delete the old one? So overwrite instead of new + delete.

server/src/domain/user/dto/update-user.dto.ts
martabal: A rest of my previous implementation, I have reverted it

server/src/domain/user/user.service.ts
martabal: I don't follow. How do you determine if a user has set its profile picture ?

martabal: Yeah it should work. Should we use the user id for the static path ?

martabal: Maybe we should change the profileImagePath to be a boolean

jrasm91: Ah, hold on. I guess the extension may not always be the same, so you will have to potentially delete it afterwards, so maybe it is fine like you have it.

jrasm91: I did get confused with user vs updated user and the user.profilePath is the old path, but referenced after you updated to the new one. Maybe you could do something like `const { profilePath: oldPath } = await this.findOrFail(...)`; to make it a little more clear.

martabal: Done

alextran1502:
Hello, thank you for working on this mechanism! 

I think we can improve the slow of not clicking many buttons by adding the edit profile image in the user profile picture from the popup panel. What do you think?

image

martabal:
It looks good. Should it be in addition or replacement ?

alextran1502:
> It looks good. Should it be in addition or replacement ?

I would say a replacement

martabal:
Done 

https://github.com/immich-app/immich/assets/74269598/2a133113-190c-49b6-8dea-2ddf113507c8

shenlong-tanwen:
Pushed the mobile changes to support displaying the avatar colors assigned from the web. It is still not possible to set a color from the mobile app, but that can be added in a separate PR.

>  [!Warning]
> The mobile app and server has to be in the same version. Only updating the mobile app will prevent the user from logging in due to changes in the dto.

server/src/domain/user/response-dto/user-response.dto.ts
```diff
@@ -1,10 +1,17 @@
...
+  const randomIndex = Math.floor(Math.random() * values.length);
+  return values[randomIndex] as UserAvatarColor;
+};
 
 export class UserDto {
   id!: string;
   name!: string;
   email!: string;
   profileImagePath!: string;
+  avatarColor!: UserAvatarColor;
```
jrasm91, line 14: This is missing api property decorator for a named enum.

server/src/domain/user/response-dto/user-response.dto.ts
```diff
@@ -1,10 +1,17 @@
-import { UserEntity } from '@app/infra/entities';
+import { UserAvatarColor, UserEntity } from '@app/infra/entities';
+
+export const getRandomAvatarColor = (): UserAvatarColor => {
+  const values = Object.values(UserAvatarColor);
+  const randomIndex = Math.floor(Math.random() * values.length);
+  return values[randomIndex] as UserAvatarColor;
+};
```
jrasm91, line 7: Why don't we just make this deterministic? I don't think need need to store it or anything like that. Empty implies random and if it was deterministic this could be accomplished in the mapper itself.

server/src/domain/user/user.core.ts
```diff
@@ -98,7 +99,7 @@ export class UserCore {
     if (payload.storageLabel) {
       payload.storageLabel = sanitize(payload.storageLabel);
     }
-
+    payload.avatarColor = getRandomAvatarColor();
```
jrasm91, line 102: I would prefer making this deterministic without having to explicitly save it in the database on user create.

web/src/lib/components/shared-components/navigation-bar/avatar-selector.svelte
```diff
@@ -0,0 +1,41 @@
...
+  import { createEventDispatcher } from 'svelte';
+  import { UserDtoAvatarColorEnum, UserResponseDto } from '@api';
+  import CircleIconButton from '$lib/components/elements/buttons/circle-icon-button.svelte';
+  import FullScreenModal from '../full-screen-modal.svelte';
+  import UserAvatar from '../user-avatar.svelte';
+
+  export let user: UserResponseDto;
+
+  const dispatch = createEventDispatcher();
+  const colors: Array<UserDtoAvatarColorEnum> = Object.values(UserDtoAvatarColorEnum);
```
jrasm91, line 12:
```suggestion
  const colors: UserDtoAvatarColorEnum[] = Object.values(UserDtoAvatarColorEnum);
```

jrasm91 (approved):
Looks pretty good!

server/src/domain/user/response-dto/user-response.dto.ts
martabal: I don't follow. The goal is to create a random avatar color only on user creation.

server/src/domain/user/response-dto/user-response.dto.ts
```diff
@@ -1,10 +1,28 @@
...
 
 export class UserDto {
   id!: string;
   name!: string;
   email!: string;
   profileImagePath!: string;
+  @IsEnum(UserAvatarColor)
+  @Optional()
+  @ApiProperty({ enumName: 'UserAvatarColor', enum: UserAvatarColor })
+  avatarColor!: UserAvatarColor | null;
```
jrasm91, line 25: Under what condition will this be null? Line 46 says "if null pick a value". Is this just to avoid some mobile compat issue?

server/src/domain/user/user.service.ts
```diff
@@ -93,10 +93,25 @@ export class UserService {
     authUser: AuthUserDto,
     fileInfo: Express.Multer.File,
   ): Promise<CreateProfileImageResponseDto> {
+    const { profileImagePath: oldpath } = await this.findOrFail(authUser.id, { withDeleted: false });
     const updatedUser = await this.userRepository.update(authUser.id, { profileImagePath: fileInfo.path });
+    if (oldpath !== '') {
+      const files = [oldpath];
+      await this.jobRepository.queue({ name: JobName.DELETE_FILES, data: { files } });
```
jrasm91, line 100:
```suggestion
      await this.jobRepository.queue({ name: JobName.DELETE_FILES, data: { files: [oldPath] } });
```

server/src/domain/user/user.service.ts
```diff
@@ -93,10 +93,25 @@ export class UserService {
...
   }
 
+  async deleteProfileImage(authUser: AuthUserDto): Promise<void> {
+    const user = await this.findOrFail(authUser.id, { withDeleted: false });
+    if (user.profileImagePath !== '') {
+      await this.userRepository.update(authUser.id, { profileImagePath: '' });
+      const files = [user.profileImagePath];
+      await this.jobRepository.queue({ name: JobName.DELETE_FILES, data: { files } });
+    }
+    return;
```
jrasm91, line 112: I don't think this is necessary, is it?

server/src/domain/user/user.service.ts
```diff
@@ -93,10 +93,25 @@ export class UserService {
...
+    if (oldpath !== '') {
+      const files = [oldpath];
+      await this.jobRepository.queue({ name: JobName.DELETE_FILES, data: { files } });
+    }
     return mapCreateProfileImageResponse(updatedUser.id, updatedUser.profileImagePath);
   }
 
+  async deleteProfileImage(authUser: AuthUserDto): Promise<void> {
+    const user = await this.findOrFail(authUser.id, { withDeleted: false });
+    if (user.profileImagePath !== '') {
```
jrasm91, line 107: I think an http error make be more appropriate if the user doesn't have a profile picture. The client can ignore it instead.

jrasm91 (approved):
Looks pretty good!

server/src/domain/user/response-dto/user-response.dto.ts
martabal: Else there's a type error with `const providedPassword = await ask(admin);` https://github.com/immich-app/immich/blob/ac7e8bcdf469bfbe9810c8ccfc39ebb1e6b37b45/server/src/domain/user/user.service.ts#L108-L120

jrasm91: That would be because it is returning an user entity not a user response. Can you change it to `await ask(mapUser(admin))` instead? And then remove the null type on the response dto.

martabal: Damn, I didn't think about this trick. Thanks !

PR #4792, opened 2023-11-02
fix(web): unstacking issues

brighteyed:
## Description

This PR fixes 2 issues in the unstacking code

### Issue 1
`handleUnstack` function doesn't leave the asset store in the consistent state. As a result, after unstacking, (1) child assets are invisible for `Shift + Click` selection, (2) forward/backward navigation in certain cases also doesn't work correctly.

### Issue 2
After unstacking all the child assets are rendered as squares. This is because they don't have `exifInfo` set

## Before/after

https://github.com/immich-app/immich/assets/71479/e71ab71e-bb58-47aa-82b5-df90bd24652e

https://github.com/immich-app/immich/assets/71479/c23a275b-f426-4e2b-bab5-610042c299cd

### DIFF
--- a/server/src/immich/api-v1/asset/asset-repository.ts
+++ b/server/src/immich/api-v1/asset/asset-repository.ts
@@ -109,7 +109,9 @@ export class AssetRepository implements IAssetRepository {
         faces: {
           person: true,
         },
-        stack: true,
+        stack: {
+          exifInfo: true,
+        },
       },
       // We are specifically asking for this asset. Return it even if it is soft deleted
       withDeleted: true,
--- a/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
+++ b/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
@@ -32,7 +32,7 @@
   export let showDownloadButton: boolean;
   export let showDetailButton: boolean;
   export let showSlideshow = false;
-  export let hasStackChildern = false;
+  export let hasStackChildren = false;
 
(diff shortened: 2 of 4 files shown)


server/src/immich/api-v1/asset/asset-repository.ts
```diff
@@ -109,7 +109,9 @@ export class AssetRepository implements IAssetRepository {
         faces: {
           person: true,
         },
-        stack: true,
+        stack: {
+          exifInfo: true,
```
jrasm91, line 113: For this need other relations? Should it essentially match assets? So exif, people, etc?

jrasm91 (approved):

server/src/immich/api-v1/asset/asset-repository.ts
brighteyed: Hmm, I think you are right. But to be 100% sure, I think the necessity of each relation should be examined

jrasm91: Do people work before/after stacking and unstacking an asset? I assume if stacked, and then refresh the screen, and then unstack and view detail people would not show up unless they are lazy loaded in the asset-viewer itself.

brighteyed: I see no issues with `people`. The `DetailPanel` component invokes the `getAssetById` function for the asset, including the case when we click the thumbnail of the child in the stack
So it seems we don't need to add faces/person relation here now. However, we definitely should remember that child assets returned by `getAssetById` contain less information comparent to the parent asset

PR #4796, opened 2023-11-02
feat(web): global activity

martabal:
## Changes made in this PR

Following #4682, this PR adds a global activity tab on the album page, where you can comment the whole album and see all the comments for each assets

## Screenshots

https://github.com/immich-app/immich/assets/74269598/2226d41c-eeaf-4d75-a12d-6456881b1f11


web/src/lib/components/asset-viewer/activity-viewer.svelte
```diff
@@ -51,9 +51,11 @@
   }
 
   $: {
-    if (previousAssetId != assetId) {
+    if (assetId && previousAssetId != assetId) {
       getReactions();
       previousAssetId = assetId;
+    } else if (assetId === undefined) {
+      getReactions();
```
danieldietzler, line 58: I think I would prefer
```typescript
if (assetId) {
    getReactions();
}
if (previousAssetId !== assetId) {
    previousAssetId = assetId;
}
```
This should work as well because `undefined !== null`.

web/src/lib/stores/activity.store.ts
```diff
@@ -0,0 +1,12 @@
+import { writable } from 'svelte/store';
+
+export const numberOfComments = writable<number | undefined>(undefined);
+
+export const setNumberOfComments = (number: number) => {
+  numberOfComments.set(number);
+};
+
+export const updateNumberOfComments = (addOrRemove: boolean) => {
```
danieldietzler, line 9: What does it mean if addOrRemove is true? Without looking at the code this isn't clear.
I would prefer to have something like e.g. `addOrRemove: 'add' | 'remove'` or even just let them specify the delta `delta: 1 | -1`.

web/src/routes/(user)/albums/[albumId]/+page.svelte
```diff
@@ -384,176 +433,225 @@
...
+            {/if}
+          </svelte:fragment>
+        </ControlAppBar>
+      {/if}
+
+      {#if viewMode === ViewMode.SELECT_ASSETS}
+        <ControlAppBar on:close-button-click={handleCloseSelectAssets}>
+          <svelte:fragment slot="leading">
+            <p class="text-lg dark:text-immich-dark-fg">
+              {#if $timelineSelected.size == 0}
```
Comment (danieldietzler) [inline web/src/routes/(user)/albums/[albumId]/+page.svelte:471]: This should be `===`. Doesn't matter in that case because `$timelineSelected.size` cannot be undefined or otherwise falsy. However I think it is good practice to always check for the explicit type, if that's the intention of the expression.

web/src/routes/(user)/albums/[albumId]/+page.svelte
```diff
@@ -384,176 +433,225 @@
...
+          isSelectionMode={viewMode === ViewMode.SELECT_THUMBNAIL}
+          singleSelect={viewMode === ViewMode.SELECT_THUMBNAIL}
+          on:select={({ detail: asset }) => handleUpdateThumbnail(asset.id)}
+          on:escape={handleEscape}
+        >
+          {#if viewMode !== ViewMode.SELECT_THUMBNAIL}
+            <!-- ALBUM TITLE -->
+            <section class="pt-24">
+              <input
+                on:keydown={(e) => e.key == 'Enter' && titleInput.blur()}
```
Comment (danieldietzler) [inline web/src/routes/(user)/albums/[albumId]/+page.svelte:526]: This should be `e.key === 'Enter'` as well.

web/src/routes/(user)/albums/[albumId]/+page.svelte
```diff
@@ -384,176 +433,225 @@
...
-            >
-              <span class="text-text-immich-primary dark:text-immich-dark-primary"
-                ><Icon path={mdiPlus} size="24" />
-              </span>
-              <span class="text-lg">Select photos</span>
-            </button>
-          </div>
-        </section>
+      {#if album.sharedUsers.length > 0 && !$showAssetViewer}
+        <div class="z-[9999] absolute bottom-0 right-0 mb-6 mr-6 justify-self-end">
```
Comment (danieldietzler) [inline web/src/routes/(user)/albums/[albumId]/+page.svelte:620]: Is that z-index avoidable? I hate it when we do that and I would really like to (slowly) get away from this in all places.

Comment (martabal) [reply web/src/routes/(user)/albums/[albumId]/+page.svelte]: Yes this one is.

jrasm91:
Can you change the UX to behave similar to google photos and the current info side panel? IE the album buttons are still available, even when the activity side panel is open. Also, can we add a comments icon to the navbar which toggles opening/closing the activity side panel?

web/src/routes/(user)/albums/[albumId]/+page.svelte
```diff
@@ -77,6 +82,12 @@
   let isCreatingSharedAlbum = false;
   let currentAlbumName = '';
   let contextMenuPosition: { x: number; y: number } = { x: 0, y: 0 };
+  let isShowActivity = false;
+  let isLiked: ActivityResponseDto | null = null;
+  let reactions: ActivityResponseDto[] = [];
+  let user = data.user;
+  let GlobalWidth: number;
```
Comment (alextran1502) [inline web/src/routes/(user)/albums/[albumId]/+page.svelte:89]: should this be named `globalWidth`?

web/src/routes/(user)/albums/[albumId]/+page.svelte
```diff
@@ -77,6 +82,12 @@
...
+  let AssetGridWidth: number;
```
Comment (alextran1502) [inline web/src/routes/(user)/albums/[albumId]/+page.svelte:90]: should this be named `assetGridWidth`?

web/src/lib/components/asset-viewer/activity-viewer.svelte
```diff
@@ -15,6 +15,14 @@
 
   const units: Intl.RelativeTimeFormatUnit[] = ['year', 'month', 'week', 'day', 'hour', 'minute', 'second'];
 
+  const shouldGroup = (currentDate: string, nextDate: string): boolean => {
+    return (
+      luxon.DateTime.fromISO(currentDate).toRelative({ unit: 'hours' }) ===
+        luxon.DateTime.fromISO(nextDate).toRelative({ unit: 'hours' }) ||
+      luxon.DateTime.fromISO(currentDate).toRelative() === luxon.DateTime.fromISO(nextDate).toRelative()
```
danieldietzler, line 22: I'm not 100% sure I understand this correctly. If what you're trying to achieve is to check if currentDate and nextData are equal concerning a specific unit, I think you would want `luxon.DateTime.fromISO(currentDate).hasSame(luxonDateTime.fromISO(nextDate), 'hour')`
Also, I think I would prefer to have two variables representing the respective DateTime objects and then use these in the expression.

martabal: This groups all the comments by units. The issue with `luxon.DateTime.fromISO(currentDate).toRelative() === luxon.DateTime.fromISO(nextDate).toRelative()` is that it correctly groups days, hours, years... But it does not correctly group reactions posted less than an hour ago

danieldietzler: Ok, I think you could then keep the `currentDateTime.toRelative() === nextDateTime.toRelative()` and the other comparison would be `currentDateTime.hasSame(nextDateTime, 'hour')`

alextran1502 (approved):

PR #4830, opened 2023-11-03
feat(web): loop videos

martabal:
## Changes made in this PR

This PR adds the ability to loop a video. By default it's disabled and can be enabled in the user account settings.

## Screenshots

https://github.com/immich-app/immich/assets/74269598/a0222ff4-2b08-4413-bde1-a1eca64e5d5e


jrasm91 (commented):
Can we add the control on the video player itself?

martabal:
AFAIK, not without a custom video player

jrasm91:
Can a user right click and select loop? Does that fire an event? I just feel like the menu is a bit full already and adding a loop video option seems less than ideal. Maybe just having it be in the user settings would be better, albeit less accessible.

alextran1502:
I will have to agree with Jason on this feature that it is already available when right click on the video. Adding this into Immich would be redundant.
image

PR #4831, opened 2023-11-03
chore(mobile): change version mismatch text

Wingysam:
The message for when the server is outdated reads as a bit strange. This PR changes it to read more naturally to me.


jrasm91:
What about just

Server is out of date please consider updating to the latest minor/major version. ...etc

Wingysam:
Yeah that's better, one second

jrasm91 (approved):
Lgtm

PR #4865, opened 2023-11-06
feat(server): make is favorite optional on asset upload

jrasm91:


danieldietzler (approved):
The diff for one char changed lmao

PR #4867, opened 2023-11-06
docs: python upload guide

jrasm91:


danieldietzler (approved):
Nice!

docs/docs/guides/python-file-upload.md
```diff
@@ -0,0 +1,42 @@
...
+    stats = os.stat(file)
+
+    headers = {
+        'Accept': 'application/json',
+        'x-api-key': API_KEY
+    }
+
+    data = {
+        'deviceAssetId': f'{file}-{stats.st_mtime}',
+        'deviceId': 'python',
```
bo0tzz, line 24: Do we still require these?

jrasm91: oh yeah

jrasm91: they aren't even nullable in the database

jrasm91: gotta tell those mobile developers to stop using them already

PR #4870, opened 2023-11-06
feat(server, web): Album's options

martabal:
## Changes made in this PR

Currently, Activity is enabled on all shared albums, this PR adds the ability to the album owner to disable the activity tabs on its albums. Once activity is disabled, previous comments are still displayed and the activity status (the like and comment icons on the bottom right) is still displayed if the viewed asset has still comments. If an album has no comments anymore, the activity status is not displayed

## Screenshots

https://github.com/immich-app/immich/assets/74269598/41564ae5-c2e1-47e4-a37f-79e1fe1c156f

### DIFF
--- a/server/src/domain/access/access.core.ts
+++ b/server/src/domain/access/access.core.ts
@@ -140,7 +140,8 @@ export class AccessCore {
       case Permission.ACTIVITY_CREATE:
         return (
           (await this.repository.album.hasOwnerAccess(authUser.id, id)) ||
-          (await this.repository.album.hasSharedAlbumAccess(authUser.id, id))
+          (await this.repository.album.hasSharedAlbumAccess(authUser.id, id)) ||
+          (await this.repository.activity.hasAlbumUpdate(authUser.id, id))
         );
 
       // uses activity id
--- a/server/src/domain/access/access.core.ts
+++ b/server/src/domain/access/access.core.ts
@@ -138,11 +138,7 @@ export class AccessCore {
     switch (permission) {
       // uses album id
       case Permission.ACTIVITY_CREATE:
-        return (
-          ((await this.repository.album.hasOwnerAccess(authUser.id, id)) ||
-            (await this.repository.album.hasSharedAlbumAccess(authUser.id, id))) &&
-          (await this.repository.album.hasActivityAccess(authUser.id, id))
-        );
+        return await this.repository.activity.hasCreateAccess(authUser.id, id);
 
       // uses activity id
       case Permission.ACTIVITY_DELETE:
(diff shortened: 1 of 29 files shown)


server/src/domain/access/access.core.ts
```diff
@@ -139,8 +139,9 @@ export class AccessCore {
       // uses album id
       case Permission.ACTIVITY_CREATE:
         return (
-          (await this.repository.album.hasOwnerAccess(authUser.id, id)) ||
-          (await this.repository.album.hasSharedAlbumAccess(authUser.id, id))
+          ((await this.repository.album.hasOwnerAccess(authUser.id, id)) ||
+            (await this.repository.album.hasSharedAlbumAccess(authUser.id, id))) &&
+          (await this.repository.album.hasActivityAccess(authUser.id, id))
```
jrasm91, line 144: Hmm, I don't think we want this. We can just search for an activity enabled album in the access implementation.

server/src/infra/repositories/access.repository.ts
```diff
@@ -219,6 +219,14 @@ export class AccessRepository implements IAccessRepository {
...
       });
     },
+    hasActivityAccess: (_: string, albumId: string): Promise<boolean> => {
+      return this.albumRepository.exist({
+        where: {
+          id: albumId,
+          isActivityEnabled: true,
+        },
+      });
+    },
```
jrasm91, line 229: Can you just merge this into the exist access checks? It can become something like:

```typescript
album: {
  id: albumId,
  isActivityEnabled: true,
}
```

jrasm91 (commented):
If people have commented, but then comments are turned off. I would expect them to be hidden, not show up as read only.

server/src/infra/repositories/access.repository.ts
martabal: for `album.hasOwnerAccess` and `album.hasSharedAlbumAccess` ?

jrasm91: Oh right, I guess that is used in other places as well. You're right!

jrasm91 (approved):
Code LGTM

web/src/routes/(user)/albums/[albumId]/+page.svelte
```diff
@@ -128,6 +138,24 @@
...
+          isActivityEnabled: !album.isActivityEnabled,
+        },
+      });
+      album = data;
+      notificationController.show({
+        type: NotificationType.Info,
+        message: `Activity is ${album.isActivityEnabled ? 'enabled' : 'disabled'}`,
+      });
+    } catch (error) {
+      handleError(error, "Can't change favorite for asset");
```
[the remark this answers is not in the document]
Comment (martabal) [reply web/src/routes/(user)/albums/[albumId]/+page.svelte:155]: Oops

Issue #4841, opened 2023-11-04
[BUG] Immich has separate first and last name fields and both are mandatory


ToxicFrog:
### The bug

When creating or editing a user account, you are _required_ to provide both a first and last name. Failure to do so will not let you complete the user creation flow.  If you only have one name, or a name that doesn't readily decompose into two parts, or if those parts don't map cleanly to "first" and "last", you are out of luck.

I'm not sure if this is a design bug (Immich's data model assumes that people have names that come in at least two parts called "first" and "last") or an implementation bug (Immich's data model permits freeform names but the UI is incorrectly wired up with two name fields and both are marked mandatory).

Ideally, the name field should be a single free text "display name" field. If Immich is commited to binomial names, at least one of the fields should be optional.

### The OS that Immich Server is running on

docker on nixos

### Version of Immich Server

v1.84.0

### Version of Immich Mobile App

N/A

### Platform with the issue

- [X] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

I'm not using docker-compose.yml, but a nixos config equivalent to [the official docker-compose](https://github.com/immich-app/immich/blob/main/docker/docker-compose.yml) but with redis and postgres running on the host.

### Your .env content

```nix
environment = {
    DB_HOSTNAME = "host.docker.inernal";
    DB_USERNAME = [redacted];
    DB_PASSWORD = [redacted];
    DB_DATABASE_NAME = "immich";

    REDIS_HOSTNAME = "host.docker.internal";
    REDIS_PASSWORD = [redacted];

    TYPESENSE_ENABLED = "true";
    TYPESENSE_HOST = "immich-typesense";
    TYPESENSE_API_KEY = [redacted];
    TYPESENSE_DATA_DIR = "/data";

    IMMICH_WEB_URL = "http://immich-web:3000";;
    IMMICH_SERVER_URL = "http://immich-server:3001";
    IMMICH_MACHINE_LEARNING_URL = "http://immich-ml:3003";
  };
```

### Reproduction steps

1. As an admin, click "add user".
2. Fill in the email and password but leave one (or both) of the name fields blank.
3. Click "create".

### Additional information

Setting the last name to a single space seems to appease it and has not _yet_ visibly broken anything.

[Falsehoods Programmers Believe about Names](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/) is 13 years old at this point and still as relevant as ever.

[...]

PR #4881, opened 2023-11-07
refactor(server): auth dtos

jrasm91:
- Consolidate auth dtos
- Remove admin sign up response (in favor of user response dto)
- Deprecate `OAuthConfigResponseDto` (for the mobile devs)

Tested via:
- All e2e still pass
- Clean install + sign up works as expected


server/src/domain/auth/auth.dto.ts
```diff
@@ -0,0 +1,132 @@
...
+export class SignUpDto extends LoginCredentialDto {
+  @IsString()
+  @IsNotEmpty()
+  @ApiProperty({ example: 'Admin' })
+  firstName!: string;
+
+  @IsString()
+  @IsNotEmpty()
+  @ApiProperty({ example: 'Doe' })
+  lastName!: string;
```
danieldietzler, line 70: We don't want to do #4841 in this PR I'm assuming?

jrasm91: Correct.

server/src/domain/auth/auth.dto.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export class ChangePasswordDto {
+  @IsString()
+  @IsNotEmpty()
+  @ApiProperty({ example: 'password' })
+  password!: string;
+
+  @IsString()
+  @IsNotEmpty()
+  @MinLength(8)
```
danieldietzler, line 81: Do we want to have a more strict password policy?

danieldietzler (approved):

server/src/domain/auth/auth.dto.ts
jrasm91: I think it is ok for now

PR #4888, opened 2023-11-07
feat(server): typesense api key as docker secret

patrickap:
The `TYPESENSE_API_KEY_FILE` variable has been added to be able to use Docker secrets.


bo0tzz:
Does the typesense container support this as well?

patrickap:
Currently not directly but same for Redis. You can override the entrypoint in docker-compose.yml and be able to set it as `--api-key $$(cat $$TYPESENSE_API_KEY_FILE)`.

bo0tzz:
Can you add a footnote about that, like the one that's there for redis?

patrickap:
> It is possible that modifications might also be necessary for immich-server and immich-microservices. 

Once my Pull Request is merged, there’s no need to overwrite the commands on either the immich-server or the immich-microservices. The only requirement is to set the new variable, `TYPESENSE_API_KEY_FILE`. In terms of implementation, I’ve followed the pattern of existing variables, such as `REDIS_PASSWORD_FILE.

jrasm91:
Thank you for the PR, but we've officially dropped Typesense so this change is no longer needed.

PR #4889, opened 2023-11-07
feat(server,web): Add private album feature

jarvis2f:
I have added a private option to the album, so that some photos that you don't want to make public can be placed on the album.
On the album interface, you can set the album to private, so that when the password of the private album is not entered, all the photos under this album are hidden in all places.

## How to set
* On the album card view, list view, or album details page, set the album to private
* Set the password for the private album in the personal settings
* On the album page, press: ↑↑↓↓←→←→baba (this is the secret code for Contra), a dialog box will pop up to enter the password, and then you can view the private album
* Click the eye icon in the lower right corner of the page to quickly close the display of the private album

(Mobile devices currently do not support setting and viewing private photo albums)

## How to implement
* Added a field: isPrivate, on the album entity to indicate whether the album is a private album. Other resources, faces, photos, etc. are judged as private resources by associating with the album.
  After entering the password for the private album, a request is made to `/user/validate-private-album-password` to verify if the password is correct. If it is correct, the `${userId}-${password}` hashed is saved to the cookie.
  Subsequent request processing is consistent with `accessToken`, verifying whether `${userId}-${password}` is correct. If it is correct, the `isShowPrivateAlbum` field in `AuthUserDto` is set to `true`.
* Added the `isPrivate` field in `album.schema.ts` and `asset.schema.ts`, used to indicate whether the album and resources are private in searches.

I have tested most scenarios, including sharing, the map page, and private albums and resources in searches, and they will not be displayed. If there are any omissions, please feel free to point them out.

## Screenshot

|private password input dialog| private album setting btn | private mode close btn |account private password setting|
|---|---|---|---|
|img| img_1 | img_2 |img_3|


alextran1502:
What do you think about the new Locked Folder approach in GPhotos?

jarvis2f:
> What do you think about the new Locked Folder approach in GPhotos?
> 
> image

I think it's better to hide the album entrance. Directly using the login password to unlock is also not a good option.

shenlong-tanwen:
We are not using album specific passwords but a per user password to access multiple different private albums?

>On the album page, press: ↑↑↓↓←→←→baba (this is the secret code for Contra), a dialog box will pop up to enter the >password, and then you can view the private album

This sounds fun! But would it be possible to not have hidden shortcuts, but make it more obvious to show / hide private albums. Maybe with a toggle? 😅

jrasm91:
This seems like an attempt implement finer-grained access controls, starting with private album access, but it is implemented by adding additional cookies client-side to keep track of state. What do you think about moving this state server-side instead? Some something like:
- Adding a new column to the user token to track private album access (per session), so like `privateAlbums!: Date`
- This column to a boolean value on the auth user (`authUser.showPrivateAlbums`) by checking if it is less than 15 minutes old (or whatever threshold)
- Modify the endpoint you have to instead set `privateAlbums: new Date()` on the token itself, rather than the cookie.
- etc.

I think this would be a cleaner solution. Basically moving from additional cookies to state on the server-side token instead.

yodatak:
Thanks for this awesome proposition did you think it can also change the xmp file or sidecard to make them platform agnostic exemple in photoprism digikam or many app ?

alextran1502:
I think a private view makes sense to have those media in a centralized place, like GPhotos implementation. 

I do not like obscuring the behavior to get to the private view. You can create a private password for that view that differs from your login password. 

The goal is to have the flow as natural as possible.

I hope that answers and gives some directions for the design choice.

jarvis2f:
> This seems like an attempt to implement finer-grained access controls, starting with private album access, but it is implemented by adding additional cookies client-side to keep track of state. What do you think about moving this state server-side instead? Some something like:
> 
> * Adding a new column to the user token to track private album access (per session), so like `privateAlbums!: Date`
> * Map this column to a boolean value on the auth user (`authUser.showPrivateAlbums`) by checking if it is less than 15 minutes old (or whatever threshold)
> * Modify the endpoint you have to instead set `privateAlbums: new Date()` on the token itself, rather than the cookie.
> * etc.
> 
> I think this would be a cleaner solution. Basically moving from additional cookies to state on the server-side token instead.

Thank you, this is indeed a better solution.

jarvis2f:
> I think a private view makes sense to have those medias in a centralized place, like GPhotos implementation.
> 
> I do not like obscuring the behavior to get to the private view. You can create a private password for that view that differs from your login password.
> 
> The goal is to have the flow as natural as possible.
> 
> I hope that answers and gives some directions for the design choice.

If we use a similar approach to Google Photos where all private albums are placed under a single entry, does that mean that we won't be able to view these photos in searches, people, maps, and other such features even if we have already entered the password for the private albums? I believe that there are times when we still need to use these features to access private albums or photos.

In some scenarios, if someone is aware of the existence of this entry, even a password may not guarantee its privacy, such as your spouse.

jarvis2f:
> We are not using album specific passwords but a per user password to access multiple different private albums?
> 
> > On the album page, press: ↑↑↓↓←→←→baba (this is the secret code for Contra), a dialog box will pop up to enter the >password, and then you can view the private album
> 
> This sounds fun! But would it be possible to not have hidden shortcuts, but make it more obvious to show / hide private albums. Maybe with a toggle? 😅

Yes, having a switch or the ability to configure how one can access this private album would be useful.

jarvis2f:
> Thanks for this awesome proposition did you think it can also change the xmp file or sidecard to make them platform agnostic exemple in photoprism digikam or many app ?

I'm sorry, I didn't quite understand your meaning.

bo0tzz:
> In some scenarios, if someone is aware of the existence of this entry, even a password may not guarantee its privacy, such as your spouse.

If you need to hide photos to that degree, Immich probably isn't the tool to use.

PR #4905, opened 2023-11-08
feat(build): add docker compose project name

Mansour-J:
I have recently started using immich and loving it. I do run quite a few containers already and immich project always comes up with a generic "docker" which makes it hard to read and find amon many other containers or projects.

image

So I decided to raise a PR for the fix:

image


jrasm91 (commented):
do you mind reverting non `name: <name>` line changes? Possibly saving the file without formatting it?

PR #4916, opened 2023-11-09
fix(server,web): like in global activity

martabal:
## Changes made in this PR

Fix an issue where the like status retrieved for the global activity is the like status of an asset.

## Screenshots

| Before | After |
| :---: | :---: |
| <video src="https://github.com/immich-app/immich/assets/74269598/acefa7cf-3f5e-430f-b95b-583a3dfbcde1">  | <video src="https://github.com/immich-app/immich/assets/74269598/e55f49bd-a0c5-4c27-ac26-4bd437eda4b9"> |

### DIFF
--- a/cli/src/api/open-api/api.ts
+++ b/cli/src/api/open-api/api.ts
@@ -125,12 +125,6 @@ export interface ActivityCreateDto {
     'comment'?: string;
     /**
      * 
-     * @type {boolean}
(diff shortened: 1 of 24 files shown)


jrasm91 (approved):
Seems fine. I'm wondering if there is a better name for the query parameter though.

Maybe subtype=album, or assetId=null.

Issue #1485, opened 2023-01-30
[Tech-Debt] Restify Asset Endpoints


jrasm91:
We should simplify (and restify) the set of API endpoints related to assets.

All endpoints will return a new header,  indicating total item count, which can be used for display totals. A `HEAD` to the endpoint will return the header without the items.

Search:
- [x] `GET` `/search/<metadata|smart>` - get a list of assets with search/sort filters, including `isFavorite`, created before/after/on, albumId, etc.

Single:
- [x] `POST` `/`: - create an asset (with file upload)
- [x] `GET` `/:id` - get asset

Multiple:
- [x] `POST` `/download/archive` - download a list of asset by ids (as an archive)
- [ ] `POST` `/check` - check a list of asset metadata to determine which ones to upload
- [x] `PUT` `/` -update multiple assets by ids
- [x] `DELETE` `/` -delete multiple assets by ids

Serve:
- [ ] `GET` `/media/thumb?assetId=<id>` - get thumbnail
- [ ] `GET` `/media/original?assetId=<id>` - get thumbnail
- [ ] `GET` `/media/resize?assetId=<id>` - get dynamic/resized version

Remove:
- [x] `GET` `/download/:id`
- [x] `GET` `/download-library`
- [ ] `GET` `/:deviceId`
- [x] `GET` `/count-by-time-bucket`
- [x] `GET` `/count-by-user-id`
- [ ] `POST` `/check`
- [ ] `POST` `/exist`
- [x] `POST` `/shared-link` (move to `/shared`)
- [x] `PATCH` `/shared-link` (move to `/shared`)
- [x] `GET` `/assetById/:assetId` (renamed)
- [x] `GET` `/curated-objects`
- [x] `GET` `/curated-locations`
- [x] `GET` `/search-terms`
- [x] `POST` `/search`

agross:
Will there be at some point a new version of `immich-sdk` to reflect these changes for projects using Immich's API? @alextran1502 Context: https://github.com/agross/immich-duplicates/issues/16#issuecomment-1989279506

jrasm91:
I believe all of the asset endpoints are basically fixed, so closing. (at least CRUD seems to be done)

PR #4931, opened 2023-11-09
feat(server): asset search endpoint

jrasm91:
Related to #1485

Adds a new endpoint (not used anywhere yet), that has a lot of support for querying the assets table, which should hopefully make it possible to consolidate some of the existing asset endpoints. Also, it should allow for things like:
- Alternative `AssetGrid` implementation that uses normal pagination and lazy loading
- Easier API consumption by other clients
- Search for assets in a specific date range (uploaded, taken, deleted, etc.)

### Basic
- type (`type=IMAGE`, `type=VIDEO`)
- assetId (`id=<uuid>`)
- isFavorite (`sFavorite=true`, `isFavorite=false`)
- isArchived (`isArchived=true`, `isArchived=false`)
- isReadOnly (`isReadOnly=true`, `isReadOnly=false`)
- checksum:base64 (`checksum=6JonAXqHXsW6tqP0QV8JVsTUMuw=`)
- checksum:hex (`checksum=e89a27017a875ec5bab6a3f4415f0956c4d432ec`)

### Dates  
eg. `field=2023-11-14T03:35:30.921Z`

- takenBefore
- takenAfter
- updatedBefore
- updatedAfter
- trashedBefore
- trashedAfter

### Pagination
- page (`page=1` - defaults to 1)
- size (`size=250` - defaults to 250)

### Sort
- descending (`order=desc`)
- ascending (`order=asc`)

### File path (full and exact match)  
- originalFileName
- originalPath
- resizePath
- webpPath
- encodedVideoPath

### Exif Metadata
- city
- state
- country
- make
- model
- lensModel

#### TODO (future PRs)
- Sort by other date fields (defaults to `fileCreatedAt`)
- Fuzzy search by filenames and paths


server/src/infra/repositories/asset.repository.ts
```diff
@@ -50,6 +53,134 @@ export class AssetRepository implements IAssetRepository {
...
+        state,
+        country,
+        make,
+        model,
+        lensModel,
+      },
+      _.isUndefined,
+    );
+
+    const withExif = Object.keys(exifWhere).length > 0 || _withExif !== undefined;
```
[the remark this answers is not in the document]
danieldietzler: Oh and also, although not really relevant, we could use `!_isEmpty(exifWhere)` since we're already using lodash excessively ;)

server/test/e2e/asset.e2e-spec.ts
```diff
@@ -31,6 +32,9 @@ const user2Dto = {
   name: 'User 2',
 };
 
+const today = DateTime.fromObject({ year: 2023, month: 11, day: 3 });
```
danieldietzler, line 35: Immich' birthday? Please? :)

etnoy (approved):

server/test/e2e/asset.e2e-spec.ts
```diff
@@ -49,83 +53,498 @@ const makeUploadDto = (options?: { omit: string }): Record<string, any> => {
...
+        const { status, body } = await request(server)
+          .get('/assets')
+          .query(query)
+          .set('Authorization', `Bearer ${user1.accessToken}`);
+
+        expect(status).toBe(200);
+        expect(body.length).toBe(assets.length);
+        for (let i = 0; i < assets.length; i++) {
+          expect(body[i]).toEqual(expect.objectContaining({ id: assets[i].id }));
+        }
```
danieldietzler, line 543: Why not just map assets to `[expect.objectContaining({id: ...}), expect.objectContaining({id: ...}), ...]`? And then `expect(body).toEqual(map)`?

danieldietzler (approved):
Awesome! Really good work. Only some little nitpicks ;)

server/src/domain/asset/asset.service.ts
```diff
@@ -91,6 +93,34 @@ export class AssetService {
...
+  search(authUser: AuthUserDto, dto: AssetSearchDto) {
+    let checksum: Buffer | undefined = undefined;
+
+    if (dto.checksum) {
+      const encoding = dto.checksum.length === 28 ? 'base64' : 'hex';
+      checksum = Buffer.from(dto.checksum, encoding);
+    }
+
+    const enumToOrder = { [AssetOrder.ASC]: 'ASC', [AssetOrder.DESC]: 'DESC' } as const;
+    const order = dto.order ? enumToOrder[dto.order] : undefined;
```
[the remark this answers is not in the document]
jrasm91: I have a personal preference of being explicit rather than implicit in these situations. I think better transparency leads to less confusion/bugs. Plus, this uses typing, while string manipulation probably doesn't.

server/src/domain/asset/asset.service.ts
```diff
@@ -91,6 +93,34 @@ export class AssetService {
     this.configCore = SystemConfigCore.create(configRepository);
   }
 
+  search(authUser: AuthUserDto, dto: AssetSearchDto) {
+    let checksum: Buffer | undefined = undefined;
```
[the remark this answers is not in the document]
jrasm91: I have no idea, but that's not valid syntax. I think option can only be on properties on objects, not in front of a variable declaration.

server/src/infra/repositories/asset.repository.ts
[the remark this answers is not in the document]
jrasm91: I prefer not to use lodash when the functionality is available natively. In this case Object.keys works just as easily IMO.

server/test/e2e/asset.e2e-spec.ts
jrasm91: This also checks the order and makes the diff easier to read when there is a mismatch.

server/test/e2e/asset.e2e-spec.ts
jrasm91: feel free to open a PR to change it :-). We really need a constant for it, I can't be bothered to look it up all the time.

danieldietzler: Well now it's in the milestones at least. So easier to look up than scrolling through PRs

server/test/e2e/asset.e2e-spec.ts
danieldietzler: The order should be checked by my approach as well (not using `expect.arrayContaining`). The diff argument makes sense though

PR #4953, opened 2023-11-11
feat(web): uniform random asset selection in slideshow mode

brighteyed:
## Description
Currently we randomly select a bucket and then randomly select an asset from that bucket. This approach is not truly uniform because it doesn't take into account the number of assets in each  bucket. As a result, assets in buckets with more assets are less likely to be selected.
This PR takes into account the asset count in each bucket. This ensures that each asset has an equal chance of being selected.

### DIFF
--- a/web/src/lib/stores/assets.store.ts
+++ b/web/src/lib/stores/assets.store.ts
@@ -32,7 +32,7 @@ export class AssetBucket {
    */
   bucketHeight!: number;
   bucketDate!: string;
-  totalAssets!: number;
+  bucketCount!: number;
   assets!: AssetResponseDto[];
   cancelToken!: AbortController | null;
   position!: BucketPosition;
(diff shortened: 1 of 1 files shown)


web/src/lib/stores/assets.store.ts
```diff
@@ -158,6 +159,7 @@ export class AssetStore {
       return {
         bucketDate: bucket.timeBucket,
         bucketHeight: height,
+        totalAssets: bucket.count,
```
jrasm91, line 162: Can we call it bucket count instead?

web/src/lib/stores/assets.store.ts
```diff
@@ -313,16 +316,20 @@ export class AssetStore {
...
 
   async getRandomAsset(): Promise<AssetResponseDto | null> {
-    const bucket = this.buckets[Math.floor(Math.random() * this.buckets.length)] || null;
-    if (!bucket) {
-      return null;
-    }
+    let index = Math.floor(Math.random() * this.buckets.reduce((acc, bucket) => acc + bucket.totalAssets, 0));
+    for (const bucket of this.buckets) {
+      if (index < bucket.totalAssets) {
+        if (bucket.assets.length === 0) {
```
jrasm91, line 322: Load bucket already has this check

jrasm91 (approved):
Looks really good!

PR #4955, opened 2023-11-11
chore: re-enable renovate

bo0tzz:


jrasm91 (commented):
I think match path has been replaced with a glob and called matchFilePaths or something now

jrasm91 (approved):

PR #5001, opened 2023-11-13
chore(ml): use strict mypy

mertalev:
## Description

This PR enables strict typing for mypy, and makes various typing improvements in the codebase to be compliant with it. The GH workflow now also checks the export folder, which was an oversight.

# How Has This Been Tested?

I checked that `mypy --strict app/ export/` passes, and tested image classification, CLIP, and facial recognition with queries to the ML service.

### DIFF
--- a/machine-learning/app/models/base.py
+++ b/machine-learning/app/models/base.py
@@ -140,7 +140,7 @@ class InferenceModel(ABC):
 
 
 # HF deep copies configs, so we need to make session options picklable
-class PicklableSessionOptions(ort.SessionOptions): # type: ignore[misc]
+class PicklableSessionOptions(ort.SessionOptions):  # type: ignore[misc]
     def __getstate__(self) -> bytes:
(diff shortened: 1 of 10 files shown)


machine-learning/app/models/base.py
```diff
@@ -139,11 +140,11 @@ def clear_cache(self) -> None:
 
 
 # HF deep copies configs, so we need to make session options picklable
-class PicklableSessionOptions(ort.SessionOptions):
+class PicklableSessionOptions(ort.SessionOptions):  # type: ignore[misc]
```
mertalev, line 143: I'd love to get rid of this comment, but the typing here is unfortunate. onnxruntime doesn't provide type stubs for this class so mypy defaults it to `Any`. Under strict typing, `Any` is not a valid base class to extend from.

jrasm91 (approved):
Thanks for the improvements

PR #5009, opened 2023-11-13
fix(web): better year labels

waclaw66:
This is a rework of year label calculation. It's faster, simplier and more precise.
It shloud give maximum possible count of year labels without their overlapping.
Please post your results to compare. Thanks.

|1.85.0|PR|
|-|-|
|obrazek|obrazek|

### DIFF
--- a/web/src/lib/components/shared-components/scrollbar/scrollbar.svelte
+++ b/web/src/lib/components/shared-components/scrollbar/scrollbar.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
-  import type { AssetStore } from '$lib/stores/assets.store';
+  import type { AssetStore, AssetBucket } from '$lib/stores/assets.store';
+  import type { DateTime } from 'luxon';
   import { fromLocalDateTime } from '$lib/utils/timeline-util';
   import { createEventDispatcher } from 'svelte';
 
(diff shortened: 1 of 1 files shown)


web/src/lib/components/shared-components/scrollbar/scrollbar.svelte
```diff
@@ -23,25 +24,35 @@
...
-    (bucket) =>
-      new Segment({
-        count: bucket.assets.length,
-        height: toScrollY(bucket.bucketHeight),
-        timeGroup: bucket.bucketDate,
-      }),
-  );
+  const calculateSegments = (buckets: AssetBucket[]) => {
+    var height: number = 0;
+    var prev!: Segment;
```
danieldietzler, line 36: Those should use let instead of var

web/src/lib/components/shared-components/scrollbar/scrollbar.svelte
```diff
@@ -23,25 +24,35 @@
...
+    return buckets.map((bucket) => {
+      var segment = new Segment();
```
danieldietzler, line 38: This can be const

web/src/lib/components/shared-components/scrollbar/scrollbar.svelte
```diff
@@ -23,25 +24,35 @@
...
-
-    constructor({ count = 0, height = 0, timeGroup = '' }) {
-      this.count = count;
-      this.height = height;
-      this.timeGroup = timeGroup;
-    }
+    public count: number = 0;
+    public height: number = 0;
+    public timeGroup: string = '';
+    public date!: DateTime;
```
danieldietzler, line 30: I don't think `!:` is necessary here, is it?

web/src/lib/components/shared-components/scrollbar/scrollbar.svelte
danieldietzler: `!:` isn't necessary here either, correct?

web/src/lib/components/shared-components/scrollbar/scrollbar.svelte
```diff
@@ -23,25 +24,35 @@
...
+    var height: number = 0;
+    var prev!: Segment;
+    return buckets.map((bucket) => {
+      var segment = new Segment();
+      segment.count = bucket.assets.length;
+      segment.height = toScrollY(bucket.bucketHeight);
+      segment.timeGroup = bucket.bucketDate;
+      segment.date = fromLocalDateTime(segment.timeGroup);
+
+      if (prev?.date.year != segment.date.year && (!prev || height > 16)) {
```
danieldietzler, line 44: What is that '16'? Also, I believe you can use strict type checking here: `!==`.

web/src/lib/components/shared-components/scrollbar/scrollbar.svelte
```diff
@@ -23,25 +24,35 @@
...
+        segment.hasLabel = true;
+        height = 0;
+      }
+
+      height! += segment.height;
```
danieldietzler, line 49: Since `height` has a default value (0), it should never be undefined, right? So '!' is unnecessary.

web/src/lib/components/shared-components/scrollbar/scrollbar.svelte
waclaw66: Yes, it is.

web/src/lib/components/shared-components/scrollbar/scrollbar.svelte
waclaw66: 16 is minimal distance from previous year label.

jrasm91 (approved):
LGTM

PR #5076, opened 2023-11-16
fix(web): reorder jobs in concurrency view to match job overview

traktuner:
On the admin page, the order of the job overview is different from the "manage concurrency" view.
This reorders the concurrency view to have a better overview, when users jump between the two sections to optimize their job concurrency settings.


jrasm91:
This file is auto generated by the immich server when running in dev mode. You will need to fix it at the source.

traktuner:
> This file is auto generated by the immich server when running in dev mode. You will need to fix it at the source.

sorry about that. it's not reordered in the api entities. hope that's correct now!

jrasm91:
Just change it here:

https://github.com/immich-app/immich/blob/main/web/src/lib/components/admin-page/settings/job-settings/job-settings.svelte#L20

jrasm91:
Closing in favor of https://github.com/immich-app/immich/pull/5386

PR #5099, opened 2023-11-17
fix(mobile): update local deleted assets in sync

fyfrey:
fixes the issue introduced in the recent release that locally (outside of Immich app) removed assets that are backed up are gone from the app until a full refresh is made


alextran1502 (approved):
Thank you for the quick fix, Fynn

alextran1502 (approved):
Thank you for the quick fix, Fynn

PR #5105, opened 2023-11-17
fix(mobile): update password change description text to use user name

shenlong-tanwen:
#### Changes made in the PR:

- Password change form description translation text still had the old `{firstName}` and `{lastName}` placeholders which are now renamed to `{name}`

### DIFF
--- a/mobile/assets/i18n/ca.json
+++ b/mobile/assets/i18n/ca.json
@@ -119,7 +119,7 @@
   "cache_settings_tile_title": "Local Storage",
   "cache_settings_title": "Configuració de la memòria cau",
   "change_password_form_confirm_password": "Confirma la contrasenya",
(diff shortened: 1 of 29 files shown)


jrasm91:
this was in the initial PR, but I think we need to make the change in localizely instead.

shenlong-tanwen:
> this was in the initial PR, but I think we need to make the change in localizely instead.

`en-US` file to which new texts and edits are usually made is also not updated. Also, since this is the placeholder and not the actual text to be translated, I guess we can handle this directly?

jrasm91:
I guess my point is just that I think non-english files might be overwritten when new translations are pulled from the platform (except for english probably).

alextran1502 (approved):

PR #5107, opened 2023-11-17
feat(web): add (lower) thumbnail resolution options 

davidacampos:
Added 200p and 720p as options for small and large thumbnails resolutions respectively.

For scenarios in which users have many assets (like for instance over 750GB of pictures/videos) the generated thumbnails consume considerable storage and time to get processed. The proposed options could alleviate users in such scenarios.

I have tried this and other features like face matching do not seem to be hardly impacted with the reduced size of the "large" thumbnails. Mobile apps looks practically the same, and the Web UI shows a minimal quality degradation, but also a reduced bandwidth consumption in both scenarios.

In my tests (i.e. forcing the proposed values directly via the API), it went from about ~60GB to ~20GB in storage and 12 hours to 5 hours of processing (14 cores with 8GB RAM).

### DIFF
--- a/web/src/lib/components/admin-page/settings/thumbnail/thumbnail-settings.svelte
+++ b/web/src/lib/components/admin-page/settings/thumbnail/thumbnail-settings.svelte
@@ -91,6 +91,7 @@
               { value: 720, text: '720p' },
               { value: 480, text: '480p' },
               { value: 250, text: '250p' },
+              { value: 200, text: '200p' },
             ]}
             name="resolution"
             isEdited={thumbnailConfig.webpSize !== savedConfig.webpSize}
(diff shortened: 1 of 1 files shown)


web/src/lib/components/admin-page/settings/thumbnail/thumbnail-settings.svelte
```diff
@@ -105,6 +106,7 @@
             options={[
               { value: 2160, text: '4K' },
               { value: 1440, text: '1440p' },
+              { value: 720, text: '720p' },
```
jrasm91, line 109: do you want to add 1080p here as well?

jrasm91 (approved):
LGTM

web/src/lib/components/admin-page/settings/thumbnail/thumbnail-settings.svelte
davidacampos: Good point! Just added it.

mertalev:
Face detection resizes images to 640x640 anyway, so 720p shouldn't cause much degradation, if any.

aviv926:
Do you know of a negative effect with object detection and clip when the resolution of the assisted image decreases?

PR #5141, opened 2023-11-19
fix(web): improve year label position

Funk66:
The year label is shown at the top of the segment - i.e. at the end of the year. Not only it makes more sense to have it at the bottom, at least in my opinion, but that's also how Google Photos does it.
Immich:
Screenshot 2023-11-18 at 21 25 33 (2)
Google Photos:
Screenshot 2023-11-18 at 21 23 50 (2)


jrasm91 (approved):

waclaw66:
Destroys the scrollbar...
obrazek

PR #5179, opened 2023-11-20
chore(dev): add pre-commit hooks

mertalev:
## Description

This PR adds a number of pre-commit hooks and configures Renovate to automatically update them. Besides being a general DX improvement, the hope is that Renovate will use the `poetry-lock` hook to produce correct lock files for ML.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,28 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+      - id: check-yaml
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+      - id: mixed-line-ending
+        args: [ --fix=lf ]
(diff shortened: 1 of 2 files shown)


bo0tzz:
Is this expected to fix errors such as https://github.com/immich-app/immich/pull/5180#issuecomment-1818435984 ?

.pre-commit-config.yaml
jrasm91, line 1: Everything else uses `.yml` instead of `.yaml` can this file as well, or does renovate expect `yaml`?

jrasm91 (approved):

.pre-commit-config.yaml
mertalev: It actually needs to be `.yaml` since this is what the git hook itself looks for.

jrasm91:
This didn't actually resolve the poetry lock problem and I'm so-so on these pre-commit hooks in general. Feel free to reopen if you still want to implement something like this.

PR #5199, opened 2023-11-20
fix(server): override date via xmp

YFrendo:
The date have priority if it come from xmp
Related to #4502 and come from #5066

This ensures a date will be used from a sidecar file, if present. Prerequisite for #5066

### DIFF
--- a/server/src/domain/metadata/metadata.service.ts
+++ b/server/src/domain/metadata/metadata.service.ts
@@ -25,6 +25,18 @@ import {
 import { StorageCore } from '../storage';
 import { FeatureFlag, SystemConfigCore } from '../system-config';
 
+/** look for a date from these tags (in order) */
+const EXIF_DATE_TAGS: Array<keyof Tags> = [
+  'SubSecDateTimeOriginal',
+  'DateTimeOriginal',
+  'SubSecCreateDate',
+  'CreationDate',
+  'CreateDate',
+  'SubSecMediaCreateDate',
+  'MediaCreateDate',
+  'DateTimeCreated',
+];
+
 interface DirectoryItem {
   Length?: number;
   Mime: string;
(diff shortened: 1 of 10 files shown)


server/src/domain/metadata/metadata.service.ts
```diff
@@ -340,6 +340,31 @@ export class MetadataService {
...
+      // remove time from mediaTags, since sidecarTags has a more accurate time
+
+      mediaTags.SubSecDateTimeOriginal = undefined;
+      mediaTags.DateTimeOriginal = undefined;
+      mediaTags.SubSecCreateDate = undefined;
+      mediaTags.CreationDate = undefined;
+      mediaTags.CreateDate = undefined;
+      mediaTags.SubSecMediaCreateDate = undefined;
+      mediaTags.MediaCreateDate = undefined;
+      mediaTags.DateTimeCreated = undefined;
```
jrasm91, line 365: Settings things to undefined when they are being merged might lead to weird results in the future.

```typescript
const a = { CreateDate: 'placeholder' };
const b = { CreateDate: undefined };
const c = { ...a, ...b };
```
In this example c.CreateDate will actually be undefined, which I feel like is confusing and misleading. What do you think about just deleting from mediaTags list itself? So, `delete mediaTags.CreateDate` instead?

jrasm91: This is actually the same list of tags as above, so I think these need to be kept in sync. Maybe we can pass a list above, and then loop over this list here and delete the values.

YFrendo: > Settings things to undefined when they are being merged might lead to weird results in the future.
> 
> ```ts
> const a = { CreateDate: 'placeholder' };
> const b = { CreateDate: undefined };
> const c = { ...a, ...b };
> ```
> 
> In this example c.CreateDate will actually be undefined, which I feel like is confusing and misleading. What do you think about just deleting from mediaTags list itself? So, `delete mediaTags.CreateDate` instead?

look like a better idea indeed !

jrasm91 (approved):

jjspierx:
How much work would it take for a PR that does this exact thing using the Google Photos JSON sidecars as well?  Support for these JSON sidecars would be very beneficial for those of using Google Takeout to export our photos.

PR #5223, opened 2023-11-21
chore(server): Prepare access interfaces for bulk permission checks

adamantike:
This change adds the `AccessCore.getAllowedIds` method, to evaluate permissions in bulk, along with some other `getAllowedIds*` private methods.

The added methods still calculate permissions by id, and are not optimized to reduce the amount of queries and execution time, which will be implemented in separate pull requests.

Services that were evaluating permissions in a loop have been refactored to make use of the bulk approach.

### DIFF
--- a/server/src/domain/album/album.service.ts
+++ b/server/src/domain/album/album.service.ts
@@ -144,8 +144,8 @@ export class AlbumService {
     await this.access.requirePermission(authUser, Permission.ALBUM_READ, id);
 
     const existingAssetIds = await this.albumRepository.getAssetIds(id, dto.ids);
-    const notPresentAssetIds = new Set(dto.ids.filter((id) => !existingAssetIds.has(id)));
-    const allowedAssetIds = await this.access.getAllowedIds(authUser, Permission.ASSET_SHARE, notPresentAssetIds);
+    const notPresentAssetIds = dto.ids.filter((id) => !existingAssetIds.has(id));
+    const allowedAssetIds = await this.access.checkAccess(authUser, Permission.ASSET_SHARE, notPresentAssetIds);
 
     const results: BulkIdResponseDto[] = [];
     for (const assetId of dto.ids) {
@@ -183,12 +183,10 @@ export class AlbumService {
     await this.access.requirePermission(authUser, Permission.ALBUM_READ, id);
 
     const existingAssetIds = await this.albumRepository.getAssetIds(id, dto.ids);
-    const assetIdsWithRemoveAccess = await this.access.getAllowedIds(
-      authUser,
-      Permission.ALBUM_REMOVE_ASSET,
-      existingAssetIds,
-    );
-    const assetIdsWithShareAccess = await this.access.getAllowedIds(authUser, Permission.ASSET_SHARE, existingAssetIds);
+    const allowedAssetIds = new Set([
+      ...(await this.access.checkAccess(authUser, Permission.ALBUM_REMOVE_ASSET, existingAssetIds)),
+      ...(await this.access.checkAccess(authUser, Permission.ASSET_SHARE, existingAssetIds)),
+    ]);
 
     const results: BulkIdResponseDto[] = [];
     for (const assetId of dto.ids) {
@@ -198,7 +196,7 @@ export class AlbumService {
         continue;
       }
 
-      const hasAccess = assetIdsWithRemoveAccess.has(assetId) || assetIdsWithShareAccess.has(assetId);
+      const hasAccess = allowedAssetIds.has(assetId);
       if (!hasAccess) {
         results.push({ id: assetId, success: false, error: BulkIdErrorReason.NO_PERMISSION });
         continue;
--- a/server/src/domain/album/album.service.ts
+++ b/server/src/domain/album/album.service.ts
@@ -144,6 +144,8 @@ export class AlbumService {
     await this.access.requirePermission(authUser, Permission.ALBUM_READ, id);
 
     const existingAssetIds = await this.albumRepository.getAssetIds(id, dto.ids);
+    const notPresentAssetIds = new Set(dto.ids.filter((id) => !existingAssetIds.has(id)));
+    const allowedAssetIds = await this.access.getAllowedIds(authUser, Permission.ASSET_SHARE, notPresentAssetIds);
 
     const results: BulkIdResponseDto[] = [];
     for (const assetId of dto.ids) {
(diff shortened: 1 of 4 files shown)


jrasm91 (commented):
What I liked about how this worked before was how clean and intuitive the interfaces were for checking access. It is a bit more complicated now.

What do you think about adding a new method `checkAccess` that returns back a map of IDs mapped to whether or not they have access.

Then the change everywhere else is removing the lookup from inside the loop, moving it to check access at the top, and replacing the line in the loop with `access[id]` instead.

adamantike:
I think the approaches we are considering are very similar, if I improve the current `getAllowedIds` function name to `checkAccess`, as you suggest.

I considered returning a map, but found a few inconveniences that would make the solution less performant and more complex for some scenarios:
* The current logic for `hasPermission` requires returning a boolean depending on whether the user has access to **all provided ids**. By returning a map, we would need to iterate over all its values, looking for any `false`. With a set, we only validate its size is equal to the amount of received ids.
* Same scenario for `hasAny`, which would also involve iterating over the map values, and only returning early when `true` is found.
* I found some checks where the same id is evaluated for multiple permissions, and if any of them is valid, the check is successful. Having a map for each of those checks would require, in the worst scenario, `numAssets * numPermissions` lookups, as we would need `access1[id] || access2[id] || ...` for each of them. I'm thinking about running a single set union in those scenarios, and reduce the complexity to a single lookup per asset.

jrasm91:
Maybe I am just being picky, but it seems like you have added a lot of boilerplate code to work with sets, specifically on the calling side. I just want the code to be clean, concise, and easy to read, and I feel like these changes, right now at least, make it less simple.

Maybe a few small changes would help improve things:
- Rename `hasPermission` to `checkAccess`
- Update checkAccess to accept a list instead of a set
- `requirePermission` throws an error if the `hasPermission` length mismatches`
- Remove the `hasAny` implementation and swap it out with two separate calls to `checkAccess` (it is only used in one place)

What do you think?

adamantike:
Please, I really appreciate the feedback!

> it seems like you have added a lot of boilerplate code to work with sets, specifically on the calling side.

I'm not very familiar with TypeScript, but ideally this would be an `Iterable<string>` to handle any structure that can be iterated over. I didn't go the Protocol-based way, as I didn't find a clean approach to get that, while excluding `string` from being considered an iterable too.

I have changed the signature to `Set<string> | string[]` to support both at the moment, as I think the small addition to support both is worth it in case we want to chain calls to `checkAccess`.

> Maybe a few small changes would help improve things

Great suggestions! I have implemented all of them, so now there's only `checkAccess` and `requirePermission` as the public permissions API.

server/src/domain/person/person.service.ts
```diff
@@ -375,10 +375,11 @@ export class PersonService {
...
     const results: BulkIdResponseDto[] = [];
 
-    for (const mergeId of mergeIds) {
-      const hasPermission = await this.access.hasPermission(authUser, Permission.PERSON_MERGE, mergeId);
+    const allowedIds = await this.access.checkAccess(authUser, Permission.PERSON_MERGE, mergeIds);
 
-      if (!hasPermission) {
+    for (const mergeId of mergeIds) {
+      const hasAccess = allowedIds.has(mergeId);
+      if (!hasAccess) {
```
jrasm91, line 382: This looks good to me.

server/src/domain/shared-link/shared-link.service.ts
```diff
@@ -119,15 +119,19 @@ export class SharedLinkService {
...
     }
 
+    const existingAssetIds = new Set(sharedLink.assets.map((asset) => asset.id));
+    const notPresentAssetIds = dto.assetIds.filter((assetId) => !existingAssetIds.has(assetId));
+    const allowedAssetIds = await this.access.checkAccess(authUser, Permission.ASSET_SHARE, notPresentAssetIds);
+
     const results: AssetIdsResponseDto[] = [];
     for (const assetId of dto.assetIds) {
-      const hasAsset = sharedLink.assets.find((asset) => asset.id === assetId);
+      const hasAsset = existingAssetIds.has(assetId);
```
jrasm91, line 128: We should change this to not require loading all the assets into memory for this to work, but I think this is the best we can do in the meantime.

server/src/domain/album/album.service.ts
```diff
@@ -181,6 +183,10 @@ export class AlbumService {
     await this.access.requirePermission(authUser, Permission.ALBUM_READ, id);
 
     const existingAssetIds = await this.albumRepository.getAssetIds(id, dto.ids);
+    const allowedAssetIds = new Set([
+      ...(await this.access.checkAccess(authUser, Permission.ALBUM_REMOVE_ASSET, existingAssetIds)),
+      ...(await this.access.checkAccess(authUser, Permission.ASSET_SHARE, existingAssetIds)),
+    ]);
```
jrasm91, line 189: This works, it might be a bit cleaner if these were two separate sets, like `canRemove` or `canShare` and then it was used below like `const hasAccess = canRemove.has(id) || canShare.has(id)`, but overall this looks good too.

jrasm91 (approved):
This change is pretty small, we can merge it as is, or because it is pretty small we can look at porting some calls to bulk here as well.

server/src/domain/shared-link/shared-link.service.ts
adamantike: For this an other places where we get `existingAssetIds` in memory, I think we can have a separate PR where we chunk the ids to limit the amount of them in memory. There are currently three places where the same pattern could be improved, so I'd prefer to maintain it separate from this change.

server/src/domain/album/album.service.ts
adamantike: Do you find this more readable, while we still maintain a single map lookup per id?

```javascript
const canRemove = await this.access.checkAccess(authUser, Permission.ALBUM_REMOVE_ASSET, existingAssetIds);
const canShare = await this.access.checkAccess(authUser, Permission.ASSET_SHARE, existingAssetIds);
const allowedAssetIds = new Set([...canRemove, ...canShare]);
```

server/src/domain/shared-link/shared-link.service.ts
jrasm91: Yes, agreed

server/src/domain/album/album.service.ts
jrasm91: That looks better to me, yes. Thanks

jrasm91 (approved):

Issue #4907, opened 2023-11-08
[BUG] web: album sorting method display problem


waclaw66:
### The bug

Album sorting method is shown always as "Album title" once Albums page is reloaded, however sorting method is kept.

![obrazek](https://github.com/immich-app/immich/assets/15554561/d38310d5-a29c-4d40-841f-4fb7bc9b35a4)

### The OS that Immich Server is running on

Fedora 38

### Version of Immich Server

v.1.85.0

### Version of Immich Mobile App

v.1.85.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

```
1. set album sort to e.g. "Most recent photo"
2. reload page by F5
3. albums are still sorted as most recent, but "Album title" method is displayed
```

waclaw66:
Fixed by https://github.com/immich-app/immich/pull/5233

PR #5233, opened 2023-11-21
fix(web): sorting options for albums

martabal:
Re-implementation of #5127 without the VirtualColumn for better perfomances


server/src/domain/album/dto/album.dto.ts
```diff
@@ -6,5 +6,5 @@ export class AlbumInfoDto {
   @Optional()
   @IsBoolean()
   @Transform(toBoolean)
-  withoutAssets?: boolean;
+  withoutAssets?: boolean = false;
```
jrasm91, line 9: Why does this need to default to false here? Isn't undefined falsy already?

server/src/domain/album/album.service.ts
```diff
@@ -69,18 +70,28 @@ export class AlbumService {
     // Get asset count for each album. Then map the result to an object:
     // { [albumId]: assetCount }
     const albumsAssetCount = await this.albumRepository.getAssetCountForIds(albums.map((album) => album.id));
-    const albumsAssetCountObj = albumsAssetCount.reduce((obj: Record<string, number>, { albumId, assetCount }) => {
-      obj[albumId] = assetCount;
-      return obj;
-    }, {});
+    const albumsAssetCountObj: Record<string, AlbumAssetCount> = albumsAssetCount.reduce(
```
jrasm91, line 73: If we're going to update this stuff, we should rename it to `getMetadataForIds` instead, and call it "album metadata" everywhere.

waclaw66:
Seems it fixes https://github.com/immich-app/immich/issues/4907

alextran1502:
When I navigate to the album, I see this 

image

The add photos button is visible, are you seeing this as well?

waclaw66:
> When I navigate to the album, I see this
> 
> The add photos button is visible, are you seeing this as well?

Can confirm...
obrazek

server/src/domain/album/album.service.ts
```diff
@@ -90,7 +101,8 @@ export class AlbumService {
   async get(authUser: AuthUserDto, id: string, dto: AlbumInfoDto): Promise<AlbumResponseDto> {
     await this.access.requirePermission(authUser, Permission.ALBUM_READ, id);
     await this.albumRepository.updateThumbnails();
-    return mapAlbum(await this.findOrFail(id, { withAssets: true }), !dto.withoutAssets);
+    const withAssets = dto.withoutAssets === undefined ? true : !dto.withoutAssets;
```
jrasm91, line 104: If this isn't specified it should use `false`

jrasm91 (approved):

martabal:
> > When I navigate to the album, I see this
> > The add photos button is visible, are you seeing this as well?
> 
> Can confirm. Own album acts like album shared with me.
> actual result 	expected
> obrazek 	obrazek

Not really, they act like if they were empty. Thanks for the report, should have been fixed :smile:

martabal:
Mmmh, it seems there's an issue with the shared links ...

web/src/lib/components/sharedlinks-page/shared-link-card.svelte
```diff
@@ -60,18 +61,28 @@
   class="flex w-full gap-4 border-b border-gray-200 py-4 transition-all hover:border-immich-primary dark:border-gray-600 dark:text-immich-gray dark:hover:border-immich-dark-primary"
 >
   <div>
-    {#await getAssetInfo()}
-      <LoadingSpinner />
-    {:then asset}
+    {#if link?.album?.albumThumbnailAssetId || link.assets.length > 0}
```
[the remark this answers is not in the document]
martabal: This was already here in https://github.com/immich-app/immich/pull/5127. If your album does not have a thumbnail (empty album for example), the client makes a useless API request (`/api/asset/thumbnail/undefined?format=WEBP`). It's not direclty related to the albums, I can open a PR just for this if you want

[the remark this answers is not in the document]
jrasm91: Oh, I didn't realize this was specifically for the thumbnail. Do you mind updating the method to be like get thumbnail data/info or something else?

jrasm91 (approved):

jrasm91 (approved):
I haven't tested this, but the code looks good assuming the SQL for the new queries are performant.

Can you show the before and after SQL for the affected queries and their explain plans?

martabal:
> I haven't tested this, but the code looks good assuming the SQL for the new queries are performant.
> 
> Can you show the before and after SQL for the affected queries and their explain plans?

SQL query before : 
```sql
SELECT
    "album"."id" AS "album_id",
    COUNT("albums_assets"."assetsId") AS "asset_count"
FROM
    "albums" "album"
    LEFT JOIN "albums_assets_assets" "albums_assets" ON "albums_assets"."albumsId" = "album"."id"
WHERE
    "album"."id" IN (
        '733ca3ac-a14a-494b-843e-1857c97ce8eb',
        'b0bab205-55b0-479a-99a3-c4a7ef2e17ec',
        '70bf5246-031c-4c91-97cc-627bbfe3eb7f',
        '644bcfae-fd76-45cd-8b72-a649874d1ef1'
    )
    AND "album"."deletedAt" IS NULL
GROUP BY
    "album"."id";
```

SQL query after : 
```sql
SELECT
    "album"."id" AS "album_id",
    MIN("assets"."fileCreatedAt") AS "start_date",
    MAX("assets"."fileCreatedAt") AS "end_date",
    COUNT("album_assets"."assetsId") AS "asset_count"
FROM
    "albums" "album"
    LEFT JOIN "albums_assets_assets" "album_assets" ON "album_assets"."albumsId" = "album"."id"
    LEFT JOIN "assets" "assets" ON "assets"."id" = "album_assets"."assetsId" AND "assets"."deletedAt" IS NULL
WHERE
    "album"."id" IN (
        '733ca3ac-a14a-494b-843e-1857c97ce8eb',
        'b0bab205-55b0-479a-99a3-c4a7ef2e17ec',
        '70bf5246-031c-4c91-97cc-627bbfe3eb7f',
        '644bcfae-fd76-45cd-8b72-a649874d1ef1'
    )
    AND "album"."deletedAt" IS NULL
GROUP BY
    "album"."id";
```

<details><summary>Explain plan before</summary>

```sql
GroupAggregate  (cost=32.64..32.67 rows=1 width=24)
  Group Key: album.id
  ->  Sort  (cost=32.64..32.65 rows=2 width=32)
        Sort Key: album.id
        ->  Hash Right Join  (cost=17.10..32.63 rows=2 width=32)
              Hash Cond: (albums_assets."albumsId" = album.id)
              ->  Seq Scan on albums_assets_assets albums_assets  (cost=0.00..13.53 rows=753 width=32)
              ->  Hash  (cost=17.09..17.09 rows=1 width=16)
                    ->  Bitmap Heap Scan on albums album  (cost=8.62..17.09 rows=1 width=16)
                          Recheck Cond: (id = ANY ('{733ca3ac-a14a-494b-843e-1857c97ce8eb,b0bab205-55b0-479a-99a3-c4a7ef2e17ec,70bf5246-031c-4c91-97cc-627bbfe3eb7f,644bcfae-fd76-45cd-8b72-a649874d1ef1}'::uuid[]))
                          Filter: ("deletedAt" IS NULL)
                          ->  Bitmap Index Scan on "PK_7f71c7b5bc7c87b8f94c9a93a00"  (cost=0.00..8.62 rows=4 width=0)
                                Index Cond: (id = ANY ('{733ca3ac-a14a-494b-843e-1857c97ce8eb,b0bab205-55b0-479a-99a3-c4a7ef2e17ec,70bf5246-031c-4c91-97cc-627bbfe3eb7f,644bcfae-fd76-45cd-8b72-a649874d1ef1}'::uuid[]))
```
</details>

<details><summary>Explain plan after</summary>

```sql
GroupAggregate  (cost=33.66..33.70 rows=1 width=40)
  Group Key: album.id
  ->  Sort  (cost=33.66..33.67 rows=2 width=40)
        Sort Key: album.id
        ->  Nested Loop Left Join  (cost=17.38..33.65 rows=2 width=40)
              ->  Hash Right Join  (cost=17.10..32.63 rows=2 width=32)
                    Hash Cond: (album_assets."albumsId" = album.id)
                    ->  Seq Scan on albums_assets_assets album_assets  (cost=0.00..13.53 rows=753 width=32)
                    ->  Hash  (cost=17.09..17.09 rows=1 width=16)
                          ->  Bitmap Heap Scan on albums album  (cost=8.62..17.09 rows=1 width=16)
                                Recheck Cond: (id = ANY ('{733ca3ac-a14a-494b-843e-1857c97ce8eb,b0bab205-55b0-479a-99a3-c4a7ef2e17ec,70bf5246-031c-4c91-97cc-627bbfe3eb7f,644bcfae-fd76-45cd-8b72-a649874d1ef1}'::uuid[]))
                                Filter: ("deletedAt" IS NULL)
                                ->  Bitmap Index Scan on "PK_7f71c7b5bc7c87b8f94c9a93a00"  (cost=0.00..8.62 rows=4 width=0)
                                      Index Cond: (id = ANY ('{733ca3ac-a14a-494b-843e-1857c97ce8eb,b0bab205-55b0-479a-99a3-c4a7ef2e17ec,70bf5246-031c-4c91-97cc-627bbfe3eb7f,644bcfae-fd76-45cd-8b72-a649874d1ef1}'::uuid[]))
              ->  Index Scan using "PK_da96729a8b113377cfb6a62439c" on assets  (cost=0.28..0.51 rows=1 width=24)
                    Index Cond: (id = album_assets."assetsId")
                    Filter: ("deletedAt" IS NULL)
```
</details>

jrasm91:
Looks good!

alextran1502 (approved):

waclaw66:
Is it expected that browser freezes for ~7s while switching between order options and consumes a lot of cpu power (~50%)?

martabal:
Interesting, I have some questions :
Do you experience that behavior with the list mode and the cover mode ? Do you experience that freeze with every sorts ?

waclaw66:
List mode is almost instant, all modes take about the same time in cover mode. The issue concerns latest release as well.

martabal:
From your Firefox recording, I think it's related to the cover animation. 250 albums is quiet big :sweat_smile:

waclaw66:
Is there any solution for that? Shall I create an issue?

waclaw66:
> 250 albums is quiet big 😅

Not really, I expect to have much more, just started transition from GP ;)
Therefore https://github.com/immich-app/immich/discussions/5078 would be really helpful.

jrasm91:
Maybe it is related to loading full sized thumbnails for the albums in card view.

martabal:
Animating 250 images can be overwhelming. I think the solution would be moving from loading the entire album collection to partial loading with pagination / infinite scrolling. This would be ideal for large album collections. Not really an issue IMO.

waclaw66:
> Maybe it is related to loading full sized thumbnails for the albums in card view.

Don't think so, all cover JPEGs are already downloaded and cached, there are no additional requests while switching sorts.
It's fully client side (JS or CSS) related.

waclaw66:
Worth to mention, that the first change of sort method after Album page refresh is always fast, the others take severals seconds. That's a bit weird.

martabal:
Yes, it's probably due to the flip animation

PR #5288, opened 2023-11-24
Update libraries.md - External Library step by step set up guide added

jcianci12:
When setting up immmich external libraries, I found some missing details around the relationship between external paths and import paths. By providing an example, I am hoping it helps new users.


danieldietzler (approved):
Sounds like a welcome addition to me!

alextran1502 (approved):

PR #5290, opened 2023-11-24
chore(server): Check album permissions in bulk

adamantike:
Modify Access repository, to evaluate `album` permissions in bulk. Queries have been validated to match what they currently generate for single ids.

Queries:

* Owner access:

```sql
-- Before
SELECT 1 AS "row_exists" FROM (SELECT 1 AS dummy_column) "dummy_table" WHERE EXISTS (
  SELECT 1
  FROM "albums" "AlbumEntity"
  WHERE
    "AlbumEntity"."id" = $1
    AND "AlbumEntity"."ownerId" = $2
    AND "AlbumEntity"."deletedAt" IS NULL
)
LIMIT 1

-- After
SELECT
  "AlbumEntity"."id" AS "AlbumEntity_id"
FROM "albums" "AlbumEntity"
WHERE
  "AlbumEntity"."id" IN ($1, $2)
  AND "AlbumEntity"."ownerId" = $3
  AND "AlbumEntity"."deletedAt" IS NULL
```

* Shared link access:

```sql
-- Before
SELECT 1 AS "row_exists" FROM (SELECT 1 AS dummy_column) "dummy_table" WHERE EXISTS (
  SELECT 1
  FROM "shared_links" "SharedLinkEntity"
  WHERE
    "SharedLinkEntity"."id" = $1
    AND "SharedLinkEntity"."albumId" = $2
)
LIMIT 1

-- After
SELECT
  "SharedLinkEntity"."albumId" AS "SharedLinkEntity_albumId",
  "SharedLinkEntity"."id" AS "SharedLinkEntity_id"
FROM "shared_links" "SharedLinkEntity"
WHERE
  "SharedLinkEntity"."id" = $1
  AND "SharedLinkEntity"."albumId" IN ($2, $3)
```

* Shared album access:

```sql
-- Before
SELECT 1 AS "row_exists" FROM (SELECT 1 AS dummy_column) "dummy_table" WHERE EXISTS (
  SELECT 1
  FROM "albums" "AlbumEntity"
    LEFT JOIN "albums_shared_users_users" "AlbumEntity_AlbumEntity__AlbumEntity_sharedUsers"
      ON "AlbumEntity_AlbumEntity__AlbumEntity_sharedUsers"."albumsId"="AlbumEntity"."id"
    LEFT JOIN "users" "AlbumEntity__AlbumEntity_sharedUsers"
      ON "AlbumEntity__AlbumEntity_sharedUsers"."id"="AlbumEntity_AlbumEntity__AlbumEntity_sharedUsers"."usersId"
      AND "AlbumEntity__AlbumEntity_sharedUsers"."deletedAt" IS NULL
  WHERE
    "AlbumEntity"."id" = $1
    AND "AlbumEntity__AlbumEntity_sharedUsers"."id" = $2
    AND "AlbumEntity"."deletedAt" IS NULL
)
LIMIT 1

-- After
SELECT
  "AlbumEntity"."id" AS "AlbumEntity_id"
FROM "albums" "AlbumEntity"
  LEFT JOIN "albums_shared_users_users" "AlbumEntity_AlbumEntity__AlbumEntity_sharedUsers"
    ON "AlbumEntity_AlbumEntity__AlbumEntity_sharedUsers"."albumsId"="AlbumEntity"."id"
  LEFT JOIN "users" "AlbumEntity__AlbumEntity_sharedUsers"
    ON "AlbumEntity__AlbumEntity_sharedUsers"."id"="AlbumEntity_AlbumEntity__AlbumEntity_sharedUsers"."usersId"
    AND "AlbumEntity__AlbumEntity_sharedUsers"."deletedAt" IS NULL
WHERE
  "AlbumEntity"."id" IN ($1, $2)
  AND "AlbumEntity__AlbumEntity_sharedUsers"."id" = $3
  AND "AlbumEntity"."deletedAt" IS NULL
```

### DIFF
--- a/server/src/domain/access/access.core.ts
+++ b/server/src/domain/access/access.core.ts
@@ -99,6 +99,24 @@ export class AccessCore {
   }
 
   private async checkAccessSharedLink(authUser: AuthUserDto, permission: Permission, ids: Set<string>) {
+    const sharedLinkId = authUser.sharedLinkId;
+    if (!sharedLinkId) {
+      return new Set();
+    }
+
+    switch (permission) {
+      case Permission.ASSET_UPLOAD:
+        return authUser.isAllowUpload ? ids : new Set();
+
+      case Permission.ALBUM_READ:
+        return await this.repository.album.checkSharedLinkAccess(sharedLinkId, ids);
+
+      case Permission.ALBUM_DOWNLOAD:
+        return !!authUser.isAllowDownload
+          ? await this.repository.album.checkSharedLinkAccess(sharedLinkId, ids)
+          : new Set();
+    }
+
     const allowedIds = new Set();
     for (const id of ids) {
       const hasAccess = await this.hasSharedLinkAccess(authUser, permission, id);
@@ -126,18 +144,9 @@ export class AccessCore {
       case Permission.ASSET_DOWNLOAD:
         return !!authUser.isAllowDownload && (await this.repository.asset.hasSharedLinkAccess(sharedLinkId, id));
 
-      case Permission.ASSET_UPLOAD:
-        return authUser.isAllowUpload;
-
       case Permission.ASSET_SHARE:
         // TODO: fix this to not use authUser.id for shared link access control
         return this.repository.asset.hasOwnerAccess(authUser.id, id);
-
-      case Permission.ALBUM_READ:
-        return this.repository.album.hasSharedLinkAccess(sharedLinkId, id);
-
-      case Permission.ALBUM_DOWNLOAD:
-        return !!authUser.isAllowDownload && (await this.repository.album.hasSharedLinkAccess(sharedLinkId, id));
 
       default:
         return false;
@@ -145,6 +154,32 @@ export class AccessCore {
   }
 
   private async checkAccessOther(authUser: AuthUserDto, permission: Permission, ids: Set<string>) {
+    switch (permission) {
+      case Permission.ALBUM_READ: {
+        const isOwner = await this.repository.album.checkOwnerAccess(authUser.id, ids);
+        const isShared = await this.repository.album.checkSharedAlbumAccess(authUser.id, ids);
+        return new Set([...isOwner, ...isShared]);
+      }
+
+      case Permission.ALBUM_UPDATE:
+        return this.repository.album.checkOwnerAccess(authUser.id, ids);
+
+      case Permission.ALBUM_DELETE:
+        return this.repository.album.checkOwnerAccess(authUser.id, ids);
+
+      case Permission.ALBUM_SHARE:
+        return this.repository.album.checkOwnerAccess(authUser.id, ids);
+
+      case Permission.ALBUM_DOWNLOAD: {
+        const isOwner = await this.repository.album.checkOwnerAccess(authUser.id, ids);
+        const isShared = await this.repository.album.checkSharedAlbumAccess(authUser.id, ids);
+        return new Set([...isOwner, ...isShared]);
+      }
+
+      case Permission.ALBUM_REMOVE_ASSET:
+        return this.repository.album.checkOwnerAccess(authUser.id, ids);
+    }
+
     const allowedIds = new Set();
     for (const id of ids) {
       const hasAccess = await this.hasOtherAccess(authUser, permission, id);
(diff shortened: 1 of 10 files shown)


server/src/domain/access/access.core.ts
```diff
@@ -126,25 +144,42 @@ export class AccessCore {
...
       default:
         return false;
     }
   }
 
   private async checkAccessOther(authUser: AuthUserDto, permission: Permission, ids: Set<string>) {
+    switch (permission) {
+      case Permission.ALBUM_READ: {
+        const isOwner = await this.repository.album.checkOwnerAccess(authUser.id, ids);
+        const isShared = await this.repository.album.checkSharedAlbumAccess(authUser.id, ids);
```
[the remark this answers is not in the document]
adamantike: Good catch! I prefer to maintain this simpler API instead of starting to add more complex queries. That's the main reason I'm splitting these changes in multiple PRs, to validate that the queries match and we don't introduce any security issues.

I have added some set utils, and we now call the second check with the ids that weren't already approved by the first one :)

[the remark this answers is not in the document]
jrasm91: Oh perfect!

server/src/domain/access/access.core.ts
```diff
@@ -126,25 +145,42 @@ export class AccessCore {
...
       default:
         return false;
     }
   }
 
   private async checkAccessOther(authUser: AuthUserDto, permission: Permission, ids: Set<string>) {
+    switch (permission) {
+      case Permission.ALBUM_READ: {
+        const isOwner = await this.repository.album.checkOwnerAccess(authUser.id, ids);
+        const isShared = await this.repository.album.checkSharedAlbumAccess(authUser.id, setDifference(ids, isOwner));
```
jrasm91, line 161: Did this still execute a query with zero records?

jrasm91 (commented):
Looks pretty good, we're still exercising two queries for albums I think, even if there are zero records. Other than that it looks good.

server/src/domain/access/access.core.ts
```diff
@@ -1,4 +1,5 @@
 import { BadRequestException, UnauthorizedException } from '@nestjs/common';
+import { setDifference, setUnion } from '..';
```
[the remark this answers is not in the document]
adamantike: Fixed!

server/src/domain/access/access.core.ts
adamantike: All access repository methods I'm adding check for set size, and return early if the provided set is empty. I think encapsulating the logic there will simplify the code and reduce the need for each caller to validate beforehand.

server/src/domain/domain.util.ts
```diff
@@ -150,3 +150,15 @@ export function Optional({ nullable, ...validationOptions }: OptionalOptions = {
...
+// NOTE: The following Set utils have been added here, to easily determine where they are used.
+//       They should be replaced with native Set operations, when they are added to the language.
+//       Proposal reference: https://github.com/tc39/proposal-set-methods
+
+export const setUnion = <T>(setA: Set<T>, setB: Set<T>): Set<T> => {
+  return new Set([...setA, ...setB]);
+};
+
+export const setDifference = <T>(setA: Set<T>, setB: Set<T>): Set<T> => {
+  return new Set([...setA].filter((x) => !setB.has(x)));
```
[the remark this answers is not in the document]
adamantike: I have changed the implementation to match the [snippets](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#implementing_basic_set_operations) provided by MDN docs, which rely more on iterating over sets without building temporary arrays.

server/src/infra/repositories/access.repository.ts
```diff
@@ -209,33 +209,60 @@ export class AccessRepository implements IAccessRepository {
...
+      return this.albumRepository
+        .find({
+          select: { id: true },
+          where: {
+            id: In([...albumIds]),
+            ownerId: userId,
+          },
+        })
+        .then((albums) => {
+          return new Set(albums.map((album) => album.id));
```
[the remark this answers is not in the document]
adamantike: Good call!

server/src/domain/access/access.core.ts
jrasm91: The case I am talking about is for a single album id. It will wait for is owned and then (, even if that is already true), it will wait for is shared to run. I am just mentioning that this is different than how it used to work and the number of calls for the album checks is now 2 sequential _always_, instead of one and _sometimes_ two calls.

jrasm91: This is only applicable to cases that need one of many options to be fulfilled for it to be true. Maybe this is good enough for now. In the future I may look at combining these cases into a single database request instead of two

jrasm91 (approved):

server/src/domain/access/access.core.ts
adamantike: If `ids` has a single id, and `checkOwnerAccess` validates that the id is accessible, then `isOwner` will also be a set with that single id.

In that scenario, `setDifference(ids, isOwner)` will return an empty set. `checkSharedAlbumAccess` will return early without actually making a second query, because of its first conditional `if (albumIds.size === 0) { return new Set(); }`. That should be matching the same behavior as we have today

jrasm91: Ah, I misunderstood that by access checks you were referring to the repo calls as well. Yes, this works great!

PR #5293, opened 2023-11-24
feat(web, server): Allow use of external geocoder service

l0nax:
This PR adds the ability to use an external service to do the reverse geocoding work.
The reason to allow an external service instead of extending the current implementation is that accurate geocoding is computationally expensive. I. e., self-hosting [Nominatim](https://nominatim.org/) would require at least 64 GiB of RAM.
The other reason is that it allows users to use any service they want without having Immich to support a particular service.

The _Maps & GPS Settings_ part looks like this:
screenshot_2023-11-24_16-59_immich-settings-custom-geocode-server

The drawback of this implementation is, that after the configuration has been changed, the server must be restarted.

Please note that I am not familiar with nestjs. If I violated some paradigms or principles this project follows I am more than happy to rewrite the solution.
This is especially true about the factory to return the different implementations.

An example (PoC) service can be found here: https://gitlab.com/l0nax/immich-external-reverse-geocode

Fixes #2684

### DIFF
--- a/server/immich-openapi-specs.json
+++ b/server/immich-openapi-specs.json
@@ -9070,13 +9070,21 @@
           "citiesFileOverride": {
             "$ref": "#/components/schemas/CitiesFile"
           },
+          "customEndpoint": {
+            "type": "string"
+          },
           "enabled": {
+            "type": "boolean"
+          },
+          "useCustomService": {
             "type": "boolean"
           }
         },
         "required": [
           "citiesFileOverride",
-          "enabled"
+          "enabled",
+          "useCustomService",
+          "customEndpoint"
         ],
         "type": "object"
       },
--- a/server/src/domain/metadata/metadata.service.ts
+++ b/server/src/domain/metadata/metadata.service.ts
@@ -103,7 +103,7 @@ export class MetadataService {
     }
 
     const { reverseGeocoding } = await this.configCore.getConfig();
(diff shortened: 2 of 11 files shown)


bo0tzz:
nb: The existing reverse geocoding implementation is currently undergoing a rework to utilize postgres for storing and querying the data. (see https://github.com/immich-app/immich/tree/feat/postgres-reverse-geocoding, cc @zackpollard)

Is the API contract with the external geocoding service here a standard interface, or is it only compatible with the example you linked?

zackpollard:
Before undergoing this work we had a big discussion internally around using an external implementation and decided against it. Our new implementation will undergo improvements to make it accurate down to specific landmarks with only a couple gigabytes of disk space and basically no memory footprint impact.

zackpollard:
My initial response to this PR, given that we are undergoing work to improve the reverse geocoding to improve accuracy, memory footprint and reliability is that we won't be accepting this PR. It seems like very niche use case and not something we would want to support going forwards. I hope this doesn't put you off contributing to Immich in the future however, but to avoid this kind of thing going forwards we are very active in discord or responding to gihub discussions, so it is always a good idea to run upcoming PRs by the team before investing the time, nobody likes to see wasted effort. Hopefully the upcoming improvements live up to what you wanted from this implementation, have a great weekend! 😊

PR #5301, opened 2023-11-25
feat: postgres reverse geocoding

zackpollard:
This PR removes the dependency on local-reverse-geocoder and instead ingests the geonames data for cities500 directly into postgres. This vastly reduces the memory utilisation of the microservices container and also should fix a lot of bugs people have had with local-reverse-geocoder during download and setup of the data.

TODO:

- [x] Update base image used to the latest once the build has finished, required for the geonames data to exist in the image


server/src/domain/metadata/metadata.service.ts
```diff
@@ -258,8 +251,9 @@ export class MetadataService {
     }
 
     try {
-      const { city, state, country } = await this.repository.reverseGeocode({ latitude, longitude });
-      Object.assign(exifData, { city, state, country });
+      const reverseGeocode = await this.repository.reverseGeocode({ latitude, longitude });
+      if(!reverseGeocode) return
+      Object.assign(exifData, { ...reverseGeocode });
```
jrasm91, line 256: Don't need to deconstruct the reverse geocoder object here

server/src/infra/repositories/metadata.repository.ts
```diff
@@ -1,77 +1,179 @@
...
-          dumpDirectory: REVERSE_GEOCODING_DUMP_DIRECTORY,
-          ...options,
-        },
-        resolve,
-      );
+  async init(): Promise<void> {
+    this.logger.log('Initializing metadata repository');
+    const geodataDate = await readFile('../resources/geodata-date.txt', 'utf8');
+
+    await this.geodataPlacesRepository.query('SELECT pg_advisory_lock(69420)');
```
jrasm91, line 37: What is this lock id, can we make it a constant?

server/src/infra/repositories/metadata.repository.ts
```diff
@@ -1,77 +1,179 @@
...
+      this.logger.fatal('Error importing geodata', e);
+      await queryRunner.rollbackTransaction();
+      throw e;
+    } finally {
+      await queryRunner.release();
+    }
+
+    await this.systemMetadataRepository.set(SystemMetadataKey.REVERSE_GEOCODING_STATE, {
+      lastUpdate: geodataDate,
+      lastImportFileName: 'cities500.txt',
```
jrasm91, line 69: This filename is used a few times can we make it a constant?

server/src/infra/repositories/metadata.repository.ts
```diff
@@ -1,77 +1,179 @@
...
+
+  private async loadAdmin1(queryRunner: QueryRunner) {
+    await this.loadGeodataToTableFromFile<GeodataAdmin1Entity>(
+      queryRunner,
+      (lineSplit: string[]) =>
+        this.geodataAdmin1Repository.create({
+          key: lineSplit[0],
+          name: lineSplit[1],
+        }),
+      '../resources/admin1CodesASCII.txt',
```
jrasm91, line 130: It might be better to make these absolute paths

server/src/infra/repositories/metadata.repository.ts
```diff
@@ -1,77 +1,179 @@
...
+      .getOne();
+
+    if (!response) {
+      this.logger.warn(
+        `Response from database for reverse geocoding latitude: ${point.latitude}, longitude: ${point.longitude} was null`,
+      );
+      return null;
+    }
+
+    this.logger.debug(`Raw: ${JSON.stringify(response, null, 2)}`);
```
jrasm91, line 171: Verbose mightt be better

jrasm91 (approved):
Looks pretty good. We don't normally include a lot of "logic" on repos, but I'm not sure how much of the reverse geo stuff we care about having in the core codebase anyway. Looks pretty good to me.

jrasm91:
Is there still a system config setting for precision that we no longer use?

danieldietzler:
What is GeodataAdmin1 and GeodataAdmin2?

zackpollard:
> What is GeodataAdmin1 and GeodataAdmin2?

It is further information that an be referenced by multiple places, stuff like counties/states, wider cities, etc

PR #5327, opened 2023-11-27
refactor(mobile): log asyncvalue errors

shenlong-tanwen:
#### Changes made in the PR:

- Scaffold body which had the common error and loading indicator are migrated to use the custom extension on `AsyncValue` which shows the `ImmichLoadingIndicator` when the value is loading and logs any errors to the Immichlogger and displays a generic error widget when the value is error


fyfrey:
Nice way to remove some of the frequent code duplication!

alextran1502 (approved):

PR #5340, opened 2023-11-27
feat(web): prefer higher GiB values

jrasm91:
Cap units at GiB. So show `1234 GiB` instead of `1 TiB`, especially on the server info page.

| Before | After |
| - | - |
| image | image |


alextran1502 (approved):

jrasm91:
Yeah, I think I will revisit this again.

PR #5344, opened 2023-11-27
feat(mobile): Add haptic feedback to asset grid

l0nax:
Add haptic feedback when jumping from one month to another in the asset grid view.
The feedback is similar to the one implemented in the Google Photos app.

Closes #5342

### DIFF
--- a/mobile/lib/modules/home/ui/asset_grid/immich_asset_grid_view.dart
+++ b/mobile/lib/modules/home/ui/asset_grid/immich_asset_grid_view.dart
@@ -1,19 +1,22 @@
 import 'dart:collection';
+import 'dart:developer';
 import 'dart:math';
 
 import 'package:collection/collection.dart';
 import 'package:easy_localization/easy_localization.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:immich_mobile/extensions/build_context_extensions.dart';
+import 'package:immich_mobile/extensions/collection_extensions.dart';
 import 'package:immich_mobile/modules/asset_viewer/providers/scroll_notifier.provider.dart';
 import 'package:immich_mobile/modules/home/ui/asset_grid/thumbnail_image.dart';
 import 'package:immich_mobile/shared/models/asset.dart';
-import 'package:immich_mobile/extensions/collection_extensions.dart';
 import 'package:scrollable_positioned_list/scrollable_positioned_list.dart';
+
 import 'asset_grid_data_structure.dart';
-import 'group_divider_title.dart';
 import 'disable_multi_select_button.dart';
 import 'draggable_scrollbar_custom.dart';
+import 'group_divider_title.dart';
 
 typedef ImmichAssetGridSelectionListener = void Function(
   bool,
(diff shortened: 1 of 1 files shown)


martyfuhry:
It looks like the added code will run inside of a function which is responsible for building a widget component.  I would instead, attach a listener to a relevant notifier and move this functionality inside of the listener.

For example, a `ScrollView` will have a `ScrollController` which notifies listeners whenever it scrolls. For this functionality, I would use the `ItemPositionsListener`:

```dart
void initState() {
  ...
  _itemPositionsListener.itemPositions.addListener(_addHapticsWhenChangingMonth);
}

...

void _addHapticsWhenChangingMonth() {
  // Put the code here
}
```

mobile/lib/modules/home/ui/asset_grid/immich_asset_grid_view.dart
```diff
@@ -326,6 +328,15 @@ class ImmichAssetGridViewState extends State<ImmichAssetGridView> {
 
     final date = widget.renderList.elements[pos % maxLength].date;
 
+    // add haptic feedback only when scrolling through the asset list
+    if (_scrolling &&
```
fyfrey, line 332: scrolling is the fast drag scrolling using the scroll handle. Is this indented?

For large collections, drag scrolling will vibrate constantly because the months change all the time.

Either disable for drag scrolling (thus only active when slowly scrolling to a new month) or add some throttle (e.g. no more than every 500ms etc.)

l0nax: Yes, this is intended. The haptic feedback "notifies" the user when a new month has been selected.

Debouncing the haptic feedback would only create confusion on the user side.

fyfrey: Have you tested the situation (quickly scrolling through a large collection where you jump multiple months a second)?
The dart-java communication does not cope well with being trashed with calls and would just not perform important actions anymore - ultimately leading to worse user experience. We had that issue with something else...

l0nax: Yes, I did test this scenario, and the app worked well (on a Pixel 6 Pro).
But for certain (older) phones this might cause problems.

I think _500ms_ debounce is too long. What do you think of _50ms_ or _100ms_?

fyfrey: Alright, that's good to hear!

50ms would already be 20 times per second.
I wonder, can you really differentiate / feel the difference between 10 and 20 haptic feedback per second?
I would throttle it to the value that you can feel a difference when you slightly decrease the rate per second.
If this is 20 times per second, go with 50ms.
If this is 10 times per second, go with 100ms.

l0nax: > Just to make sure, I think you need to use throttle - not debounce.

Yeah, I mixed them up, sorry.


The difference between 50ms and 100ms is noticeable but can be ignored.

alextran1502 (approved):

PR #5346, opened 2023-11-27
chore(server): sql versioning

jrasm91:
1. Call whitelisted methods that have the `@GenerateSql(...)` decorator
2. Capture query sql
3. Save as a generated sql file
4. GitHubAction checks that these files are kept up to date.

This should give us nice diffs in the future when a sql changes


server/src/infra/repositories/api-key.repository.ts
```diff
@@ -3,9 +3,10 @@ import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
 import { APIKeyEntity } from '../entities';
+import { DummyValue, GenerateSql } from '../infra.util';
 
 @Injectable()
-export class APIKeyRepository implements IKeyRepository {
+export class ApiKeyRepository implements IKeyRepository {
```
zackpollard, line 9: If we are going to rename this to API should we also do the same for the entity and other other places we have this? Personally I don't mind whether it is API or Api

zackpollard (approved):
Looks great, will be really useful

server/src/infra/repositories/api-key.repository.ts
jrasm91: True, I changed it because I take the name and convert it to `api.repository.sql`. The algorithm didn't work if it was `API`. I can look at changing the other references in the future to match this one.

server/src/infra/sql-generator/index.ts
```diff
@@ -0,0 +1,121 @@
...
+    imports: [
+      TypeOrmModule.forRoot({
+        ...databaseConfig,
+        entities: databaseEntities,
+        logging: ['query'],
+        logger: sqlLogger,
+      }),
+      TypeOrmModule.forFeature(databaseEntities),
+    ],
+    providers: [...repositories],
```
mertalev, line 63: You can just pass `repositories` here.

server/src/infra/sql-generator/index.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  console.log('Done');
+};
+
+main()
+  .then(() => process.exit(0))
+  .catch((error) => {
+    console.error(error);
+    process.exit(1);
```
mertalev, line 120: Is there anything that should be cleaned up before exiting on an error?

mertalev (approved):

server/src/infra/repositories/album.repository.ts
```diff
@@ -36,6 +38,7 @@ export class AlbumRepository implements IAlbumRepository {
     return this.repository.findOne({ where: { id }, relations, order });
   }
 
+  @GenerateSql({ params: [DummyValue.UUID] })
```
danieldietzler, line 41: I think this should be `[[DummyValue.UUID]]`, correct?

server/src/infra/repositories/library.repository.ts
```diff
@@ -69,6 +76,7 @@ export class LibraryRepository implements ILibraryRepository {
     });
   }
 
+  @GenerateSql({ params: [false] })
```
danieldietzler, line 79: Why not just provide no parameters when we pass `false` anyways?

danieldietzler (approved):
Awesome, I like how simple it is to "track" a query!

server/src/infra/sql-generator/index.ts
jrasm91: I refactored the code to make sure it always cleans up now.

server/src/infra/repositories/album.repository.ts
jrasm91: Good catch

danieldietzler (approved):
Amazing!

PR #5350, opened 2023-11-28
chore(docs): Added guide on remote access

thariq-shanavas:
A brief guide on different ways to access immich remotely. I use Wireguard to connect to my server - apologies for any technical mistakes in what I've written about reverse proxy. Let me know and I'll be happy to fix it!

### DIFF
--- a/docs/docs/guides/remote-access.md
+++ b/docs/docs/guides/remote-access.md
@@ -35,17 +35,14 @@ If you are unable to open a port on your router for Wireguard or OpenVPN to your
 - Tailscale is a paid service. However, there is a generous [free tier](https://tailscale.com/pricing/) that permits up to 3 users and up to 100 devices.
 - Tailscale needs to be installed and running on both server-side and client-side.
 
-## Option 3: Self-signed certificate
-Using a [self-signed certificate](https://en.wikipedia.org/wiki/Self-signed_certificate) protects you against man-in-the-middle attacks, but exposes your Immich instance to the web.
+## Option 3: Reverse Proxy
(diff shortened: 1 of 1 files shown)


docs/docs/guides/remote-access.md
```diff
@@ -0,0 +1,56 @@
...
+- Tailscale needs to be installed and running on both server-side and client-side.
+
+## Option 3: Self-signed certificate
+Using a [self-signed certificate](https://en.wikipedia.org/wiki/Self-signed_certificate) protects you against man-in-the-middle attacks, but exposes your Immich instance to the web.
+
+### Pros
+- No additional software needs to be installed client-side
+
+### Cons
+- Both the Immich web interface and API may be exposed to the internet. Immich is under very active developement and the existence of severe security vulnerabilities cannot be ruled out.
```
bo0tzz, line 45: A self-signed cert is not technically a standalone remote access option, but would be a part of a remote proxy setup. I would suggest removing this section and instead adding a note to the reverse proxy section, indicating that with a remote proxy a certificate should be used for https. The options there are to get a free certificate via Let's Encrypt (preferred), or to use a self signed one.

docs/docs/guides/remote-access.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Option 4: Reverse Proxy
+A reverse proxy is a service that sits between web servers and clients. A reverse proxy increases security by hiding the server IP address, which makes targeted attacks like [DDoS](https://www.cloudflare.com/learning/ddos/what-is-a-ddos-attack/) harder.
```
bo0tzz, line 48: A reverse proxy does not necessarily hide the server IP/make targeted attacks harder. This is only the case if you're using something like cloudflare, but not when you're running a reverse proxy like nginx on your server. Maybe cloudflare and a local proxy server should be separate sections?

bo0tzz (changes requested):
This is pretty much exactly what I had in mind, thank you!

docs/docs/guides/remote-access.md
thariq-shanavas: Thank you, that clarifies some lingering confusion I had about reverse proxies. I made the changes.

bo0tzz (approved):

jrasm91:
This looks pretty good. Any thoughts on making the pro's and con's list start with ✅ and ❌ or similar?

thariq-shanavas:
Thank you for the formatting fix! I was just about to get to it.

PR #5356, opened 2023-11-28
feat(web): Lazy load thumbnails on the people page

l0nax:
My browser freezes for a short amount of time when I open the people page (approx. 600+ non-hidden faces).
I managed to reproduce this behavior every time with
- Google Chrome
- Thorium (Chromium)
- Firefox

This PR ensures that the thumbnails are lazy loaded with the native `loading` property, which has been supported for a while and thus should not cause any problems.
To prevent increasing the LCP, the first _20_ images are eagerly loaded (approx. amount of images shown on a 1920x1080 display).

### DIFF
--- a/web/src/lib/components/assets/thumbnail/image-thumbnail.svelte
+++ b/web/src/lib/components/assets/thumbnail/image-thumbnail.svelte
@@ -17,12 +17,14 @@
   export let circle = false;
   export let hidden = false;
   export let border = false;
+  export let priority = true;
   let complete = false;
 
   export let eyeColor: 'black' | 'white' = 'white';
 </script>
 
 <img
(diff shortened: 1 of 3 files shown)


web/src/lib/components/assets/thumbnail/image-thumbnail.svelte
```diff
@@ -17,12 +17,14 @@
   export let circle = false;
   export let hidden = false;
   export let border = false;
+  export let priority = false;
```
bo0tzz, line 20: Should this be defaulted to false? ImageThumbnail is used in a few different places.

bo0tzz (approved):

jrasm91 (approved):
Looks good. Maybe the prop could be "preload" instead of "priority"

PR #5363, opened 2023-11-28
fix(web): storage decimals

jrasm91:
Follow up to #5340.

Restore old behavior, and then change it to show decimals on the stats page when usage units are TiB or higher.

image

### DIFF
--- a/web/src/lib/components/admin-page/server-stats/server-stats-panel.svelte
+++ b/web/src/lib/components/admin-page/server-stats/server-stats-panel.svelte
@@ -21,7 +21,8 @@
     return '0'.repeat(zeroLength);
   };
 
-  $: [statsUsage, statsUsageUnit] = getBytesWithUnit(stats.usage, 0);
+  const TiB = 1024 ** 4;
+  $: [statsUsage, statsUsageUnit] = getBytesWithUnit(stats.usage, stats.usage > TiB ? 2 : 0);
 </script>
 
 <div class="flex flex-col gap-5">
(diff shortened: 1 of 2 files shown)


web/src/lib/components/admin-page/server-stats/server-stats-panel.svelte
```diff
@@ -21,7 +21,8 @@
     return '0'.repeat(zeroLength);
   };
 
-  $: [statsUsage, statsUsageUnit] = getBytesWithUnit(stats.usage, 0);
+  const TiB = 1024 ** 4;
```
danieldietzler, line 24: Maybe this could be a constant in byte-units?

zackpollard (approved):
LGTM 😊

web/src/lib/utils/byte-units.ts
```diff
@@ -9,7 +9,7 @@
  * @returns size (number) and unit (string)
  */
 export function getBytesWithUnit(bytes: number, maxPrecision = 1): [number, string] {
-  const units = ['B', 'KiB', 'MiB', 'GiB'];
+  const units = ['B', 'KiB', 'MiB', 'GiB', 'TiB', 'PiB', 'EiB'];
```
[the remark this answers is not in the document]
jrasm91: It was there before lol

web/src/lib/components/admin-page/server-stats/server-stats-panel.svelte
jrasm91: Maybe, if I use it somewhere else 😁

PR #5366, opened 2023-11-28
chore(web): handle signals for faster restart

jrasm91:
`npm` doesn't correctly handle signals, so avoid using in when running inside docker. Restarting or killing the `immich-web` container now exits cleanly after a few seconds, instead of waiting 10 seconds for docker to then kill it.


danieldietzler (approved):
Amazing, thanks! :)

PR #5368, opened 2023-11-29
chore: add renovate grouping for docs and consolidate schedule

jrasm91:
Set the schedule for taco Tuesday and add another group for the docs.


mertalev (approved):
Taco Tuesday lmao

PR #5374, opened 2023-11-29
fix(docs): build

jrasm91:
Fix the bad link and build the docs in GitHub Actions to prevent similar issues from getting through in the future.


mertalev (approved):
LGTM

PR #5375, opened 2023-11-29
chore: update job names

jrasm91:


.github/workflows/test.yml
```diff
@@ -188,7 +188,7 @@ jobs:
           poetry run pytest --cov app
 
   generated-api-up-to-date:
-    name: Check generated files are up-to-date
+    name: Open API Clients
```
mertalev, line 191: No space between Open and API

mertalev (approved):

.github/workflows/test.yml
jrasm91: Thanks.

PR #5401, opened 2023-11-30
fix(web): whitespace in person name

### LINKED ISSUE #5397: [BUG] Suggest people when typing a name doesn't handle full names
### The bug

If I am typing a name the suggestions disappear as soon as a whitespace is typed. For example if I have a person with the name `"Foo Bar"` while I type `"Foo"` the suggestion is present, but when I type `"Foo "` the suggestion goes away until I remove the whitespace. Typing `"Foo B"` doesn't suggest anything either.

There are no errors in the console or in the server logs.

A quick look suggests it might be because of [this](https://github.com/immich-app/immich/pull/4566/commits/79e89823d8e2903a90db3b13db2a8c8647ffd325#diff-185e73078772a18812f575108c4dc368aa375a7bd750c97c2d1798d3fedcf431R116-R122) and the solution would be to split the search term into whitespace delineated tokens.

Ubuntu 22.04

### Version of Immich Server

1.89.0

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1. Type a name and see suggestions
2. Type a whitespace and suggestions go away
```

### Additional information

_No response_

martabal:
## Changes made in this PR

This PR adds the ability to search people with white spaces in their names.
Fixes #5397

### DIFF
--- a/web/src/lib/components/faces-page/merge-face-selector.svelte
+++ b/web/src/lib/components/faces-page/merge-face-selector.svelte
@@ -17,6 +17,7 @@
   import CircleIconButton from '../elements/buttons/circle-icon-button.svelte';
   import { cloneDeep } from 'lodash-es';
   import LoadingSpinner from '../shared-components/loading-spinner.svelte';
+  import { searchNameLocal } from '$lib/utils/person';
 
   export let person: PersonResponseDto;
   let people: PersonResponseDto[] = [];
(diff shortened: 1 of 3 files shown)


jrasm91 (approved):
Can we share this code between the two pages?

martabal:
> Can we share this code between the two pages?

Done

PR #5407, opened 2023-11-30
feat(mobile): unify asset grid multiselect actions

### LINKED ISSUE #5494: [BUG] no option on multi-select on local album in app (android)
### The bug

There is no options when multi-selecting pictures in local albums.

Debian

### Version of Immich Server

1.89.0

1.89.0

### Platform with the issue

- [ ] Server
- [ ] Web
- [X] Mobile

```YAML
Same as example
```

### Your .env content

```Shell
Same as example
```

### Reproduction steps

```bash
1. Click on library at bottom of app
2. Select folder under "Albums on device"
3. Long press on a picture
4. Procede click on other pictures to multi-select.
5. No option to do anything with selected pictures.
```

### Additional information

I would love the ability to multi-delete to cleanup unneeded screenshots
Screenshot_20231205-110832

fyfrey:
### unify multiselect UI across home, partner assets and albums

* extracted code for multiselect handling from home tab to new file
* partner asset view and album view now provide the fancy bottom sheet previously only found on home tab
* provide all allowed batch actions per view
* e.g. users can now select assets in an album or partner view and add these to another album
* multiselect favorite now also allows to unfavorite assets (when all select assets are already favs)
* lots of code cleanup like new `currentAlbumProvider` and removal of isOwner & sharedAlbumId from Thumbnail (was only used to pass to gallery)
* error-prone provider invalidation for album changes (like add asset) replaced with proper DB watcher
* reduced code duplication e.g. sharing a selection  is now unified as well (previously album had its own stuff)
* net reduction in LOC: -200
* maybe more that I forgot :-)  

|Screen | Old | New |
|-----------| ----------|---------|
| Home | image | image |
| Partner | image | image |
| Album   | image | image|
| Favorites | image | image |
| Archive | image | image |


tomikonieczny:
Nice to see improvements here. 

What about keeping assets selection while switching between different views (ex. main list view and details view)? I aslo discovered the same issue while changing screen orientation - selection is gone.

fyfrey:
> What about keeping assets selection while switching between different views (ex. main list view and details view)?

It's not possible to switch to another view while multi select it active.

> I aslo discovered the same issue while changing screen orientation - selection is gone.

That's an issue that could already be fixed by this PR, or at least make it a lot easier to fix once this is merged. The selection state is now stored in a provider (that can "survive" an orientation change / rebuild)

fyfrey:
>  So temporary solution could be multi select feature in details view too (easier to select proper asset having it in a full screen).

Multi select in a single image view makes no sense. Or you are speaking about some other detail view?
To make images larger/easier to select the right asset, you can already use pinch-to-zoom making the thumbnails up to half of the screen width.

I think this topic has nothing to do with this specific PR, please open a separate discussion for it in order to keep the conversation in this PR focused. Thank you.

tomikonieczny:
> Multi select in a single image view makes no sense. Or you are speaking about some other detail view?
> To make images larger/easier to select the right asset, you can already use pinch-to-zoom making the thumbnails up to half of the screen width.

Last comment here: pitch-to-zoom would be good enough when one item per row with original ratio

alextran1502 (approved):

PR #5410, opened 2023-11-30
fix(web): status box rendering

danielktaylor:
This PR caches the ServerInfo response in a new Svelte Store so that the loading spinner does not appear in the StatusBox every time a page loads.

(I'm new to Svelte and open to all suggestions for improving this PR.)

### DIFF
--- a/web/src/lib/stores/server-info.store.ts
+++ b/web/src/lib/stores/server-info.store.ts
@@ -1,0 +1,4 @@
+import { writable } from 'svelte/store';
+import type { ServerInfoResponseDto } from '../../api/open-api';
+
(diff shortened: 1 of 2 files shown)


web/src/lib/stores/server-info.store.ts
```diff
@@ -0,0 +1,4 @@
+import { writable } from 'svelte/store';
+import type { ServerInfoResponseDto } from '../../api/open-api';
```
martabal, line 2:
```suggestion
import type { ServerInfoResponseDto } from '@api';
```

alextran1502 (approved):

PR #5416, opened 2023-12-01
fix(web): delete modal z-index

### LINKED ISSUE #4954: [BUG] web: trashing asset doen't shift to next asset and browsing does't work anymore
### The bug

When an asset is trashed while browsing and viewing using arrow keys it doen't shift to next asset reliably and browsing by arrow keys stops working. It does in more than ~90% of cases. Next asset is shown and arrow keys are working very rarely.

Fedora 38

### Version of Immich Server

v.1.85.0

not relevant

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

```
1. view an asset
2. trash it using Trash button
3. "Moved to trash" toast appears and trashed asset is still shown
5. arrow keys or left/right buttons doesn't work anymore
6. there is needed to go back and view next asset again to restore browsing
```

### Additional information

_No response_

danielktaylor:
This fixes #4954 

When trashing a photo, it should move to the next photo automatically. If there is no next photo, then the previous photo, and otherwise close the asset viewer. This matches Google Photos behavior.

The current code does not work because the deleted photo's next asset reference is null after the `deleteAssets(...)` call.

I'm new to Svelte, all code suggestions are welcome.

### DIFF
--- a/web/src/lib/components/shared-components/full-screen-modal.svelte
+++ b/web/src/lib/components/shared-components/full-screen-modal.svelte
@@ -12,7 +12,7 @@
 <section
   in:fade={{ duration: 100 }}
   out:fade={{ duration: 100 }}
-  class="fixed left-0 top-0 z-[990] flex h-screen w-screen place-content-center place-items-center bg-black/40"
+  class="fixed left-0 top-0 z-[9990] flex h-screen w-screen place-content-center place-items-center bg-black/40"
 >
   <div
     class="z-[9999]"


danielktaylor:
Just added a fix for deleting photos, too, not just trashing.

I increased the z-index on the delete confirmation modal, because it's currently behind the photo:

Screenshot from 2023-12-01 21-48-46

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -372,9 +372,18 @@
...
       await api.assetApi.deleteAssets({ assetBulkDeleteDto: { ids: [asset.id] } });
 
-      await navigateAssetForward();
+      if (nextAsset) {
+        assetViewingStore.setAssetId(nextAsset);
+      } else if (previousAsset) {
+        assetViewingStore.setAssetId(previousAsset);
+      } else {
+        closeViewer();
+      }
```
jrasm91, line 386: Is there a reason why we cannot put this logic inside of navigateAssetForward()?

jrasm91:
It turns out the issue is because of the websocket event, which I fixed in #5493 instead. Also, the code already had logic for `next || previous || close viewer` in the asset-grid component #5488 re-uses that same logic for trash and delete.

The delete modal is a good catch, so I've rebased this PR to include that fix.

jrasm91 (approved):

PR #5420, opened 2023-12-01
feat(web): allow showing hidden people in image asset details view

brokeh:
This makes it possible to easily find people/faces that have accidentally been hidden. Unhiding them still requires clicking on the person to go to their page to unhide them.

 

This adds the show/hide button to the people section header if the photo has hidden people
image

And clicking the button shows the hidden people
image

And the view for when there are no hidden people remains unchanged, with no toggle button.
image

### DIFF
--- a/web/src/lib/components/asset-viewer/detail-panel.svelte
+++ b/web/src/lib/components/asset-viewer/detail-panel.svelte
@@ -19,8 +19,11 @@
     mdiImageOutline,
     mdiMapMarkerOutline,
     mdiInformationOutline,
+    mdiEye,
+    mdiEyeOff,
   } from '@mdi/js';
   import Icon from '$lib/components/elements/icon.svelte';
+  import CircleIconButton from '../elements/buttons/circle-icon-button.svelte';
   import Map from '../shared-components/map/map.svelte';
   import { websocketStore } from '$lib/stores/websocket';
   import { AppRoute } from '$lib/constants';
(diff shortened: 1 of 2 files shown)


web/src/lib/components/asset-viewer/detail-panel.svelte
```diff
@@ -176,25 +180,38 @@
...
+          <CircleIconButton
+            title="Show hidden people"
+            icon={showingHiddenPeople ? mdiEyeOff : mdiEye}
+            padding="0"
+            on:click={() => (showingHiddenPeople = !showingHiddenPeople)}
+          />
+        </div>
+      </div>
 
       <div class="mt-4 flex flex-wrap gap-2">
```
martabal, line 195: What do you think about this ?
```suggestion
      <div class="flex h-10 w-full items-center justify-between">
        <h2>PEOPLE</h2>
        {#if people.some((person) => person.isHidden)}
          <CircleIconButton
            title="Show hidden people"
            icon={showingHiddenPeople ? mdiEyeOff : mdiEye}
            padding="1"
            on:click={() => (showingHiddenPeople = !showingHiddenPeople)}
          />
        {/if}
      </div>

      <div class="mt-2 flex flex-wrap gap-2">
```

martabal (approved):
Thanks for the PR !

web/src/lib/components/asset-viewer/detail-panel.svelte
brokeh: Yep, that looks cleaner. I don't do much web dev so I don't really know what I'm doing here.

I've given it a test locally and it works the same so merged this into the PR.

Issue #4297, opened 2023-10-02
[BUG] mobile: new album icon has different height to existing album cover


waclaw66:
### The bug

|portrait|landscape|
|-|-|
|![obrazek](https://github.com/immich-app/immich/assets/15554561/8b209b8f-e0b7-4268-abae-2a8716237bae)|![obrazek](https://github.com/immich-app/immich/assets/15554561/12f809aa-4bf8-403e-839d-a5cf445cd412)|

### The OS that Immich Server is running on

Fedora 38

### Version of Immich Server

v1.79.1

### Version of Immich Mobile App

v1.79.0

### Platform with the issue

- [ ] Server
- [ ] Web
- [X] Mobile

### Your docker-compose.yml content

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

```bash
not relevant
```

DarhkVoyd:
@alextran1502 @waclaw66  Hello, I'm a newcomer to this project and eager to help.Is this issue still open, and if so, how can I assist? What's the best way for me to get started?

shenlong-tanwen:
@waclaw66 Is this still an issue? If so, can you please share your device details?

waclaw66:
Hi @shenlong-tanwen. Unfortunatelly it is, tested on Samsung Galaxy S20 FE and Galaxy A51, same problem.

shenlong-tanwen:
@DarhkVoyd This issue is still up if you want to give it a try. You can start by looking at the difference between the widget returned from `library_page.dart#buildCreateAlbumButton` method, which is for the new album method and the one from `AlbumThumbnailCard`

shenlong-tanwen:
Fixed with #5422

PR #5422, opened 2023-12-01
fix(mobile):  new album icon has different height to existing album cover

Skyquek:
## Description

Fix the new album icon height issue with existing album cover.

## Related Issue

#4297


waclaw66:
Works perfectly, tested on mobile and tablet.

alextran1502 (approved):

PR #5455, opened 2023-12-02
fix(server): video orientation

martabal:
## Changes made in this PR

With this PR, video rotation metadata is extracted with ffprobe using the first video stream. It fixes incorrect ratios for vertical videos on the asset grid. With ffprobe, rotation metadata is saved as a number corresponding to an angle in degrees. This PR translates the angle into exiftool rotation values:

```
1 = Horizontal (normal)
2 = Mirror horizontal
3 = Rotate 180
4 = Mirror vertical
5 = Mirror horizontal and rotate 270 CW
6 = Rotate 90 CW
7 = Mirror horizontal and rotate 90 CW
8 = Rotate 270 CW
```

## Screenshots

The asset is a vertical video, taken with an iphone : 

Screenshot from 2023-12-02 21-28-46

| Before | After |
| :---: | :---: |
|Screenshot from 2023-12-02 21-06-03 | Screenshot from 2023-12-02 21-06-19 |

### DIFF
--- a/server/src/domain/metadata/metadata.service.ts
+++ b/server/src/domain/metadata/metadata.service.ts
@@ -14,6 +14,7 @@ import {
   IAssetRepository,
   ICryptoRepository,
   IJobRepository,
+  IMediaRepository,
   IMetadataRepository,
   IMoveRepository,
   IPersonRepository,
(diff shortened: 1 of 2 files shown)


server/src/domain/metadata/metadata.service.ts
```diff
@@ -182,6 +184,27 @@ export class MetadataService {
...
+            exifData.orientation = '6';
+            break;
+          case 90:
+            exifData.orientation = '8';
+            break;
+          case 180:
+            exifData.orientation = '3';
+            break;
+          case 0:
+            exifData.orientation = '1';
```
danieldietzler, line 202: Can we have an enum for those values? I think this would be more intuitive.

jrasm91 (approved):
Lgtm

danieldietzler (approved):
Thanks for making it an enum! LGTM

PR #5471, opened 2023-12-03
fix(server): send upload_success notification only for non hidden assets

shenlong-tanwen:
#### Change made in the PR

- `UPLOAD_SUCCESS` event is only send to assets that are not hidden. 

Fixes the issue where the live part of a live photo is displayed as a separate asset in the mobile app


jrasm91 (approved):
This is a fix for that specific case. We have others that are still broken though.

PR #5486, opened 2023-12-04
feat(web): navigate with keyboard on person page

martabal:
## Changes made in this PR

With this PR, users can navigate through suggested people with Tab and Arrow keys. On the suggestion merge modal, pressing the Tab key will only switch between the `No` and `Yes` buttons.

Resolves #5400 and #5398


jrasm91 (commented):
Can you use svelte:document instead please?

https://svelte.dev/docs/special-elements#svelte-document

web/src/lib/components/elements/buttons/button.svelte
```diff
@@ -26,6 +26,7 @@
   export let fullwidth = false;
   export let border = false;
   export let title: string | undefined = '';
+  export let ref: HTMLButtonElement | null = null;
```
jrasm91, line 29: What is this used for?

web/src/lib/components/faces-page/merge-suggestion-modal.svelte
```diff
@@ -8,27 +8,57 @@
...
 
+  const dispatch = createEventDispatcher<{
+    reject: void;
+    confirm: [PersonResponseDto, PersonResponseDto];
+    close: void;
+  }>();
+
+  const handleKeyboardPress = (event: KeyboardEvent) => {
+    if (!$showAssetViewer) {
+      event.stopPropagation();
```
jrasm91, line 33: Why are you stopping propagation, especially if you aren't handling the keyboard event?

web/src/lib/components/faces-page/merge-suggestion-modal.svelte
```diff
@@ -8,27 +8,57 @@
...
+      switch (event.key) {
+        case 'Tab':
```
jrasm91, line 35: Tabbing through focusable elements is something that is built into the browser. We don't need to re-invent the wheel here.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

martabal: Yeah I know but I couldn't find a way to limit the focus to the suggested people only.

web/src/lib/components/elements/buttons/button.svelte
martabal: I used it to focus on the button from the parent component.

web/src/lib/components/faces-page/merge-suggestion-modal.svelte
jrasm91: tabindex=-1 tells the browser to skip it.

martabal: Yeah I tried it but the issue is that you want to navigate only through the suggested people, not all the buttons on the page. Do you have any ideas :thinking: ?

jrasm91: Does using an actual browser model still let you focus on background elements?

martabal: What's an actual browser model ?

martabal: Mmmh, just tried it, does it work only for dialogs ?

martabal: I don't think we want dialogs to display suggested people :sweat_smile:

jrasm91:
I think it is fine to add keyboard shortcuts, but we should do it in a way that doesn't make the site less accessible (which is the outcome of overriding the tab behavior or preventing other default key behaviors)

martabal:
> I think it is fine to add keyboard shortcuts, but we should do it in a way that doesn't make the site less accessible (which is the outcome of overriding the tab behavior or preventing other default key behaviors)

I agree 100% with you. But in this case we want to navigate with the tabs and arrow keys which is a little unusual. I don't know what's the good way for that case :sweat_smile:

jrasm91 (approved):
LGTM

PR #5488, opened 2023-12-04
refactor(web): asset viewer actions

jrasm91:
Clean up asset viewer => asset grid => asset store interactions.

### DIFF
--- a/web/src/lib/constants.ts
+++ b/web/src/lib/constants.ts
@@ -4,7 +4,9 @@ export enum AssetAction {
   FAVORITE = 'favorite',
   UNFAVORITE = 'unfavorite',
   TRASH = 'trash',
-  RESTORE = 'restore',
+  DELETE = 'delete',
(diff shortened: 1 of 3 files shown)


web/src/lib/constants.ts
```diff
@@ -4,7 +4,9 @@ export enum AssetAction {
   FAVORITE = 'favorite',
   UNFAVORITE = 'unfavorite',
   TRASH = 'trash',
-  RESTORE = 'restore',
+  DELETE = 'delete',
+  // RESTORE = 'restore',
```
danieldietzler, line 8: We don't need restore anymore?

danieldietzler (approved):
Nice one! Looks pretty clean.

web/src/lib/constants.ts
jrasm91: It was never used originally

PR #5504, opened 2023-12-05
fix: notify mobile app when live photos are linked

shenlong-tanwen:
#### Changes made in the PR:

- A new event `on_asset_hidden` is added to notify the clients when an asset is hidden in the server
- The new end point is used to notify the clients when motion part of a live photo is hidden so that the clients in turn can handle them
- The `on_asset_hidden` event is handled in the mobile app by removing the asset from the local db
- The websocket pending changes code is also refactored. A new ID field is added to each pending change and processed changes are removed based on the ID instead of removing them based on the type which could result in cases where an event might be missed

- The changes in `album_thumbnail_listtile.dart` is not related to this PR, but all it does is to wrap the children in an `Expanded` widget to prevent overflows in albums having large title

### DIFF
--- a/server/src/domain/job/job.service.ts
+++ b/server/src/domain/job/job.service.ts
@@ -188,15 +188,6 @@ export class JobService {
         break;
 
       case JobName.LINK_LIVE_PHOTOS:
-        // Notify clients to hide the linked live photo asset
-        const [asset] = await this.assetRepository.getByIds([item.data.id]);
-        if (asset) {
-          const motionId = asset.type == AssetType.VIDEO && !asset.isVisible ? asset.id : asset.livePhotoVideoId;
-          if (motionId) {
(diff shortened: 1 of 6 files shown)


server/src/domain/job/job.service.ts
```diff
@@ -188,6 +188,15 @@ export class JobService {
         break;
 
       case JobName.LINK_LIVE_PHOTOS:
+        // Notify clients to hide the linked live photo asset
```
jrasm91, line 191: I don't think this is correct. This will run and emit ids for _every_ live motion asset whenever you re-run metadata extraction (for all assets).

jrasm91: This stuff should go in `metadata.service.ts` line 150-160-ish instead.

shenlong-tanwen: Okay that was way easier than I thought it'd be. Updated now. Thank you!

jrasm91 (approved):

PR #5510, opened 2023-12-05
refactor(mobile): album sort

### LINKED ISSUE #5432: [BUG] mobile: dropdown list of albums is not sorted
### The bug

There is shown a list of albums while trying to add an asset to album...
List of all albums should be sorted in the same way as list of shared albums. Sorting method should be taken from Library.

Fedora 39

### Version of Immich Server

v1.89.0

v1.89.0

### Platform with the issue

- [ ] Server
- [ ] Web
- [X] Mobile

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

```
1. select an asset
2. swipe up to show list of albums
3. list of all albums is sorted randomly, list of shared albums as well
```

### Additional information

_No response_

### LINKED ISSUE #5489: [BUG] mobile: Albums are not sorted correctly
### The bug

In the mobile app, when using the "Most recent photo" sort order on the Albums page, albums are not sorted by their most recent photos.

The web app fixed this in 1.89.0 (see #4680). #5432 might be related.

Debian

### Version of Immich Server

v1.89.0

v1.89.0

### Platform with the issue

- [ ] Server
- [ ] Web
- [X] Mobile

```YAML
(not relevant here)
```

### Your .env content

```Shell
(not relevant here)
```

### Reproduction steps

```bash
1. Open the mobile app
2. Open the Album page
3. Sort albums by Most recent photo
```

### Additional information

_No response_

shenlong-tanwen:
Closes #5489, #5432 

#### Changes made in the PR

- Album sort and related handling has been migrated to a separate class with a riverpod provider
- Reverse sorting the album option has also been added (Tap the same option twice the reverse the sorting direction)
- The remaining sorting options from the web are added:
    - Asset count
    - Oldest photo
- The sort option ordering is also modified to reflect the same order used in the web
- The shared albums in the bottom scroll sheet are now displayed using `ListView`
- Add to album list in bottom sheet now uses the sort order specified in the library page 

#### TODO
- [x] Test cases for the sort functions


waclaw66:
This PR causes, that sorting method "Most recent photo" doesn't work correctly. The order seems like random totaly.

waclaw66:
The following block is to blame...
```dart
.sorted((a, b) {
      if (a.startDate == null) return 1;
      if (b.startDate == null) return -1;
      return 0;
    });
```

waclaw66:
This fixes it...
```dart
    final sorted = albums
        .where((a) => a.isRemote)
        //.sortedBy((album) => album.endDate ?? DateTime(1000))
        .sorted((a, b) {
      if (a.endDate != null && b.endDate != null) return a.endDate!.compareTo(b.endDate!);
      if (a.endDate == null) return 1;
      if (b.endDate == null) return -1;
      return 0;
    });
```

shenlong-tanwen:
> This fixes it...

Weird, I could not reproduce the issue where the first album is swapped with a different one though. I get the same results that I get from the web version. But I did refactor it like the one from your comment because it is much cleaner and executes the sort in a single iteration instead of sorting them twice. Thank you!

mobile/lib/modules/album/providers/album_sort_by_options.provider.dart
```diff
@@ -0,0 +1,135 @@
...
+
+typedef AlbumSortFn = List<Album> Function(List<Album> albums, bool isReverse);
+
+class _AlbumSortHandlers {
+  const _AlbumSortHandlers._();
+
+  static const AlbumSortFn created = _sortByCreated;
+  static List<Album> _sortByCreated(List<Album> albums, bool isReverse) {
+    final sorted =
+        albums.where((a) => a.isRemote).sortedBy((album) => album.createdAt);
```
fyfrey, line 17: where(..is remote) is used everywhere in this file - can we simply leave it out?
Device albums could also be sorted with this.

fyfrey (commented):
Nicely done! This is a handy feature

mobile/lib/modules/album/providers/album_sort_by_options.provider.dart
fyfrey: Filtering device out means they won't show at all, so no need to apply any of these functions on them.
If the sorting functions are only to be used on remote albums, we don't need the where(is remote) part.

shenlong-tanwen: Ah, yes. It should be made that way and the sorting functions should not be filtering the albums passed to it. Will update it. Thank you!

shenlong-tanwen: Updated!

alextran1502 (approved):

PR #5511, opened 2023-12-05
feat(web,server): unassign faces (3)

### LINKED ISSUE #5603: [BUG] ctl+f keyboard shortcut is used to mark a photo as a favorite
### The bug

ctl+f is a browser shortcut for finding text. Maybe it was used because who would be searching for text in a photo, but I've been using it as a workaround for https://github.com/immich-app/immich/issues/5571. Now I have to go unfavorite all of those photos.

Ubuntu 22.04

### Version of Immich Server

v1.90.2

v1.90.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1.Press ctl+f when viewing a photo
```

### Additional information

_No response_

martabal:
## Changes made in this PR

This PR introduces the ability to un-assign a face from an asset. Once un-assigned, the face can be re-assigned to a new or an existing person.

Fix #5603 too

To do:
- [x] Global edit
- Get feedbacks about the flow and change what can be improved.

Next PR will add the ability to manually tag faces within an image.

## Screenshots

https://github.com/immich-app/immich/assets/74269598/35eda975-4b27-413f-ade7-acc95faea97b


alextran1502:
I think the menu is a bit hidden. Instead of clicking on a face, would it be more straightforward to show the delete icon next to the blue icon when entering the person edit mode?

image

martabal:
> I think the menu is a bit hidden. Instead of clicking on a face, would it be more straightforward to show the delete icon next to the blue icon when entering the person edit mode?

Ahah, this is exactly what I did at first but I thought having two icons is a bit confusing 
- it's hard to tell at first glance what button does what (I tried with different colors and different icons)
- it makes navigation a bit harder, I miss-clicked multiple times and make one icon on the left and one on the right takes a lot of width

martabal:
> Many thanks for this feature, I am already enjoying it very much!

Glad to hear that !

server/src/domain/person/person.service.ts
```diff
@@ -127,22 +145,47 @@ export class PersonService {
...
-    const person = await this.findOrFail(personId);
-
-    await this.repository.reassignFace(face.id, personId);
-    if (person.faceAssetId === null) {
-      await this.createNewFeaturePhoto([person.id]);
+  async unassignFace(authUser: AuthDto, id: string): Promise<AssetFaceResponseDto> {
+    let face = await this.repository.getFaceById(id);
+    await this.access.requirePermission(authUser, Permission.PERSON_CREATE, face.id);
+    if (face.personId) {
+      await this.access.requirePermission(authUser, Permission.PERSON_WRITE, face.personId);
```
jrasm91, line 152: Why do you need these permissions to unassign a face?

server/src/domain/person/person.service.ts
```diff
@@ -127,22 +145,47 @@ export class PersonService {
...
+    const results: BulkIdResponseDto[] = [];
+
+    for (const data of dto.data) {
+      const faces = await this.repository.getFacesByIds([{ personId: data.personId, assetId: data.assetId }]);
+
+      for (const face of faces) {
+        await this.access.requirePermission(authUser, Permission.PERSON_CREATE, face.id);
+        if (face.personId) {
+          await this.access.requirePermission(authUser, Permission.PERSON_WRITE, face.personId);
+        }
```
jrasm91, line 174: You should use the bulk access checks instead.

server/src/domain/person/person.dto.ts
```diff
@@ -99,18 +105,18 @@ export class AssetFaceResponseDto extends AssetFaceWithoutPersonResponseDto {
   person!: PersonResponseDto | null;
 }
 
+export class FaceDto {
+  @ValidateUUID()
+  id!: string;
+}
```
jrasm91, line 111: We're `PUT`-ing to a route with an id and the dto is an id but they are different types. I think it might make more sense for this to be `personId`, which is what I believe it actually is. If this is a personId then the dto should probably be called ReassignFaceDto or something

server/src/domain/person/person.service.ts
```diff
@@ -98,16 +98,34 @@ export class PersonService {
     return this.repository.create({ ownerId: auth.user.id });
   }
 
-  async reassignFaces(auth: AuthDto, personId: string, dto: AssetFaceUpdateDto): Promise<PersonResponseDto[]> {
-    await this.access.requirePermission(auth, Permission.PERSON_WRITE, personId);
+  async reassignFace(authUser: AuthDto, personId: string, dto: FaceDto): Promise<PersonResponseDto> {
```
jrasm91, line 101: Wait, I'm confused. Isn't `personId` the parameter from `PUT /face/:id`? Why is it called personId not faceId? The id in the route should correspond to a face.

server/src/domain/person/person.service.ts
```diff
@@ -98,16 +98,34 @@ export class PersonService {
...
+    const face = await this.repository.getFaceById(dto.id);
+    const person = await this.findOrFail(personId);
+
+    await this.repository.reassignFace(face.id, personId);
+    if (person.faceAssetId === null) {
+      await this.createNewFeaturePhoto([person.id]);
+    }
+    if (face.person && face.person.faceAssetId === face.id) {
+      await this.createNewFeaturePhoto([face.person.id]);
+    }
```
jrasm91, line 114: I am confused a bit. The `createNewFeaturePhoto` argument takes in a list of faces, but you call it two separate times here. Shouldn't you instead start with a list like `refreshIds = []` and then add to the list in the if statements and then finally call `this.createNewFeaturePhoto` with the list?

server/src/domain/person/person.service.ts
```diff
@@ -98,16 +98,34 @@ export class PersonService {
...
+
+    return await this.findOrFail(personId).then(mapPerson);
```
jrasm91, line 116: There is no need to await a method in a return statement

server/src/domain/person/person.service.ts
```diff
@@ -98,16 +98,34 @@ export class PersonService {
...
+    await this.access.requirePermission(authUser, Permission.PERSON_WRITE, personId);
     const person = await this.findOrFail(personId);
     const result: PersonResponseDto[] = [];
     const changeFeaturePhoto: string[] = [];
     for (const data of dto.data) {
       const faces = await this.repository.getFacesByIds([{ personId: data.personId, assetId: data.assetId }]);
 
       for (const face of faces) {
-        await this.access.requirePermission(auth, Permission.PERSON_CREATE, face.id);
+        await this.access.requirePermission(authUser, Permission.PERSON_CREATE, face.id);
```
jrasm91, line 128: These repository methods have inputs lists/sets so you can do a single call in case like this, but you aren't using their that functionality.

jrasm91 (approved):
I looked at the server code and it looks pretty good. I think the endpoints are pretty close and the implementations just need a little bit of clean-up/optimization. I think there are some database calls that are called over and over again, and you could do a single call instead, since they take list inputs.

server/src/immich/controllers/face.controller.ts
```diff
@@ -18,11 +18,16 @@ export class FaceController {
   }
 
   @Put(':id')
-  reassignFacesById(
-    @Auth() auth: AuthDto,
+  reassignFace(
```
jrasm91, line 21: Is this endpoint used in the UI anywhere? What is it's purpose?

martabal: This is used when you assign a face from one person to another on the person-side-panel

server/src/domain/person/person.service.ts
martabal: Yes, this is incorrect, I will change that

server/src/domain/person/person.service.ts
jrasm91: And can you rename it back to auth?

server/src/immich/controllers/face.controller.ts
jrasm91: I did not find it in the code can you link it for me?

martabal: https://github.com/immich-app/immich/pull/5511/files#diff-0748192911c047330e6448297fb246113a246e8fb459d6ba0af512387e7abf0dR242-R245

server/src/domain/person/person.service.ts
martabal: I always check if the user has access to the face and the person he wants to change. I should maybe create new permissions for that

prometheanfire:
If you could link to this issue/pr from the new pr when you make it that'd be appreciated.  Removing faces for things that are not faces (bad matches) or from strangers would be nice.
Thanks

PR #5515, opened 2023-12-06
Fix not awaiting `sendFile`

jameskeane:
resolves #4762 #4754 

This change fixes the intermittent EPIPE errors that myself and others are seeing.

By explicitly returning a promise we ensure the caller correctly waits until the `sendFile` is complete before potentially closing or cleaning the socket. This is the most likely bug that would cause EPIPE errors.

To confirm the fix, I manually edited the file in a running docker container. The EPIPE errors were especially noticeable on Firefox with videos and after this patch, they are completely gone.

### DIFF
--- a/server/src/immich/api-v1/asset/asset.service.ts
+++ b/server/src/immich/api-v1/asset/asset.service.ts
@@ -336,14 +336,18 @@ export class AssetService {
 
     res.set('Cache-Control', 'private, max-age=86400, no-transform');
     res.header('Content-Type', mimeTypes.lookup(filepath));
-    res.sendFile(filepath, options, (error: Error) => {
-      if (!error) {
-        return;
-      }
+    return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


jrasm91 (commented):
You need to resolve or reject the promise in all paths. What happens if it is an aborted request?

jameskeane:
@jrasm91 not sure, we could either 1. have a unit test mocking a `sendFile` error or 2. make it lower impact and make it `resolve` either way and rely on just the error log message. What're your thoughts?

mertalev:
Does this mean it'll continue streaming the video after a user stops viewing it?

jameskeane:
> Does this mean it'll continue streaming the video after a user stops viewing it?

No, this simply means that the calling function will properly await until the `sendFile` operation is completed. Either because the whole file was sent, or because the client closed the socket, or any other reason.

mertalev (approved):

jrasm91 (approved):

PR #5569, opened 2023-12-08
feat(server): use postgres-adapter for websockets

jrasm91:
With the goal of dropping another container (redis), we will need to replace:
1. Websocket adapter (this PR)
2. Bull Queues (follow-up PR)

Tested by connecting two separate sessions to two different immich-server instances, and then triggering a websocket event from one (asset delete) and verifying it was received in the second instance.

image


danieldietzler (approved):

server/src/infra/websocket.adapter.ts
```diff
@@ -0,0 +1,28 @@
...
+import { DataSource } from 'typeorm';
+import { PostgresDriver } from 'typeorm/driver/postgres/PostgresDriver.js';
+
+const CREATE_ATTACHMENT_TABLE = `
+  CREATE TABLE IF NOT EXISTS socket_io_attachments (
+      id          bigserial UNIQUE,
+      created_at  timestamptz DEFAULT NOW(),
+      payload     bytea
+  );
+`;
```
bo0tzz, line 14: Should this just go in our migrations instead?

bo0tzz (approved):

server/src/infra/websocket.adapter.ts
jrasm91: Maybe although then it will prompt to delete it cause we don't have an entity. We could make an entity and add it to the migration, idk.

zackpollard: I think you could add it to migrations with no entity. If it does try to drop it, it would still try doing it this way when we generate migrations.

jrasm91: It actually doesn't. You can see all the checks passed. TypeORM only generates migrations for entities or knows about.

zackpollard: That's what I mean, so if you put it in the migrations folder, it should also do nothing.

jrasm91: So I can move it to one instead.

riedel:
Hi I just wanted to inform that this PR introduced regular (every second) disk writes  (see https://github.com/immich-app/immich/discussions/8009) as the pub sub mechanism of postgres in contrast to redis currently requires WAL (disk) commits to acquire a transaction id (see https://stackoverflow.com/a/78178571/148337)

PR #5577, opened 2023-12-09
chore(server): faster shutdown

jrasm91:
`npm run` doesn't handle signals from docker compose for container shutdown, and so it waits ten seconds before killing the container forcefully instead. This PR migrates to using `tini` and starting nest/node directly, which does respond correctly to signals. As a result, the shutdown time is significantly faster.

Before (~25 seconds):
image

After (~6 seconds):
image


alextran1502 (approved):

danieldietzler (approved):
Finally. Nice!

PR #5590, opened 2023-12-09
Fix critical missing hyphen docker-compose.md

jonhnet:
"docker compose up -d" looks like a command, but it's not.


bo0tzz:
`docker-compose` is deprecated, and is nowadays integrated into the main docker command as `docker compose`, so this is actually correct.

PR #5594, opened 2023-12-10
docs: add a walk-through guide for External Libraries

jonhnet:
I've left a few clarification questions in-line.


jonhnet:
@jrasm91 I'm not sure why there's no knob to request a review, so I'm attempting to do so here by @ing you. :) Please let me know if there's a preferred way.

alextran1502:
Hi @jonhnet I was going to add some more pictures to the PR but I haven't gotten around that. If you can help to add some screenshots to the guide, that would be great! Otherwise I can do it this weekend

jonhnet:
> Hi @jonhnet I was going to add some more pictures to the PR but I haven't gotten around that. If you can help to add some screenshots to the guide, that would be great! Otherwise I can do it this weekend

Oh that's a good idea. I'll try to get some put together.

jonhnet:
Okay, I added a bunch of screenshots.

alextran1502 (approved):

alextran1502:
Thank you for your contribution!

aviv926:
@alextran1502 😅
I think we have an unresolved question here
image

aviv926:
In addition, the guide explains how to use external libraries with a named variable
`EXTERNAL_PATH` which is added in the `.env` file
I think it might be a bit confusing because it doesn't appear in the official external libraries documentation.
In addition, using EXTERNAL_PATH is not necessary in my opinion because the full path can be specified already in the `docker-compose.yml` file. Or am I wrong?
It worked for me as follows:
image
ה

aviv926:
In addition, the guide explains how to use external libraries with a named variable
EXTERNAL_PATH which is added in the .env file
I think it might be a bit confusing because it doesn't appear in the official external libraries documentation.
In addition, using EXTERNAL_PATH is not necessary in my opinion because the full path can be specified already in the docker-compose.yml file. Or am I wrong?
It worked for me as follows:
image

jonhnet:
The unresolved question is pending a fix in https://github.com/immich-app/immich/pull/5733

I made a design choice to put the EXTERNAL_PATH into .env rather than put two copies of the absolute path in the yml file. I know that if I have two copies of the same path that have to be kept in sync, then someday later I'll move the data, update only one of the paths, and then suddenly everything will break mysteriously.

aviv926:
> The unresolved question is pending a fix in #5733
> 
> I made a design choice to put the EXTERNAL_PATH into .env rather than put two copies of the absolute path in the yml file. I know that if I have two copies of the same path that have to be kept in sync, then someday later I'll move the data, update only one of the paths, and then suddenly everything will break mysteriously.

Yes I see, interesting way.
In any case, it might be worth adding in the guide an example of adding another path in this way (creating another variable), I think this is something that can help those who are not familiar with Docker.

PR #5595, opened 2023-12-10
fix(server): connection aborted logging

jrasm91:
When the client terminates the connection one of two cases seem to happen server-side:
1. "connection aborted request" or 
2. Error write EPIPE

Error write EPIPE implies writing to a closed connection, which was already indicated via the connection aborted error. As far as I can tell, throwing an error in the service method, prevents nestjs from trying to write to the socket. That, combined with ignoring (not logging errors) for connection aborted requests seem to be the behavior we are looking for.

### DIFF
--- a/server/src/domain/domain.util.ts
+++ b/server/src/domain/domain.util.ts
@@ -21,6 +21,8 @@ export type Options = {
   each?: boolean;
 };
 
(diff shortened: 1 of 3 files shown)


mertalev (approved):

Zhan85:
This does not appear to be fixed.  Received this error just today.

PR #5608, opened 2023-12-10
fix: missing environment variable for machine learning

martabal:
The docs does not mention the ability to edit  machine learning with environment variables.
Env added : `IMMICH_MACHINE_LEARNING_ENABLED` and `IMMICH_MACHINE_LEARNING_URL`

### DIFF
--- a/docs/docs/install/environment-variables.md
+++ b/docs/docs/install/environment-variables.md
@@ -126,16 +126,18 @@ Redis (Sentinel) URL example JSON before encoding:
 
 ## Machine Learning
 
-| Variable                                         | Description                                                       |       Default       | Services         |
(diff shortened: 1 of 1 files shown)


danieldietzler (approved):

jrasm91:
I think we omitted this on purpose. The plan was to actually get rid of it, since it can be set via a config file now. But we kept it because you use it in the all in one still. We should have a discussion on this first, since this is the only one that you can set both ways at the same time.

martabal:
> I think we omitted this on purpose. The plan was to actually get rid of it, since it can be set via a config file now. But we kept it because you use it in the all in one still. We should have a discussion on this first, since this is the only one that you can set both ways at the same time.

I think having multiple way to set those is great and every environment variables should be documented. What do you think about documenting the order of priority ? (I think it's config file > config in db > env)

PR #5635, opened 2023-12-11
feat(web): re-add open graph tags for public share links

jrasm91:
Serve a dynamic index.html page, which includes appropriate `<meta>` tags for the shared link route.


alextran1502 (approved):
Nicely done! I assume this has been tested in production build?

bo0tzz (approved):

server/src/immich/app.service.ts
```diff
@@ -28,4 +59,47 @@ export class AppService {
...
+        excludePaths.find((item) => req.url.startsWith(item))
+      ) {
+        return next();
+      }
+
+      const targets = [
+        {
+          regex: /^\/share\/(.+)$/,
+          onMatch: async (matches: RegExpMatchArray) => {
+            const key = matches[1];
```
danieldietzler, line 79: Named capture groups <3

server/src/immich/app.service.ts
```diff
@@ -28,4 +59,47 @@ export class AppService {
...
+            if (meta) {
+              html = render(index, meta);
+            }
+
+            break;
+          }
+        }
+      } catch {}
+
+      res.type('text/html').header('Cache-Control', 'no-store').send(html);
```
danieldietzler, line 102: I thought disabling the cache was only for testing?

server/src/immich/app.service.ts
jrasm91: Those are too advanced for me. You should fix it :-)

server/src/immich/app.service.ts
jrasm91: Oh no, I think it is actually better to just never cache the index page, since it will be dynamic sometimes.

server/src/immich/app.service.ts
danieldietzler: Always let the new one do the work... :D

PR #5637, opened 2023-12-11
chore(server): fix ssr in dev

jrasm91:


alextran1502 (approved):

server/src/immich/app.service.ts
```diff
@@ -61,7 +61,12 @@ export class AppService {
   }
 
   ssr(excludePaths: string[]) {
-    const index = readFileSync('/usr/src/app/www/index.html').toString();
+    let index = '';
+    try {
+      index = readFileSync('/usr/src/app/www/index.html').toString();
```
martabal, line 66: Can we use an env or something for this ? It breaks the AIO image

jrasm91: I'm fine with adding an `IMMICH_WEB_ROOT` that is set to /usr/src/app/www`

martabal: Should it be docs ?

PR #5640, opened 2023-12-11
fix(web): grid on people page

martabal:
## Changes made in this PR

With this PR, the people grid always takes up the full available width and the size of the thumbnails changes depending on the screen size.

## Screenshots

| Before | After |
| :---:  | :---: |
|<video src="https://github.com/immich-app/immich/assets/74269598/7ba9e77d-60b6-4308-b7d3-708f5b6b27b6"> | <video src="https://github.com/immich-app/immich/assets/74269598/305d76a2-a390-4994-84d6-5041f3054392">  |


jrasm91 (approved):
This seems OK, but you can actually accomplish all of this with css using flex, not with javascript that has to run in the browser and may cause performance issues on slower devices, like phones.

martabal:
> This seems OK, but you can actually accomplish all of this with css using flex, not with javascript that has to run in the browser and may cause performance issues on slower devices, like phones.

The thumbnail width is re-calculated only when the width of the div containing all the thumbnails change. So could be an issue only on screen rotation

martabal:
Changed to use grid instead

jrasm91:
It's the video updated too or is that the old one?

jrasm91 (approved):
Lgtm

martabal:
> It's the video updated too or is that the old one?

I just changed it

PR #5645, opened 2023-12-11
feat: use <a> tag for albums in list view

martabal:
## Changes made in this PR

With this PR, the body of the table can be opened in a new tab with click wheel or Right click > Open in a new tab. The albums are preloaded just like with the cover view.

It reduces the number of API requests to '/api/me' by using the user value stored in a svelte store too

### DIFF
--- a/web/src/routes/(user)/albums/+page.svelte
+++ b/web/src/routes/(user)/albums/+page.svelte
@@ -22,7 +22,7 @@
   import { flip } from 'svelte/animate';
   import Dropdown from '$lib/components/elements/dropdown.svelte';
   import ConfirmDialogue from '$lib/components/shared-components/confirm-dialogue.svelte';
-  import { dateFormats } from '$lib/constants';
+  import { AppRoute, dateFormats } from '$lib/constants';
   import { locale, AlbumViewMode } from '$lib/stores/preferences.store';
   import {
     notificationController,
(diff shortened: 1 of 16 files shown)


web/src/routes/(user)/albums/+page.svelte
```diff
@@ -317,46 +324,48 @@
...
-                {/if}</td
-              >
-              <td class="text-md text-ellipsis text-center hidden xl:block xl:w-[15%] 2xl:w-[12%]"
-                >{#if album.startDate}
-                  {dateLocaleString(album.startDate)}
-                {:else}
-                  &#10060;
-                {/if}</td
-              >
+              <a data-sveltekit-preload-data="hover" class="flex w-full" href={`${AppRoute.ALBUMS}/${album.id}`}>
```
jrasm91, line 330: Can you change the href to use svelte templating instead of es6 string literals?

web/src/routes/+page.svelte
```diff
@@ -12,7 +13,7 @@
     <h1 class="font-immich-title text-4xl font-bold text-immich-primary dark:text-immich-dark-primary">
       Welcome to IMMICH Web
     </h1>
-    <Button size="lg" rounded="lg" on:click={() => goto('/auth/register')}>
+    <Button size="lg" rounded="lg" on:click={() => goto(AppRoute.AUTH_REGISTER)}>
```
jrasm91, line 16: Can this be changed to a link as well?

jrasm91 (approved):
Looks really good. We should be able to remove user references from everywhere and always use the store instead in the future, right?

martabal:
> Looks really good. We should be able to remove user references from everywhere and always use the store instead in the future, right?

Yes that's the goal

PR #5669, opened 2023-12-13
feat(server): server-side events

jrasm91:
Reliably send events to connected immich-server and immich-microservice containers when the system-config changes. Replaces old solution of using a single BullMQ job.

### DIFF
--- a/server/src/domain/system-config/system-config.service.spec.ts
+++ b/server/src/domain/system-config/system-config.service.spec.ts
@@ -11,14 +11,9 @@ import {
   VideoCodec,
 } from '@app/infra/entities';
 import { BadRequestException } from '@nestjs/common';
-import { newCommunicationRepositoryMock, newJobRepositoryMock, newSystemConfigRepositoryMock } from '@test';
-import { JobName, QueueName } from '../job';
-import {
-  ICommunicationRepository,
-  IJobRepository,
-  ISmartInfoRepository,
-  ISystemConfigRepository,
-} from '../repositories';
+import { newCommunicationRepositoryMock, newSystemConfigRepositoryMock } from '@test';
+import { QueueName } from '../job';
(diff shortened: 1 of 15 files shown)


server/src/domain/system-config/system-config.service.spec.ts
```diff
@@ -269,13 +262,14 @@ describe(SystemConfigService.name, () => {
...
+    it('should update the config and emit client and server events', async () => {
       configMock.load.mockResolvedValue(updates);
 
       await expect(sut.updateConfig(updatedConfig)).resolves.toEqual(updatedConfig);
 
+      expect(communicationMock.broadcast).toHaveBeenCalled();
+      expect(communicationMock.sendServerEvent).toHaveBeenCalledWith(ServerEvent.CONFIG_UPDATE);
       expect(configMock.saveAll).toHaveBeenCalledWith(updates);
-      expect(jobMock.queue).toHaveBeenCalledWith({ name: JobName.SYSTEM_CONFIG_CHANGE });
     });
```
bo0tzz, line 273: Can this test just connect to the channel and assert it receives the message?

bo0tzz (approved):

server/src/domain/system-config/system-config.service.spec.ts
zackpollard: If this is a unit test I think this is the right level to mock and assert. We should have more tests on the event emitters themselves to ensure that we are confident that code works.

danieldietzler (approved):
That's nice!

server/src/domain/system-config/system-config.service.spec.ts
jrasm91: It is a unit test and I think it makes sense like we have it. e2e will definitely take some more effort to check in an automated fashion, but I have at least manually verified that this is working correctly.

PR #5670, opened 2023-12-13
fix(immich-admin): in dev mode

jrasm91:
This adds the `bin/` to the path _in dev_, which makes the `immich` and `immich-admin` commands available out of the box, similar to production.


danieldietzler (approved):
I got confused by this a little while ago when I wanted to use the cli and it wasn't there lol

PR #5689, opened 2023-12-14
fix(ml): more graceful shutdown

mertalev:
## Description

This PR addresses some of the issues with the ML service's automatic restart.

1. The service now performs more checks before beginning a shutdown. 
    - In particular, it now checks that there are no active requests and that no model is being loaded/downloaded (i.e. the model loading lock is not acquired).
2. The service does more cleanup.
    - It now shuts down the thread pool and closes log handlers.
    - It lets Uvicorn perform its own cleanup by sending a shutdown signal instead of canceling async tasks and stopping the loop.
3. Because of the use of a shutdown signal, it presents an INFO log instead of an ERROR log when restarting.
4. It sets the `LD_BIND_NOW` flag, which results in more deterministic RAM usage.

## How Has This Been Tested?

All tasks work as normal and the server is able to restart after idling. RAM usage as inspected from htop is normal as well, including after restart.

### DIFF
--- a/machine-learning/app/config.py
+++ b/machine-learning/app/config.py
@@ -1,12 +1,16 @@
 import logging
 import os
+import sys
 from pathlib import Path
+from socket import socket
 
-import gunicorn
 import starlette
+from gunicorn.arbiter import Arbiter
 from pydantic import BaseSettings
 from rich.console import Console
 from rich.logging import RichHandler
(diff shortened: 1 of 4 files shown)


jrasm91 (approved):
Nice!

bo0tzz (approved):

PR #5693, opened 2023-12-14
feat(web): increase map max zoom

waclaw66:
This PR increases max zoom of the map, to be able see detail location and mainly set gps location to an asset precisely.

|before (14)|PR (18)|
|-|-|
|obrazek|obrazek|

### DIFF
--- a/mobile/lib/modules/map/providers/map_state.provider.dart
+++ b/mobile/lib/modules/map/providers/map_state.provider.dart
@@ -44,7 +44,7 @@ class MapStateNotifier extends StateNotifier<MapState> {
       state.mapStyle != null && state.mapStyle!.rasterTileProvider != null;
 
   double get maxZoom =>
-      (isRaster ? state.mapStyle!.rasterTileProvider!.maximumZoom : 14)
+      (isRaster ? state.mapStyle!.rasterTileProvider!.maximumZoom : 18)
           .toDouble();
 
   void switchTheme(bool isDarkTheme) {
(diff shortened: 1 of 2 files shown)


bo0tzz:
Can you also take a look at the maxZoom in the app?

waclaw66:
> Can you also take a look at the maxZoom in the app?

Done.

danieldietzler:
@bo0tzz didn't we 14 should be the max because the map provider does not support a higher max zoom?

waclaw66:
> @bo0tzz didn't we say 14 should be the max because the map provider does not support a higher max zoom?

Map provider obviously supports 18 as you can see from pictures. 14 is not enought for set a gps location.

bo0tzz:
The provider only supports 14 (although they're pushing out 15 soon). However since they're vector maps, you can still keep zooming beyond that, there just won't be any extra detail. That caused a bit of confusion in the past which is why the zoom level was limited to 14 before.

jrasm91 (approved):

bo0tzz (approved):

danieldietzler (approved):

egutierrez-ar:
Hi folks, I'm not sure if this is the right place to reopen this discussion or if I should create a new ticket, but could this setting be configured at the system-settings page?

`admin/system-settings?isOpen=location+map`

image

Such that if we have a custom map provider that can support higher zoom levels, we can set that up.

danieldietzler:
> Hi folks, I'm not sure if this is the right place to reopen this discussion or if I should create a new ticket, but could this setting be configured at the system-settings page?
> 
> `admin/system-settings?isOpen=location+map`
> 
> image
> 
> Such that if we have a custom map provider that can support higher zoom levels, we can set that up.

Hey, at first thought this sounds reasonable to me. Feel free to open a feature request for it :) (Discussion => Feature Request)

PR #5703, opened 2023-12-14
feat(web,server): search people

martabal:
## Changes made in this PR

This PR adds an option to search people on the /people page. It adds a title on `People` to know how many people the user has on its libraries

## Screenshots

https://github.com/immich-app/immich/assets/74269598/5ed14860-e31e-470a-8d1a-e06250bfa4ea


pstuermlinger:
When you're on the people page (web) and rename a person, then it seems like the original people and peopleCopy get out of sync.

Steps:
1. Search for a name
2. Change name of any result to "foobar"

Effect: name change not visible until you delete the search string

web/src/routes/(user)/people/+page.svelte
```diff
@@ -45,14 +52,31 @@
...
+  let peopleCopy = data.people.people;
+
+  for (const person of peopleCopy) {
     initialHiddenValues[person.id] = person.isHidden;
-  });
+  }
+
+  $: {
+    if (searchName) {
+      people = searchNameLocal(searchName, searchPeopleCopy, 10);
```
pstuermlinger, line 63:
```suggestion
      people = searchNameLocal(searchName, searchPeopleCopy, 100);
```
The result set is limited to 10. Can you raise that to...let's say 100?
I think this is more appropriate for the big people list.

martabal: I increased it to 20; the API to search people returns only 20 people

Batwam:
Rather that a tooltip showing the number, it feels like the info would be more visible if you wrote "People (77)" and updated the qty to match the number of people in the list?

web/src/lib/constants.ts
```diff
@@ -59,6 +60,32 @@ export const dateFormats = {
   },
 };
 
+export enum QueryParameter {
```
jrasm91, line 63: In the future, this would be a good change to do in a separate pull request. If you find yourself refactoring stuff in order to make the current task easier, considering doing it in a separate pull request, as it is easier to review, test, etc. (easier to review means it gets merged faster :smile:)

web/src/lib/components/layouts/user-page-layout.svelte
```diff
@@ -37,7 +38,12 @@
       <div
         class="absolute flex h-16 w-full place-items-center justify-between border-b p-4 dark:border-immich-dark-gray dark:text-immich-dark-fg"
       >
-        <p class="font-medium">{title}</p>
+        <div class="flex gap-2 items-center">
+          <div class="font-medium">{title}</div>
+          {#if secondTitle}
```
jrasm91, line 43: Let's rename this to `description` or something like that, especially if it is small text in a p tag.

jrasm91 (approved):
There are quite a few changes on the web side of things, and I am not familiar enough with the person/face/merge code to know exactly what is going on.

In the future breaking some of the unrelated changes into a smaller, separate PR would be helpful.

The server changes seem fine as far as I can tell as well. I haven't tested the changes, but I assume that you have :smile:

web/src/lib/constants.ts
martabal: Refactoring is too good :stuck_out_tongue_winking_eye: 

All joking aside, yes I should keep separate changes in separate PRs, but every time I start refactoring something I'm like "it's a small change, it's not worth a whole PR for so little change"

For this case, I limited myself to "only" change the way query parameters are handled. I plan to refactor the routing through the webapp.

aviv926:
I'm a little late, but maybe you can make the search something that's just an icon and when you click on it it opens to a search box? I think it's more aesthetic. In any case, this is a great addition, thank you 0:

PR #5730, opened 2023-12-15
fix(web): never ungroup map markers. ever.

danieldietzler:


jrasm91 (approved):
Seems lame to never ungroup them, but I guess it is better than a million photos.

danieldietzler:
> Seems lame to never ungroup them, but I guess it is better than a million photos.

Yeah we could probably do some fancy stuff like the old implementation with a circle that exposes 5 assets or something. But for that I'd need come along with CSS and I certainly don't lol

PR #5733, opened 2023-12-15
fix(docs): remove inline dev inquiries

jonhnet:
* clean up my [inline questions] that should have been resolved in the original conversation
* fixed a relative path link to features/libraries that was broken

### DIFF
--- a/docs/docs/guides/external-library.md
+++ b/docs/docs/guides/external-library.md
@@ -1,6 +1,6 @@
 # External Library
 
-This guide walks you through adding an [External Library](../features/libraries#external-libraries).
+This guide walks you through adding an [External Library](../../features/libraries#external-libraries).
 This guide assumes you are running Immich in Docker and that the files you wish to access are stored
 in a directory on the same machine.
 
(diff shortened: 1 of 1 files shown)


jonhnet:
@jrasm91 review+label por favor

jrasm91 (approved):

docs/docs/guides/external-library.md
```diff
@@ -1,6 +1,6 @@
 # External Library
 
-This guide walks you through adding an [External Library](../features/libraries#external-libraries).
+This guide walks you through adding an [External Library](../../features/libraries#external-libraries).
```
jrasm91, line 3: I don't think this is right, why did you change it?

jonhnet: It seems wrong to me too, but look at the rendered version of the current version. The link doesn't go where it should.

jonhnet:
I've removed the relative path problem from this PR; some deeper mystery needs plumbing. Now this PR is uncontroversial.

PR #5754, opened 2023-12-16
chore(web): migrate to vitest

jorge-dev:
This change fixes an error while trying to run the `npm run dev` command and serve the application.
This was due to a missing alias in the vite config file.

The change is to re-enable the existing unit tests that were failing after the update to Svelte 4.
Most of the tests were failing due to Jest trying to import packages using require instead of import since it was running in CSJ mode instead of ESM mode that Svelte 4 requires.
I spent 2 days trying to configure jest to fix this issue but nothing worked.

That was the main reason to try vitest for this project.
Vitest works with almost no extra configuration and it was built using ESM so  all the tests were able to be fixed with very minor modifications since vitest's api is almost fully compatible with Jest API (with some exceptions) 

These changes only affect the testing environment
image


jrasm91 (approved):
LGTM. Thanks for re-enabling the tests again.

PR #5767, opened 2023-12-16
fix(server): stricter dim size check for pgvecto.rs migration

mertalev:
## Description

This PR changes the migration to now check the clip model being used and determine the dim size from that. It then filters to only migrate the embeddings that have this dim size.

The current implementation looks at the dim size of the embedding in the first row of the smart_info table, which can cause issues if the dim size is (for some reason) inconsistent across rows. It also checks for the presence of NULLs in the embeddings since there seem to be reports of this happening.

## How Has This Been Tested?

I created a 1.90.2 instance, changed the ML model to `XLM-Roberta-Large-Vit-L-14` (which has 768 dims rather than the default models' 512) and upgraded to 1.91 successfully.

### DIFF
--- a/server/src/infra/migrations/1700713871511-UsePgVectors.ts
+++ b/server/src/infra/migrations/1700713871511-UsePgVectors.ts
@@ -31,7 +31,9 @@ export class UsePgVectors1700713871511 implements MigrationInterface {
         INSERT INTO smart_search("assetId", embedding)
         SELECT si."assetId", si."clipEmbedding"
         FROM smart_info si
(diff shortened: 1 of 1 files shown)


bo0tzz:
Will changing a migration like this cause any problems for future upgrades?

zackpollard:
> Will changing a migration like this cause any problems for future upgrades?

Looking at the changes, all it's really doing is checking which model is active and then filtering the inserts so we only insert values that match the current dimsize. So I wouldn't expect this to cause any issues later, as currently anyone experiencing issues that this fixes will be in a broken boot loop

jrasm91 (approved):

PR #5771, opened 2023-12-16
fix(docs): remove spurious hyphen in docker compose cmd

jonhnet:
The command example is correct, but the text just before it still references the old docker-compose command.


jonhnet:
@jrasm91 Boy this hyphen has been nefarious.

jrasm91 (approved):

PR #5806, opened 2023-12-17
fix(server): add more conditions to smart search

mertalev:
## Description

Smart search filters by owner ID, but doesn't check for asset visibility or archive status. This PR adds those conditions.

## How Has This Been Tested?

I did a search, archived the top result and confirmed that it didn't show up when I repeated the search.


alextran1502 (approved):

stevenwcarter:
Should this allow for users to choose if they want to search their archived photos? The UI could default to the current behavior, but I'd think searching archived photos should be possible via the API at least.

Issue #5804, opened 2023-12-17
Startup checks


bo0tzz:
When starting up, the server/microservices should perform some basic checks to make sure everything is in order, and refuse to start with a clear error message if it isn't:

1. [X] Make sure the loaded pgvecto.rs version is as expected (currently v0.1.11)
2. [X] Make sure the Postgres version is at least 14
3. [ ] Make sure that the process can read and write `IMMICH_MEDIA_LOCATION`.

bo0tzz:
Sidenote: seems like pgvecto.rs 0.1.11 incorrectly has the internal version set as 0.1.1

PR #5815, opened 2023-12-18
chore(server): startup check for pgvecto.rs

mertalev:
## Description

This PR checks that the pgvecto.rs extension exists and that it is at the expected version. There are three paths here:

1. The extension is created and the right version, i.e. the check is successful.
2. The extension hasn't been created yet. It tries to create the extension and throws an error if it was unsuccessful.
3. The extension is created, but it isn't the right version. It throws an error in this case.


bo0tzz:
related: #5804

bo0tzz (approved):

PR #5838, opened 2023-12-18
feat(cli): Add existing assets to album and allow album name

etnoy:
* Add assets to existing albums without uploading
* Allow user to specify album name

Thanks to the work by @Willena and adapted from #5843 and #5861

Added e2e tests for these scenarios.


Willena:
Just my thoughts here. 
We should keep the hash check in any situation (unless explicitly disabled by the user). It makes no sense to re-upload the same asset if in the end all we want is adding the asset to an album.

Adding to album should be independent from uploading the asset. 
We could be using the checkBulkUpload response to determine the assetId in case it is a duplicate. 

(FYI I was about to implement this, but I saw your PR)

etnoy:
> Just my thoughts here. We should keep the hash check in any situation (unless explicitly disabled by the user). It makes no sense to re-upload the same asset if in the end all we want is adding the asset to an album.
> 
> Adding to album should be independent from uploading the asset. We could be using the checkBulkUpload response to determine the assetId in case it is a duplicate.
> 
> (FYI I was about to implement this, but I saw your PR)

Thanks for your feedback, maybe I will take another look at this and implement a separate code path for only adding to the album.

Willena:
@etnoy PR #5861 implements my previous comment.

etnoy:
> @etnoy PR #5861 implements my previous comment.

Thanks, I'm currently writing code, too. Unfortunately we can no longer use the old deviceasset id checks as they are not reliable and will be deprecated. After some testing we likely have to upload assets every time in order to match between client and server in a reliable way.

Willena:
> Unfortunately we can no longer use the old deviceasset id checks as they are not reliable and will be deprecated

I'm not sure to understand: the bulk-upload-check api is checking the presence of the asset by comparing the sha1 computed on the server with a sha1 that the cli provides.
https://github.com/immich-app/immich/blob/b40859551b2e792afafdfe4057de6ef1276ad47c/server/src/immich/api-v1/asset/asset-repository.ts#L192-L204
It does not seeam to be using the deviceAssetId to make the comparison; 

The upload api is using the same method to detect a duplicate.
https://github.com/immich-app/immich/blob/b40859551b2e792afafdfe4057de6ef1276ad47c/server/src/immich/api-v1/asset/asset.service.ts#L100

Each time the returned asset id correspond to the uuid generated for that asset the first time it got uploaded; This uuid is independant of the deviceAssetId.

etnoy:
> > Unfortunately we can no longer use the old deviceasset id checks as they are not reliable and will be deprecated
> 
> I'm not sure to understand: the bulk-upload-check api is checking the presence of the asset by comparing the sha1 computed on the server with a sha1 that the cli provides.
> 
> https://github.com/immich-app/immich/blob/b40859551b2e792afafdfe4057de6ef1276ad47c/server/src/immich/api-v1/asset/asset-repository.ts#L192-L204
> 
> 
> It does not seeam to be using the deviceAssetId to make the comparison;
> The upload api is using the same method to detect a duplicate.
> 
> https://github.com/immich-app/immich/blob/b40859551b2e792afafdfe4057de6ef1276ad47c/server/src/immich/api-v1/asset/asset.service.ts#L100
> 
> Each time the returned asset id correspond to the uuid generated for that asset the first time it got uploaded; This uuid is independant of the deviceAssetId.

You are right, my mistake. I misunderstood before and your implementation is correct as it does not use deviceAssetId, only hash. Can I adapt the implementation from your PR into this one, merging all album-related tasks into one PR?

cli/src/commands/upload.ts
```diff
@@ -75,22 +79,36 @@ export default class Upload extends BaseCommand {
...
-            const formData = asset.getUploadFormData();
-            const res = await this.uploadAsset(formData);
+            if (!skipUpload) {
+              const formData = asset.getUploadFormData();
+              const res = await this.uploadAsset(formData);
+              existingAssetId = res.data.id;
+            }
 
-            if (options.album && asset.albumName) {
+            if (asset.albumName !== undefined) {
```
Willena, line 111: Not sure about this check alone: `asset.albumName` is always defined to parent folder name when checked after `asset.process()`. 
`options.album` and `options.albumName` are not tested, it means the cli is always going to create and add the asset to an album. I think the expected behavior is to only add assets to an album if `-A` or `-a` is defined. 
I would use something like  `if ((options.album || options.albumName) && asset.albumName !== undefined) {`

cli/src/commands/upload.ts
```diff
@@ -75,22 +79,36 @@ export default class Upload extends BaseCommand {
         });
 
         let skipUpload = false;
-        if (!options.skipHash) {
+
+        let skipAsset = false;
+        let existingAssetId: string | undefined = undefined;
+
+        if (!options.skipHash && !options.album) {
```
Willena, line 86: Not uploaded assets can now be added to albums without issues. I think it can probably be simplified to 
```suggestion
        if (!options.skipHash) {
```

cli/src/commands/upload.ts
etnoy: Thanks, added another test to catch this case and fixed the code

cli/src/commands/upload.ts
etnoy: Done!

etnoy:
All done, should be ready to merge

alextran1502 (approved):

PR #5843, opened 2023-12-18
feat(cli): Set destination album name in CLI

Willena:
This PR adds a way for the user to specify the destination album name in CLI.


etnoy:
Thanks, I'll likely incorporate this into the related #5838 PR that also handles albums in the CLI

PR #5845, opened 2023-12-18
fix(server): remove shared link with removed asset

### LINKED ISSUE #5842: [BUG] Unable to remove Individual Share item after removing one item from that Individual Share
### The bug

Error when trying to remove an individual share item where 1 or more pictures from that individual share item are removed to trash. When restoring the deleted item(s), the shared link can be removed.

> Failed to remove shared link
> (Immich Server Error)

demo server

### Version of Immich Server

v1.91.3

/

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
/
```

### Your .env content

```Shell
/
```

### Reproduction steps

```bash
1. Select some pictures from your library 
2. Share and create link
3. Remove 1 or more of those pictures
4. Go to "Sharing"->"Shared links"
5. When trying to remove the Individual Share, you are unable to and get an error
```

### Additional information

_No response_

martabal:
fixes #5842

### DIFF
--- a/server/src/infra/entities/shared-link.entity.ts
+++ b/server/src/infra/entities/shared-link.entity.ts
@@ -52,7 +52,7 @@ export class SharedLinkEntity {
   @Column({ type: 'boolean', default: true })
   showExif!: boolean;
 
-  @ManyToMany(() => AssetEntity, (asset) => asset.sharedLinks)
+  @ManyToMany(() => AssetEntity, (asset) => asset.sharedLinks, { onDelete: 'CASCADE', onUpdate: 'CASCADE' })
   assets!: AssetEntity[];
 
   @Index('IDX_sharedlink_albumId')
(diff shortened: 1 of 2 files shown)


jrasm91 (commented):
This is wrong. Deleting an asset will delete _all_ shared links that include it, even if other assets are still shared.

martabal:
> This is wrong. Deleting an asset will delete _all_ shared links that include it, even if other assets are still shared.

I can't experience this behavior, is it a background task or something ? :thinking:

jrasm91:
Did you test adding multiple assets to a single shared link and then deleting one asset, from what I understand the cascade should try to delete the shared link that still contains other assets. Is this what happens?

martabal:
Am I missing something ?

https://github.com/immich-app/immich/assets/74269598/86c0a781-8b5f-4c7f-8f51-e6fb6988470a

jrasm91 (approved):
Ah, I misread the migration originally. This looks good.

PR #5847, opened 2023-12-19
fix(web): open image in new tab with memories on firefox

### LINKED ISSUE #5751: [BUG] When viewing memories in Firefox, the images cannot be saved with the right click menu
### The bug

Normally, for all images in Immich (f.e. in the timeline, from the map, in albums), you can save them by right clicking them and selecting "Save image as". However, in Firefox this does not work with the memories slideshow and instead it just shows the default non-image right click menu with options like "Save page as" and "Select all"

Arch

### Version of Immich Server

v1.91.0

1.91.0 build.115 (but irrelevant to the issue)

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
Same as example
```

### Your .env content

```Shell
Same as example
```

### Reproduction steps

```bash
1. Open Firefox (Tested with 120.0.1 (64-bit) on Arch Linux) and load the frontpage of your immich installation
2. At the top of the page click a memory (f.e. "2 years since"), a slideshow of multiple images of the memory opens
3. Right click on an image to save it

Expected behavior: See right click menu with options like "Save image", "Open image in new tab"
Actual behavior: Right click menu does not have any image-related options, only "Save page as", "Select all", ...
```

### Additional information

Also tested with Chromium Version 119.0.6045.159 and this does show the correct "save image" right click menu.

By inspecting the element in developer tools, I believe the issue is that the div containing the controls (left and right arrows) is displaying on top of the image.
`<div class="absolute flex h-full w-full justify-between">...</div>`

 Deleting this div makes the "save image" right click menu show as expected.

martabal:
fixes #5751

### DIFF
--- a/web/src/lib/components/memory-page/memory-viewer.svelte
+++ b/web/src/lib/components/memory-page/memory-viewer.svelte
@@ -182,14 +182,14 @@
           <div class="h-full w-full rounded-2xl bg-black">
             <!-- CONTROL BUTTONS -->
             <div class="absolute flex h-full w-full justify-between">
-              <div class="ml-4 flex h-full flex-col place-content-center place-items-center">
+              <div class="relative z-[1] ml-4 flex h-full flex-col place-content-center place-items-center">
                 <div class="inline-block">
                   {#if canGoBack}
                     <CircleIconButton icon={mdiChevronLeft} backgroundColor="#202123" on:click={toPrevious} />
                   {/if}
                 </div>
               </div>
-              <div class="mr-4 flex h-full flex-col place-content-center place-items-center">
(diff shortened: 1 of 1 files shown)


jrasm91:
Whhhhhhhhy do we need more z ones changes....

martabal:
You want to be able to click on the button when navigating your memories. I think we also use z indexes for navigation buttons on the asset-viewer.

jrasm91:
Then why can't we draw them on top? Z-index is for overwriting the default stacking order. Why does the default order not work? We should try to fix that if possible instead.

jrasm91 (approved):
I can't test it myself, but the code seems better.

martabal:
Like you suggested, buttons come after the image to avoid using z-index

Issue #5776, opened 2023-12-17
[BUG] Existing assets are not uploaded to album


yaitzme114:
### The bug

This is a clone of https://github.com/immich-app/CLI/pull/49 - it appears that the new CLI has inherited the bug but not the fix.

### The OS that Immich Server is running on

Ubuntu

### Version of Immich Server

v1.90.2

### Version of Immich Mobile App

not relevant

### Platform with the issue

- [ ] Server
- [ ] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
Not relevant
```

### Your .env content

```Shell
Not relevant
```

### Reproduction steps

```bash
1. Upload photos to Immich.
2. Upload an album to Immich via Immich CLI. eg: `immich upload -a <folder>`, where some photos are already in server.
3. See that these photos are not included in the created album.
```

yaitzme114:
I checked a bit further - this does not happen to ALL assets. I am trying to upload my google photos albums.. And only the "portrait mode pictures" (i.e. the ones `*.PORTRAIT.jpg`  files seem to have this problem. When I run the new CLI:

```
immich upload -a $ALBUM

Successfully uploaded 33 assets (31.3 MB)
```

At this point my album had 33 assets


When I use the old CLI:

```
immich upload --key <key>--server <endpoint> -d $ALBUM -t 16 -al

Warning: deprecated option --directory used, this will be removed in a future release. Please specify paths with --recursive instead
Checking connectivity with Immich instance...
Server status: OK
Checking credentials...
Login status: OK
Successful authentication for user <user>
Indexing local assets...
Indexing complete, found 52 local assets
Comparing local assets with those on the Immich instance...
A total of 52 assets will be uploaded to the server
A total of 52 assets will be added to album(s).
NOTE: some assets may already be associated with the album, this will not create duplicates.
Do you want to start upload now? (y/n) y
Start uploading...
Upload Progress | ████████████████████████████████████████ | 100% || 52/52 || Current file
Creating albums...
Upload Progress | ████████████████████████████████████████ | 100% || 1/1 || Current file [{filepath}]
```

At this point, the album has 52 assets.

boydthomson:
I'm getting similar on Debian with 1.91.3

When I try to upload a single file, I get: 
$immich upload amc02568_52458970544_o.jpg 
No assets found, exiting
$

boydthomson:
Trying to upload a dir/album yields considerably more output:

$ immich upload -a Logos/
[...]

PR #5861, opened 2023-12-19
feat(cli): Add assets to album when detected as duplicates

Willena:
When upload is skipped because of a duplicate asset, allow the asset to be added to the album. 

The solution proposed in PR #5838 disable the hash check before upload. We should keep it and use the response from `checkBulkUpload` to determine the assetId to add to the album. 

This is fixing https://github.com/immich-app/immich/issues/5776


etnoy:
Superseded by #5838

PR #5869, opened 2023-12-19
fix(mobile): use cronet / cupertino http clients

shenlong-tanwen:
#### Changes made:

- Updates the mobile app to use [cupertino_http](https://pub.dev/packages/cupertino_http) on iOS and [cronet_http](https://pub.dev/packages/cronet_http) on Android devices which should hopefully respect the system proxy and certs

### DIFF
--- a/mobile/ios/Podfile.lock
+++ b/mobile/ios/Podfile.lock
@@ -2,6 +2,8 @@ PODS:
   - connectivity_plus (0.0.1):
     - Flutter
     - ReachabilitySwift
+  - cupertino_http (0.0.1):
+    - Flutter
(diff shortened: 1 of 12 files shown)


fyfrey:
I think this breaks the f-droid build and APK install on devices without Google.Play Services. We might need to use cronet_embedded for these cases

shenlong-tanwen:
> I think this breaks the f-droid build and APK install on devices without Google.Play Services. We might need to use cronet_embedded for these cases

Yes, we can easily override the cronet dep with cronet_embedded for the f-droid build. I'm not actually sure that this works though, so waiting on few users to actually test the change and confirm it

PR #5903, opened 2023-12-20
feat(server,ml): remove image tagging

mertalev:
## Description

This job has been effectively deprecated for some time and is currently disabled by default. The removal of it will be good in several respects:

1. Many dependencies can be removed from the ML service
2. Crashes on ARM devices will no longer occur due to the classification model being exported at runtime
3. Dependency updates can be automated through Renovate
4. Avoids user confusion -- users often confuse it with the more useful smart search feature

The image size has been shrunk from 1.66gb to 762mb.

The ML service's RAM usage at startup has decreased from 320mb to 215mb.

## How Has This Been Tested?

Tested that the server boots up and that the ML endpoints still work.


bo0tzz (approved):

jrasm91:
Why did we change the behavior of incorrect values in the config file? Did we decide to make them warnings instead?

mertalev:
It behaves the same for incorrect values, but now warns instead of throwing for extras. I think this is less disruptive; it doesn't need to error there as long as it lets the admin be aware that the option wasn't used. It's not like setting an unused env will throw either.

jrasm91:
Fair enough.

PR #5917, opened 2023-12-21
feat: storage template file move hardening

zackpollard:
Storage template moving can sometimes fail and cause data corruption when the upload folder and library folder are not the same filesystem. This PR aims to address the existing problem and add validation of the file operations to ensure the moved file data matches the original asset data by the way of file size checks and optional hash checks.

TODO:
- [x] When file is copied rather than renamed, validate filesize in database matches filesize on disk
- [x] When file is copied rather than renamed, validate hash in database matches hash of copied asset
- [x] Change job concurrency to one and remove configuration options for this
- [x] Add setting to enable and disable storage template system entirely, default storage template system to off
- [ ] ~~Change structure of default upload folder to split GUIDs into seperate folders to improve performance~~ **Will be done in seperate PR**

Optional TODO:
- [ ] ~~Have the filename itself indicate if it is partial or not (.tmp suffix?)~~ **This actually substantially increases complexity with the move table, not going ahead with this for now**
- [ ] ~~Keep an audit log of moves/file operations in the database~~ **Could be done in seperate PR**
- [ ] ~~Use database locks to force template migration concurrency to one, avoiding race conditions where two processes move a file with the same name to the same location, not appending a +n~~ Done in #6054


docker/docker-compose.dev.yml
```diff
@@ -16,6 +16,7 @@ x-server-build: &server-common
   volumes:
     - ../server:/usr/src/app
     - ${UPLOAD_LOCATION}/photos:/usr/src/app/upload
+    - ${UPLOAD_LOCATION}/photos/upload:/usr/src/app/upload/upload
     - /usr/src/app/node_modules
```
jrasm91, line 20: Why are you changing this? This seems like a noop

server/src/infra/database.config.ts
```diff
@@ -20,6 +20,7 @@ export const databaseConfig: PostgresConnectionOptions = {
   subscribers: [__dirname + '/subscribers/*.{js,ts}'],
   migrationsRun: false,
   connectTimeoutMS: 10000, // 10 seconds
+  parseInt8: true,
```
jrasm91, line 23: This should be a separate PR imo.

server/package.json
```diff
@@ -28,7 +28,7 @@
     "typeorm:migrations:generate": "typeorm migration:generate -d ./dist/infra/database.config.js",
     "typeorm:migrations:run": "typeorm migration:run -d ./dist/infra/database.config.js",
     "typeorm:migrations:revert": "typeorm migration:revert -d ./dist/infra/database.config.js",
-    "typeorm:schema:drop": "typeorm schema:drop -d ./dist/infra/database.config.js",
+    "typeorm:schema:drop": "typeorm query -d ./dist/infra/database.config.js 'DROP schema public cascade; CREATE schema public;'",
```
jrasm91, line 31: Depending how long this PR takes this would also be good to just fix on main immediately.

docker/docker-compose.dev.yml
bo0tzz: I'm guessing that having them as two separate mounts means they get treated as different filesystems, hence the rename operation needs to become a move, causing that to get implicitly tested during dev?

server/src/infra/database.config.ts
zackpollard: I require the change for this PR, unless I want to change the type back to string. We only have one usage of bigint and that's all this change should affect.

server/package.json
zackpollard: Feel free to just cherry pick this onto main

docker/docker-compose.dev.yml
zackpollard: Exactly this, I didn't actually plan to include this change, but thinking about it, it might be worth it so we're always testing the copy mechanism in dev

server/src/infra/database.config.ts
jrasm91: That's fine. I think we have more than one use of this though and it is just easier to test as a separate change. This returns all _numbers_ as numbers instead of strings, right?

server/package.json
jrasm91: Cool, sounds good.

server/src/domain/storage-template/storage-template.service.spec.ts
```diff
@@ -236,12 +271,47 @@ describe(StorageTemplateService.name, () => {
...
+        id: '123',
+        entityId: assetStub.image.id,
+        pathType: AssetPathType.ORIGINAL,
+        oldPath: assetStub.image.originalPath,
+        newPath: 'upload/library/user-id/2023/2023-02-23/asset-id.jpg',
+      });
+      // @ts-expect-error we only use the size property, no point implementing all the other properties
+      when(storageMock.stat).calledWith('upload/library/user-id/2023/2023-02-23/asset-id.jpg').mockResolvedValue({
+        size: 100,
+      });
```
danieldietzler, line 291: I would prefer `.mockResolvedValue({ size: 100 } as Stats)` here instead of having this annotation

server/src/domain/storage/storage.core.ts
```diff
@@ -140,29 +143,74 @@ export class StorageCore {
 
     let move = await this.moveRepository.getByEntity(entityId, pathType);
     if (move) {
+      console.log('Found existing move record for', entityId, pathType, oldPath, newPath);
```
danieldietzler, line 146: You're probably removing removing this anyways, just wanting to make sure it doesn't get overlooked when finalizing the PR :)

server/src/domain/storage/storage.core.ts
```diff
@@ -140,29 +143,74 @@ export class StorageCore {
...
+        if (!(await this.isNewPathSameSize(move.oldPath, move.newPath, assetInfo))) {
+          this.logger.fatal(`Skipping move due to file size mismatch, old file is missing and new file is different`);
+          return;
+        }
+      }
 
       move = await this.moveRepository.update({ id: move.id, oldPath: actualPath, newPath });
     } else {
       move = await this.moveRepository.create({ entityId, pathType, oldPath, newPath });
     }
```
danieldietzler, line 169: We should probably flip this to decrease the overall indentation. So `if (!move) { ... } else { ... }`

server/src/domain/storage/storage.core.ts
```diff
@@ -140,29 +143,74 @@ export class StorageCore {
...
+        }
+        this.logger.debug(`Unable to rename file. Falling back to copy, verify and delete`);
+        await this.repository.copyFile(move.oldPath, newPath);
+
+        if (!(await this.isNewPathSameSize(move.oldPath, newPath, assetInfo))) {
+          this.logger.warn(`Skipping move due to file size mismatch`);
+          return;
+        }
+
+        await this.repository.unlink(move.oldPath);
```
danieldietzler, line 195: What if deleting the file fails?

danieldietzler (approved):
Looking good so far! :)

server/src/infra/database.config.ts
zackpollard: My understanding is it parses all 8 byte number types into a number in typeorm. Postgres number types are variable length and to my understanding will still be strings. This also only affects integer types, so double precision won't be affects either. It should just be bigint and bigserial.

server/src/domain/storage-template/storage-template.service.spec.ts
zackpollard: Ah yes, forgot you can do that

server/src/domain/storage/storage.core.ts
zackpollard: Yea, already have that removed locally actually :laughing:

server/src/domain/storage/storage.core.ts
zackpollard: I think at that point we leave it to become orphaned? Would we really want to fail the entire move for that? I guess we should at least try catch around that though

danieldietzler: Yeah that's what I was thinking.

server/src/domain/storage/storage.core.ts
danieldietzler: Btw nvm this. This would only apply if we were to move all of this to a separate method and early return from it

zackpollard:
Ready for review and merge :smile:

danieldietzler (commented):
Have you thought about nesting the files by default in a `ab/cd/abcdxx-xxxx-xxxxx` structure?

server/src/domain/storage/storage.core.ts
```diff
@@ -143,26 +168,93 @@ export class StorageCore {
...
 
-      this.logger.log(`Found file at ${actualPath === move.oldPath ? 'old' : 'new'} location`);
+      const fileAtNewLocation = actualPath === move.newPath;
+      this.logger.log(`Found file at ${fileAtNewLocation ? 'new' : 'old'} location`);
+
+      if (fileAtNewLocation) {
+        if (!(await this.verifyNewPathContentsMatchesExpected(move.oldPath, move.newPath, assetInfo))) {
+          this.logger.fatal(
+            `Skipping move as file verification failed, old file is missing and new file is different to what was expected`,
+          );
```
bo0tzz, line 184: Is there any more action we can take in this case? Do assets that hit here implicitly also show up as offline on the repair page, or do we want to do something to make that happen?

server/src/domain/storage/storage.core.ts
```diff
@@ -143,26 +168,93 @@ export class StorageCore {
...
+          return;
+        }
+        this.logger.debug(`Unable to rename file. Falling back to copy, verify and delete`);
+        await this.repository.copyFile(move.oldPath, newPath);
+
+        if (!(await this.verifyNewPathContentsMatchesExpected(move.oldPath, newPath, assetInfo))) {
+          this.logger.warn(`Skipping move due to file size mismatch`);
+          return;
+        }
```
bo0tzz, line 216: In this case the asset still uses the old path, right? Should we do any cleanup of the new one?

server/src/domain/storage/storage.core.ts
```diff
@@ -143,26 +168,93 @@ export class StorageCore {
...
+            'base64',
+          )}`,
+        );
+        return false;
+      }
+      this.logger.debug(`File checksum check: ${newChecksum.toString('base64')} === ${checksum.toString('base64')}`);
+    }
+    return true;
+  }
+
```
bo0tzz, line 257: This will always fail in cases where users have changed the file on-disk behind Immich's back (but maybe that's OK, just good to be aware of)

bo0tzz (commented):
Good stuff. Did you not end up adding the locking to this?

server/src/domain/storage/storage.core.ts
danieldietzler: I think at this point you're pretty much screwed. Even with manual intervention I don't think you can do anything to fix this given that the original file is lost. Notifying the user that a file got lost and that they restore it from a backup rather quickly seems like a good idea in general. Not sure if the repair page is a good place for this though. People may not look at it (at all).

server/src/domain/storage/storage.core.ts
danieldietzler: I think that's a good call because otherwise 180 would still be true in subsequent tries and we'd be running in 181/182. Thus, the file never will be fixed.

alextran1502:
I just tested the PR, it seems like the motion part of LivePhotos aren't moved to the library

server/src/domain/storage-template/storage-template.service.ts
```diff
@@ -93,6 +106,10 @@ export class StorageTemplateService {
 
   async handleMigration() {
     this.logger.log('Starting storage template migration');
+    if ((await this.configCore.getConfig()).storageTemplate.enabled === false) {
```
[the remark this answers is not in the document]
zackpollard: I specifically put false because with these huge if statement's it's easy to miss the !, but happy to switch it out :P

server/src/domain/storage/storage.core.ts
zackpollard: At this point the old file is somehow gone and the new file is corrupt, so you're f**ked. That's why I log it as fatal. The file will show up as orphaned. It's worth noting that with all these extra checks, it should actually be impossible to get to this state now.

server/src/domain/storage/storage.core.ts
zackpollard: Yea we could attempt a cleanup, I just wanted to avoid doing even more stuff in here tbh

server/src/domain/storage/storage.core.ts
zackpollard: That's not something we support and they shouldn't be doing it, so I think failing here is fine.

zackpollard:
> I just tested the PR, it seems like the motion part of LivePhotos aren't moved to the library

We have explicit tests that check this behaviour so they should be. I haven't actually changed any of that logic, it's just two seperate calls to move each seperate part.

zackpollard:
> Good stuff. Did you not end up adding the locking to this?

No, I will only add locking if the PR refactoring the locking mechanism gets merged before this PR does.

zackpollard:
> Have you thought about nesting the files by default in a `ab/cd/abcdxx-xxxx-xxxxx` structure?

Forgot about that, can change it to that structure, yea

server/src/domain/storage-template/storage-template.service.ts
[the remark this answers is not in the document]
danieldietzler: Gotcha. It just reads wrong to me lol... I've had this discussion in a different case (not instanceof) with Jason as well :D

I guess it comes down to personal preference after all...

danieldietzler:
> > Have you thought about nesting the files by default in a `ab/cd/abcdxx-xxxx-xxxxx` structure?
> 
> Forgot about that, can change it to that structure, yea

Ty :)

server/src/domain/storage-template/storage-template.service.ts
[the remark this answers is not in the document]
zackpollard: I would usually always default to ! but it just seems to hide amongst all the brackets. I can pull it out to a variable and then ! the variable to make it more readable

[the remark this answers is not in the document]
danieldietzler: Good idea, I like that!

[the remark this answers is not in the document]
danieldietzler: Maybe also just extract the config to a variable and then have `!config.storageTemplate.enabled`.

server/src/domain/storage/storage.core.ts
zackpollard: I added a cleanup but if it fails we just error out 😄

danieldietzler (approved):
Great stuff 🚀

alextran1502:
@zackpollard This condition prevent moving the video part of the LivePhotos upload

image

zackpollard:
Hmmm, do you know why we aren't extracting this info?

zackpollard:
> @zackpollard This condition prevents moving the video part of the LivePhotos upload
> 
> image
> 
> image
> 
> I tested with `main` on a small set of LivePhotos, and they are moved correctly.
> 
> If I remember correctly, we are not extracting EXIF and add a record in EXIF table for the motion part of LivePhotos
> 

Hmmm, do you know why we aren't extracting this info?

zackpollard:
Changed exif extraction to run for all assets, including assets marked isVisible = false. This should resolve the issues with live photos not moving correctly.

alextran1502 (approved):
LGTM

PR #5953, opened 2023-12-23
chore(server): refactor locks

mertalev:
## Description

This PR moves locking functionality to the new database repository. Specifically, it adds three methods and replaces existing lock usages with them:

* `withLock` takes a lock type and an async callback and acquires an async lock as well as a Postgres advisory lock. It wraps the callback in a try / finally clause that ensures the advisory lock is released.
* `isBusy` is a synchronous check to see if a lock is being used without actually acquiring it
* `wait` asynchronously waits for the lock to be released by acquiring and immediately releasing it
  * It is more lightweight than `withLock` for this purpose as it only relies on the async lock, not a Postgres advisory lock

# How Has This Been Tested?

The server booted, running a Smart Search job worked, and changing the CLIP model while jobs were running had the same behavior (i.e. the active jobs waited for the update to complete, then errored because they had different size embeddings, while later jobs completed successfully).


server/src/infra/repositories/metadata.repository.ts
```diff
@@ -47,6 +48,17 @@ export class MetadataRepository implements IMetadataRepository {
 
     this.logger.log('Importing geodata to database from file');
 
+    this.databaseRepository.withLock(DatabaseLock.GeodataImport, this.importGeodata);
```
jrasm91, line 51: This should be awaited, the callback should be () => this.importGeodata()

mertalev: Oops, forgot to await that. I think passing the method works fine though, any particular reason to wrap it like that?

jrasm91: Otherwise you can't use "this" in the method correctly. It probably isn't working correctly right now.

mertalev: Oh... yeah, that seems like a good reason to change it.

server/src/infra/repositories/database.repository.ts
```diff
@@ -25,4 +28,34 @@ export class DatabaseRepository implements IDatabaseRepository {
   async runMigrations(options?: { transaction?: 'all' | 'none' | 'each' }): Promise<void> {
     await this.dataSource.runMigrations(options);
   }
+
+  async withLock<R>(lock: DatabaseLock, callback: () => Promise<R>): Promise<R> {
+    let res;
+    await this.asyncLock.acquire(DatabaseLock[lock], async () => {
```
jrasm91, line 34: Why can't this just be "lock"?

jrasm91 (approved):

server/src/infra/repositories/database.repository.ts
mertalev: There are two kinds of locks being used here, so I thought it'd be good to specify the kind in the name. Not really needed though since it's already an `AsyncLock`.

jrasm91: I meant you are unexpected into an enum with an enum

mertalev: Oh, that was a typing thing: `DatabaseLock is not assignable to parameter of type 'string | string[]`

jrasm91: Hmm. We probably want to just cast it in that case.

Issue #5873, opened 2023-12-19
Clean up unnecessarily transcoded videos


bo0tzz:
When an admin changes the video transcoding settings, it's possible to end up with assets that still have a transcoded video file even though according to the new settings they shouldn't. When running the transcoding job, if it is determined that [a transcode is not required](https://github.com/immich-app/immich/blob/main/server/src/domain/media/media.service.ts#L246-L249), the code should clean up the existing transcode file if there is one and remove the path from the database.

Batwam:
I would personally prefer to keep already transcoded videos and only skip transcoding for future assets. This would be consistent with the way pictures are only rearranged when "Storage Template Migration" is clicked.

Therefore, rather than deleting all transcoded videos right away, I believe that the correct behaviour would be to only delete previously transcoded videos if the user voluntarily clicks on "ALL" under Transcoded Videos"?

bo0tzz:
> the correct behaviour would be to only delete previously transcoded videos if the user voluntarily clicks on "ALL" under Transcoded Videos"

That would be the behaviour with this change. The file would only get deleted when running the job, which in turn would only happen if you click "ALL"

Batwam:
I was wondering actually, are transcoded videos currently deleted if a video is no longer in the library (either because they have been moved by the user or deleted within the app)?

bo0tzz:
Yes: 
https://github.com/immich-app/immich/blob/dea1063b17ef24e3a9616953d5279e6f2a19425f/server/src/domain/asset/asset.service.ts#L489-L496

PR #5958, opened 2023-12-24
Edit and update of FAQ+new database GUI guide

aviv926:
Added edit and update of FAQ:
- Division by subject
- Editing of irrelevant things (object detection) + cloudflare 100M limit
- Adding **many** repetitive questions to FAQ
- Adding a "comparison" page in the overview for people from different platforms
- Added all kinds of explanations about how ML works

**Instructors:**
  Added database login guide using PgAdmin (GUI).

**Note:**
My English is not 100% and I use a translation, maybe you should follow grammar and words that probably weren't translated properly, I did my best so that there are no situations where there is a spelling mistake

**Note 2:**
i want to use docusaurus with
import TOCInline from '@theme/TOCInline';

https://docusaurus.io/docs/markdown-features/toc#inline-table-of-contents

So that I can display all the TOC from all MD pages in FAQ.mdx
For a quick search for a question among all FAQ departments

Do you think there is a way I can make this happen?

**It's not finished yet, but I'd like you to take a look at it**


danieldietzler:
You should be able to `npm i @theme/TOCInline` in the docs/ folder and then the import in the faq.mdx file should work. (You can only use react/ts stuff in *.mdx files)

danieldietzler (commented):
First off, content-wise it's already really good and *very* helpful! 

Sorry if seeing 38 review comments is a little slaying. It ended up being a lot of little nitpicks but I think that's what you wanted, especially given that you used a translator for it. 

I've sometimes skipped sections that felt WIP.
Generally, you should pay attention to capitalization and missing periods (there are just multiple sentences without a '.')
I hope this is somewhat helpful though! :) Really appreciate your hard work on this (so far)!

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,125 @@
...
+([Sources](https://discord.com/channels/979116623879368755/1071165397228855327/1187088114599071855))
+
+### I don't speak English, is there a way I can use the Smart Serach in my language?
+
+Yes, you can get such an option, to do this you need to check what is the best model according to your language, you can do it by checking the [table](https://github.com/mlfoundations/open_clip/blob/main/docs/openclip_multilingual_retrieval_results.csv).
+
+:::note
+If you found a suitable model for your language and you want to use it you may not be able to because Immich didn't add it to their list of models in [Huggingface](https://huggingface.co/immich-app), you can make it happen by send a [pull request](https://github.com/immich-app/immich/pulls) via Github to add this model to Immich.
```
mertalev, line 117: Just say "feel free to make a feature request if there's a model you want to use that isn't in this list".

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,125 @@
...
+
+It is important to check the level of accuracy of the model and the amount of use of processing power Flops(B) in [this table](https://github.com/mlfoundations/open_clip/blob/main/docs/openclip_results.csv) so you can know it will work well with your system.
```
mertalev, line 119: This is too abstract to be helpful. I'd just talk about using a base size model vs. large, huge, etc.

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,125 @@
...
+:::
+
+### Does Immich support Facial Recognition in video ?
+
+Although it seems that insightface is able to [recognize faces in video](https://github.com/deepinsight/insightface#arcface-video-demo) as of now the option is not yet implemented in Immich.
```
mertalev, line 124: Just say "this is not yet implemented in Immich".

docs/docs/FAQ/Performance-FAQ.md
```diff
@@ -0,0 +1,48 @@
...
+3. jobs
+4. On the left side Manage Concurrency
+5. Change the settings as needed
+
+:::danger
+On a normal machine, 2 or 3 concurrent jobs can probably max the CPU, so if you're not hitting those maximums with, say, 30 jobs,
+Also, it is important to know that the Storage should have INPUT & OUTPUT at high speed in order to handle all of this
+
+For reference I never went above 32/job while testing with an RTX 4090 GPU
+Or for example if machine learning is enabled by a CPU such as I7 8700 a bit higher than default here would be like 16/job.
```
mertalev, line 39: Remove these.

docs/docs/FAQ/Performance-FAQ.md
```diff
@@ -0,0 +1,48 @@
...
+
+Do not exaggerate with the amount of jobs because you're probably thoroughly overloading the database.
```
mertalev, line 41: Change "overloading the database" to "overloading the CPU".

docs/docs/FAQ/Performance-FAQ.md
```diff
@@ -0,0 +1,48 @@
...
+
+more info [here](https://discord.com/channels/979116623879368755/994044917355663450/1174711719994605708)
+:::
+
+
+### When you run machine learning my processor is 100% usge is it noraml?
```
mertalev, line 47: The "how can I reduce CPU usage" section already goes into this, so I don't think this question is necessary.

aviv926:
> First off, content-wise it's already really good and *very* helpful! 
> 
> Sorry if seeing 38 review comments is a little slaying. It ended up being a lot of little nitpicks but I think that's what you wanted, especially given that you used a translator for it. 
> 
> I've sometimes skipped sections that felt WIP.
> Generally, you should pay attention to capitalization and missing periods (there are just multiple sentences without a '.')
> I hope this is somewhat helpful though! :) Really appreciate your hard work on this (so far)!

Wow, that's great!  I will work on the fixes soon!  Thanks for the corrections and your time :)

mertalev (dismissed):
Thanks for working on this! My feedback is generally about making the FAQ as useful as possible for the average user. When in doubt, keep it simple. Also, don't link to spreadsheets, Discord or Reddit. The information in the FAQ should stand on its own and be presented in an accessible way.

mertalev:
> Thanks for working on this! My feedback is generally about making the FAQ as useful as possible for the average user. When in doubt, keep it simple. Also, don't link to spreadsheets, Discord or Reddit. The information in the FAQ should stand on its own and be presented in an accessible way.

docs/docs/FAQ/Assets-FAQ.md
```diff
@@ -0,0 +1,63 @@
...
+
+### How can I backup data from Immich?
+
+See [backup and restore](/docs/administration/backup-and-restore.md).
+
+### Does Immich supports reading faces tags from the Exif?
+
+not at this moment
+
+### Does Immich support filtering of images defined as NSFW?
```
mertalev, line 31: Change "filtering of images defined as NSFW" to "NSFW images".

docs/docs/FAQ/Assets-FAQ.md
```diff
@@ -0,0 +1,63 @@
...
+
+As of now, this option is not implemented, but it seems that there is an [open discussion about it
```
mertalev, line 33: Remove "it seems that"

docs/docs/FAQ/Assets-FAQ.md
```diff
@@ -0,0 +1,63 @@
...
+On Github](https://github.com/immich-app/immich/discussions/2451) you can submit a pull request or vote for the discussion
```
mertalev, line 34: Add a period before the "you", capitalize to "You" and add a period after "discussion".

docs/docs/FAQ/Docker-FAQ.md
```diff
@@ -0,0 +1,62 @@
...
+
+```bash title="Remove Immich (containers and volumes)"
+docker compose down -v
+```
+
+After removing the containers and volumes, the **Files** can be cleaned up (if necessary) from the `UPLOAD_LOCATION` by simply deleting an unwanted files or folders.
+
+## Docker errors
+
+### I am getting `Could not find image processor` ERROR class in ML containers what can i do?
```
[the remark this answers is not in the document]
aviv926: Yes, as I mentioned in :::note, this is a problem that is no longer relevant, I decided to add it if people have problems with it and maybe during the next month or two it can be removed

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,125 @@
+---
+sidebar_position: 6
+---
+# Machine Learning
+
+### What is the technology that Immich use for machine learning?
+
+- For search function (smart search) - Immich uses Clip Models, for more information about Clip and its capabilities can be [read here](https://openai.com/research/clip) and [experience here](https://huggingface.co/models?pipeline_tag=zero-shot-image-classification&sort=trending).
```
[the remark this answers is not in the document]
aviv926: In fact it can give people another point of view on how different models react to different images (if someone wants to test a model that is more trained to "recognize" types of vehicles)

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,125 @@
...
+### What is the best order for ML to best work
+
+Because Immich is relies on thumbnail generation
+For machine learning [[1](http://localhost:3005/docs/developer/architecture#:~:text=For%20example%2C%20Smart%20Search%20and%20Facial%20Recognition%20relies%20on%20thumbnail%20generation)], it is recommended to wait for the thumbnail generation job to finish and then run the machine learning jobs (Smart Search and Facial Recognition)
+
+* Let all jobs run except the machine learning jobs and wait for them to finish
+* Run Smart Search
+* Run Facial Recognition
+
+### What are the best ML models I can use for Immich
```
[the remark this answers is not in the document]
aviv926: This is something I thought about when I wrote it, on the one hand I didn't want to make it too cumbersome but a short answer for the simple user.

Therefore everything written in "why is it?" I put in the label "section" if the user clicks on it it opens, and only if the user is interested in knowing why the model was chosen

If you think it's something less relevant, maybe I should add all this information on a separate page called "Extended/Advanced?" Or just remove it?

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,125 @@
...
+The [export script](https://github.com/immich-app/immich/tree/main/machine-learning/export) in Immich Github page is meant for us to be able to automate exporting and uploading to Huggingface you can send a [pull request](https://github.com/immich-app/immich/pulls) via Github to add this model to Immich check if Immich already have this model in the list of models in [Huggingface](https://huggingface.co/immich-app) before you send pull request.
+([Sources](https://discord.com/channels/979116623879368755/1071165397228855327/1187088114599071855))
+
+### I don't speak English, is there a way I can use the Smart Serach in my language?
+
+Yes, you can get such an option, to do this you need to check what is the best model according to your language, you can do it by checking the [table](https://github.com/mlfoundations/open_clip/blob/main/docs/openclip_multilingual_retrieval_results.csv).
```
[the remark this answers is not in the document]
aviv926: I thought of listing the name of a model with a high percentage of accuracy in Multilingual but I could not find the models
LABSE-Vit-L-14
and
XLM-Roberta-Large-Vit-L-14

In the open_clip performance report. That's why I left a link to the table so that people can check for themselves what would be the best option.
In any case if you think it is less relevant I will remove it.

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,125 @@
...
+5. rerun the jobs of ML to take effect
+
+</details>
+
+:::note
+Changing a model does not delete the previous model, pay attention to delete the old model to save space in drive.
+Models are stored in IMMICH_MODEL-CACHE.
+:::
+
+### I would like to use the Clip model i created, how can i do that?
```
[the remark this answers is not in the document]
aviv926: I've seen a few recurring questions about it on Discord at different times, I assumed that maybe it could go into the FAQ (maybe another new "advanced" page?)

docs/docs/FAQ/Machine-Learning-FAQ.md
aviv926: hahaha and I also thought of providing a link to a page that can provide information about Flops(B) etc. but I guess it's not that much needed for a simple user

docs/docs/FAQ/Performance-FAQ.md
aviv926: In fact, I intentionally wrote overloading the database because it was information I received in [Discord](https://discord.com/channels/979116623879368755/994044917355663450/1174713157034451015) when I increased the amount to a value that was too high
Is this no longer a relevant problem in the context of the database?

docs/docs/FAQ/Performance-FAQ.md
aviv926: The question here is a little different because it answers whether it is something normal for Immich to do.
A lot of people complain that their processor reaches 100% while running machine learning and they want to know if this is a normal situation, and not necessarily to lower the amount of cores that Immich has

docs/docs/FAQ/Performance-FAQ.md
danieldietzler: I guess you can easily end up with I/O bottlenecks, yes. If you have a slow drive this may become an issue. So yes, the database may also be "overloaded" but the cause of that is a slow drive.

mertalev: The root cause of the issue is ultimately the CPU. Postgres can easily handle thousands of requests, but choking the CPU will also choke the database.

docs/docs/FAQ/Performance-FAQ.md
mertalev: That's fair. Maybe it'd be better to split the lowering CPU question into "Why is Immich using so much of my CPU?" and move the explanation there. I think this is the question people would have, not so much why machine learning is using so much, but why the CPU is at 100%.

docs/docs/FAQ/Machine-Learning-FAQ.md
mertalev: Yeah, I think this will just leave people with more questions than they started with.

docs/docs/FAQ/Machine-Learning-FAQ.md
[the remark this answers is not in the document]
mertalev: The models in that link can't actually be used for Immich, so I think it will just lead to confusion. The answer here should just be about what Immich actually does.

docs/docs/FAQ/Machine-Learning-FAQ.md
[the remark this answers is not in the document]
mertalev: Those models are from M-CLIP, not OpenCLIP. I think this is a useful question, but it should be about what Immich supports. 

Also, the question is kinda funny. If they don't know English, they can't read the answer either 😄 

You can change it to "I want to be able to search in other languages besides English. How can I do that?" with the answer:

"This is supported. You can change to a multilingual model listed [here](https://huggingface.co/collections/immich-app/multilingual-clip-654eb08c2382f591eeb8c2a7) by going to Administration > Machine Learning Settings > Smart Search and replacing the name of the model. You can then search in over 100 languages."

docs/docs/FAQ/Machine-Learning-FAQ.md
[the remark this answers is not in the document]
mertalev: In that case, change the question to "Can I use a custom model?" with the answer "No, this is not supported. Only models listed in the [Hugging Face org](https://huggingface.co/immich-app) are compatible. Feel free to make a feature request if there's a model not listed here that you think should be added."

docs/docs/FAQ/Machine-Learning-FAQ.md
mertalev: Given the above question, this note is redundant.

docs/docs/FAQ/Machine-Learning-FAQ.md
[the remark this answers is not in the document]
mertalev: I think you can just remove it.

docs/docs/FAQ/Docker-FAQ.md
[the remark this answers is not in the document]
mertalev: I thin it'd be better to make this question more generally about worker crashes. Change the question to "Why does the machine learning service report workers crashing?" with this answer: 

:::note
If the error says the worker is *exiting*, then this is normal. This is a feature intended to reduce RAM consumption when the service isn't being used.

There are a few reasons why this can happen.

If the error mentions `SIGKILL` or error code 137, it most likely means the service is running out of memory. Consider either increasing the server's RAM or moving the service to a server with more RAM.

If it mentions `SIGILL` (note the lack of a K) or error code 132, it most likely means your server's CPU is incompatible. This is unlikely to occur on version 1.92.0 or later. Consider upgrading if your version of Immich is below that.

If your version of Immich is below 1.92.0 and the crash occurs after logs about tracing or exporting a model, consider either upgrading or disabling the Tag Objects job.

docs/docs/FAQ/Performance-FAQ.md
bo0tzz: > The root cause of the issue is ultimately the CPU

I don't think this is necessarily true. Immich does a LOT of I/O, and starving the disk can easily grind a database to a halt as well.

mertalev: Fair enough, in some cases it might be an IO issue. Maybe just say "overloading the server".

aviv926:
I will soon make all the necessary corrections.
I thought I would add more questions to the FAQ, I would like to know if it could be relevant in your opinion.
@danieldietzler @mertalev 

**Under Assets -> How to delete transcoded assets if I don't need them anymore?**
(Actually I'm not 100% sure about the answer because I saw in Discord the answer that ZACK answered https://discord.com/channels/979116623879368755/994044917355663450/1186359061265010748 but I couldn't figure out if it's something that exists or not).

**Under Assets -> What happens when 2 users have access to load from the same library?**
A. All machine learning jobs and thumbnail images are recreated.

**Under Assets -> Is it possible to use item compression like in App-Which-Must-Not-Be-Named.?**
A. No. There was a discussion about this in the past but it was rejected, there is an unofficial way to achieve this.
https://gist.github.com/JamesCullum/6604e504318dd326a507108f59ca7dcd.

**Under performance -> I disabled machine learning but I still experience high CPU usage why?**
A.The process of creating the thumbnail images is based on the CPU, so the CPU usage can still be high until the creation of the thumbnails is finished.

**Under  ML -> Does Immich have animals recognition?**
A.No.

**Under Mobile application -> I can't log into the application after the server update, what can I do?**
A. update the app.

danieldietzler:
1) That's currently not a thing. You'd need to do it manually by querying the database and comparing it to the local files I guess. So imo not really feasible. Easiest for (not too large) libraries would probably be to just delete the folder and re-generate all files. Since this isn't really a good approach either I'd vote for leaving this point out for now.

2) I'm not sure if it's *that* helpful but it can't hurt either
3) Sounds good (I guess?)
4) Sounds good
5) Sounds good ("animal", not "animals")
6) Maybe also explain that app store updates sometimes take longer because the stores (play store; google and app store; apple) need to approve the update first which may take some time

aviv926:
> 1
Great, I'll start working on everything soon,
If you have any other suggestions I will add them as well.

mertalev:
> Under Assets -> How to delete transcoded assets if I don't need them anymore?

This would be a good addition, but only after we actually implement #5873

> Under Assets -> What happens when 2 users have access to load from the same library?

This is good, but reword to "What happens if an asset exists in more than one account?".

> Under Assets -> Is it possible to use item compression like in App-Which-Must-Not-Be-Named.?

Eh sure, but just say it isn't supported.

> Under performance -> I disabled machine learning but I still experience high CPU usage why?

I think this is redundant if we have a "Why is Immich using so much of my CPU?" question that explains which jobs are CPU-intensive.

> Under ML -> Does Immich have animals recognition?

This is good.

> Under Mobile application -> I can't log into the application after the server update, what can I do?

This is also good.

docs/docs/FAQ/External Library-FAQ.md
```diff
@@ -0,0 +1,20 @@
+---
+sidebar_position: 5
+---
+
+# External Library
+
+### Can I add an external library while keeping the existing album structure?
+
+We haven't put in an official mechanism to create album from your external library yet, but there are some [workarounds from the community](https://github.com/immich-app/immich/discussions/4279) which you can find here to help you achieve that.
```
danieldietzler, line 9: "albums" (plural)

docs/docs/FAQ/External Library-FAQ.md
```diff
@@ -0,0 +1,20 @@
...
+
+### I got duplicate files in the upload library and the external library
+
+It is important to remember that Immich does not filter duplicate files from [external Library](/docs/features/libraries) at the moment.
+
+Upload library and external library different mechanism to detect duplicates:
```
danieldietzler, line 15: I missed a "use" between "external library" and "different mechanism" in my first comment, sorry!
So "[...] external library use different mechanisms [...]"

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,61 @@
+---
+sidebar_position: 6
+---
+# Machine Learning
+
+### How does smart search work?
+
+Immich uses CLIP models, for more information about CLIP and its capabilities can be [read here](https://openai.com/research/clip).
```
danieldietzler, line 8: "[...] capabilities read about it here"

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,61 @@
...
+
+### How does facial recognition work?
```
danieldietzler, line 10: I believe the question you're actually answering is "Which models are being used?"

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,61 @@
...
+### Can I use a custom CLIP model?
+
+No, this is not supported. Only models listed in the [Huggingface](https://huggingface.co/immich-app) are compatible. Feel free to make a feature request if there's a model not listed here that you think should be added.
+
+### I want to be able to search in other languages besides English. How can I do that?
+
+You can change to a multilingual model listed [here](https://huggingface.co/collections/immich-app/multilingual-clip-654eb08c2382f591eeb8c2a7) by going to Administration > Machine Learning Settings > Smart Search and replacing the name of the model. Be sure to re-run Smart Search on all assets after this change. You can then search in over 100 languages.
+
+:::note
+feel free to make a feature request if there's a model you want to use that isn't in Immich Huggingface list.
```
danieldietzler, line 51: Again, this is a new sentence so the first word should be capitalized

danieldietzler (commented):
Nice, some more minor things. Also, it there are a few open issues still from my last review :)

docs/docs/FAQ/Machine-Learning-FAQ.md
mertalev: This was actually my suggestion. I was thinking we could flesh it out some more and have that be the question. Do you think it'd be better for them to be separate questions?

danieldietzler: I basically don't know anything about machine learning stuff and reading the respective text did not explain to me how facial recognition works. I don't think I have a preference on either making it two questions or extending the current text. All I'm saying is that the question right now isn't being answered by the text :)

docs/docs/FAQ/Assets-FAQ.md
```diff
@@ -0,0 +1,81 @@
...
+
+Yes, with an [external library](/docs/features/libraries.md).
+
+### What happens to existing files after I choose a new [Storage Template](/docs/administration/storage-template.mdx)?
+
+Template changes will only apply to new assets. To retroactively apply the template to previously uploaded assets, run the Storage Migration Job, available on the [Jobs](/docs/administration/jobs.md) page.
+
+### Why are only photos and not videos being uploaded to Immich?
+
+This often happens when using a reverse proxy or cloudflare tunnel in front of Immich. Make sure to set your reverse proxy to allow large POST requests. In `nginx`, set `client_max_body_size 50000M;` or similar. Also check the disk space of your reverse proxy, in some cases proxies cache requests to disk before passing them on, and if disk space runs out the request fails.
```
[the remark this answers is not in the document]
aviv926: > Technically cloudflare tunnel is a reverse proxy :P

The truth is that this is a question that I did not add, this is how it is written in the original.
I just rearranged it.
I wasn't quite able to understand what change I need to apply here?

docs/docs/FAQ/Machine-Learning-FAQ.md
aviv926: Actually I can try to expand on this, but if we're going to expand on it, maybe you should put it in one of these:

https://immich.app/docs/features/facial-recognition/
or here
https://immich.app/docs/administration/jobs
or
https://immich.app/docs/developer/architecture

In any case, I think that an extended explanation would definitely be something welcome, it would allow contributors to the code or people to understand how the models work in Immich. I understand that it may not be of interest to everyone, but people can be directed to read a research article if they are interested in the depth of things

Here is my suggestion:

Immich uses InsightFace to perform
face detection and recognition.
Every time a new property is uploaded to Immich, thumbnails are created for it.
InsightFace uses machine learning to identify faces in thumbnails, for each face a unique identifier is created, also called an embed, and it belongs to the Postgers database under the face_asset column where the embed values of all the faces in the image + the position of the face in the image are saved, this is saved as a vector value, Immich performs a comparison of :

The embed created for the face from thumbnails of the uploaded asset
Compared to the embeds that exist in a database

Before the additional association there are 2 values that Immich is set to check according to:

1. If a value greater than the MIN DETECTION SCORE defined by the user is received, then Immich will not consider it as a face to add

2. If a value equal to or higher than the MAX RECOGNITION DISTANCE defined by the user is received, then Immich will insert the embed found in the database into the embed of an existing person (if any) if there is no existing person, Immich will create a new person.

For each new embed, Immich performs a test of the 100 **closest** embeds to the new embed, so that there is no unnecessary load on the system for searching among all the embeds in the database.


@mertalev  What do you think of this explanation?
I hope the translation was able to keep my explanation clear enough.

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,125 @@
...
+For more info see [here](https://github.com/immich-app/immich/pull/5903)
+
+### What is the best order for ML to best work
+
+Because Immich is relies on thumbnail generation
+For machine learning [[1](http://localhost:3005/docs/developer/architecture#:~:text=For%20example%2C%20Smart%20Search%20and%20Facial%20Recognition%20relies%20on%20thumbnail%20generation)], it is recommended to wait for the thumbnail generation job to finish and then run the machine learning jobs (Smart Search and Facial Recognition)
+
+* Let all jobs run except the machine learning jobs and wait for them to finish
+* Run Smart Search
+* Run Facial Recognition
```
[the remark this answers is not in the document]
aviv926: I read that some people who ran everything together caused errors in face detection and CLIP so I thought maybe explain that the best order would be to let all the jobs finish and then run the LM jobs.
Of course there is another reason that LM depends on THUMBNAILS thumbnails.

what do you think?

docs/docs/FAQ/Performance-FAQ.md
```diff
@@ -0,0 +1,48 @@
...
+Yes, you can do it by increasing the amount of job concurrency, as the amount increases, the computer will work on several assets at the same time.
+You can do it by
+1. Admin user login
+2. Administration
+3. jobs
+4. On the left side Manage Concurrency
+5. Change the settings as needed
+
+:::danger
+On a normal machine, 2 or 3 concurrent jobs can probably max the CPU, so if you're not hitting those maximums with, say, 30 jobs,
```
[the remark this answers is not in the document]
aviv926: LOL, this was Zack's answer in [Discord](https://discord.com/channels/979116623879368755/994044917355663450/1174713816072863827), I edited and put a period after the 30.

docs/docs/FAQ/Machine-Learning-FAQ.md
aviv926: Yes, Mertalev also mentioned it in my previous commit, I'm new to everything related to Git and just now started to understand a little how to work with it with help from my brother.
It's Done in the new commitment :)

docs/docs/FAQ/Assets-FAQ.md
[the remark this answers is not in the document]
danieldietzler: I would write something like this I guess:
```suggestion
This often happens when using a reverse proxy (such as nginx or Cloudflare tunnel) in front of Immich. Make sure to set your reverse proxy to allow large POST requests. In `nginx`, set `client_max_body_size 50000M;` or similar. Also check the disk space of your reverse proxy, in some cases proxies cache requests to disk before passing them on, and if disk space runs out the request fails.
```

docs/docs/FAQ/Assets-FAQ.md
```diff
@@ -0,0 +1,81 @@
...
+
+See [backup and restore](/docs/administration/backup-and-restore.md).
+
+### Does Immich supports reading faces tags from exif?
+
+For now, it doesn't.
+
+### Does Immich support filtering of NSFW images?
+
+It seems that, this option is not implemented, but it seems that there is an [open discussion about it
```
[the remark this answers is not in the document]
aviv926: Yes, Mertalev also mentioned it in my previous commit, I'm new to everything related to Git and just now started to understand a little how to work with it with help from my brother.
It's Done in the new commitment :)

docs/docs/FAQ/Machine-Learning-FAQ.md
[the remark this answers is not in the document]
danieldietzler: Hm ok, if that's the case it's good as is. For me personally it's quite obvious but that's probably some bias I have there

docs/docs/FAQ/Assets-FAQ.md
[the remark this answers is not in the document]
danieldietzler: Ty :)

docs/docs/FAQ/Machine-Learning-FAQ.md
aviv926: Shouldn't it be with a dot between them?

"[...] capabilities. read about it here."

danieldietzler: Imo no, because it still references the "for". Maybe we can rephrase it though: "For more information [...], continue reading here"

aviv926: yes my mistake.
fixed.
Thanks!

aviv926:
Do you think this could be an essential addition?
**under Assets**

### How and where are assets stored in Immich?

**1. User-Specific Folders:**
- Each user has a unique string representing them.
    - The main user is "Admin" (but only for `\library\library\`)
    - Other users have different string identifiers.
- You can find your user ID in Account  Account Settings > Account > User ID.

**2. Asset Types and Storage Locations:**

- **Source Assets:**
    - Original assets uploaded through the browser interface&mobile&CLI.
    - Stored in `\library\library\<userID>`.
- **Avatar Images:**
    - User profile images.
    - Stored in `\library\profile\<userID>`.
- **Thumbs Images:**
    - Preview images (blurred, small, large) for each asset and thumbnails for recognized faces.
    - Stored in `\library\thumbs\<userID>`.
- **Encoded Assets:**
    -  By default, unless otherwise specified re-encoded video assets for wider compatibility .
    - Stored in `\library\encoded-video\<userID>`.
- **Files in Upload Queue (Mobile):**
    - Files uploaded through mobile apps.
    - Temporarily located in `\library\upload\<userID>`.
    - Transferred to `\library\library\<userID>` upon successful upload.

:::danger
Do not touch the files inside these folders under any circumstances except taking a backup, changing or removing an asset can cause untracked and missing files.
You can think of it as App-Which-Must-Not-Be-Named, the only access to viewing, changing and deleting assets is only through the mobile or browser interface.
:::

### How can I hide photos from the timeline?
You can archive them.


### What happens if I changed or canceled vdeo transcoding ?
If you chose to cancel the vdeo transcoding, the transcoded videos will be deleted, if you chose to change the encoding format, the newly encoded videos will overwrite the old ones.
---- ---- ---- 

**under Machine Learning**

### The immich_model-cache volume takes up a lot of space, what could be the problem?

If you installed several models and chose not to use some of them, it might be worth deleting the old models that are in immich_model-cache.
To be able to do this you can run:
`docker run -it --rm -v immich_model-cache:/mnt ubuntu bash`
`cd mnt`
`ls`
and delete unused models with `rm -r <model_name>`.

mertalev:
The first question is out of scope for a FAQ, but the other two are good.

aviv926:
> The first question is out of scope for a FAQ, but the others are good. The transcoding answer is wrong though because it won't delete on cancel.

Doesn't your PR https://github.com/immich-app/immich/pull/6027 deal with the removal of irrelevant encoded videos/after deactivation?  The question I wrote came from a conversation in [Discord](https://discord.com/channels/979116623879368755/1174440583188709477) and after I saw the PR you created and merged into the project.

mertalev:
1. Transcodes are only deleted if they already existed beforehand for a job and the job found transcoding to be unnecessary. 

Canceling transcoding jobs won't have this effect. Canceling active jobs is not possible, so they will still process as normal. Canceling queued jobs is the same as never queueing them to begin with.

2. That seems like a good place for it.

aviv926:
> 1. Transcodes are only deleted if they already existed beforehand for a job and the job found transcoding to be unnecessary. Canceling transcoding jobs won't have this effect. Canceling active jobs is not possible, so they will still process as normal. Canceling queued jobs is the same as never queueing them to begin with.
> 2. That seems like a good place for it.

That means a user can't decide to delete the conversions if he wants to right now?
And if the user decides to delete it manually, it will create "not tracked" on the repair page

If so, the answer to this question will be that as long as the asset is not deleted through the browser interface, the transcode created will continue to exist. Right?

mertalev:
> > 1. Transcodes are only deleted if they already existed beforehand for a job and the job found transcoding to be unnecessary. Canceling transcoding jobs won't have this effect. Canceling active jobs is not possible, so they will still process as normal. Canceling queued jobs is the same as never queueing them to begin with.
> 
> > 2. That seems like a good place for it.
> 
> 
> 
> That means a user can't decide to delete the conversions if he wants to right now?
> 
> And if the user decides to delete it manually, it will create "not tracked" on the repair page
> 
> 
> 
> If so, the answer to this question will be that as long as the asset is not deleted through the browser interface, the transcode created will continue to exist. Right?

No, that's not right. A user can delete a given existing transcode if they set a transcode policy that makes it unnecessary and then run a transcoding job for it. This can be done on a per-asset basis by starting a transcoding job for an asset, or for all assets by running transcoding jobs for all assets.

aviv926:
> > > 1. Transcodes are only deleted if they already existed beforehand for a job and the job found transcoding to be unnecessary. Canceling transcoding jobs won't have this effect. Canceling active jobs is not possible, so they will still process as normal. Canceling queued jobs is the same as never queueing them to begin with.
> > 
> > 
> > > 2. That seems like a good place for it.
> > 
> > 
> > That means a user can't decide to delete the conversions if he wants to right now?
> > And if the user decides to delete it manually, it will create "not tracked" on the repair page
> > If so, the answer to this question will be that as long as the asset is not deleted through the browser interface, the transcode created will continue to exist. Right?
> 
> No, that's not right. A user can delete a given existing transcode if they set a transcode policy that makes it unnecessary and then run a transcoding job for it. This can be done on a per-asset basis by starting a transcoding job for an asset, or for all assets by running transcoding jobs for all assets.

Good to know, I'll add that.

In your opinion it would be necessary to specify at [libraries](https://immich.app/docs/features/libraries)
that if you want to add a folder that contains spaces in its name, it must have a `-` instead of a space? Or is this something that Immich can be made to do automatically and will be fixed in future versions? [According to Jrasm](https://discord.com/channels/979116623879368755/1184258493948117084/1191879841906167909) spaces are problematic in general when it comes to programming, so I'm wondering if it should be noted in the documentation.

mertalev:
It'd be good to document the behavior of spaces, but I'm not sure how it works currently. Spaces in the path should probably work since the full path is a string and not tokenized.

aviv926:
> It'd be good to document the behavior of spaces, but I'm not sure how it works currently. Spaces in the path should probably work since the full path is a string and not tokenized.

I tried yesterday but without success to import a path with a space in the folder.
Maybe @jrasm91 can tell which answer would be best to put in this part of the documentation

After that this PR is ready :)

mertalev:
It looks like we use the external path for a regex. Maybe that's not behaving right with spaces? 

Either way, if it isn't working now, then a note that spaces in the path aren't supported would be fine.

aviv926:
> It looks like we use the external path for a regex. Maybe that's not behaving right with spaces?
> 
> Either way, if it isn't working now, then a note that spaces in the path aren't supported would be fine.

done :)

docs/docs/FAQ/Assets-FAQ.md
```diff
@@ -0,0 +1,92 @@
...
+### What happens if an asset exists in more than one account?
+
+All machine learning jobs and thumbnail images are recreated.
+
+### What happens if I changed or disabled video transcoding ?
+
+If user choose to transcode policy that makes it unnecessary and then run a transcoding job, it will delete unnecessary transcodes.
+This can be done on a per-asset basis by starting a transcoding job for an asset, or for all assets by running transcoding jobs for all assets.
+
+If user choose to changed transcode policy and rerun transcoding job for an asset, the newly encoded videos will overwrite the old ones.
```
mertalev, line 53:
```suggestion
### How can I delete transcoded videos without deleting the original?

The transcode of an asset can be deleted by setting a transcode policy that makes it unnecessary, then running a transcoding job for that asset. This can be done on a per-asset basis by starting a transcoding job for an asset (with the _Refresh encoded videos_ button in the asset viewer options, or for all assets by running transcoding jobs for all assets.

To update the transcode policy, navigate to Administration > Video Transcoding Settings > Transcoding Policy and select a policy from the drop-down. This policy will determine whether an existing transcode will be deleted or overwritten in the transcoding job. If a video should be transcoded according to this policy, an existing transcode is overwritten. If not, then it is deleted.

:::note
For example, say you have existing transcodes with the policy "Videos higher than normal resolution or not in the desired format" and switch to a narrower policy: "Videos not in the desired format". If an asset was only transcoded due to its resolution, then running a transcoding job for it will now delete the existing transcode. This is because resolution is no longer part of the transcode policy and the transcode is unnecessary as a result. Likewise, if you set the policy to "Don't transcode any videos" and run transcoding jobs for all assets, this will delete all existing transcodes as they are all unnecessary.
:::
```

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,73 @@
...
+For more info see [here](https://github.com/immich-app/immich/pull/5903)
+
+### What is the best order for ML to best work
+
+Because Immich is relies on thumbnail generation
+For machine learning [[1](/docs/developer/architecture#:~:text=For%20example%2C%20Smart%20Search%20and%20Facial%20Recognition%20relies%20on%20thumbnail%20generation)], it is recommended to wait for the thumbnail generation job to finish and then run the machine learning jobs (Smart Search and Facial Recognition)
+
+* Let all jobs run except the machine learning jobs and wait for them to finish
+* Run Smart Search
+* Run Facial Recognition
```
mertalev, line 40: I think you can remove this.

docs/docs/FAQ/Machine-Learning-FAQ.md
```diff
@@ -0,0 +1,73 @@
...
+
+:::note
+Feel free to make a feature request if there's a model you want to use that isn't in [Immich Huggingface list](https://huggingface.co/immich-app).
+:::
+
+### Does Immich support Facial Recognition for videos ?
+
+This is not currently implemented, but may be in the future.
+
+On the other hand, Immich uses a thumbnail image created from the video in order to scan it for a face, if there is a face in the thumbnail image of the video, Immich will try to recognize a face in thumbnail.
```
mertalev, line 58:
```suggestion
On the other hand, Immich does scan video thumbnails for faces, so it can perform recognition if the face is clear in the video thumbnail.

File: docs/docs/FAQ/Machine-Learning-FAQ.md [hunk 27]
```diff
@@ -0,0 +1,73 @@
...
+
+### Does Immich have animal recognition?
+
+No.
+
+### The immich_model-cache volume takes up a lot of space, what could be the problem?
+
+If you installed several models and chose not to use some of them, it might be worth deleting the old models that are in immich_model-cache.
+
+To be able to do this you can run:
```
mertalev, line 68:
```suggestion
To do this you can run:
```

docs/docs/FAQ/Mobile-App-FAQ.md
```diff
@@ -0,0 +1,32 @@
...
+
+### IOS app
+
+### No albums&photos seen by Immich on Iphone
+
+there are some reporting issue with iPhoneXR regarding the problem with scanning the album,we haven't yet to find the cause [see more info here](https://discord.com/channels/979116623879368755/1107718197655908412)
+
+:::note
+If you think you know what the solution to the problem might be, we'd be happy to discuss it on [Discord](https://discord.com/channels/979116623879368755/1071165397228855327)/[Github](https://github.com/immich-app/immich/discussions)
+:::
```
mertalev, line 24: I'm not sure if this is actually helpful. We don't have a solution for this issue, and it can point people in the wrong direction if the reason they don't see images is unrelated.

docs/docs/FAQ/Performance-FAQ.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Can I boost machine learning speed?
+
+Yes, you can by increasing the job concurrency. With higher concurrency, the host will work on more assets in parallel.
+You can do it by:
+1. Admin user login
+2. Administration
+3. jobs
+4. On the left side Manage Concurrency
+5. Change the settings as needed
```
mertalev, line 34:
```suggestion
### How can I boost machine learning speed?
:::note
This advice increases throughput, not latency. This is to say that it will make Smart Search jobs process more quickly, but it won't make searching faster.
:::

You can increase throughput by increasing the job concurrency for machine learning jobs (Smart Search, Recognize Faces). With higher concurrency, the host will work on more assets in parallel. You can do this by navigating to Administration > Settings > Job Settings and increasing concurrency as needed.
```

docs/docs/FAQ/Privacy-FAQ.md
```diff
@@ -0,0 +1,14 @@
...
+---
+import TOCInline from '@theme/TOCInline';
+
+# Privacy
+
+### Is all data local?
+Everything stays local, the models get downloaded from a repository the first time you run it.
+Machine learning is also done locally,
+Machine learning URL isn't pointing to the external page. It is the DNS of the container within the docker's network context.
+Model retrievals are done automatically from the [hugging face repository](https://huggingface.co/immich-app). more info [here](https://www.reddit.com/r/immich/comments/17obeyj/is_all_data_local/).
```
mertalev, line 12: Not that this is wrong, but we need to be more precise with the wording. I think the answer to this question should be in the form of a privacy policy and a TL;DR linking to it. Probably better to add this separately.

docs/docs/developer/setup.md
```diff
@@ -1,9 +1,18 @@
 ---
 sidebar_position: 2
 ---
-
 # Setup
 
+:::note
+if there is something you're planning to work on just give us a heads up in [Discord](https://discord.com/channels/979116623879368755/1071165397228855327) of what it is so we can
```
mertalev, line 7:
```suggestion
If there's a feature you're planning to work on, just give us a heads up in [Discord](https://discord.com/channels/979116623879368755/1071165397228855327) so we can:
```

docs/docs/features/libraries.md
```diff
@@ -145,6 +148,14 @@ Only an admin can do this.
 - Navigate to `Administration > Users` page on the web.
 - Click on the user edit button.
 - Set `/mnt/media` to be the external path. This folder will only contain the three folders that we want to import, so nothing else can be accessed.
+  :::note
+  spaces in the path aren't supported.
```
mertalev, line 152:
```suggestion
Spaces in the internal path aren't currently supported.
```

docs/docs/guides/database-GUI-view.md
```diff
@@ -0,0 +1,42 @@
+# Access database (GUI)
+
+:::note
+in order to connect to the database the immich_postgres container **must be on**.
```
mertalev, line 4:
```suggestion
In order to connect to the database, the immich_postgres container **must be running**.
```

mertalev (commented):
It's getting closer!

docs/docs/FAQ/Assets-FAQ.md
```diff
@@ -0,0 +1,81 @@
...
+No. There was a discussion about this in the past but it was [rejected](https://github.com/immich-app/immich/pull/1242), there is an [unofficial way to achieve this](https://gist.github.com/JamesCullum/6604e504318dd326a507108f59ca7dcd).
+
+:::danger
+If you do choose to use unofficial way, it's important to be aware of the following security risks:
+
+- The golang:1.19.4-alpine3.17 base image was released a year ago and has [18 vulnerabilities for OpenSSL](https://hub.docker.com/layers/library/golang/1.19.4-alpine3.17/images/sha256-8b532e4f43b6ccab31b2542d132720aa6e22f6164e0ed9d4885ef2d7c8b87aa5?context=explore)
+- The image `ghcr.io/jamescullum/multipart-upload-proxy:main` was last released long time ago and has vulnerable versions of OpenSSL and libwebp.
+- The vips-dev package relies on libwebp, which had a major CVE a few months ago.
+there may be other vulnerabilities as well.
+:::
```
mertalev:
```suggestion
### Is it possible to only store compressed images instead of the original?

No. Our golden rule is that the original assets should always be untouched, so we don't think this feature is a good fit for Immich.
```

docs/docs/FAQ/Assets-FAQ.md
```diff
@@ -0,0 +1,92 @@
...
+No. There was a discussion about this in the past but it was [rejected](https://github.com/immich-app/immich/pull/1242), there is an [unofficial way to achieve this](https://gist.github.com/JamesCullum/6604e504318dd326a507108f59ca7dcd).
+
+:::danger
+If you do choose to use the unofficial way, it's important to be aware of the following security risks:
+
+- The golang:1.19.4-alpine3.17 base image was released a year ago and has [18 vulnerabilities for OpenSSL](https://hub.docker.com/layers/library/golang/1.19.4-alpine3.17/images/sha256-8b532e4f43b6ccab31b2542d132720aa6e22f6164e0ed9d4885ef2d7c8b87aa5?context=explore)
+- The image `ghcr.io/jamescullum/multipart-upload-proxy:main` was last updated a long time ago and has vulnerable versions of OpenSSL and libwebp.
+- The vips-dev package relies on libwebp, which had a major CVE a few months ago.
+There may be other vulnerabilities as well.
+:::
```
mertalev, line 67:
```suggestion
No. Our golden rule is that the original assets should always be untouched, so we don't think this feature is a good fit for Immich.
```

jrasm91:
It might be faster to just encourage developers to push to the branch directly instead of going back and forth on all the changes. It seems most will be accepted as is.

docs/docs/FAQ/Mobile-App-FAQ.md
aviv926: I understand, I agree with that, in my opinion it might be worthwhile to create a place where the problems referred to the development team would be, this could prevent a lot of duplication in "issues".

Maybe I'll see if there is a bot that can look for duplicates in issues and creates a summary of the things that repeat themselves a lot or something like that (it would be interesting to know if there is something like that).

docs/docs/FAQ/Privacy-FAQ.md
aviv926: I think it's worth mentioning somewhere, I've seen such a question repeated several times on Reddit.

Because currently Immich does not have a section that talks about a privacy policy and according to Alex's words on Reddit, I don't know if when there is a privacy policy what information will be included in it if it is so short?
I mean it should be something like.. we don't collect anything

aviv926:
If it's possible, I think it's worth adding a question mark button above where the sun and moon button is, with a link to the FAQ

image

to this

docs/docs/FAQ/Privacy-FAQ.md
mertalev: I know what you mean, but I just want to be sure we're not exposing ourselves legally here. I have some reservations about a blanket statement like this.

Are we 100% sure there's nothing being sent, even indirectly? For instance, what if we have a dependency that defaults to sending telemetry? Even if we don't right now, what if this happens in the future?

What does local mean? Different services can run in different locations, and we share data across these services. More broadly, given how many environments Immich can be deployed in, what do we include and exclude when we're talking about data being shared? Does inserting into a cloud provider's database count as sharing with a third party?

Do requests to e.g. GitHub or Hugging Face need to be mentioned? We're technically sharing IP address with them.

We also have OAuth integration, so we'd need to mention that too.

mertalev:
Eh, I don't like having that in the main UI. But a link to the FAQ in the new admin onboarding pop-up would be nice.

mertalev:
That can be a separate PR, though. This one should just be about updating the FAQ itself.

aviv926:
> Eh, I don't like having that in the main UI. But a link to the FAQ in the new admin onboarding pop-up would be nice.

Yes, the admin panel feels like a better place for that.

aviv926:
> That can be a separate PR, though. This one should just be about updating the FAQ itself.

agree.

In your opinion under Comparison.md it would be necessary to add

::: Tip
If you have chosen to move definitively from Photoprism to Immich, you may want to check how to perform [migrating favorites and albums from Photoprism](https://github.com/immich-app/immich/discussions/4701) to Immich
:::

?

docs/docs/FAQ/Privacy-FAQ.md
aviv926: Yes I understand that completely, in fact it is a big issue that requires understanding on the subject, I guess Immich could draw some essential knowledge from the [Photoprism](https://www.photoprism.app/privacy) project regarding the matter of privacy policy and implement an adapted privacy policy accordingly.

I will remove this section.

jrasm91:
I think it would be better if all the FAQs where on a single page. We can still group them into sections (on a single page) and make them collapsible, which I think makes the content more accessible rather than having to page through 10 different pages. Also, existing FAQ links would all be broken.

jrasm91:
If you move everything back into a single page, I'm happy to help look into some react component or format that would make it more usable.

aviv926:
> If you move everything back into a single page, I'm happy to help look into some react component or format that would make it more usable.


I was thinking about putting it all on the same page as it was before I started working on it, like I mentioned in the first message of this PR.
My idea was to use the TOC that support by [Docusaurus]( https://docusaurus.io/docs/markdown-features/toc#inline-table-of-contents
) The thing is, I won't able to use `<TOCInline toc={toc} />` as `<TOCInline toc={example.md} />`
If it worked, I would try to import all TOCS from other md pages...but it dosn't /:

So my final idea was to do something that was not very smart... but it works. 
The idea is to copy all TOCS from the right side and also the link (with JS code) and manually put it on the first page.
Unless you think of a better way.

jrasm91:
I think it would be better if they were all in the same page.

danieldietzler:
If we may want to add a TOC for other things I could look into this but I agree with Jason in that the FAQ should be just one page - with maybe an "accordion" component or something, so that single sections can collapse.

aviv926:
> If we may want to add a TOC for other things I could look into this but I agree with Jason in that the FAQ should be just one page - with maybe an "accordion" component or something, so that single sections can collapse.

I managed to create a code that will do the job so that the previous comment is not relevant, now it is generated automatically.

For example:
image



But if it's not part of what works for Immich I'll change it all to one page.

jrasm91:
I don't think we only want a table of contents. We want the actual content all together in one place, on the same page.

aviv926:
> I don't think we only want a table of contents. We want the actual content all together in one place, on the same page.

Got it, I'll upload a one page version for everything soon.

aviv926:
@jrasm91 I moved everything to the configuration of AIO
I think it's a bit messy because after the new additions the page contains a lot of information.
image

jrasm91 (approved):
Thanks for working on this! The additional information is great and we can iterate on this page over time to continue to fine tune and improve it.

I think this actually looks really good on a single page. You can easily search for everything and content is very discoverable (plus, all the old links still work :tada:).

aviv926:
> Thanks for working on this! The additional information is great and we can iterate on this page over time to continue to fine tune and improve it.
> 
> I think this actually looks really good on a single page. You can easily search for everything and content is very discoverable (plus, all the old links still work 🎉).

Thanks for the merge :)

Regarding the GUI guide of the database, I saw that you removed the EXTRA part, it is mainly for Ubuntu users, there is no GUI as software, it is a GUI through the browser, so I added information for those who want to find out where the link to connect to the database via PgAdmin is

jrasm91:
Are you talking about those running a headless server?

aviv926:
> Are you talking about those running a headless server?

I mean those running Immich on a Linux system like Ubuntu, where there is no GUI for the pgAdmin software and you have to use a browser for a user interface. This is explained in the manual in the [EXTRA section](https://github.com/immich-app/immich/pull/5958/commits/83315804be42f9d7d6d0a93b75e56e22dd33989e#diff-10364a1cf773aaf0e15103cf5b117648023d3a93baae669390d3a7039e94b477R31).

In addition, I added up-to-date information regarding backup in a Windows-based system and information about the new storage format on new and old machines.
Is this required in your opinion?

jrasm91:
I'm not sure I understand. I use pgAdmin and I'm on ubuntu and it works fine.

aviv926:
> I'm not sure I understand. I use pgAdmin and I'm on ubuntu and it works fine.

Maybe it was just me...
When I tried to develop a new feature in Immich I had to check the database in Ubuntu, it took me too much time than expected because in Windows it was easy, just run the software and connect to the database.
In Ubuntu the interface of the software was through the browser and it took me a while to figure it out
In any case, it's probably not that important, so we'll leave the guide as it is (just need to fix the links to the pgAdmin website, I've created a commit https://github.com/aviv926/immich/commit/3e01667f5743389c4652bc85953d4b256f81599c)

PR #5966, opened 2023-12-25
feat(server): search across own+partner assets

fyfrey:
search now works across own and partner assets (only for partners visible in the combined timeline). Thus, user experience is improved (seeing partner assets -> searching partner assets; not seeing partner assets -> searching only own assets).


server/src/infra/repositories/asset.repository.ts
```diff
@@ -805,9 +805,13 @@ export class AssetRepository implements IAssetRepository {
   }
 
   @GenerateSql({ params: [DummyValue.STRING, DummyValue.UUID, { numResults: 250 }] })
```
danieldietzler, line 807: You'd need to adapt this since there is now an array of user ids => `params: [DummyValue.STRING, [DummyValue.UUID], { numResults: 250 }]`

server/src/infra/repositories/smart-info.repository.ts
```diff
@@ -44,7 +44,7 @@ export class SmartInfoRepository implements ISmartInfoRepository {
   @GenerateSql({
     params: [{ ownerId: DummyValue.UUID, embedding: Array.from({ length: 512 }, Math.random), numResults: 100 }],
```
danieldietzler, line 45: Same thing here. It's now an array of userIds instead of a single ownerId

server/src/infra/repositories/asset.repository.ts
fyfrey: Thanks, missed that!

server/src/infra/repositories/smart-info.repository.ts
fyfrey: Thanks, missed that!

danieldietzler (approved):
LGTM! :)

jrasm91 (approved):

PR #5974, opened 2023-12-25
feat(server): Enqueue jobs in bulk

adamantike:
The Job Repository now has a `queueAll` method, that enqueues messages in bulk (using BullMQ's
[`addBulk`](https://docs.bullmq.io/guide/queues/adding-bulks)), improving performance when many jobs must be enqueued within the same operation.

Primary change is in `server/src/infra/repositories/job.repository.ts`, and other services have been refactored to use `queueAll` when useful.

As a simple local benchmark, triggering a full thumbnail generation process over a library of ~1,200 assets and ~350 faces went from **~600ms** to **~250ms**.

### DIFF
--- a/server/src/domain/media/media.service.ts
+++ b/server/src/domain/media/media.service.ts
@@ -86,7 +86,7 @@ export class MediaService {
 
     const people = force ? await this.personRepository.getAll() : await this.personRepository.getAllWithoutThumbnail();
 
-    const idsToGenerateThumbnail = new Set<string>();
+    const jobs: JobItem[] = [];
     for (const person of people) {
       if (!person.faceAssetId) {
         const face = await this.personRepository.getRandomFace(person.id);
@@ -97,12 +97,10 @@ export class MediaService {
         await this.personRepository.update({ id: person.id, faceAssetId: face.assetId });
       }
 
-      idsToGenerateThumbnail.add(person.id);
+      jobs.push({ name: JobName.GENERATE_PERSON_THUMBNAIL, data: { id: person.id } });
     }
 
-    await this.jobRepository.queueAll(
-      Array.from(idsToGenerateThumbnail).map((id) => ({ name: JobName.GENERATE_PERSON_THUMBNAIL, data: { id } })),
-    );
+    await this.jobRepository.queueAll(jobs);
 
     return true;
   }
--- a/server/src/domain/media/media.service.ts
+++ b/server/src/domain/media/media.service.ts
@@ -20,6 +20,7 @@ import {
   IPersonRepository,
   IStorageRepository,
   ISystemConfigRepository,
+  JobItem,
   VideoCodecHWConfig,
   VideoStreamInfo,
   WithoutProperty,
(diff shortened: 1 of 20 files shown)


jrasm91 (approved):
LGTM

server/src/domain/media/media.service.ts
```diff
@@ -65,22 +66,27 @@ export class MediaService {
...
+          jobs.push({ name: JobName.GENERATE_THUMBHASH_THUMBNAIL, data: { id: asset.id } });
         }
       }
+
+      await this.jobRepository.queueAll(jobs);
     }
 
     const people = force ? await this.personRepository.getAll() : await this.personRepository.getAllWithoutThumbnail();
 
+    const idsToGenerateThumbnail = new Set<string>();
```
danieldietzler, line 89: Why do you use a set of strings here and not just an array of `{name, data}` objects like in the other places?

server/src/domain/person/person.service.ts
```diff
@@ -144,6 +144,8 @@ export class PersonService {
     this.logger.debug(
       `Changing feature photos for ${changeFeaturePhoto.length} ${changeFeaturePhoto.length > 1 ? 'people' : 'person'}`,
     );
+
+    const idsToGenerateThumbnail = new Set<string>();
```
danieldietzler, line 148: Here the same question

server/src/domain/user/user.service.ts
```diff
@@ -129,12 +129,9 @@ export class UserService {
...
   async handleUserDeleteCheck() {
     const users = await this.userRepository.getDeletedUsers();
-    for (const user of users) {
-      if (this.isReadyForDeletion(user)) {
-        await this.jobRepository.queue({ name: JobName.USER_DELETION, data: { id: user.id } });
-      }
-    }
-
+    await this.jobRepository.queueAll(
+      users.filter(this.isReadyForDeletion).map((user) => ({ name: JobName.USER_DELETION, data: { id: user.id } })),
```
danieldietzler, line 133: I'd optimize it to:
```suggestion
      users.flatMap((user) => this.isReadyForDeletion(user) ? [{ name: JobName.USER_DELETION, data: { id: user.id } }] : [])),
```
this removes one iteration through the array

server/src/infra/repositories/job.repository.ts
```diff
@@ -116,12 +116,31 @@ export class JobRepository implements IJobRepository {
...
+      acc[queueName] = acc[queueName] || [];
+      acc[queueName].push(item);
+      return acc;
+    }, {});
+
+    for (const [queueName, items] of Object.entries(itemsByQueue)) {
+      const queue = this.getQueue(queueName as QueueName);
+      const jobs = items.map((item) => ({
+        name: item.name,
+        data: (item as { data?: any })?.data || {},
```
danieldietzler, line 135: What's the reason you don't just do `data: item.data ?? {}`?

danieldietzler (changes requested):
That's some great performance improvements!
And actually, imo even the code looks cleaner, nice

server/src/infra/repositories/job.repository.ts
jrasm91: I think this was just existing code and I wrote || over ?? Since I usually prefer it.

danieldietzler: Even then, why would you do `(item as { data: any }).data`?

jrasm91: Not all items have a data property.

danieldietzler: Every job item has (at least an optional) data property, yes?

jrasm91: I think that may be right, but this is existing code, which doesn't need to be modified as part of this pull request.

danieldietzler (approved):
LGTM! 🚀

server/src/domain/media/media.service.ts
adamantike: Good call! I used this way in case we could be receiving duplicate ids, but I can see that's not possible in any of these two places, so building the array of jobs directly will be simpler.

server/src/domain/user/user.service.ts
adamantike: Thanks!

PR #5983, opened 2023-12-26
feat(web): add keyboard shortcut to stack selected photos

Funk66:
This is my attempt at implementing https://github.com/immich-app/immich/discussions/5143
I should say that I have zero experience with Svelte. I would be nice to reduce the duplication with https://github.com/immich-app/immich/blob/3edc87f6840a96b04e318f1f1d823c81e642bc20/web/src/lib/components/photos-page/actions/stack-action.svelte#L15 but I'm not familiar with the import hierarchy and event propagation conventions, so I would appreciate some guidance here.
As a side note, it may seem confusing to some that the help dialog shows action shortcuts that only work in the asset-viewer page, and now this new one that only works in the asset-grid page. Perhaps these should be shown under different categories.


ThomasConrad:
When can this get merged?

martabal:
The logic to stack assets already exist in https://github.com/immich-app/immich/blob/main/web/src/lib/components/photos-page/actions/stack-action.svelte.

Can you use the same logic for the menu option and the shortcut action ?

gga-kialo:
I tried deduplicating the logic, but I don't know if my changes follow best practices.

Funk66:
I tried deduplicating the code, but I don't know if my changes follow best practices.

flightmansam:
This would be so handy! How close is a merge?

isaacolsen94:
Any chance this can get merged? It would be crazy helpful!!

jrasm91:
@Funk66 can you resolve the merge conflicts and then we can look at merging it.

isaacolsen94:
Thank you so much @funk66, @jrasm91 is this good to go now? Sorry to pester! Just could really use this 😛

jrasm91 (approved):
LGTM

PR #5992, opened 2023-12-26
fix(documentation): Update to latest default config

jansohn:
The default configuration seems to be outdated. I compared it to the demo app as well as a newly spun-up local instance.

### DIFF
--- a/docs/docs/install/config-file.md
+++ b/docs/docs/install/config-file.md
@@ -38,6 +38,9 @@ The default configuration looks like this:
     "metadataExtraction": {
       "concurrency": 5
     },
+    "objectTagging": {
+      "concurrency": 2
+    },
     "recognizeFaces": {
       "concurrency": 2
     },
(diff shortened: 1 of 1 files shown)


danieldietzler (approved):
It will be outdated with 1.92 again lol
It's fine with me to merge this though

mertalev:
The next release will have object tagging removed, so if we merge this PR we'll have to revert it very soon. (Possibly *very* soon)

jansohn:
> The next release will have object tagging removed, so if we merge this PR we'll have to revert it very soon. (Possibly _very_ soon)

Every release that changes the settings configuration file should update the documentation, too in my opinion. Maybe we can somehow reduce the effort by linking to the file content instead...

mertalev:
It was updated (which is why those fields aren't there anymore), but the issue is that the docs on the website are based on main, not the latest release.

jansohn:
Makes sense, so I'll close this again.

PR #6016, opened 2023-12-27
feat(mobile): multiselect for search & person page

fyfrey:
this PR enables multiselect on search result, person asset result and videos page.

It allows users to add assets from search to albums etc.


mobile/lib/modules/search/views/search_result_page.dart
```diff
@@ -163,15 +154,15 @@ class SearchResultPage extends HookConsumerWidget {
...
-          );
-        }
+        return MultiselectGrid(
+          renderListProvider: searchRenderListProvider,
+          archiveEnabled: true,
+          deleteEnabled: true,
+          editEnabled: true,
+          favoriteEnabled: true,
+          stackEnabled: false,
+          onRefresh: refresh,
```
shenlong-tanwen, line 164: This is great. Using the refresh in `MultiSelectGrid` to actually refresh the search

mobile/lib/modules/asset_viewer/providers/render_list.provider.dart
```diff
@@ -16,6 +16,15 @@ final renderListProvider =
   );
 });
 
+final renderListProviderWithGrouping =
```
shenlong-tanwen, line 19: Nit. Maybe we can migrate the `renderListProviderWithGrouping` and the existing `renderListProvider` to the riverpod_generator and named argument for `GroupAssetsBy`? We can leave them as it is and can always migrate them later as well

shenlong-tanwen (approved):
LGTM

mobile/lib/modules/asset_viewer/providers/render_list.provider.dart
fyfrey: let's migrate later :-)

PR #6031, opened 2023-12-28
fix(server): Fix incorrect sorting with search

### LINKED ISSUE #5916: [BUG] Mobile: Wrong date sorting on metadata search result
### The bug

1.Click on the search button 
2.search any metadata using m:you-search-keyword
3.Check the date sorting (attached screenshoot in the additional information)

Results: 
The order is random 

Expected: 
The order should be the same as the other screens on the app. (From the newest to the oldest)

P.S: You can see this behaviour also by opening any place to see the assets. 

N/A

### Version of Immich Server

v1.91.4

v1.91.V

### Platform with the issue

- [ ] Server
- [ ] Web
- [X] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1.Click on the search button 
2.search any metadata using m:you-search-keyword
3.Check the date sorting (attached screenshoot in the additional information)

Results: 
The order is random 

Expected: 
The order should be the same as the other screens on the app. (From the newest to the oldest)

P.S: You can see this behaviour also by opening any place to see the assets.
```

### Additional information

WhatsApp Image 2023-12-21 at 23 13 08

l0nax:
Fixes #5916

### DIFF
--- a/server/src/infra/repositories/asset.repository.ts
+++ b/server/src/infra/repositories/asset.repository.ts
@@ -822,6 +822,7 @@ export class AssetRepository implements IAssetRepository {
         @@ PLAINTO_TSQUERY('english', :query)`,
         { query },
       )
+      .addOrderBy('asset.fileCreatedAt', 'DESC')
       .limit(numResults)
       .getRawMany();
 
--- a/server/src/infra/sql/asset.repository.sql
+++ b/server/src/infra/sql/asset.repository.sql
@@ -693,5 +693,7 @@ WHERE
     ) @@ PLAINTO_TSQUERY('english', $3)
   )
   AND ("asset"."deletedAt" IS NULL)
+ORDER BY
+  "asset"."fileCreatedAt" DESC
 LIMIT
   250


server/src/infra/repositories/smart-info.repository.ts
```diff
@@ -61,6 +61,7 @@ export class SmartInfoRepository implements ISmartInfoRepository {
         .andWhere('a.fileCreatedAt < NOW()')
         .leftJoinAndSelect('a.exifInfo', 'e')
         .orderBy('s.embedding <=> :embedding')
+        .orderBy('a.fileCreatedAt DESC')
```
mertalev, line 64: I don't think this is a good idea. For one thing, this code doesn't work as expected because it replaces the previous order by instead of using `addOrderBy`. But it wouldn't work even if changed. This is because vector search does comparisons on float values that are very unlikely to be equal.

It's possible to make this query a CTE and order it by date in the second pass, but this is also not great. Instead of sorting by relevance, it can now show less relevant images first. Additionally, sorting by date would be a real headache once we add pagination for this query.

l0nax: You're right.
I removed it.

mertalev (approved):
LGTM

PR #6037, opened 2023-12-28
Age prediction

aviv926:
In trying to implement [5133 ](https://github.com/immich-app/immich/discussions/5133) I worked on a better attempt to access the implementation of the new feature. Contrary to my initial idea in 5133 to use a machine learning model for age detection, I decided to use what already exists in Immich.

The function I created goes through all the existing assets of `personId` and checks the vector with the lowest distance for it in the database, when a match of `maxDistance` less than 0.3 is found, then it matches the `dateTimeOriginal` of the image it relied on to find the vector with the lowest `maxDistance` , at the end he adds to the image without the `dateTimeOriginal` the `dateTimeOriginal` of the image he relied on.
When several faces are identified in the image, it looks for the `personId` with the lowest `maxDistance` level and according to that it matches the `dateTimeOriginal` to the image.

Additionally, I added an option for images that do not have a `dateTimeOriginal` to be associated by searching for a similar CLIP vector from the images in the library.

**Known problems:**
The CLIP function doesn't work the best because it won't know how to associate the new photos with the ones that necessarily match the situation, which will mean that if the user uploads birthday pictures **without a face** they will be associated with a birthday party with a similar CLIP vector, but it's not The real `dateTimeOriginal` that the photo was taken on. (Any suggestion would be welcome).

**Help me:**
At the moment I choose to put it as a draft because after checking with the Discord development team I saw that cosine_ops is used and it may be prevented for better detection because if I set `maxDistance` to less than 0.3 the function fails in prediction.

I would like to know if there is something deeper than `cosine_ops` that I can use to get deeper values which will result in even better prediction results

**Example (The original date of the image is the **name of the image**):**

[Screencast from 28-12-23 19:31:13.webm](https://github.com/immich-app/immich/assets/51673860/125b5a31-b932-47f9-814c-ac8682a25064)


mertalev:
The CLIP approach isn't going to work. The issue isn't that it uses cosine similarity, but the fact that the model isn't trained for facial recognition, let alone age.

The facial recognition approach is more viable, but there are caveats. To start, I would want to know how well clustering by a person's age actually works without false positives. I think the max distance needs to be set low for this, low enough that only the same person around the same age could meet the threshold. Maybe 0.1 or 0.15.

If the core idea works, it would need to differentiate between actual EXIF metadata and the prediction by writing to a different column. It should also be opt-in.

aviv926:
> The CLIP approach isn't going to work. The issue isn't that it uses cosine similarity, but the fact that the model isn't trained for facial recognition, let alone age.

I may not have written it, but I will add, the CLIP model is designed for images that do not have faces, so only when there is no face in the image CLIP will try to search for an image with a similar vector and adjust the image accordingly in the timeline.
That is, my use of CLIP is intended to predict the situation that the image fits in the timeline and it is not related to the identification of a face or age.

> The facial recognition approach is more viable, but there are caveats. To start, I would want to know how well clustering by a person's age actually works without false positives. I think the max distance needs to be set low for this, low enough that only the same person around the same age could meet the threshold. Maybe 0.1 or 0.15.

I would like to use maxDistance 0.1 or 0.15, but as I mentioned everything I try below 0.3 does not find matches even when the `personId` has over 1000 photos from 2014 to 2023, I suspect `cosine_ops` is not "deep" enough to do this.

> If the core idea works, it would need to differentiate between actual EXIF metadata and the prediction by writing to a different column. It should also be opt-in.

Regarding the EXIF I guess it's because Immich isn't meant to touch the original image, so we'll have to work with a database approach for that, right?




Let me know what you think

mertalev:
CLIP will only find similar scenes. There's nothing to suggest it would correlate to date.

For max distance, you may want to wait for the facial clustering rework. An issue with the current approach is that search won't actually have all the embeddings since each job also inserts embeddings. If we cluster after all the embeddings are added, the search will have a higher chance of finding a match.

aviv926:
> CLIP will only find similar scenes. There's nothing to suggest it would correlate to date.

Yes, actually I think it's something that should be left to the user to decide whether to associate the imported images together with images with face and prediction

> For max distance, you may want to wait for the facial clustering rework. An issue with the current approach is that search won't actually have all the embeddings since each job also inserts embeddings. If we cluster after all the embeddings are added, the search will have a higher chance of finding a match.



I'm not sure what you mean by "wait for the facial clustering rework", is this something in the future that will be in Immich? Is this an existing process?
Do you have any suggestions on how I can improve this? As far as I looked at the code, I understood that Immich does not build a cluster for a person, it looks for a vector related to an existing person, if there is none, it creates a new person in the system

mertalev:
> Yes, actually I think it's something that should be left to the user to decide whether to associate the imported images together with images with face and prediction

I don't think it's worth adding a feature that won't work well, even if it's up to the user to use it.

> I'm not sure what you mean by "wait for the facial clustering rework", is this something in the future that will be in Immich? Is this an existing process?

Yes, I'm referencing PR #5598.

> Do you have any suggestions on how I can improve this? As far as I looked at the code, I understood that Immich does not build a cluster for a person, it looks for a vector related to an existing person, if there is none, it creates a new person in the system

That's how the clustering works. Creating a new person is creating a new cluster, finding an existing person grows that cluster.

aviv926:
> I don't think it's worth adding a feature that won't work well, even if it's up to the user to use it.

The thing is, if someone uploads birthday photos from the past and they are without EXIF and there are some photos where the cake is taken, it will not be associated with the date when photos with the face received a prediction.
I mean it's pretty clear that something is better than nothing, but maybe if the team wants to get this feature it will be possible to upgrade this feature and find a solution to this problem together

> Yes, I'm referencing PR #5598.

Interesting, I actually saw that you talked about it on Discord a few times, it would be interesting to see it integrated into Immich, do you think the PR you created will eventually be merged into Immich?

> That's how the clustering works. Creating a new person is creating a new cluster, finding an existing person grows that cluster.

True, the point is that as far as I understood Immich does not build one large cluster per person. Each time it tries to find the 100 embeds (of all people) that are closest to the embed created by the facial recognition model and thus it tries to match the embed to the personId with the closest embed (or maybe I'm misunderstanding?).

aisbergde:
In General, the concept here is fine and I will subscribe.
Waiting for #5598 is a good idea.
But I could support only regarding SQL.

aviv926:
> In General, the concept here is fine and I will subscribe. Waiting for #5598 is a good idea. But I could support only regarding SQL.

Right now my line of thought would be to create a query that tries to create a vector average for all the assets from each month of the year and then try to compare the vector created against those averages, in order to getting as high an accuracy as possible.
I believe I will sit on it soon but it might be a better idea to wait for the implementation of https://github.com/immich-app/immich/pull/5598

aviv926:
I'm closing it, hopefully I'll have time to update it soon with all the new changes and get a higher prediction accuracy.

PR #6040, opened 2023-12-28
feat(mobile): Focus search on doubletap nav button

justinvdk:
Focuses search when tapping on the 'search' navigational button when search page already active.

Small demo:

https://github.com/immich-app/immich/assets/9844285/1f10bbf5-3b0a-4b63-8186-08f78c88a365

This was my first change in a Flutter codebase. The way I declare a closure to capture the `WidgetRef` inside `ImmichSearchBar` feels a bit dirty. Would love to hear what would be a better approach.

Tested on "tablet size"-screens by hardcoding the conditional to use that layout.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1701680307,
+        "narHash": "sha256-kAuep2h5ajznlPMD9rnQyffWG8EM/C73lejGofXvdM8=",
+        "owner": "numtide",
+        "repo": "flake-utils",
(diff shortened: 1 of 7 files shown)


bo0tzz:
❤🧡💛

bo0tzz:
FYI for the future, you almost never need to start a new PR to fix something like that :) https://sethrobertson.github.io/GitFixUm/fixup.html is a helpful resource for fixing up branches.

justinvdk:
> FYI for the future, you almost never need to start a new PR to fix something like that :) https://sethrobertson.github.io/GitFixUm/fixup.html is a helpful resource for fixing up branches.

Thanks for the link. I'm ok with git :); but I wasn't aware I wouldn't be able to re-open my own PR; which I closed before thinking about that.

PR #6045, opened 2023-12-29
feat(web): quick hide person on detail panel

martabal:
## Changes made in this PR

With this PR, you can quickly hide/unhide a person from the detail panel by right clicking on the image thumbnail of the person.


web/src/lib/components/asset-viewer/detail-panel.svelte
```diff
@@ -75,13 +79,37 @@
...
+  const handleHidePerson = async (index: number) => {
+    const timeout = setTimeout(() => (isShowLoadingUpdatePerson = true), 100);
+    try {
+      await api.personApi.updatePerson({
+        id: people[index].id,
+        personUpdateDto: { isHidden: !people[index].isHidden },
+      });
+      showingHiddenPeople = true;
+      people[index].isHidden = !people[index].isHidden;
+      notificationController.show({ message: 'SPerson visibility changed', type: NotificationType.Info });
```
danieldietzler, line 103:
```suggestion
      notificationController.show({ message: 'Person visibility changed', type: NotificationType.Info });
```

web/src/lib/components/asset-viewer/detail-panel.svelte
```diff
@@ -75,13 +79,37 @@
...
       asset = assetUpdate;
     }
   });
 
+  const handleRightClickPeople = (index: number) => {
+    showActionPerson[index] = !showActionPerson[index];
+  };
+
+  const handleHidePerson = async (index: number) => {
+    const timeout = setTimeout(() => (isShowLoadingUpdatePerson = true), 100);
```
danieldietzler, line 95: Having a timeout seems rather ugly... Does it behave weirdly if we don't wait those 100ms?

web/src/lib/components/asset-viewer/detail-panel.svelte
```diff
@@ -247,19 +278,36 @@
...
+                      class="absolute rounded-xl items-center bg-gray-300 dark:bg-slate-100 py-3 px-6 text-left text-sm font-medium text-immich-fg hover:bg-gray-200 focus:outline-none focus:ring-2 focus:ring-inset dark:text-immich-dark-bg dark:hover :bg-gray-200 transition-colors"
+                      use:clickOutside
+                      on:outclick={() => (showActionPerson[index] = false)}
+                      on:click|preventDefault|stopPropagation={() => handleHidePerson(index)}
+                    >
+                      {#if person.isHidden}
+                        Unhide
+                      {:else}
+                        Hide
+                      {/if}
```
danieldietzler, line 308: Maybe an eye icon or something for the "emojification-factor"? ;)

web/src/lib/components/asset-viewer/detail-panel.svelte
martabal: Yep, if you don't have this timeout, the loading spinner flashes even if you have a low latency.

web/src/lib/components/asset-viewer/detail-panel.svelte
martabal: I thought about that, but I don't know what look the best between having a text or an icon for that one :


| | Hide | Unhide |
|  :---: | :---: | :---: |
| Icons | Screenshot from 2023-12-29 12-42-15 | Screenshot from 2023-12-29 12-42-08 |
| Text | Screenshot from 2023-12-29 12-42-38 | Screenshot from 2023-12-29 12-42-29 |

web/src/lib/components/asset-viewer/detail-panel.svelte
martabal: Exactly

web/src/lib/components/asset-viewer/detail-panel.svelte
danieldietzler: Hm yeah you're right... In my imagination it looked better lol

web/src/lib/components/asset-viewer/detail-panel.svelte
martabal: It's clearly not needed, but I like to show that the request is being processed when you update something.
For exemple, in the case where the user is disconnected and tries to update a person visibility, if he hits the "Hide"/"Unhide" button, he won't be surprised if the person visibility does not change.

danieldietzler: Fair! Then I'm fine with leaving it as is :)

danieldietzler (approved):
Thank you! :)

alextran1502:
I tested the PR and found that the left click is unclear to the user since we haven't used the left click mechanism anywhere in the app. 

Is it better to show a button to hide/unhide on top of the picture when you enter edit mode?

jrasm91:
The way this works on the people is you go to hide/show faces and then you see eye open vs eye close on each person thumbnail. Clicking on the thumbnail toggles the visibility and then you can hit save afterwards. I think that would be fine here.

alextran1502:
> > Is it better to show a button to hide/unhide on top of the picture when you enter edit mode?
> 
> Where should be that button ?

image

Right here with the `eye` icon

martabal:
> Right here with the `eye` icon

That's for the faces, I don't think it's a great idea to mix faces / people together

Maybe the easiest solution would be to have a new shortcut to quickly hide a person from person page

PR #6048, opened 2023-12-29
feat(mobile): Focus search on doubletap nav button

justinvdk:
Superseeds #6040 

Focuses search when tapping on the 'search' navigational button when search page already active.

Small demo:

https://github.com/immich-app/immich/assets/9844285/1f10bbf5-3b0a-4b63-8186-08f78c88a365

This was my first change in a Flutter codebase. The way I declare a closure to capture the `WidgetRef` inside `ImmichSearchBar` feels a bit dirty. Would love to hear what would be a better approach.

Tested on "tablet size"-screens by hardcoding the conditional to use that layout.

### DIFF
--- a/mobile/lib/modules/search/providers/search_focus_notifier.provider.dart
+++ b/mobile/lib/modules/search/providers/search_focus_notifier.provider.dart
@@ -1,9 +1,0 @@
-import 'package:flutter/material.dart';
-
-final searchFocusNotifierProvider = SearchFocusNotifier();
-
-class SearchFocusNotifier with ChangeNotifier {
-  void requestFocus() {
-    notifyListeners();
-  }
-}
--- a/mobile/lib/modules/search/providers/search_focus_notifier.provider.dart
+++ b/mobile/lib/modules/search/providers/search_focus_notifier.provider.dart
@@ -1,0 +1,9 @@
+import 'package:flutter/material.dart';
+
+final searchFocusNotifierProvider = SearchFocusNotifier();
+
+class SearchFocusNotifier with ChangeNotifier {
+  void requestFocus() {
+    notifyListeners();
+  }
+}
--- a/mobile/lib/modules/search/ui/immich_search_bar.dart
+++ b/mobile/lib/modules/search/ui/immich_search_bar.dart
@@ -4,7 +4,6 @@ import 'package:flutter_hooks/flutter_hooks.dart';
 import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:immich_mobile/extensions/build_context_extensions.dart';
 import 'package:immich_mobile/modules/search/providers/search_page_state.provider.dart';
-import 'package:immich_mobile/modules/search/providers/search_focus_notifier.provider.dart';
 
 class ImmichSearchBar extends HookConsumerWidget
     implements PreferredSizeWidget {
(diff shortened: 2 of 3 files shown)


justinvdk:
@bo0tzz Would you be so kind to apply the mobile tag for me ?

mobile/lib/modules/search/providers/search_focus_notifier.provider.dart
shenlong-tanwen, line 1: This is not actually a riverpod provider. Can we move this class to a different file instead?

Maybe the search file itself or a new utils folder and keep this there instead?
`mobile/lib/modules/search/ui/immich_search_bar.dart`

Also, since this is not a provider, we can just name it as `searchFocusNotifier`

mobile/lib/modules/search/ui/immich_search_bar.dart
```diff
@@ -21,6 +22,28 @@ class ImmichSearchBar extends HookConsumerWidget
...
+
+      searchFocusNode.requestFocus();
+    }
+    onSearchFocusRequest() {
+      focusSearch();
+    }
+
+    useEffect(
+      () {
+        searchFocusNotifierProvider.addListener(onSearchFocusRequest);
```
shenlong-tanwen, line 39: I guess we can use the `focusSearch` method directly as the callback to the notifier?

mobile/lib/modules/search/ui/immich_search_bar.dart
```diff
@@ -41,12 +64,7 @@ class ImmichSearchBar extends HookConsumerWidget
         focusNode: searchFocusNode,
         autofocus: false,
         onTap: () {
```
shenlong-tanwen, line 66:
```suggestion
        onTap: focusSearch
```

shenlong-tanwen (approved):

mobile/lib/modules/search/providers/search_focus_notifier.provider.dart
justinvdk: I have updated this. I chose to put it in the file instead over a separate file for now as that is the first option you proposed.

mobile/lib/modules/search/ui/immich_search_bar.dart
justinvdk: Correct. I think its cleaner. I was having a hard time getting used to the codebase and used this naming to keep things readable, intent-wise :)

mobile/lib/modules/search/providers/search_focus_notifier.provider.dart
justinvdk: For my own sanity; `lib/modules/asset_viewer/providers/scroll_notifier.provider.dart` has the same issue right? I was 'inspired' by that pattern.

shenlong-tanwen: > For my own sanity; `lib/modules/asset_viewer/providers/scroll_notifier.provider.dart` has the same issue right? I was 'inspired' by that pattern.

Ah, we might need to refactor it to a different place as well then. I'm thinking of a module level utils folder to group utility classes / methods for that module. Maybe in a separate refactor 😅

justinvdk: Makes sense 😄 
Thank you for your review.

Is it ok for me to self-resolve feedback or do we let the reviewer do that?

shenlong-tanwen: > Is it ok for me to self-resolve feedback or do we let the reviewer do that?

I've resolved them myself in few of my PRs and let the reviewer do it in other few. But I guess the best is to let the reviewer resolve it.

alextran1502 (approved):

PR #6054, opened 2023-12-29
feat: storage template locking + fix for database locks

zackpollard:
This PR fixes database locking so that it always uses the same session for unlocking as was used for locking.
It also adds locking to the storage template migration solution so only a single file is moved at a time, regardless of how many microservice instances you have running.

PR based on top of #5917


danieldietzler (approved):
LGTM! Do you plan on making newPath unique in this PR or should that be a third one?

zackpollard:
> LGTM! Do you plan on making newPath unique in this PR or should that be a third one?

Different PR

jrasm91 (approved):

PR #6094, opened 2024-01-01
fix(docs): update broken link

m1yon:
I noticed a broken link towards the bottom of the [Config File page](https://immich.app/docs/install/config-file) in the docs.


danieldietzler (approved):
Oh nice, thanks! I actually stumbled across that earlier but didn't come around to fix it yet. Could you use just a path though? `/docs/install/environment-variables` should work

m1yon:
> Oh nice, thanks! I actually stumbled across that earlier but didn't come around to fix it yet. Could you use just a path though? `/docs/install/environment-variables` should work

Whoops, updated!

jrasm91:
Do you mind changing the link in `remote-access.md` as well do use an internal url instead of an external one? (`/docs/administration/reverse-proxy`)

jrasm91 (approved):

PR #6103, opened 2024-01-01
chore: svelte-kit-2

jrasm91:
Upgrade to svelte kit 2 and vite 5.
- Redirects no longer need to be thrown
- Type imports are a lot more strict

### DIFF
--- a/web/src/lib/components/shared-components/navigation-bar/navigation-bar.svelte
+++ b/web/src/lib/components/shared-components/navigation-bar/navigation-bar.svelte
@@ -30,7 +30,11 @@
   const logOut = async () => {
     resetSavedUser();
     const { data } = await api.authenticationApi.logout();
-    goto(data.redirectUri || '/auth/login?autoLaunch=0');
+    if (!data.redirectUri.startsWith('/')) {
+      window.location.href = data.redirectUri;
+    } else {
+      goto(data.redirectUri);
+    }
   };
 </script>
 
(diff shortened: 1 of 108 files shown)


martabal (approved):

web/src/lib/components/shared-components/navigation-bar/navigation-bar.svelte
```diff
@@ -30,7 +30,11 @@
   const logOut = async () => {
     resetSavedUser();
     const { data } = await api.authenticationApi.logout();
-    goto(data.redirectUri || '/auth/login?autoLaunch=0');
+    if (!data.redirectUri.startsWith('/')) {
+      window.location.href = data.redirectUri;
+    } else {
+      goto(data.redirectUri);
+    }
```
danieldietzler, line 37: Why is this necessary? Do we sometimes have an URI that's like `auth/login` and `goto` then does not understand that it's still the same host URL?

danieldietzler (approved):
LGTM! Can we have a linting rule that enforces type imports if we only use the import as a type and not as a value?

web/src/lib/components/shared-components/navigation-bar/navigation-bar.svelte
jrasm91: Goto only works for internal urls now

alextran1502 (approved):

PR #6108, opened 2024-01-01
feat(mobile): use new asset device endpoint

jrasm91:
The server has had this endpoint for a few releases now, so we can start using it. A few releases later we can go ahead and remove the old endpoint.


danieldietzler:
Finally! 🎉

shenlong-tanwen (approved):

Issue #4264, opened 2023-09-28
[BUG] time zone displacement out of range


Zutart:
### The bug

After moving to v1.79.1, I started getting a generic "Error - Internal server error (Immich server error)" messages. The logs are as follows, and my deployment strategy has been to use the docker compose file within unRAID v6.12.4

`immich_postgres          | 2023-09-28 17:41:26.350 UTC [58] ERROR:  time zone displacement out of range: "+127136-10-01T00:00:00.000Z"

immich_postgres          | 2023-09-28 17:41:26.350 UTC [58] CONTEXT:  unnamed portal parameter $3 = '...'

[...]

PR #6117, opened 2024-01-02
fix(web): allow future timebuckets

### LINKED ISSUE #5578: [BUG] Images with timestamps in the future don't show up in the web app
### The bug

For some reason, my Open Camera has been taking pictures with a timestamp that is off by one hour. They show up in the app and are backed up with no issue, but the web app does not show them until an hour later, when the timestamp is in the past.

It's not a huge issue for me, but it's at least worth documenting :)

Thanks for all the hard work!

Gentoo

### Version of Immich Server

v1.90.2

v1.89.0 build 113

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
version: "3.8"

services:
  immich-server:
    volumes:
    env_file:
      - .env
    environment:
      - NODE_ENV=production
    depends_on:
      - redis
      - database
    restart: always
    ports:
      - 2283:3001

  immich-microservices:
    volumes:
    env_file:
      - .env
    environment:
      - NODE_ENV=production
    depends_on:
      - redis
      - database
    restart: always

  #immich-machine-learning:
  #  image: bertmelis1/immich-machine-learning-noavx:release
  #  entrypoint: ["/bin/sh", "./entrypoint.sh"]
  #  volumes:
  #    - ${UPLOAD_LOCATION}:/usr/src/app/upload
  #  env_file:
  #    - .env
  #  environment:
  #    - NODE_ENV=production
  #  depends_on:
  #    - database
  #  restart: always

  redis:
    container_name: immich_redis
    image: redis:6.2
    restart: always

  database:
    image: postgres:14
    env_file:
      - .env
    environment:
      POSTGRES_USER: ${DB_USERNAME}
    volumes:
    restart: always

  typesense:
    image: typesense/typesense:0.24.0
    environment:
      - TYPESENSE_DATA_DIR=/data
    logging:
      driver: none
    volumes:
      - tsdata:/data
    restart: always
volumes:
  pgdata:
  tsdata:
  model-cache:
```

### Your .env content

```Shell
DB_HOSTNAME=immich_postgres
DB_USERNAME=postgres
DB_PASSWORD=postgres
DB_DATABASE_NAME=immich
REDIS_HOSTNAME=immich_redis
UPLOAD_LOCATION=/var/lib/immich/data
JWT_SECRET=hi
PUBLIC_LOGIN_PAGE_MESSAGE=
TYPESENSE_API_KEY=meep.
TYPESENSE_ENABLED=true
```

### Reproduction steps

```bash
1. Use the android app to upload a picture with a timestamp in the future
2. Check web app for the picture - it does not show up
3. Wait until the timestamp is in the past
4. The picture shows up in the web app
```

### Additional information

_No response_

m1yon:
fixes #5578

The original cause of the issue was a fix (#4540) intended for #4264 and #4538, which are Postgres "Time zone displacement out of range" errors caused by invalid dates.

[Postgres only supports date values from 4713 BC to 5874897 AD.](https://www.prisma.io/dataguide/postgresql/date-types#:~:text=The%20range%20of%20values%20for,in%20PostgreSQL%20for%20inserting%20data.)

I'm having a hard time replicating the original issues because using any date past 9999 AD in the `.xmp` files seems to reset the postgres date to the current date.

I've tested a smaller date range here though, and it works as expected.

### DIFF
--- a/server/src/infra/repositories/asset.repository.ts
+++ b/server/src/infra/repositories/asset.repository.ts
@@ -760,7 +760,7 @@ export class AssetRepository implements IAssetRepository {
     let builder = this.repository
       .createQueryBuilder('asset')
       .where('asset.isVisible = true')
-      .andWhere('asset.fileCreatedAt < NOW()');
+      .andWhere("asset.fileCreatedAt BETWEEN '4713-01-01 BC' AND '294276-12-31 AD'");
     if (assetType !== undefined) {
       builder = builder.andWhere('asset.type = :assetType', { assetType });
     }
--- a/server/src/infra/repositories/asset.repository.ts
+++ b/server/src/infra/repositories/asset.repository.ts
@@ -760,7 +760,7 @@ export class AssetRepository implements IAssetRepository {
     let builder = this.repository
       .createQueryBuilder('asset')
       .where('asset.isVisible = true')
-      .andWhere("asset.fileCreatedAt BETWEEN '4713-01-01 BC' AND '294276-12-31 AD'");
+      .andWhere("asset.fileCreatedAt BETWEEN '4713-01-01 BC' AND '9999-12-31 AD'");
     if (assetType !== undefined) {
       builder = builder.andWhere('asset.type = :assetType', { assetType });
     }
--- a/server/src/infra/sql/asset.repository.sql
+++ b/server/src/infra/sql/asset.repository.sql
@@ -555,7 +555,7 @@ FROM
 WHERE
   (
     "asset"."isVisible" = true
-    AND "asset"."fileCreatedAt" < NOW()
+    AND "asset"."fileCreatedAt" BETWEEN '4713-01-01 BC' AND '9999-12-31 AD'
   )
   AND ("asset"."deletedAt" IS NULL)
 GROUP BY
@@ -668,7 +668,7 @@ FROM
 WHERE
   (
     "asset"."isVisible" = true
-    AND "asset"."fileCreatedAt" < NOW()
+    AND "asset"."fileCreatedAt" BETWEEN '4713-01-01 BC' AND '9999-12-31 AD'
     AND (
       date_trunc(
         'month',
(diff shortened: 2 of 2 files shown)


server/src/infra/repositories/asset.repository.ts
```diff
@@ -761,7 +761,7 @@ export class AssetRepository implements IAssetRepository {
     let builder = this.repository
       .createQueryBuilder('asset')
       .where('asset.isVisible = true')
-      .andWhere('asset.fileCreatedAt < NOW()');
+      .andWhere("asset.fileCreatedAt BETWEEN '4713-01-01 BC' AND '294276-12-31 AD'");
```
jrasm91, line 764: Hmm, this seems a little insane lol. What if we just did less than 9999?

m1yon: yeah it is a bit crazy haha.

We could just do less than 9999, but we would need a lower constraint on it too since it could also be out of range in that direction.

**Counter argument though:**
Allowing such a wide range would allow users to view any assets with weird dates and be able to change those dates through the UI manually. It would make troubleshooting easier and potentially prevent future related issues.

jrasm91: > it could also be out of range

Maybe, but it hasn't happened yet at least. If we are going to change this query to using `BETWEEN` can you show the results of running an `EXPLAIN ANALYZE` on the resulting query? This is a pretty critical query and we should check that it doesn't introduce any performance impact for instances that have lots of assets. Alternatively, we already know that less than DATETIME already works fine.

jrasm91: Let's rebase this after #6156 is merged so we can get a nice delta for the generated SQL.

bo0tzz: > What if we just did less than 9999

I think we've seen a few broken assets with 5-digit years in the past.

bo0tzz:
@jrasm91 #6156 has been merged, how do we get the SQL diff?

jrasm91:
Rebase this PR on main and then run `make sql`

jrasm91 (approved):

jrasm91:
Closing in favor of #6457.

bo0tzz:
@m1yon thank you for your effort to fix this, and for prompting us to dig a little deeper and find out what the actual underlying problem was!

PR #6146, opened 2024-01-03
feat(web,server): external domain setting

### LINKED ISSUE #6147: [Feature] Configurable Public URL for Share Links
### Discussed in https://github.com/immich-app/immich/discussions/4852

<sup>Originally posted by **NicholasFlamy** November  5, 2023</sup>
### The feature

So this was mentioned on the discord:
https://discord.com/channels/979116623879368755/994044917355663450/1170824736352260156

There should be a configurable option in either the admin panel (more likely this), or the user settings that allows changing the base URL that shows up in a share link when you create or view share links.

In the case of having one IP/domain for logged-in users who make the share link but want to share it with friends who are not users who have access through a different IP/domain then the share link would need the public URL rather than the private one that will show up because the user who is viewing the share link is accessing the page through that URL.

### Platform

- [X] Server
- [X] Web
- [ ] Mobile</div>

jrasm91:
Closes #6147.

Add the option to set the instance's "external domain", which currently is used as an override for `window.origin` when constructing the shared link url to copy to the clipboard.

### Admin Settings
image

## Shared Links
image

### DIFF
--- a/web/src/api/utils.ts
+++ b/web/src/api/utils.ts
@@ -19,6 +19,10 @@ export const copyToClipboard = async (secret: string) => {
   }
 };
 
+export const makeSharedLinkUrl = (externalDomain: string, key: string) => {
+  return `${externalDomain || window.location.origin}/share/${key}`;
+};
+
 export const oauth = {
   isCallback: (location: Location) => {
     const search = location.search;
--- a/web/src/lib/components/shared-components/create-share-link-modal/create-shared-link-modal.svelte
+++ b/web/src/lib/components/shared-components/create-share-link-modal/create-shared-link-modal.svelte
@@ -5,7 +5,7 @@
   import SettingSwitch from '$lib/components/admin-page/settings/setting-switch.svelte';
   import Button from '$lib/components/elements/buttons/button.svelte';
   import { handleError } from '$lib/utils/handle-error';
-  import { api, copyToClipboard, SharedLinkResponseDto, SharedLinkType } from '@api';
+  import { api, copyToClipboard, makeSharedLinkUrl, SharedLinkResponseDto, SharedLinkType } from '@api';
   import { createEventDispatcher, onMount } from 'svelte';
   import Icon from '$lib/components/elements/icon.svelte';
   import BaseModal from '../base-modal.svelte';
@@ -13,6 +13,7 @@
   import DropdownButton from '../dropdown-button.svelte';
   import { notificationController, NotificationType } from '../notification/notification';
   import { mdiLink } from '@mdi/js';
+  import { serverConfig } from '$lib/stores/server-config.store';
 
   export let albumId: string | undefined = undefined;
   export let assetIds: string[] = [];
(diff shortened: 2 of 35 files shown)


web/src/api/utils.ts
```diff
@@ -19,6 +19,10 @@ export const copyToClipboard = async (secret: string) => {
   }
 };
 
+export const makeSharedLinkUrl = (externalDomain: string, key: string) => {
```
danieldietzler, line 22: nitpick but I'd probably prefer `getSharedLinkUrl`

web/src/lib/components/shared-components/create-share-link-modal/create-shared-link-modal.svelte
```diff
@@ -182,7 +183,7 @@
       {:else}
         <div class="text-sm">
           Individual shared | <span class="text-immich-primary dark:text-immich-dark-primary"
-            >{editingLink.description}</span
+            >{editingLink.description || ''}</span
```
danieldietzler, line 186: :)

danieldietzler (approved):
Nice!

alextran1502 (approved):

vcdwelt:
Maybe a small thing to add. The external URl must be entered without a / at the end, otherwise the links conatining a // an wont work. 
Maybe something like:
"Domain for public shared links, including http(s):// and without / at the end."

Otherwise a really great addon! :-)

danieldietzler:
> Maybe a small thing to add. The external URl must be entered without a / at the end, otherwise the links conatining a // an wont work. Maybe something like: "Domain for public shared links, including http(s):// and without / at the end."
> 
> Otherwise a really great addon! :-)

Thanks for pointing this out! Actually, we shouldn't restrict the input but instead better sanitize the link :)

devtechk:
Hello Guys, I was tryng to expose externally only the "/share" links and not the entire immich service, just to share only the foto. Someone can help me? Thank you!

danieldietzler:
@devtechk Please open a new discussion or ask this question on our Discord, not in an already merged PR. However, as a first hint; that's something your reverse proxy should take care of (accept requests to /share if coming from a subnet/port/whatever).

devtechk:
Thank you @danieldietzler I've just managed it. I used Nginx Proxy Manager and duckdns so I've allowed the /share/* location and denied the others location "/".I think I will open a new discussion to tell how I've done. Where I can find discord about immich? thank you

danieldietzler:
The discord is linked right at the top of the readme. Glad to hear it works @devtechk :)

mk12:
For anyone else reading this, note that only exposing /share/* does not work:
https://github.com/alangrainger/immich-public-proxy?tab=readme-ov-file#why-not-simply-put-immich-behind-a-reverse-proxy-and-only-expose-the-share-path-to-the-public

mdddev:
Thank you for clarifying this. Are you the author of IPP? In fact I just tried yesterday to do something similar with a self-hosted version of Nextcloud, i.e. only publicly serving the /s/ route for public share links. This failed because other js assets from /dist failed to load. Sounds like a similar use case and I already anticipated the same pain for immich. Luckily I have now found IPP. Awesome, thx.

mk12:
@mdddev No I'm not the author, just a user. However I've not found it that reliable. Just now when I try to view an album that way, many of the images show "Oops... failed to load content..." I am considering just exposing Immich directly now that it's reached a stable release. I also like the real Immich UI a lot better.

PR #6155, opened 2024-01-04
refactor(mobile): use startOAuth and server features flags

shenlong-tanwen:
#### Changes made:

- Use `startOAuth` instead of `generateOAuthConfig` to get the redirect URL
- Use oauth feature flags and button text from server features and server config, instead of from the `generateOAuthConfig` api


jrasm91 (approved):
LGTM

PR #6163, opened 2024-01-04
feat(server): add env for reverse geocoding path

martabal:
## Changes made in this PR

This PR adds an environment variable (`IMMICH_REVERSE_GEOCODING_ROOT`) to set the path of the reverse geocoding dump directory.

### DIFF
--- a/docs/docs/install/environment-variables.md
+++ b/docs/docs/install/environment-variables.md
@@ -30,14 +30,15 @@ These environment variables are used by the `docker-compose.yml` file and do **N
 
 ## General
 
-| Variable                | Description                                  |       Default       | Services                                     |
-| :---------------------- | :------------------------------------------- | :-----------------: | :------------------------------------------- |
-| `TZ`                    | Timezone                                     |                     | microservices                                |
-| `NODE_ENV`              | Environment (production, development)        |    `production`     | server, microservices, machine learning, web |
-| `LOG_LEVEL`             | Log Level (verbose, debug, log, warn, error) |        `log`        | server, microservices                        |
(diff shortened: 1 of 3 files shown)


danieldietzler (commented):
What's the purpose of this? I mean it shouldn't matter at all how we store that stuff internally. You can always just map it to whatever volume you like with docker

zackpollard (approved):

zackpollard:
LGTM, only thing might be to add documentation of this path, but as it's for very specific cases, also happy for it to be an undocumented variable

danieldietzler (approved):

danieldietzler:
Didn't even know there is an AUR package lol

martabal:
> LGTM, only thing might be to add documentation of this path, but as it's for very specific cases, also happy for it to be an undocumented variable

I added it https://feat-reverse-geocoding-env.immich.pages.dev/docs/install/environment-variables. But, yes, it's probably good too if it's undocumented

martabal:
> Didn't even know there is an AUR package lol

It's for sure not a great experience to maintain this lol https://aur.archlinux.org/cgit/aur.git/plain/metadata.repository.ts.patch?h=immich

zackpollard:
> > LGTM, only thing might be to add documentation of this path, but as it's for very specific cases, also happy for it to be an undocumented variable
> 
> I added it https://feat-reverse-geocoding-env.immich.pages.dev/docs/install/environment-variables. But, yes, it's probably good too if it's undocumented

Lol somehow I missed that file entirely...

zackpollard (approved):

jrasm91:
Maybe we can migrate to a single env for a single folder and assume everything is a relative path from there. It is kind of annoying adding a new env anytime we read any file from disk.

PR #6175, opened 2024-01-04
feat(server): in upload folder, split the files into folders based on the first four of the files uuid

zackpollard:


jrasm91 (approved):
LGTM

PR #6194, opened 2024-01-05
docs: Revert "Update libraries.md - External Library step by step set up guide added"

bo0tzz:
Reverts immich-app/immich#5288

This new guide duplicates the more comprehensive guide that is already at the bottom of the page, and also has some inaccuracies that have confused users about what paths they need to configure.


jrasm91:
I think we have a dedicated "guide" as well now.

jrasm91 (approved):

PR #6213, opened 2024-01-06
feat(server): add postgres major version check

maxer137:
The pgvecto.rs extension, though not distributed, can be built for PostgreSQL 12 and 13.
An installation of PostgreSQL 12 with the pgvecto.rs extensions installed will not be caught by immich.
This causes immich to attempt to run the database migrations without having a proper environment.
With assertPostgresql the server will throw an error if the PostgreSQL version is not within the supported range.

### DIFF
--- a/server/src/domain/database/database.service.spec.ts
+++ b/server/src/domain/database/database.service.spec.ts
@@ -28,6 +28,28 @@ describe(DatabaseService.name, () => {
   });
 
   describe('init', () => {
+    it('should return if minimum supported PostgreSQL and vectors version are installed', async () => {
+      databaseMock.getPostgresVersion.mockResolvedValueOnce(new Version(14, 0, 0));
+      databaseMock.getExtensionVersion.mockResolvedValueOnce(new Version(0, 1, 1));
+
+      await sut.init();
+
+      expect(databaseMock.getPostgresVersion).toHaveBeenCalledTimes(2);
+      expect(databaseMock.createExtension).toHaveBeenCalledWith(DatabaseExtension.VECTORS);
+      expect(databaseMock.createExtension).toHaveBeenCalledTimes(1);
+      expect(databaseMock.getExtensionVersion).toHaveBeenCalledTimes(1);
+      expect(databaseMock.runMigrations).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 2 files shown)


zackpollard:
Hey, thanks for the PR, I think I would prefer if we did this as a greater than comparison rather than an explicit list, would you be able to make that change? 😊

jrasm91 (approved):
LGTM

maxer137:
I have replaced the list with the comparison.
My original reasoning was that if PostgreSQL 17 would bring breaking changes we would also thrown an error.
But I think that this might be better.

zackpollard:
Yea, my reasoning is basically that we don't actually officially support any version other than 14 now anyway. So limiting the versions almost feels like we have tested with versions over 14 which we don't. Running anything other than 14 is at your own risk basically, but I agree we shouldn't let people run versions under 14 as it's likely to use features that don't yet exist in those versions.

zackpollard:
You have a failing test but once that is fixed I'll get this merged in 😊

maxer137:
The problem is that in the tests, we use mockResolvedValueOnce.
With my addition, the postgres version is being read out twice. 
Shall I replace the `databaseMock.getPostgresVersion.mockResolvedValueOnce(new Version(major, 0, 0)))` calls with 
`databaseMock.getPostgresVersion.mockResolvedValue(new Version(major, 0, 0)))`?

zackpollard:
> The problem is that in the tests, we use mockResolvedValueOnce. With my addition, the postgres version is being read out twice. Shall I replace the `databaseMock.getPostgresVersion.mockResolvedValueOnce(new Version(major, 0, 0)))` calls with `databaseMock.getPostgresVersion.mockResolvedValue(new Version(major, 0, 0)))`?

Works for me

zackpollard:
Would maybe be good also to add a test specifically testing this functionality, so a test that asserts the correct error is thrown when postgres version is less than 14 😄

zackpollard:
I haven't merged this yet as I wanted to double check the tests on my PC, won't be able to do that until this evening

jrasm91:
I checked with 13 and it threw and error. Then I check with 14 and it worked as expected.

PR #6216, opened 2024-01-06
fix(server): Execute query in AlbumRepository.removeAsset method

adamantike:
The current `removeAsset` implementation just builds the query but does not execute it. That also seems to be the reason the `@GenerateSql` decorator was commented out.

### DIFF
--- a/server/src/infra/repositories/album.repository.ts
+++ b/server/src/infra/repositories/album.repository.ts
@@ -177,14 +177,15 @@ export class AlbumRepository implements IAlbumRepository {
     });
   }
 
-  // @GenerateSql({ params: [DummyValue.UUID] })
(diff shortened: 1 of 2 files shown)


zackpollard:
lol, nice spot 😆

zackpollard (approved):

jrasm91:
Uhhhh what. How long has this been like this?

adamantike:
Just [since `v1.82.1`](https://github.com/immich-app/immich/commit/0994575bf3c240ab105108826831502e182f2bec#diff-f0c5e563f31d69a89325ac8f1f15160daf1abdbf76da6ae4c66218bd4f065415L172) :)

adamantike:
However, this is not `removeAssets`, which seems to be the most used method. This is only called from a single place related to live photos.

PR #6239, opened 2024-01-08
feat(web): force delete with shift key

martabal:
## Changes made in this PR

With this PR, you can delete assets without trashing them with Shift + Delete


web/src/lib/utils/actions.ts
```diff
@@ -0,0 +1,25 @@
+import { notificationController, NotificationType } from '$lib/components/shared-components/notification/notification';
+import { api } from '@api';
+import { handleError } from './handle-error';
+
+export type OnAssetDelete = (assetId: string) => void;
```
alextran1502, line 5: maybe change the name to `OnDelete` to keep the consistency with other types

jrasm91:
On my operation system, you have `delete` and `shift delete`. Shift delete only bypasses the trash, not the confirmation modal as well. I don't think it is a good idea to bypass the modal by default. If anything we should add an option to not show the confirm modal again after the first time.

martabal:
> On my ubuntu system, you have `delete` and `shift delete`. Shift delete only bypasses the trash, not the confirmation modal as well. I don't think it is a good idea to bypass the modal by default. If anything we should add an option to not show the confirm modal again after the first time.

image

What do you think about that ?
Should we add a setting in the user settings to enable / disable this modal ?

web/src/lib/components/photos-page/delete-asset-dialog.svelte
```diff
@@ -0,0 +1,52 @@
...
+    escape: void;
+  }>();
+</script>
+
+<ConfirmDialogue
+  title="Permanently Delete Asset{size > 1 ? 's' : ''}"
+  confirmText="Delete"
+  on:confirm={handleConfirm}
+  on:cancel={() => dispatch('cancel')}
+  on:escape={() => dispatch('escape')}
```
jrasm91, line 34: I don't think we need separate cancel vs escape events, just emit "close" or "cancel" both both of them.

web/src/lib/stores/preferences.store.ts
```diff
@@ -96,3 +96,5 @@ export const albumViewSettings = persisted<AlbumViewSettings>('album-view-settin
   sortDesc: true,
   view: AlbumViewMode.Cover,
 });
+
+export const showDeleteModal = persisted<boolean>('delete-confirm-dialog', true, {});
```
jrasm91, line 100: This seems fine for me. It can be _per client_.

jrasm91 (approved):

web/src/lib/stores/preferences.store.ts
martabal: Should we add a toggle in the user settings ?

jrasm91: Sureeeeeee

PR #6240, opened 2024-01-08
feat(mobile): long-press delete button to permanently delete asset

shenlong-tanwen:
### Changes made:
- Handles the mobile app side change for #6235 
- Long-pressing the delete button will display the prompt for permanently deleting assets even if trash is enabled


mobile/lib/modules/home/ui/control_bottom_app_bar.dart
```diff
@@ -103,15 +103,25 @@ class ControlBottomAppBar extends ConsumerWidget {
...
                         },
                       );
                     } else {
-                      onDelete!();
+                      onDelete?.call(true);
                     }
                   }
                 : null,
+            onLongPressed: enabled
+                ? () => showDialog(
```
fyfrey, line 116: This block is identical to the case a few lines above. Consider moving this to a helper function if this makes things better to reduce code duplication.

fyfrey (approved):

shenlong-tanwen:
We can merge this after #4505

PR #6253, opened 2024-01-08
fix(web): handle trailing slash in external domain

danieldietzler:
Addresses https://github.com/immich-app/immich/pull/6146#issuecomment-1880650789

### DIFF
--- a/web/src/api/utils.ts
+++ b/web/src/api/utils.ts
@@ -20,7 +20,8 @@ export const copyToClipboard = async (secret: string) => {
 };
 
 export const makeSharedLinkUrl = (externalDomain: string, key: string) => {
-  return `${externalDomain || window.location.origin}/share/${key}`;
+  const url = externalDomain || window.location.origin;
+  return `${url + url.endsWith('/') ? '' : '/'}share/${key}`;
 };
 
 export const oauth = {


web/src/api/utils.ts
```diff
@@ -20,7 +20,8 @@ export const copyToClipboard = async (secret: string) => {
 };
 
 export const makeSharedLinkUrl = (externalDomain: string, key: string) => {
-  return `${externalDomain || window.location.origin}/share/${key}`;
+  const url = externalDomain || window.location.origin;
+  return `${url + url.endsWith('/') ? '' : '/'}share/${key}`;
```
bo0tzz, line 24: Is it possible to use a `URL` constructor instead? I believe this should work (untested):

```suggestion
  const base = externalDomain || window.location.origin;
  return new URL(`/share/${key}`, base).href;
```

zackpollard:
I'm wondering if it would be better to just validate this in the config, rather than have to correct for it later

danieldietzler:
> I'm wondering if it would be better to just validate this in the config, rather than have to correct for it later

I was thinking about that as well. Basically you have three options:
1) validate server-side and don't accept "invalid" domains
2) sanitize server-side and store the sanitized value
3) do what I did

IMO 2) is kinda ugly because then you save something and on reload there is something else. That's confusing. And 1) could be annoying I think because technically it's not wrong and I personally hate it when I get errors that shouldn't be errors and only happen because the software is too stupid to handle those lmao

web/src/api/utils.ts
danieldietzler: Oh, that's very well possible. I only did a quick google search (less than 1 minute lol) and people were like "yea, regex" lol
I'll test it real quick

danieldietzler: Apparently this doesn't work if your domain has a path as well (so `immich.app/public/`. Not sure if that's common though...?

bo0tzz: That's a good point, though at the moment it's not possible to run Immich under a path anyways.

jrasm91: We haven't tested with it and I think there are some things that don't work with it like `goto(/photos)`

jrasm91 (approved):
LGTM

jrasm91:
I went through this same mental exercise and I think this solution works well enough.

web/src/api/utils.ts
danieldietzler: I know that there are absolute paths but I was thinking you could have proxy rewrite rules for that stuff. Or well, at that point, you could also redirect /share to /public/share, nvm lol

PR #6293, opened 2024-01-09
fix(server): check if sidecarPath exists

### LINKED ISSUE #6169: Automatically clean up removed sidecar files
Immich supports adding and editing .XMP sidecar files in the library folder, and has a few jobs to discover and sync those. However deleting those files externally is currently not supported, as once a sidecar path has been added to the database, it is expected to stay on the filesystem as well. The [sidecar discovery job](https://github.com/immich-app/immich/blob/main/server/src/domain/metadata/metadata.service.ts#L283-L298) should be updated so that when a `sidecarPath` exists in the database but not on disk, the path in the database is nulled out.

JW-CH:
Fixes #6169

### DIFF
--- a/server/src/domain/metadata/metadata.service.ts
+++ b/server/src/domain/metadata/metadata.service.ts
@@ -3,7 +3,6 @@ import { ImmichLogger } from '@app/infra/logger';
 import { Inject, Injectable } from '@nestjs/common';
 import { ExifDateTime, Tags } from 'exiftool-vendored';
 import { firstDateTime } from 'exiftool-vendored/dist/FirstDateTime';
-import fs from 'fs';
 import _ from 'lodash';
 import { Duration } from 'luxon';
 import { constants } from 'node:fs/promises';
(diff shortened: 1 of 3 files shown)


server/src/domain/metadata/metadata.service.ts
```diff
@@ -431,7 +432,18 @@ export class MetadataService {
...
-    const sidecarTags = asset.sidecarPath ? await this.repository.readTags(asset.sidecarPath) : null;
+
+    let sidecarTags = null;
+    if (asset.sidecarPath) {
+      // check if sidecar-file exists
+      if (fs.existsSync(asset.sidecarPath)) {
+        sidecarTags = await this.repository.readTags(asset.sidecarPath);
+      } else {
+        this.logger.error(
+          `Sidecar-File '${asset.sidecarPath}' for asset '${asset.id}' does not exist on the filesystem`,
```
JW-CH, line 443: Is an error/warning enough or does the db-field need to be nulled when not found?

danieldietzler, line 443: Don't import `fs` in domain/ files.

danieldietzler (changes requested):
~~What is the point of all this? `exiftool.read` should just throw an error (which we catch in the repository `readTags` method)~~
Has been discussed on Discord

jrasm91 (dismissed):
It seems like this is something better suited for the sidecar sync/discover job. Sync can probably be updated to null out the database in the case the file no longer exists.

JW-CH:
TODO: Add valuable tests 🤣

danieldietzler:
Hey @JW-CH, just wanting to check in... Would you like us to add some tests or do you plan on adding them yourself?

danieldietzler (approved):

jrasm91 (approved):

PR #4536, opened 2023-10-18
refactor(server): e2e tests

jrasm91:
Few ~~small~~ awesome(!) tweaks to improve e2e performance

PR #6307, opened 2024-01-10
refactor(server): split api and jobs into separate e2e suites

jrasm91:
In this PR:
- Split api and jobs e2e tests into two separate suites
- Add back `fsync=off` (#4536)
- Correctly teardown the app in metadata e2e so the test exists promptly

Next steps:
- Move all api based testing to e2e/api
- Add more microservices tests, including exiftool
- Clean up how the CLI e2e tests integrate with the code inside of `server/`
- Clean up (separate) e2e and unit fixtures
- Use the production build on immich-microservices for format testing


server/e2e/jobs/specs/metadata.e2e-spec.ts
```diff
@@ -33,8 +32,7 @@ describe(`${AssetController.name} (e2e)`, () => {
   });
 
   afterAll(async () => {
-    await db.disconnect();
-    await app.close();
+    await testApp.teardown();
```
jrasm91, line 35: This is what was causing the process to hang (not cleaning up exiftool-vendored)

server/e2e/api/specs/server-info.e2e-spec.ts
```diff
@@ -73,11 +73,11 @@ describe(`${ServerInfoController.name} (e2e)`, () => {
...
       expect(body).toEqual({
-        clipEncode: false,
-        configFile: true,
-        facialRecognition: false,
+        clipEncode: true,
+        configFile: false,
+        facialRecognition: true,
         map: true,
-        reverseGeocoding: false,
+        reverseGeocoding: true,
```
mertalev, line 80: Why does this need to be changed?

jrasm91: Some of the e2e tests that are now in the jobs suite, would actually run all the jobs, so it was configured with a config file that disabled it. I've reverted those changes for the api suite, so it's back to the defaults again.

mertalev: Ahh, gotcha

alextran1502 (approved):

danieldietzler (commented):
🚀

PR #6310, opened 2024-01-10
 fix(mobile): copy shared link

shenlong-tanwen:
#### Changes made:

- Fixes the mobile side handling to be in sync with the web client changes from #6309 
    - Only copy the shared link URL and not the password
    - Handle trailing slash in External Domain


jrasm91 (approved):
Thanks!

PR #6332, opened 2024-01-11
feat(server): provide the ability to search archived photos

stevenwcarter:
## Description

Adds a query parameter (`withArchived`) to the search URL parameters to allow the results to contain archived photos.

I commented on #5806 that it might be useful if we could optionally search archived photos, and this PR adds that ability. I'm not really a UI developer, so I've only added the ability to optionally perform this search with a query parameter for now. Eventually we might want to add radio options or magic search terms to enable searching archived/non-visible photos.

## How has this been tested

- Import photos
- Perform search that returns some of those photos.
- Archive some of the returned photos.
- Repeat search and confirm the archived photos are not returned.
- Add query parameter to URL `&withArchived=true`
- Confirm that the previously archived photos are now returned as part of the search.

### DIFF
--- a/server/src/infra/repositories/smart-info.repository.ts
+++ b/server/src/infra/repositories/smart-info.repository.ts
@@ -43,7 +43,7 @@ export class SmartInfoRepository implements ISmartInfoRepository {
   @GenerateSql({
     params: [{ userIds: [DummyValue.UUID], embedding: Array.from({ length: 512 }, Math.random), numResults: 100 }],
   })
-  async searchCLIP({ userIds, embedding, numResults, searchArchived }: EmbeddingSearch): Promise<AssetEntity[]> {
+  async searchCLIP({ userIds, embedding, numResults, withArchived }: EmbeddingSearch): Promise<AssetEntity[]> {
     if (!isValidInteger(numResults, { min: 1 })) {
       throw new Error(`Invalid value for 'numResults': ${numResults}`);
     }
@@ -52,14 +52,15 @@ export class SmartInfoRepository implements ISmartInfoRepository {
     await this.assetRepository.manager.transaction(async (manager) => {
       await manager.query(`SET LOCAL vectors.k = '${numResults}'`);
       await manager.query(`SET LOCAL vectors.enable_prefilter = on`);
-      let query = manager
+
+      const query = manager
         .createQueryBuilder(AssetEntity, 'a')
         .innerJoin('a.smartSearch', 's')
         .where('a.ownerId IN (:...userIds )')
         .andWhere('a.isVisible = true');
 
-      if (!searchArchived) {
-        query = query.andWhere('a.isArchived = false');
+      if (!withArchived) {
+        query.andWhere('a.isArchived = false');
       }
 
       results = await query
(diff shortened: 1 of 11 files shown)


danieldietzler (changes requested):
Where is the difference to https://immich.app/docs/api/search-assets and the `isArchived` parameter?

stevenwcarter:
Hm, fair point, that I might be able to exercise that existing option. The problem is that the changes made in #5806 explicitly set the `archived=false` check in the query, so no archived results will ever be returned without this change.

danieldietzler:
> Hm, fair point, that I might be able to exercise that existing option. The problem is that the changes made in #5806 explicitly set the `archived=false` check in the query, so no archived results will ever be returned without this change.

We will refactor search anyways and then I think the asset search endpoint will be the foundation for that. (it has already lots of good filters)
So I'm not sure "smart search" is even a thing that will be around much longer lol
I'll leave it to the others to decide :)

danieldietzler:
Generally, I believe this PR does look good though

server/src/infra/repositories/smart-info.repository.ts
```diff
@@ -52,12 +52,18 @@ export class SmartInfoRepository implements ISmartInfoRepository {
...
-        .andWhere('a.isVisible = true')
-        .andWhere('a.isArchived = false')
+        .andWhere('a.isVisible = true');
+
+      if (!searchArchived) {
+        query = query.andWhere('a.isArchived = false');
+      }
+
+      results = await query
+        .andWhere('a.isArchived = true')
```
danieldietzler, line 66: Won't you have `andWhere('a.isArchived = false')` and `andWhere('a.isArchived = true')` both in the query if `searchArchived` is false?

stevenwcarter: You are absolutely correct, and it's embarrassing that I missed that. I'm not sure how my testing archived vs. non-archived search results was returning properly with that conditional. I'm fixing and will push a new build. I've been rebasing to create a single commit, is that the preferred process you follow for PRs for Immich?

stevenwcarter: I've pushed a fix for this. I also repeated my testing, and confirmed the search is working as expected both with and without the query parameter.

danieldietzler: It's not embarrassing at all! I miss such things all the time. It happens when you're in the flow. That's what reviews are for; if everyone would make no typos/mistakes ever, we wouldn't need reviews either :D

danieldietzler (approved):
Thank you! Great work :)

stevenwcarter:
Do I need to clean up these conflicts in order for this to be merged? Won't most PRs cause a conflict with these build artifacts included in the repository?

danieldietzler:
Yes, they need to be cleaned up. Since those are just generated files you should be able to just run `make api` again (during the rebase)

jrasm91:
> Yes, they need to be cleaned up. Since those are just generated files you should be able to just run `make api` again (during the rebase)

`make open-api`

danieldietzler:
> > Yes, they need to be cleaned up. Since those are just generated files you should be able to just run `make api` again (during the rebase)
> 
> `make open-api`

Did the open api refactoring change the target?

jrasm91 (approved):

PR #6337, opened 2024-01-11
fix(server): extraction of Samsung Motionphoto videos

### LINKED ISSUE #4873: [BUG] Unable to play motion photos from One UI 6 (Android 14)
### The bug

Immich is unable to play motion photos captured from One UI 6 (Samsung's Android 14) while all other apps can play it just fine.
I'm not sure if this applies to all Android 14 devices or if it is Samsung specific.

While I am unsure of the exact cause, the following characteristics were noticed for all photos taken with that particular OS version:

One UI 5 (Android 13) has
Directory item length : 0
Directory item padding : 59

One UI 6 (Android 14) has
Directory item length : varies from file to file
Directory item padding : 111

I added one sample picture from each version.
https://drive.google.com/drive/folders/1as4MVNJsLQWUTeVFcv8AusSukWHEFaDb?usp=sharing

Support thread from discord:
https://discord.com/channels/979116623879368755/1170085304414240848

Synology Docker Compose

### Version of Immich Server

v1.84

v1.84

### Platform with the issue

- [X] Server
- [X] Web
- [X] Mobile

```YAML
version: "3.8"

services:
  immich-server:
    container_name: immich_server
    volumes:
      - /volume1/homes/user/Photos:/data/Photos:ro
    env_file:
      - .env
    depends_on:
      - redis
      - database
      - typesense
    restart: always

  immich-microservices:
    # extends:
    #   file: hwaccel.yml
    #   service: hwaccel
    volumes:
      - /volume1/homes/user/Photos:/data/Photos:ro
    env_file:
      - .env
    depends_on:
      - redis
      - database
      - typesense
    restart: always

  immich-machine-learning:
    volumes:
      - model-cache:/cache
    env_file:
      - .env
    restart: always

  immich-web:
    container_name: immich_web
    env_file:
      - .env
    restart: always

  typesense:
    environment:
      - TYPESENSE_DATA_DIR=/data
      - GLOG_minloglevel=1
    volumes:
      - tsdata:/data
    restart: always

  redis:
    container_name: immich_redis
    restart: always

  database:
    env_file:
      - .env
    environment:
      POSTGRES_USER: ${DB_USERNAME}
    volumes:
    restart: always

  immich-proxy:
    container_name: immich_proxy
    ports:
      - 2283:8080
    depends_on:
      - immich-server
      - immich-web
    restart: always

volumes:
  pgdata:
  model-cache:
  tsdata:
```

### Your .env content

```Shell

UPLOAD_LOCATION=./library

IMMICH_VERSION=release

DB_PASSWORD=postgres

DB_HOSTNAME=immich_postgres
DB_USERNAME=postgres
DB_DATABASE_NAME=immich

REDIS_HOSTNAME=immich_redis
```

### Reproduction steps

```bash
1. Take a picture with One UI 6 device
2. Upload it to Immich, doesn't matter web or mobile
3. While the image itself loads, attempting to play motion photo will result in infinite looping of video, and trying to hover the cursor on the motion photo icon on the timeline will show an error mark.
```

### Additional information

_No response_

kaysond:
Fixes #4873.

The issue is that the current video extraction relies on data length and padding from the XMP inside the EXIF. This happened to work on One UI 5 (the resulting mp4 file was technically invalid, but still worked because the extra bytes were at the end), but stopped working on One UI 6. In any case, the author of exiftool recommends not using the XMP because it doesn't have any fundamental connection to the "Trailer" of the jpeg where the video data actually resides. See https://exiftool.org/forum/index.php?topic=15565.0

Fortunately, exiftool is already able to parse the jpeg trailer data structure, and exiftool-vendored has methods to leverage that. The data structure is somewhat complicated, (see https://github.com/exiftool/exiftool/blob/master/lib/Image/ExifTool/Samsung.pm#L1532), so I think it's unwise to attempt to parse the structure ourselves.

This also adds support for extracting the video from an heic-encoded motionphoto, since it's just a different tag name. (see https://exiftool.org/forum/index.php?topic=7161)

This is a minimally-invasive but naive implementation. Creating a dedicated exiftool singleton for each video that has to be extracted is inefficient and the performance hit would probably be noticeable if you upload a lot of motionphotos at once. Fixing this pretty straightforward; it would just mean refactoring exiftool up from `metadata.repository` into `metadata.service`.

I also didn't touch any of the existing extraction code. I'm not sure if it's still needed for iOS or other motionphoto implementations (Google?), but it could potentially be removed if not.

### DIFF
--- a/server/src/domain/metadata/metadata.service.ts
+++ b/server/src/domain/metadata/metadata.service.ts
@@ -1,7 +1,7 @@
 import { AssetEntity, AssetType, ExifEntity } from '@app/infra/entities';
 import { ImmichLogger } from '@app/infra/logger';
 import { Inject, Injectable } from '@nestjs/common';
-import { ExifDateTime, Tags, exiftool } from 'exiftool-vendored';
+import { ExifDateTime, Tags } from 'exiftool-vendored';
 import { firstDateTime } from 'exiftool-vendored/dist/FirstDateTime';
 import { constants } from 'fs/promises';
 import _ from 'lodash';
@@ -394,11 +394,11 @@ export class MetadataService {
       // Samsung MotionPhoto video extraction
       // JPEG-encoded
       if (tags.EmbeddedVideoType === 'MotionPhoto_Data' && tags.EmbeddedVideo) {
-        video = await this.repository.extractBinaryTag('EmbeddedVideo', asset.originalPath);
+        video = await this.repository.extractBinaryTag(asset.originalPath, 'EmbeddedVideo');
       }
       // HEIC-encoded
       else if (tags.MotionPhotoVideo) {
-        video = await this.repository.extractBinaryTag('MotionPhotoVideo', asset.originalPath);
+        video = await this.repository.extractBinaryTag(asset.originalPath, 'MotionPhotoVideo');
       }
       // Default video extraction
       else {
--- a/server/src/domain/metadata/metadata.service.ts
+++ b/server/src/domain/metadata/metadata.service.ts
@@ -31,6 +31,7 @@ import {
 } from '../repositories';
 import { StorageCore } from '../storage';
 import { FeatureFlag, SystemConfigCore } from '../system-config';
+import { exiftool } from 'exiftool-vendored';
 
 /** look for a date from these tags (in order) */
 const EXIF_DATE_TAGS: Array<keyof Tags> = [
@@ -367,6 +368,7 @@ export class MetadataService {
     let length = 0;
     let padding = 0;
 
+    // Not needed for extractBinaryTagToBuffer()
     if (isMotionPhoto && directory) {
       for (const entry of directory) {
         if (entry.Item.Semantic == 'MotionPhoto') {
@@ -390,11 +392,22 @@ export class MetadataService {
     try {
       const stat = await this.storageRepository.stat(asset.originalPath);
       const position = stat.size - length - padding;
-      const video = await this.storageRepository.readFile(asset.originalPath, {
-        buffer: Buffer.alloc(length),
-        position,
-        length,
-      });
+      let video: Buffer;
+      if (tags.EmbeddedVideoType === "MotionPhoto_Data" && tags.EmbeddedVideo) {
+        video = await exiftool.extractBinaryTagToBuffer("EmbeddedVideoFile", asset.originalPath)
+        await exiftool.end()
+      }
+      else if (tags.MotionPhotoVideo) {
+        video = await exiftool.extractBinaryTagToBuffer("MotionPhotoVideo", asset.originalPath)
+        await exiftool.end()
+      }
+      else {
+        video = await this.storageRepository.readFile(asset.originalPath, {
+          buffer: Buffer.alloc(length),
+          position,
+          length,
+        });
+      }
       const checksum = this.cryptoRepository.hashSha1(video);
 
       const motionPath = StorageCore.getAndroidMotionPath(asset);
--- a/server/src/domain/metadata/metadata.service.ts
+++ b/server/src/domain/metadata/metadata.service.ts
@@ -434,10 +434,6 @@ export class MetadataService {
           deviceId: 'NONE',
         });
 
-        await this.storageRepository.writeFile(motionAsset.originalPath, video);
-        await this.jobRepository.queue({ name: JobName.METADATA_EXTRACTION, data: { id: motionAsset.id } });
-        await this.assetRepository.save({ id: asset.id, livePhotoVideoId: motionAsset.id });
-
         // If the asset already had an associated livePhotoVideo, delete it, because
         // its checksum doesn't match the checksum of the motionAsset we just extracted
         // (if it did, getByChecksum() would've returned non-null)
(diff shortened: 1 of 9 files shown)


server/src/domain/repositories/metadata.repository.ts
```diff
@@ -35,4 +38,5 @@ export interface IMetadataRepository {
   reverseGeocode(point: GeoPoint): Promise<ReverseGeocodeResult | null>;
   readTags(path: string): Promise<ImmichTags | null>;
   writeTags(path: string, tags: Partial<Tags>): Promise<void>;
+  extractBinaryTag(tagName: string, path: string): Promise<Buffer>;
```
jrasm91, line 41: Can you have it be `path` and then `tagName`?

jrasm91 (approved):
Looks good, minus the linting.

kaysond:
Done with the draft. Just need to actually test it now and it should be good to go.

Do we want to add unit tests for live/motionphoto video extraction?

server/src/domain/repositories/metadata.repository.ts
kaysond: Done

jrasm91:
If you could add a unit test or two that would be great.

danieldietzler (approved):
LGTM! :)

kaysond:
> LGTM! :)

Thanks. Going to make one more change so that running the metadata extraction job will fix broken videos.

server/src/domain/metadata/metadata.service.ts
```diff
@@ -381,7 +381,11 @@ export class MetadataService {
       length = videoOffset;
     }
 
-    if (!length) {
+    if (
+      length === 0 &&
+      tags.MotionPhotoVideo === undefined &&
+      (tags.EmbeddedVideoFile === undefined || tags.EmbeddedVideoType !== 'MotionPhoto_Data')
+    ) {
```
jrasm91, line 388: Can you require this as: not length or no motion tag then return? Maybe it has motion tag as a constant in the line above

server/src/domain/metadata/metadata.service.ts
```diff
@@ -423,6 +440,11 @@ export class MetadataService {
         await this.jobRepository.queue({ name: JobName.METADATA_EXTRACTION, data: { id: motionAsset.id } });
       }
 
+      if (asset.livePhotoVideoId) {
```
jrasm91, line 443: Two cases - corrupted or good extract

kaysond:
Ok this is finally ready for merging. Thanks to everyone who helped!

A few notes:

- Users can run the Metadata Extraction (All) job to fix corrupt motionphoto videos. Maybe we can mention this in the release notes @alextran1502 ?
- Because we fix corrupt videos, the above job has slowed down from before. It could be moved into an entirely separate job, but that would be a much bigger change.
- It seems that the metadata extraction job does not run on trashed assets, but it does on the trashed asset's motionphoto videos. This means errors can still show up if an image with corrupt motionphoto video is trashed. (NB: it may be worth updating the "trashing" job to also trash the associated video)

jrasm91 (commented):
I'm surprised this works because motionPath is the same for both the old and new motion assets. Saving the new video would overwrite the old motion file, but then you queue the old one for deletion.

kaysond:
> I'm surprised this works because motionPath is the same for both the old and new motion assets. Saving the new video would overwrite the old motion file, but then you queue the old one for deletion.

@jrasm91 - that's a really good catch. I'm guessing that by the time the code looked like this, my immich instance had all the right assets, so it never triggered that code.

I moved the delete before the save. There's no race condition here, right? Meaning `await this.jobRepository.queue({ name: JobName.ASSET_DELETION, data: { id: asset.livePhotoVideoId } });` will block until the job completes?

jrasm91:
No, that only queues it. Who knows when it will be deleted. Although, we can also solve this by simply changing how the original path is generated - it doesn't have to be the parent asset's id. I think we could generate a UUID ahead of time, and use that to create the motion asset and motion asset original path.

kaysond:
> No, that only queues it. Who knows when it will be deleted. Although, we can also solve this by simply changing how the original path is generated - it doesn't have to be the parent asset's id. I think we could generate a UUID ahead of time, and use that to create the motion asset and motion asset original path.

Ok this should take care of things. I've been trying to test it but I can't get the dev docker compose to start up because of some weird docker issue. Tried two different machines and same bug on both...

If you have a chance to test it please lmk.

kaysond:
This should finally be ready for merge @jrasm91 
Tested by:

1. Ran immich v1.93.2 dev instance
2. Uploaded 3 test motionphotos
3. Note that OneUI 5 works, but 6 is broken
4. Switch to PR dev instance
5. Note that OneUI 5 still works, but 6 is broken
6. Run metadata extraction job
7. Note that old video assets are deleted and new video assets are extracted correctly
8. Run job again, nothing happens as expected

kaysond:
I'm going to hold my tongue until this is actually merged :D

Formatting fixed and tests fixed and updated to have more

jrasm91 (approved):
I tested this and it worked well. The only thing I noticed was that the broken video could be cached in the browser and that could prevent it from playing successfully immediately afterwards.

PR #6344, opened 2024-01-12
fix(server): extract image description

tlvince:
Extracts the image description (aka caption).

Per an image from with a description set in Lightroom:

```
$ exiftool swans.jpg | grep Swans
Image Description               : Swans & cygnets
Caption-Abstract                : Swans & cygnets
Description                     : Swans & cygnets
```

Closes https://github.com/immich-app/immich/discussions/4220


bo0tzz:
How does this interact with a description set through the Immich UI?

bo0tzz:
Since #5066, date and location are written to an XMP sidecar file when edited inside Immich. Maybe we should do the same for description here?

jrasm91 (approved):

jrasm91:
We already write Image Description actually

tlvince:
👋🏻 @alextran1502. e2e's green locally, but seems to be pending approval to be ran in GHA

> This workflow is awaiting approval from a maintainer

PR #6352, opened 2024-01-12
feat(server): track metadata extracted at

jrasm91:
Queue and track "missing exif" via the AssetJobStatus entity. Makes it possible to merge quotas (#4471) without dealing with re-organizing the exif table right away.


danieldietzler (approved):
Nice!

zackpollard (approved):

PR #6353, opened 2024-01-13
refactor: move asset stacks to their own entity

zackpollard:
Refactor to move asset stacks to their own entity. This is the first part of this work. Bundled into this is also some initial work for auto-stacking burst assets.

To Do:
- [x] Add SQL migrations for existing stacks to move to new table format

Further Work (Different PR(s)):
- [ ] Add FK constraint so primaryAssetId must exist within the same stack
- [ ] Check test cases and add more for asset stacking
- [ ] Finish burst auto-stacking work
- [ ] Refactor stacking to have a asset-stack service and repository, rework API calls from mobile and web to use new APIs for stack operations
- [ ] Migrate live photos to use these stacks (Add stack type to stack entity)


server/src/domain/asset/asset.service.ts
```diff
@@ -414,32 +416,71 @@ export class AssetService {
...
-      await this.assetRepository.updateAll([options.stackParentId], { stackParentId: null });
-    }
+      ids.push(options.stackParentId);
+      const assets = await this.assetRepository.getByIds(ids);
+      console.log(
+        'assets',
+        assets.map((a) => a.id),
+      );
+      stackIdsToCheckForDelete.push(
+        ...new Set(assets.filter((a) => !!a.stackId && stack?.id !== a.stackId).map((a) => a.stackId!)),
```
danieldietzler, line 446: Why do we need to create a new Set from it first?

server/src/domain/asset/asset.service.ts
```diff
@@ -414,32 +416,71 @@ export class AssetService {
...
+          }),
+        });
+      } else {
+        await this.assetStackRepository.save({
+          id: stack.id,
+          primaryAssetId: primaryAsset.id,
+          assets: ids.map((id) => {
+            return { id } as AssetEntity;
+          }),
+        });
```
danieldietzler, line 465: I'm a little thrown off by the logic... We're in the branch where option has a `stackParentId`. `primaryAsset` is exactly this parent. Now we get the stack from `primaryAsset` and check if that is defined. But that should always be the case because otherwise the `primaryAsset` couldn't have been found (using the stack) in the first place, no?

server/src/domain/asset/asset.service.ts
```diff
@@ -414,32 +416,71 @@ export class AssetService {
     const { ids, removeParent, dateTimeOriginal, latitude, longitude, ...options } = dto;
     await this.access.requirePermission(auth, Permission.ASSET_UPDATE, ids);
 
+    const stackIdsToCheckForDelete: string[] = [];
     if (removeParent) {
-      (options as Partial<AssetEntity>).stackParentId = null;
+      (options as Partial<AssetEntity>).stack = null;
```
danieldietzler, line 421: Can we type options from the beginning, so in l. 416?

server/src/domain/asset/asset.service.ts
```diff
@@ -476,11 +517,20 @@ export class AssetService {
...
-    if (asset.stack && asset.stack.length != 0) {
-      const stackIds = asset.stack.map((a) => a.id);
-      const newParentId = stackIds[0];
-      await this.assetRepository.updateAll(stackIds.slice(1), { stackParentId: newParentId });
-      await this.assetRepository.updateAll([newParentId], { stackParentId: null });
+    if (asset.stack?.primaryAssetId === id) {
+      const stackIds = asset.stack.assets.map((a) => a.id);
+      if (stackIds.length > 2) {
+        console.log('Updating stack, old primary asset', id);
+        const primaryAssetId = stackIds.find((a) => a !== id)!;
```
danieldietzler, line 524: Shouldn't this be equal?

server/src/domain/asset/asset.service.ts
```diff
@@ -555,16 +605,21 @@ export class AssetService {
 
     const childIds: string[] = [];
     const oldParent = await this.assetRepository.getById(oldParentId);
+    if (!oldParent || !oldParent.stackId) {
```
danieldietzler, line 608: I believe you can simplify this to `!oldParent?.stackId`?

server/src/domain/asset/response-dto/asset-response.dto.ts
```diff
@@ -116,9 +116,13 @@ export function mapAsset(entity: AssetEntity, options: AssetMapOptions = {}): As
...
     people: peopleWithFaces(entity.faces),
     checksum: entity.checksum.toString('base64'),
-    stackParentId: entity.stackParentId,
-    stack: withStack ? entity.stack?.map((a) => mapAsset(a, { stripMetadata })) ?? undefined : undefined,
-    stackCount: entity.stack?.length ?? null,
+    stackParentId: withStack ? entity.stack?.primaryAssetId : undefined,
+    stack: withStack
+      ? entity.stack?.assets
+          .filter((a) => a.id !== entity.stack?.primaryAssetId)
+          .map((a) => mapAsset(a, { stripMetadata })) ?? undefined
```
danieldietzler, line 123: That has been there before, I know. But `?? undefined` doesn't make any sense, right?

server/src/domain/asset/asset.service.ts
zackpollard: So we don't have many duplicates of stackId, it doesn't really matter it'd just be inefficient.

server/src/domain/asset/asset.service.ts
zackpollard: On my phone, but iirc we are in the branch where options.parentId is set, which is setting a new parent. If this pare t doesn't already have a stack we must create on, make it the parent, then delete all the other stacks that were merged into this new stack.

server/src/domain/asset/asset.service.ts
zackpollard: Probably, I mostly just kept things the same where I could

server/src/domain/asset/response-dto/asset-response.dto.ts
zackpollard: Again, on phone but I would expect that this value could also be null. This syntax will set any false values to undefined.

server/src/domain/asset/asset.service.ts
zackpollard: So this code is within an if block that only executes if the asset being deleted is the primary asset of a stack. In this case, `id` is the id of the primary asset in the stack. We then check the stack has more than 2 assets (if it has 2, it'll only have 1 after this asset is deleted), and if that check passes, we find a new primary asset. This `stackIds.find` searches for an asset that has an id other than the current `id`, which will then be the new primaryAssetId. I have renamed some of the variables to make what is going on a little clearer :smile:

server/src/domain/asset/asset.service.ts
zackpollard: Yea this was correct. Options is passed to us by the client, options.stackParentId is the new parent to be assigned to all the asset ids passed in the request. So we must create a new stack if one doesn't exist, or merge all other assets into the stack that this asset is currently in. In both cases, we set the primaryAssetId of the stack, new or existing, to the options.stackParentId (but we use primaryAsset.id), these values will always be the same. After this is done we then cleanup any stacks that no longer have assets later in the function.

server/src/domain/asset/response-dto/asset-response.dto.ts
jrasm91: This only sets null to undefined, everything else passes through as is.

zackpollard: Sorry, yes, it's specifically nulls it checks for and converts those to undefined

zackpollard: Looking over the code now I am at my PC, I think you are right and here we should actually never get null, so this is unnecessary.

zackpollard: image
Typescript appears to agree with this :stuck_out_tongue:

server/src/domain/asset/asset.service.ts
danieldietzler: Got it. Thanks for the explanation! :)

server/src/domain/asset/asset.service.spec.ts
```diff
@@ -15,6 +16,7 @@ import {
   newSystemConfigRepositoryMock,
   newUserRepositoryMock,
 } from '@test';
+import { newAssetStackRepositoryMock } from '@test/repositories/asset-stack.repository.mock';
```
jrasm91, line 19: this should import from `@test` like everything else.

server/src/domain/asset/asset.service.spec.ts
```diff
@@ -175,6 +178,7 @@ describe(AssetService.name, () => {
   let communicationMock: jest.Mocked<ICommunicationRepository>;
   let configMock: jest.Mocked<ISystemConfigRepository>;
   let partnerMock: jest.Mocked<IPartnerRepository>;
+  let assetStackMock: jest.Mocked<IAssetStackRepository>;
```
jrasm91, line 181: This is fine, I'd also be cool with just "stackMock"

server/src/domain/asset/asset.service.ts
```diff
@@ -1,3 +1,4 @@
+import { IAssetStackRepository } from '@app/domain/repositories/asset-stack.repository';
```
jrasm91, line 1: should import from `'../repositories'` like everything else

server/src/domain/asset/asset.service.ts
```diff
@@ -414,32 +416,67 @@ export class AssetService {
...
     } else if (options.stackParentId) {
+      //Creating new stack if parent doesn't have one already. If it does, then we add to the existing stack
       await this.access.requirePermission(auth, Permission.ASSET_UPDATE, options.stackParentId);
-      // Merge stacks
-      const assets = await this.assetRepository.getByIds(ids);
-      const assetsWithChildren = assets.filter((a) => a.stack && a.stack.length > 0);
-      ids.push(...assetsWithChildren.flatMap((child) => child.stack!.map((gChild) => gChild.id)));
+      const primaryAsset = await this.assetRepository.getById(options.stackParentId);
+      if (!primaryAsset) {
+        throw new Error('Asset not found');
```
jrasm91, line 435: This should probably throw a bad request, not a generic error that would result in a internal server error. Unless I'm mistaken, specifying a stackParentId (uuid) that doesn't exist should be an error. Also, the whole situation is a little weird, since the access check above it would fail pretty much always for this scenario.

server/src/domain/asset/asset.service.ts
```diff
@@ -414,32 +416,67 @@ export class AssetService {
...
+      );
+      const assetsWithChildren = assets.filter((a) => a.stack && a.stack.assets.length > 0);
+      ids.push(...assetsWithChildren.flatMap((child) => child.stack!.assets.map((gChild) => gChild.id)));
+
+      if (!stack) {
+        stack = await this.assetStackRepository.create({
+          primaryAssetId: primaryAsset.id,
+          assets: ids.map((id) => {
+            return { id } as AssetEntity;
+          }),
```
jrasm91, line 452: might as well change this to `=> ({ id } as AssetEntity)`

server/src/domain/metadata/metadata.service.ts
```diff
@@ -467,6 +467,7 @@ export class MetadataService {
       latitude: validate(tags.GPSLatitude),
       lensModel: tags.LensModel ?? null,
       livePhotoCID: (tags.ContentIdentifier || tags.MediaGroupUUID) ?? null,
+      autoStackId: this.getAutoStackId(tags),
```
jrasm91, line 470: I think this is the only part of the PR I understand :-)

server/src/infra/entities/asset.entity.ts
```diff
@@ -156,14 +156,11 @@ export class AssetEntity {
...
-  stackParentId?: string | null;
+  stackId?: string | null;
 
-  @ManyToOne(() => AssetEntity, (asset) => asset.stack, { nullable: true, onDelete: 'SET NULL', onUpdate: 'CASCADE' })
-  @JoinColumn({ name: 'stackParentId' })
-  stackParent?: AssetEntity | null;
-
-  @OneToMany(() => AssetEntity, (asset) => asset.stackParent)
-  stack?: AssetEntity[];
+  @ManyToOne(() => AssetStackEntity, { nullable: true, onDelete: 'SET NULL' })
```
jrasm91, line 161: probably should keep `onUpdate: 'CASCADE'` here

server/src/infra/entities/index.ts
```diff
@@ -1,3 +1,4 @@
+import { AssetStackEntity } from '@app/infra/entities/asset-stack.entity';
```
jrasm91, line 1: should import from './asset-stack.entity';

server/src/infra/repositories/asset-stack.repository.ts
```diff
@@ -0,0 +1,41 @@
+import { IAssetStackRepository } from '@app/domain/repositories/asset-stack.repository';
+import { AssetStackEntity } from '@app/infra/entities/asset-stack.entity';
```
jrasm91, line 2:
```suggestion
import { AssetStackEntity } from '../entities';
```

server/src/infra/repositories/asset-stack.repository.ts
```diff
@@ -0,0 +1,41 @@
+import { IAssetStackRepository } from '@app/domain/repositories/asset-stack.repository';
```
jrasm91, line 1:
```suggestion
import { IAssetStackRepository } from '@app/domain';
```

server/src/domain/repositories/asset-stack.repository.ts
```diff
@@ -0,0 +1,10 @@
+import { AssetStackEntity } from '@app/infra/entities/asset-stack.entity';
+
+export const IAssetStackRepository = 'IAssetStackRepository';
+
+export interface IAssetStackRepository {
+  create(assetStack: Partial<AssetStackEntity>): Promise<AssetStackEntity>;
+  save(asset: Pick<AssetStackEntity, 'id'> & Partial<AssetStackEntity>): Promise<AssetStackEntity>;
```
jrasm91, line 7:
```suggestion
  update(asset: Pick<AssetStackEntity, 'id'> & Partial<AssetStackEntity>): Promise<AssetStackEntity>;
```
In general, the repos try to have get, create, update, and delete. Although it's fine to have create and update both call "save" internally.

jrasm91 (commented):
Looks pretty good, although I am having a hard time wrapping my head about the actual stacking/unstacking logic.

jrasm91 (approved):
It looks like an entire asset-stack CRUD api exists inside of the bulk asset update endpoint now :laughing:, besides that I think this looks great. I was able to test creating stacks with the new code, as well as migrating from main to this branch and back again, all without losing existing stacks.

danieldietzler (approved):

PR #6354, opened 2024-01-13
fix(docs) Fix relative paths leading to broken links

thariq-shanavas:
If there is a trailing slash on a documentation page, any links referenced as ../xxx/file.md do not work

e.g.,
The links on this page work: https://immich.app/docs/overview/quick-start
The links on this page don't: https://immich.app/docs/overview/quick-start/

The trailing slash is automatically added if the user refreshes the page or gets there from an external link,
Specifying absolute path fixes this.

cc @danieldietzler

### DIFF
--- a/docs/docs/administration/backup-and-restore.md
+++ b/docs/docs/administration/backup-and-restore.md
@@ -1,6 +1,6 @@
 # Backup and Restore
 
-A [3-2-1 backup strategy](https://www.backblaze.com/blog/the-3-2-1-backup-strategy/) is recommended to protect your data. You should keep copies of your uploaded photos/videos as well as the Immich database for a comprehensive backup solution. This page provides an overview on how to backup the database and the location of user-uploaded pictures and videos. A template bash script that can be run as a cron job is provided [here](../guides/template-backup-script)
+A [3-2-1 backup strategy](https://www.backblaze.com/blog/the-3-2-1-backup-strategy/) is recommended to protect your data. You should keep copies of your uploaded photos/videos as well as the Immich database for a comprehensive backup solution. This page provides an overview on how to backup the database and the location of user-uploaded pictures and videos. A template bash script that can be run as a cron job is provided [here](/docs/guides/template-backup-script.md)
 
 ## Database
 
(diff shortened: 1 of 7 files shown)


alextran1502 (approved):

jrasm91:
Can we add a rule or check to catch these in the future?

jrasm91 (approved):

thariq-shanavas:
@jrasm91The trailing slash is never added on a local deployment, which makes me think it is a bug with Docusaurus. (Also presumably why it went uncaught so long)

If you concur, I'll open an issue there instead of writing a test to catch relative paths.

bo0tzz:
From https://github.com/facebook/docusaurus/issues/2298#issuecomment-590911963:
> For the links to work, a file name (along with the extension) has to be used (e.g. `[docs 2](doc_test_2.md)`). If you do something like `[docs 2](doc_test_2)`, Docusaurus doesn't treat that link as a Docusaurus file and doesn't process the href value. The end result is it becomes a relative link and defaults to the browser behavior

I think this might be what is happening for us? Unfortunately I can't find a lint or setting to enforce that `.md` suffix.

danieldietzler (approved):

jrasm91:
Sounds like it would be better to add extensions to the links instead.

thariq-shanavas:
The really confusing bit is Docusaurus does not add a trailing slash when deployed locally.
image


From Docusaurus documentation, it appears like it might be something to do with cloudflare.
https://docusaurus.io/docs/api/docusaurus-config
image

From the Cloudflare forum:
https://community.cloudflare.com/t/removing-trailing-slash-on-static-websites/583429

The easiest fix appears to be simply specifying the absolute path. I'm open to more elegant solutions though.

bo0tzz:
The `trailingSlash` option applies to the `.md` -> `.html` link transform. The issue I flagged above is that (some of?) our links don't even end in `.md` to begin with, meaning that docusaurus does not pick up on them for transforms (but just passes them through verbatim). That also means this `trailingSlash` option would have no effect on them.

Issue #6263, opened 2024-01-08
[BUG] API repeatedly reporting 'successfully uploaded x assets'


undaunt:
### The bug

Prior to the 1.92 release, such as on 1.91.4, if I were to use the CLI to upload a folder and it stated, eg: "successfully uploaded 30 assets", on a subsequent run it would state all assets had already been uploaded. (I checked because I am nervous about missing any photos on import.)

Now, no matter how many times I execute the same CLI folder upload, the message always states it has successfully uploaded the assets. I don't appear to have duplicates being added to Immich but I'm wondering if this is a false positive message, or something with the hashing changed, or otherwise.

### The OS that Immich Server is running on

Ubuntu 22.04

### Version of Immich Server

v1.92

### Version of Immich Mobile App

N/A

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
version: "3.8"

services:
  immich-server:
    container_name: immich-server
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    restart: unless-stopped
    profiles: ["all","immich"]
    networks:
      - $PROXY_NETWORK
      - $APP_NETWORK
    security_opt:
      - no-new-privileges:true
    depends_on:
      - immich-redis
      - immich-db
    command: [ "start.sh", "immich" ]
    volumes:
      - ${UPLOAD_DIR}:/usr/src/app/upload
      - /mnt/media/photos/icloud:/icloud:ro
      - $DOCKERDIR/immich/config/auth.yml:/root/.config/immich/auth.yml:ro
      - /etc/localtime:/etc/localtime:ro
    environment:
      - DB_HOSTNAME=${POSTGRES_IMMICH_HOST}
      - DB_DATABASE_NAME=${POSTGRES_IMMICH_DB}
      - DB_USERNAME=${POSTGRES_IMMICH_USER}
      - DB_PASSWORD=${POSTGRES_IMMICH_PASS}
      - REDIS_HOSTNAME=${IMMICH_REDIS_HOSTNAME}
    labels:
      - traefik.enable=true
      - traefik.http.routers.immich.entrypoints=https
      - traefik.http.routers.immich.rule=Host(`immich.$DOMAINNAME`)
      - traefik.http.routers.immich.middlewares=secure-headers@file,crowdsec@file
      - traefik.http.routers.immich.service=immich
      - traefik.http.services.immich.loadbalancer.server.port=3001

[...]

PR #6357, opened 2024-01-13
fix(cli): uploadCounters increase only when files are uploaded

AngelPone:
## Description

Currectly the upload function increase the `uploadCounter` for every asset that can not skip, causing issue #6263. 

This PR moves `uploadCounter++;` before, increase the counter only when files are actually uploaded. 

## Checklist:

### DIFF
--- a/cli/src/commands/upload.ts
+++ b/cli/src/commands/upload.ts
@@ -107,6 +107,7 @@ export class Upload extends BaseCommand {
               const formData = asset.getUploadFormData();
               const res = await this.uploadAsset(formData);
               existingAssetId = res.data.id;
+              uploadCounter++;
             }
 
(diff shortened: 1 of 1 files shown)


cli/src/commands/upload.ts
```diff
@@ -129,7 +130,6 @@ export class Upload extends BaseCommand {
           }
 
           totalSizeUploaded += asset.fileSize;
```
jrasm91, line 132: This line needs to be moved up as well. I was just going to do it myself and merge it, but it doesn't look like you've enabled that option.

AngelPone: Done

jrasm91 (approved):

PR #6376, opened 2024-01-14
added a configuration option to select the dri node in transcoding

t4keda:
This code adds a field in SystemConfigFFmpegDto to specify the node in /dev/dri used for VAAPI hardware transcoding. 

The default value is auto, and in that case the VAAPIConfig config class works like up until now (it chooses the first available node in /dev/dri), but if the field is configured with a path to a dri node that exists (/dev/dri/renderD128 for example) it uses that node in the ffmpeg command in -init_hw_device parameter

### DIFF
--- a/server/src/domain/media/media.util.ts
+++ b/server/src/domain/media/media.util.ts
@@ -287,14 +287,17 @@ export class BaseHWConfig extends BaseConfig implements VideoCodecHWConfig {
   }
 
   getPreferredHardwareDevice(): string | null {
-    if (this.config.preferredHwDevice !== 'auto') {
-      if (!this.devices.includes(this.config.preferredHwDevice.replace('/dev/dri/', ''))) {
-        throw new Error(`Device '${this.config.preferredHwDevice}' does not exist`);
-      }
-      return this.config.preferredHwDevice;
+    const device = this.config.preferredHwDevice;
+    if (device === 'auto') {
+      return null;
     }
 
-    return null;
+    const deviceName = device.replace('/dev/dri/', '');
+    if (!this.devices.includes(deviceName)) {
+      throw new Error(`Device '${device}' does not exist`);
+    }
+
+    return device;
   }
 }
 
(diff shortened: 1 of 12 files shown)


mertalev (commented):
Thanks for working on this! We expose every config option in the  settings UI, so this should be added there as well.

t4keda:
> Thanks for working on this! We expose every config option in the settings UI, so this should be added there as well.

I don't have a lot of spare time right now (hence only doing the config file part), but I will give it a shot :)

jrasm91:
If you don't have time I'm sure one of us can help with that part, so just let us know.

danieldietzler:
I can try to get it done tomorrow if you'd like :)

t4keda:
Sorry for the delay in the response and thanks for the offering :D

 Yesterday I started looking into the code of the web and I have already the option exposed, I will try to finish tonight the work and commit it.

Also I am thinking that maybe this option can be used also por QSV and other acceleration APIs, but I have to check it first

mertalev:
It can be used for QSV, but it would need to be tested to check that the command still works. Unlike VAAPI, we don't pass a device to the QSV command and let QSV decide what to use.

server/src/domain/media/media.util.ts
```diff
@@ -521,9 +531,19 @@ export class QSVConfig extends BaseHWConfig {
...
-    return [`-init_hw_device vaapi=accel:/dev/dri/${this.devices[0]}`, '-filter_hw_device accel'];
+
+    let hwDevice = `/dev/dri/${this.devices[0]}`;
+
+    if (this.config.preferredHwDevice !== 'auto') {
+      if (!this.devices.includes(this.config.preferredHwDevice.replace('/dev/dri/', ''))) {
+        throw new Error(`Device '${this.config.preferredHwDevice}' does not exist`);
+      }
+      hwDevice = this.config.preferredHwDevice;
+    }
```
danieldietzler, line 544: This is duplicated in the QSVConfig. Do you think it makes sense to move this part to some utility method in BaseHWConfig and call it in both places?

danieldietzler (approved):
Looks good to me!

server/src/domain/media/media.util.ts
t4keda: I Will do It tomorrow, a function that retorns either a string with the node or null

jrasm91 (approved):
Nicely done!

PR #6435, opened 2024-01-16
feat(web): store accordion state in search params

danieldietzler:
This PR introduces the ability to store the state of accordions (`isOpen`?) in the URL. If an accordion component gets a key `foo`, this key will be used in the URL. (i.e. `demo.immich.app/admin/system-settings?isOpen=foo`)


web/src/lib/stores/search-params.store.ts
```diff
@@ -0,0 +1,61 @@
...
+    this.store$.subscribe(this.handleUpdate);
+  }
+
+  private handleUpdate = (values: Array<T>) => {
+    if (values.length === 0) {
+      get(page).url.searchParams.delete(this.key);
+    } else {
+      get(page).url.searchParams.set(this.key, values.join(','));
+    }
+    goto(`?${get(page).url.searchParams.toString()}`);
```
martabal, line 24: It's probably better to have something like this : 

```ts
const newUrl = new URL($page.url);
if (values.length === 0) {
  newUrl.searchParams.delete(this.key);
} else {
  newUrl.url.searchParams.set(this.key, values.join(','));
}
goto(newUrl);
```

based on this https://github.com/sveltejs/kit/issues/2785#issuecomment-967790973

I'm planning to refactor across the entire webapp

jrasm91 (approved):
If your happy with this new approach, everything works and looks good to me.

web/src/lib/components/admin-page/settings/setting-accordion.svelte
```diff
@@ -1,10 +1,23 @@
...
+
   export let title: string;
   export let subtitle = '';
-
+  export let key: string;
   export let isOpen = false;
-  const toggle = () => (isOpen = !isOpen);
+
+  const syncFromUrl = () => (isOpen = hasParamValue('isOpen', key));
+  const syncToUrl = (isOpen: boolean) => updateParamList({ param: 'isOpen', value: key, add: isOpen });
```
martabal, line 12: In a previous PR, I put all query parameters in the `QueryParameter` enum in `constants.ts`, can you add it too ?

martabal (approved):

PR #6449, opened 2024-01-17
Update remove-offline-files.md

hirenshah:
The previous Windows Powershell script didn't work - resulted in a bunch of errors due to multiple reasons.  The revised code works as expected.


jrasm91 (approved):

aviv926:
In fact, everything can be replaced with this Python code and it will work well on all systems:

https://gist.github.com/sircharlo/10ae74bf53a252a367c206715d26088d

PR #6460, opened 2024-01-17
feat(server, web): accepted codecs

Hely0n:
As mentioned in [ this feature request](#6319), I implemented two new settings under the transcoding settings, which allow the admin to specify which video and audio codecs are acceptable for him, so videos in those codecs will fall under the "desired format" (renamed to "accepted formats") in the policy and thus not be transcoded due to their codec.

image
 
If the admin changes the target video or audio codec, the appropriate "accepted codecs" selection will be cleared and only include the newly selected target codec, to keep the existing behaviour intact.


Hely0n:
Just wanted to add, that somehow my tests were failing, even before I changed any code, so this must be due to some other commits. But Lint and prettier ran as it should.

web/src/lib/components/admin-page/settings/setting-select.svelte
```diff
@@ -10,12 +10,14 @@
   export let isEdited = false;
   export let number = false;
   export let disabled = false;
+  export let onChange: () => void = () => {};
```
mertalev, line 13: This should use `dispatch`. See [setting-switch](https://github.com/immich-app/immich/blob/main/web/src/lib/components/admin-page/settings/setting-switch.svelte) for an example of that.

web/src/lib/components/admin-page/settings/ffmpeg/ffmpeg-settings.svelte
```diff
@@ -164,6 +167,34 @@
           isEdited={config.ffmpeg.transcode !== savedConfig.ffmpeg.transcode}
         />
 
+        <SettingCheckboxes
+          label="ACCEPTED VIDEO CODECS"
+          {disabled}
+          desc="Select which video codecs should not be transcoded"
```
mertalev, line 173:
```suggestion
          desc="Select which video codecs do not need to be transcoded. Only used for certain transcode policies."
```

"should not" makes it sound like it overrides other settings.

web/src/lib/components/admin-page/settings/ffmpeg/ffmpeg-settings.svelte
```diff
@@ -164,6 +167,34 @@
...
+            { value: VideoCodec.Hevc, text: 'HEVC' },
+            { value: VideoCodec.Vp9, text: 'VP9' },
+          ]}
+          isEdited={!isEqual(sortBy(config.ffmpeg.acceptedVideoCodecs), sortBy(savedConfig.ffmpeg.acceptedVideoCodecs))}
+        />
+
+        <SettingCheckboxes
+          label="ACCEPTED AUDIO CODECS"
+          {disabled}
+          desc="Select which audio codecs should not be transcoded"
```
mertalev, line 187:
```suggestion
          desc="Select which audio codecs do not need to be transcoded. Only used for certain transcode policies."
```

mertalev (commented):
Looks good! Just some minor comments

web/src/lib/components/admin-page/settings/setting-select.svelte
Hely0n: Done, thank you

server/src/domain/system-config/system-config.core.ts
```diff
@@ -248,6 +250,14 @@ export class SystemConfigCore {
       }
     }
 
+    if (!config.ffmpeg.acceptedVideoCodecs.includes(config.ffmpeg.targetVideoCodec)) {
```
jrasm91, line 253: I added this so that the target video/audio codec is always included in the accepted lists.

web/src/lib/components/admin-page/settings/admin-settings.svelte
```diff
@@ -25,16 +26,17 @@
...
+      const { data: newConfig } = await api.systemConfigApi.updateConfig({
         systemConfigDto: {
           ...savedConfig,
-          ...config,
+          ...update,
         },
       });
 
-      savedConfig = { ...result.data };
+      config = cloneDeep(newConfig);
```
jrasm91, line 38: This line makes it possible so that if the response from " update config" is different than the current config, to display it immediately. Saving "accepted codecs" that doesn't include the target codec, will automatically mark it as accepted after saving.

web/src/lib/components/admin-page/settings/ffmpeg/ffmpeg-settings.svelte
```diff
@@ -89,6 +90,21 @@
           ]}
           name="acodec"
           isEdited={config.ffmpeg.targetAudioCodec !== savedConfig.ffmpeg.targetAudioCodec}
+          on:select={() => (config.ffmpeg.acceptedAudioCodecs = [config.ffmpeg.targetAudioCodec])}
+        />
+
+        <SettingCheckboxes
```
jrasm91, line 96: I moved these checkboxes to be directly after the codec selection.

jrasm91 (approved):
Thanks for the PR! I found one issue where it was possible to remove the target codec from the accepted codec list. I fixed this by updating config.core.ts to automatically/always add the target codec to the accepted lists. This also made the migration redundant, so I deleted it.

jrasm91:
This PR is good to go, but it looks like some issue with the mobile build is related to the new enum list, which is used for the first time in this PR. We will need to fix this on main first.

danieldietzler (approved):
LGTM!

PR #6461, opened 2024-01-18
fix(server): scale transcoded videos if dimensions are odd

mertalev:
## Description

Codecs like H.264 require both height and width to be even numbers. We handle this in some cases when scaling videos to a target resolution. However, transcoding will fail if a video is not scaled and has an odd dimension. It will also fail if the odd dimension is the smaller one and this value is used for scaling, as in this case the scaled video will still have an odd dimension. This PR handles these cases to make sure transcoded videos always have valid dimensions.

### DIFF
--- a/server/src/domain/media/media.service.spec.ts
+++ b/server/src/domain/media/media.service.spec.ts
@@ -600,6 +600,66 @@ describe(MediaService.name, () => {
       );
     });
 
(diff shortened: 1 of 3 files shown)


jrasm91 (approved):
LGTM

jrasm91:
The file that caused this error before is working again.

PR #6465, opened 2024-01-18
chore: update bug template

jrasm91:
- We don't really use the triage label
- All issues are supposed to be bugs, so the bug label seems redundant
- Removing the `[Bug]` title prefix because of the above mentioned, also it should make the title nicer when non-bugs are converted to discussions.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,6 +1,5 @@
 name: Report an issue with Immich
 description: Report an issue with Immich
-title: ""
 body:
   - type: markdown
     attributes:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -1,7 +1,6 @@
 name: Report an issue with Immich
 description: Report an issue with Immich
-labels: ["bug", "need triage"]
-title: "[BUG] <title>"
+title: ""
```
bo0tzz, line 3: The title field is optional, so I believe we can just leave it out.

```suggestion
```

bo0tzz (approved):

PR #6479, opened 2024-01-19
feat(server, web): Added TranscodePolicy "Bitrate higher than max bitrate or not in accepted format"

Hely0n:
And here comes my last promise from #6319 : 
The Bitratete policy.

This is just an additional transcode policy, which works like the "Transcode if higher than target resolution or not in desired format", but it will only transcode videos where the total bitrate is higher or equal than the bitrate set under "max bitrate". 

image

This will allow admins to set the upload speed of the provider, where the server is running, as max bitrate so only videos that wouldn't be streamable due to their high bitrate will be transcoded regardless of their resolution.

This PR is build on top of my "Accepted Codecs"-PR ( #6460 ) . 
With both features, the admin will have much more control about which videos have to be transcoded and thus allow for much better storage efficiency.


server/src/infra/entities/system-config.entity.ts
```diff
@@ -105,6 +107,7 @@ export enum SystemConfigKey {
 export enum TranscodePolicy {
   ALL = 'all',
   OPTIMAL = 'optimal',
+  BY_BITRATE = 'by_bitrate',
```
mertalev, line 110: The value should be in camelCase. Could you make this `byBitrate` instead? Actually, you could just name this enum `BITRATE`.

server/src/domain/media/media.service.ts
```diff
@@ -330,7 +341,9 @@ export class MediaService {
     const allTargetsMatching = isTargetVideoCodec && isTargetAudioCodec && isTargetContainer;
     const scalingEnabled = ffmpegConfig.targetResolution !== 'original';
     const targetRes = Number.parseInt(ffmpegConfig.targetResolution);
+    const maxBitrate = Number.parseInt(ffmpegConfig.maxBitrate);
     const isLargerThanTargetRes = scalingEnabled && Math.min(videoStream.height, videoStream.width) > targetRes;
+    const isLargerThanTargetBitrate = bitrate >= maxBitrate;
```
mertalev, line 346: This should be `>` instead of `>=`. Also, this doesn't check the units. What unit does `results.format.bit_rate` use? You'd want to make sure the user input is converted to this. e.g. if it's in bps, but the user inputted in kbps ('k' at the end), you'd need to multiply the user input by 1000 to make the comparison.

mertalev (commented):
There might be a bug in the bitrate check, but otherwise looks good (only looking at the bitrate changes since the other changes are in #6460)

server/src/infra/entities/system-config.entity.ts
Hely0n: I wasn't sure, how those values will be interpreted/parsed by open-api etc, so I went the safe way here and just made it lowercase as the others. By I will change it today, thank you.

server/src/domain/media/media.service.ts
Hely0n: Okay, I will change it to >. 

I didn't know that the user can pass units. Thank you for pointing it out to me. I plan to fix everything later today.

server/src/infra/entities/system-config.entity.ts
Hely0n: Done

Hely0n: @mertalev Are there still any problems with this PR? 
I ask because the other PR was merged very quickly, while here is no activity even if the changes are simpler. 😅

mertalev: Sorry! I think you should rebase on main since the diff shows the accepted codec stuff that's already merged.

mertalev: You should also add unit tests for this.

mertalev:
@Hely0n I rebased the PR and it looks good. We can merge it once you add unit tests for the policy.

Hely0n:
> @Hely0n I rebased the PR and it looks good. We can merge it once you add unit tests for the policy.

Thank you very much. I plan to write them in the next days

Hely0n:
@mertalev I hope this unit test meets your expectations. I just oriented on the resolution policy.

mertalev (approved):

PR #6485, opened 2024-01-19
fix(web): slider

jrasm91:


danieldietzler (approved):
LGTM! Would be nice if we could move the label with the title to the slider. So `title` would be a prop on the slider component

martabal (approved):

PR #6494, opened 2024-01-19
feat: add shortcuts to asset-grid

martabal:
## Changes made in this PR

This PR adds the ability to use shortcuts on the asset grid such as `Archive`, `Favorite` and `Download`. It adds a new shortcut: Select All with `ctrl + a`.

Fixes an issue where a non-owner user can archive an asset from a shared album resulting in a server error

## Screenshots

https://github.com/immich-app/immich/assets/74269598/6bae2fe5-1a50-4a54-8591-0baa7524747b


jrasm91 (changes requested):
This can't possibly work, because the actions are also used on non asset-grid pages.

PR #6514, opened 2024-01-19
fix(web): delete user

### LINKED ISSUE #6513: Deleting a user
### The bug

when deleting a user after pressing delete again to confirm, the popup stays open and you can only cancel
if you cancel after confirming the delete, the user is still in a delete stage

Ubuntu 20.04

### Version of Immich Server

v1.93.1

1.93.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
-
```

### Your .env content

```Shell
-
```

### Reproduction steps

```bash
1. create a user (with oauth in my case)
2. delete the user
3. click delete again to confirm
...
```

### Additional information

_No response_

martabal:
fixes #6513

### DIFF
--- a/web/src/routes/admin/user-management/+page.svelte
+++ b/web/src/routes/admin/user-management/+page.svelte
@@ -130,7 +130,7 @@
       {#if shouldShowDeleteConfirmDialog}
         <DeleteConfirmDialog
           user={selectedUser}
-          on:succes={onUserDeleteSuccess}
+          on:success={onUserDeleteSuccess}
           on:fail={onUserDeleteFail}
           on:cancel={() => (shouldShowDeleteConfirmDialog = false)}
         />


danieldietzler (approved):
Lol what a stupid typo :D
Thanks!

martabal:
It didn't take too long to figure out what the problem was :laughing: 

I think this isn't the first time we've run into this kind of issue with the dispatcher, I don't know if svelte has any built in features to limit this.

jrasm91 (commented):
Why is this not throwing an error? We're using typed dispatchers so this should be validated, no?

martabal:
> Why is this not throwing an error? We're using typed dispatchers so this should be validated, no?

No, it does prevent that error

jrasm91:
Ah, maybe it would only prevent it if you used a custom event with a type that mismatched.

PR #6512, opened 2024-01-19
fix(web): album description

martabal:
## Changes made in this PR

- fixes #6510
- fixes sytle issues on the activity viewer

## Screenshots

| Before | After |
| :----: | :---: |
| Screenshot from 2024-01-19 20-25-30 | image  |

https://github.com/immich-app/immich/assets/74269598/4280b858-3e96-4fd9-8d5b-ec00e807d9d5

PR #6517, opened 2024-01-19
refactor(web): descriptions

martabal:
## What's changed

- Edit the album description directly in the textarea like the description in the detail panel
- Refactor auto-grow textarea as suggested by https://github.com/immich-app/immich/pull/6512#discussion_r1459824744
- Keep the break lines in the description when viewing the asset with a shared link as a guest
- Avoid useless API requests when clicking on the description in the detail panel when description has been focused out but not changed
- Fix the escape shortcut on the album page when opening some modals
- Update the description by pressing `Ctrl + Enter`

## Screenshots

https://github.com/immich-app/immich/assets/74269598/40fd4b96-26da-49af-a3b2-288aac7ad6be


danieldietzler (approved):

web/src/lib/components/asset-viewer/detail-panel.svelte
```diff
@@ -197,20 +200,20 @@
   {/if}
 
   <section class="mx-4 mt-10" style:display={!isOwner && description === '' ? 'none' : 'block'}>
-    {#if !isOwner || api.isSharedLink}
-      <span class="break-words">{description}</span>
-    {:else}
```
jrasm91, line 202: Why did you get rid of this span and convert it back to a placeholder? Did you test a multi-line description in a shared link, specifically going from short to long to see if it auto grows?

martabal: Yes it works fine.

Original asset:

Screenshot from 2024-01-20 02-44-51

| Before | After |
| :---: | :---: |
| Screenshot from 2024-01-20 02-48-16 | Screenshot from 2024-01-20 02-48-29 |

jrasm91 (dismissed):
This doesn't work and re-introduces back an auto grow bug. In this video I click "next" to go to another asset with a longer description. The textarea does not auto-grow until I start typing.

[Screencast from 01-19-2024 09:08:42 PM.webm](https://github.com/immich-app/immich/assets/4334196/ab87d62e-5a94-401e-82af-d1b7f7893f64)

martabal:
> This doesn't work and re-introduces back an auto grow bug. In this video I click "next" to go to another asset with a longer description. The textarea does not auto-grow until I start typing.

Oh right, I see what's the issue. Should be fixed. I removed `overflow-hidden` too so you can scroll the *really* large descriptions.

alextran1502 (commented):
Hello, thank you for your work on this. It works very well.

One bug I notice is that you cannot see the description of the non-owned asset in a shared album.

martabal:
> Hello, thank you for your work on this. It works very well.
> 
> One bug I notice is that you cannot see the description of the non-owned asset in a shared album.

Hello, thank you for testing it !
I can't reproduce this behavior. Can you precise how you discovered this bug ?

| Asset owned by the user | Asset not owned by the user |
| :----: | :---: |
| Screenshot from 2024-01-22 00-01-24 | Screenshot from 2024-01-22 00-01-31 |

alextran1502:
@martabal Here is what I see. Probably a bug with a very long description

image

martabal:
> @martabal Here is what I see. Probably a bug with a very long description

Really weird, I can't reproduce it. I tested all kind of descriptions (none, long, one line) with assets in shared albums, viewed from the owner, a non-owner and a guest and I can't reproduce this behavior :thinking: 

I changed a bit that logic in my latest commit, can you try it ?

alextran1502:
@martabal the description is still showing up as a single line if you share the album in public shared link

martabal:
Right, I forgot to update the description in the `share/+page.svelte` :sweat: 

Thanks for the info

PR #6525, opened 2024-01-20
feat(server, web): smart search filtering and pagination

mertalev:
## Description

This PR uses the new pgvecto.rs VBASE feature to implement paginated smart search. It updates the web app to take advantage of this for infinite scroll. Additionally, it refactors the custom search builder to be more reusable, integrating it into smart search. This paves the way for filtered hybrid search in the future.

Summary of changes:
- Rename smart info repository to search repository
- Move asset search method to search repository
- Make smart search repository method paginated and share the same builder (with some exclusions)
- Update web to query for the next page after reaching the bottom of the page (specifically when the second-to-last row is in view)
- Modularize typing for search builder
- Add new smart search DTO that includes most of the filters for the search builder
- Split current `/search` endpoint into `/search/metadata` and `/search/smart`
  - There's enough difference in what the two endpoints can accept that separating them makes validation etc. simpler
- Mark current search endpoint as deprecated
  - This is because the new search filtering logic isn't ready yet, and because making all of the necessary changes to actually switch everything to the new endpoints increases the scope of what is already a very large PR

## How Has This Been Tested?

Tested with both pgvecto.rs and pgvector. The latter can still somewhat benefit from this change, but with a maximum of 1000 assets that will effectively be lower in the case of filters or multiple users.

https://github.com/immich-app/immich/assets/101130780/d2b71a06-b5c6-4104-a1b4-4c1eaa91b9b6


server/src/infra/infra.utils.ts
```diff
@@ -91,3 +121,79 @@ export function ChunkedArray(options?: { paramIndex?: number }): MethodDecorator
...
+  const exifInfo = _.omitBy(_.pick(options, ['city', 'country', 'lensModel', 'make', 'model', 'state']), _.isUndefined);
+  if (Object.keys(exifInfo).length > 0) {
+    builder.leftJoin(`${builder.alias}.exifInfo`, 'exifInfo');
+    builder.andWhere({ exifInfo });
+  }
+
+  const id = _.pick(options, ['checksum', 'deviceAssetId', 'deviceId', 'id', 'libraryId', 'ownerId']);
+  builder.andWhere(_.omitBy(id, _.isUndefined));
+
+  const path = _.pick(options, ['encodedVideoPath', 'originalFileName', 'originalPath', 'resizePath', 'webpPath']);
```
danieldietzler, line 150: I guess we could technically type those array, e.g. `Array<keyof SearchPathOptions>`. Not sure if it's worth it though

web/src/lib/components/shared-components/gallery-viewer/gallery-viewer.svelte
```diff
@@ -88,7 +87,22 @@
...
+          format={assets.length < 7 ? ThumbnailFormat.Jpeg : ThumbnailFormat.Webp}
+          on:click={(e) => (isMultiSelectionMode ? selectAssetHandler(e) : viewAssetHandler(e))}
+          on:select={selectAssetHandler}
+          selected={selectedAssets.has(asset)}
+          {showArchiveIcon}
+        />
+      </div>
+    {/each}
+
+    {#each assets.slice(-1) as asset (asset.id)}
```
danieldietzler, line 105: Why do we need a for each here? This will always be just one asset, right?

web/src/routes/(user)/search/+page.svelte
```diff
@@ -107,6 +110,36 @@
   const handleSelectAll = () => {
     selectedAssets = new Set(searchResultAssets);
   };
+
+  export const loadNextPage = async () => {
+    console.log('loadPage', curPage, term);
```
danieldietzler, line 115: There are three console log left here which I presume need to get removed

danieldietzler (approved):
Really good!

server/src/domain/search/search.service.ts
```diff
@@ -55,6 +55,54 @@ export class SearchService {
...
+        orderDirection: dto.order ? enumToOrder[dto.order] : 'DESC',
+      },
+    );
+
+    return this.mapResponse(items, hasNextPage ? (page + 1).toString() : null);
+  }
+
+  async searchSmart(auth: AuthDto, dto: SmartSearchDto): Promise<SearchResponseDto> {
+    await this.configCore.requireFeature(FeatureFlag.SEARCH);
+    await this.configCore.requireFeature(FeatureFlag.SMART_SEARCH);
```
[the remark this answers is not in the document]
mertalev: Oooh that's a good point

server/src/infra/infra.utils.ts
mertalev: I typed them like that earlier. The only reason I removed that is to keep them on one line haha.

web/src/lib/components/shared-components/gallery-viewer/gallery-viewer.svelte
mertalev: Hmmm, good question. I probably just adapted the code above it. I'll change this to an `#if` block and check for a non-empty array for clarity.

server/src/domain/search/dto/search.dto.ts
```diff
@@ -1,8 +1,183 @@
...
+import { IsBoolean, IsEnum, IsInt, IsNotEmpty, IsString, Max, Min } from 'class-validator';
+import { Optional, QueryBoolean, QueryDate, ValidateUUID, toBoolean } from '../../domain.util';
 
+class BaseSearchDto {
+  @ValidateUUID({ optional: true })
+  libraryId?: string;
+
+  @IsString()
+  @Optional()
+  deviceId?: string;
```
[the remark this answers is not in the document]
mertalev: Oops!

web/src/routes/(user)/search/+page.svelte
mertalev: Oops

server/src/domain/search/search.service.ts
```diff
@@ -55,6 +55,54 @@ export class SearchService {
     }));
   }
 
+  async searchMetadata(auth: AuthDto, dto: MetadataSearchDto): Promise<SearchResponseDto> {
+    let checksum: Buffer | undefined;
```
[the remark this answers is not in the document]
mertalev: I actually don't think that syntax works since it's a variable and not just a type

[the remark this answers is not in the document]
danieldietzler: Oh true, nvm that then. Sorry

web/src/lib/components/shared-components/gallery-viewer/gallery-viewer.svelte
mertalev: Oh, now I remember why I did this. It's because `animate:flip` seems to require `each`: `An element that uses the animate directive must be the immediate child of a keyed each block`. Should I remove that flag?

danieldietzler: I have no idea what that does lol. So up to you or someone else haha

mertalev: I don't know either so I think I'll just leave it alone for now lol

mertalev: I realized I can actually get the index of an asset in the each block and just conditionally fire the intersect event based on that.

alextran1502:
Thank you for the high quality PR.

Can you explain briefly how you achieve the infinite scrolling, especially on the performance aspect? Will it cause a performance issue if you keep scrolling since the DOM will continue to add more `div` as it is scrolled?

mertalev:
> Thank you for the high quality PR.
> 
> Can you explain briefly how you achieve the infinite scrolling, especially on the performance aspect? Will it cause a performance issue if you keep scrolling since the DOM will continue to add more `div` as it is scrolled?

Existing assets aren't re-rendered when a new page of assets are added, so I don't think this slows things down much besides likely increasing RAM usage. We also only append to the list of assets - this avoids overhead of copying or traversing this list as it gets longer.

There's definitely a point where it will get slower, not just on the frontend but also because the vector search query will have to skip past more and more rows for each page. But in an earlier version, I scrolled to about 50-60 pages deep (i.e. 5-6k assets) and didn't notice a degradation, so I'm not sure what the threshold is where it would start getting slower. It might be different for a client with less RAM, on a different browser (I tested on Chrome), etc.

mertalev:
I can confirm it starts degrading for me at 7k+ assets. The scrolling starts getting jittery, the next page takes longer to load and the thumbnails don't show up as quickly. Continuously scrolling without ever stopping, RAM usage climbed to 1.5gb at 10k assets. But after leaving it alone to type this, it dropped down to around 650mb. Lastly, changing the width of the page with so many assets is smooth *if* it doesn't result in the size of the thumbnails changing. If it does, the thumbnails have to get re-rendered and this absolutely chokes the page.

mertalev:
I think we can set a 50 page limit to prevent the number of assets from getting out of hand.

web/src/routes/(user)/search/+page.svelte
```diff
@@ -27,15 +26,20 @@
   import { preventRaceConditionSearchBar } from '$lib/stores/search.store';
   import { shouldIgnoreShortcut } from '$lib/utils/shortcut';
   import { mdiArrowLeft, mdiDotsVertical, mdiImageOffOutline, mdiPlus, mdiSelectAll } from '@mdi/js';
+  import type { AssetResponseDto, SearchResponseDto } from '@immich/sdk';
+  import { authenticate } from '$lib/utils/auth';
+  import { api } from '../../../api/api';
```
jrasm91, line 31: I think you can just import from `@api`

jrasm91 (approved):
Code looks fine AFAICT. We just can't seem to be able to decide if it'll live in `search/` or `asset/` lol

PR #6543, opened 2024-01-21
feat(mobile): play motion video with long press gesture support

jzhangdev:
Supports long-press gestures to play motion videos, which can make viewing motion videos more convenient.


shenlong-tanwen:
Thanks a lot for working on this. What do you think about updating the handling to stop the motion photo when the long press event ends? i.e, play it as long as the user is pressing the screen?

zackpollard:
> Thanks a lot for working on this. What do you think about updating the handling to stop the motion photo when the long press event ends? i.e, play it as long as the user is pressing the screen?

imo the live photos are so short that it's not really necessary, I have tested this as-is and it works nicely.

zackpollard (approved):
LGTM

PR #6556, opened 2024-01-21
fix(server): only calculate quota usage for internal assets

danieldietzler:


jrasm91 (approved):
LGTM

alextran1502 (approved):

aviv926:
Thanks for the fix, it still shows an unreasonable amount of percentages in Server Stats
is this on purpose?
image

PR #6578, opened 2024-01-22
chore(server): remove old device id endpoint

jrasm91:
> [!WARNING]
> Mobile apps connecting to server versions prior to 1.92 will no longer be able to use the full device id sync endpoint

- We added a new device id endpoint (#5273) in v1.89
- We started using the new endpoint (#6108) in v1.92
- We removed it in this PR, which can be included in v1.94.


alextran1502 (approved):

danieldietzler (approved):
nice nice :)

PR #6581, opened 2024-01-22
fix(web): statusbox re-rendering and nav bar when trashing assets

### LINKED ISSUE #6570: [BUG] web: selection still shown after trash delete
### The bug

There is still shown upper selection bar after all selected items are deleted from trash using Delete button.
It keeps showing everywhere, where selection bar is, e.g. Photos

obrazek

Fedora 39

### Version of Immich Server

v1.93.3

not relevant

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
not releavant
```

### Your .env content

```Shell
not releavant
```

### Reproduction steps

```bash
see above
```

### Additional information

_No response_

martabal:
- fixes #6570 
- avoids unnecessary re-rendering of storage statistics in the `StatusBox` each time a page loads.

### DIFF
--- a/web/src/lib/components/shared-components/status-box.svelte
+++ b/web/src/lib/components/shared-components/status-box.svelte
@@ -48,10 +48,6 @@
         const { data } = await api.serverInfoApi.getServerInfo();
         $serverInfoStore = data;
       }
-      if (!$user) {
-        const { data } = await api.userApi.getMyUserInfo();
-        $user = data;
-      }
     } catch (e) {
       console.log('Error [StatusBox] [onMount]');
     }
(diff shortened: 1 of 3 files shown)


web/src/lib/components/shared-components/status-box.svelte
```diff
@@ -36,18 +36,22 @@
...
-        api.userApi.getMyUserInfo().then(({ data }) => data),
-      ]);
+      if (!$serverInfoStore) {
+        const { data } = await api.serverInfoApi.getServerInfo();
+        $serverInfoStore = data;
+      }
+      if (!$user) {
+        const { data } = await api.userApi.getMyUserInfo();
+        $user = data;
+      }
```
danieldietzler, line 54: Shouldn't this be handled by the store?

martabal: Yep, that AIP request is useless, correct me if I'm wrong, but you're always authenticated when you have the side-bar and that store is already handled by `auth.ts`

jrasm91:
This description stuff keeps popping up, can you keep it out of this PR, since it seems completely unrelated?

jrasm91 (approved):

PR #6604, opened 2024-01-23
docs: diff highlighting

jrasm91:
Add prism highlighting for diffs.

| Before | After |
| - | - |
| image | image |


bo0tzz (approved):
While you're in there, want to add `bash` as well? I think that's not enabled by default.

PR #6617, opened 2024-01-24
feat(mobile): Added support for BasicAuth via userinfo in server url

jpg0:
Added support for BasicAuth protected immich servers to mobile apps.

Allows specifying credentials in `https://user:pass@host` format in server URL. They will be sent in the `Authorization` header for all requests - in this case moving the bearer token auth to the (already supported) `x-immich-user-token` (it's not moved when BasicAuth is not used). I have added a note in the docs, and tested using Traefik.

### DIFF
--- a/docs/docs/partials/_mobile-app-login.md
+++ b/docs/docs/partials/_mobile-app-login.md
@@ -1,3 +1,5 @@
 Login to the mobile app with the server endpoint URL at `http://<machine-ip-address>:2283/api`
 
 <img src={require('./img/sign-in-phone.jpeg').default} width='50%' title='Mobile App Sign In' />
+
+Note that if the Immich server is hosted behind a proxy requiring Basic Authentication, this is supported using a server endpoint URL in the format `http://username:password@<machine-ip-address>:2283/api`
--- a/mobile/lib/shared/services/api.service.dart
+++ b/mobile/lib/shared/services/api.service.dart
@@ -35,9 +35,13 @@ class ApiService {
 
   setEndpoint(String endpoint) {
     _apiClient = ApiClient(basePath: endpoint);
+
+    getAuthHeaders(endpoint, null).forEach((k, v) => _apiClient.addDefaultHeader(k, v));
+    
     if (_authToken != null) {
       setAccessToken(_authToken!);
     }
+
     userApi = UserApi(_apiClient);
     authenticationApi = AuthenticationApi(_apiClient);
     oAuthApi = OAuthApi(_apiClient);
@@ -95,9 +99,7 @@ class ApiService {
     // we do not care if the endpoints hits an HTTP error
     try {
       await client
-          .get(
-            Uri.parse(serverUrl),
-          )
+          .get(Uri.parse(serverUrl), headers: getAuthHeaders(serverUrl, null))
           .timeout(const Duration(seconds: 5));
     } on TimeoutException catch (_) {
       return false;
@@ -111,10 +113,12 @@ class ApiService {
     final Client client = Client();
 
     try {
-      final res = await client.get(
-        Uri.parse("$baseUrl/.well-known/immich"),
-        headers: {"Accept": "application/json"},
-      );
+      String url = "$baseUrl/.well-known/immich";
+
+      Map<String, String> headers = getAuthHeaders(url, null);
+      headers["Accept"] = "application/json";
+
+      final res = await client.get(Uri.parse(url), headers: headers);
 
       if (res.statusCode == 200) {
         final data = jsonDecode(res.body);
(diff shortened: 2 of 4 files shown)


alextran1502:
what is the use case of this?

jpg0:
This is to allow deployment of immich behind a proxy implementing/enforcing
BasicAuth. My personal use-case is that I host Immich at home and want to
use the mobile app when I'm not at home, but want to follow the advice of
not opening a developing product up to the internet (and also not having to
regularly toggle a VPN on my phone). This means assuming that I trust
BasicAuth at the proxy level, no security hole in Immich can be exploited
if I open it up.

There is also a discussion thread on exactly this use-case; I'll link it
here...

jpg0:
Existing discussion: https://github.com/immich-app/immich/discussions/3204

alextran1502:
So it is currently working with the web app but not the mobile app, correct?

jpg0:
Yes. With the web app the browser handles this automatically.

rovo89:
I would really appreciate if Basic Auth was working. I have various services running on my home server, exposed via nginx with mTLS. This way, I have an extra layer of protection in case of application bugs (e.g. forgotten authorization check for an endpoint), because attackers would _additionally_ need my certificate (or a 0-day in nginx). Otherwise none of the resources are reachable.

There's a discussion saying that mTLS will probably never be supported in the app because the underlying framework doesn't support it. Fair enough.

My next attempt was Basic Auth, but while the initial connection attempt (like fetching server config) used the username/password, further requests added the Bearer token and that overwrote Basic Auth.

I currently use a separate domain with a secret path prefix (like https://app.photos.example.org/xyz). I disabled mTLS for that domain and forward only /xyz to Immich. Since the path is only transmitted within TLS, that should be save (as long as it's a bit more complex than xyz). But obviously, it's a hack and also doesn't work in the browser (which requires Immich to be served from top level).

So I'll gladly test this PR with nginx. Is there any automatically built APK?

jpg0:
I don't believe that there is an official APK built automatically (well, certainly not until the workflows are run), but I have the one that I have built and am now using. It is not using the official signing key however (as I don't have that), so you'd need to uninstall the official Immich before installing this one.

It's too big to attach here, but you can get it from my Google drive: https://drive.google.com/file/d/14t3WGoRfSqpSz7-M8JsC2sm8M2ucPf1r/view?usp=sharing

bo0tzz:
> the initial connection attempt (like fetching server config) used the username/password, further requests added the Bearer token and that overwrote Basic Auth.

This sounds to me like basic auth already 'just works', except we end up overwriting the auth header with the Immich token later on. If that is the case, it should be sufficient to just always put the Immich token in `x-immich-user-token` (a simple one line change) without needing any sort of "does this URL contain basic auth stuff" checks.

bo0tzz:
> there is no BasicAuth headers actually being sent to the server.

Then I misunderstood. I expected that whatever http client we're using already supported basic auth under the hood but we were just overwriting the headers.

rovo89:
From my nginx logs:
```
192.168.2.222 - x [15/Jan/2024:17:03:45 +0100] "GET /api HTTP/1.1" 404 66 "-" "Dart/3.1 (dart:io)" "-"
192.168.2.222 - x [15/Jan/2024:17:03:45 +0100] "GET /.well-known/immich HTTP/1.1" 200 27 "-" "Dart/3.1 (dart:io)" "-"
192.168.2.222 - - [15/Jan/2024:17:03:45 +0100] "GET /api/server-info/version HTTP/1.1" 200 32 "-" "Dart/3.1 (dart:io)" "-"
192.168.2.222 - - [15/Jan/2024:17:03:45 +0100] "GET /api/server-info/features HTTP/1.1" 200 199 "-" "Dart/3.1 (dart:io)" "-"
192.168.2.222 - - [15/Jan/2024:17:03:45 +0100] "GET /api/server-info/config HTTP/1.1" 200 162 "-" "Dart/3.1 (dart:io)" "-"
```

In that example I used something like `https://x:y@example.org` as server URL in the Android app.
Maybe the credentials get lost when constructing the API URL from the well-known manifest? Not sure if the server version requires authentication and if the token would already be sent on that request. We might have two (small?) issues preventing Basic Auth.

jpg0:
All the URLs in your log don't require authentication. Also if you set a URL with a user:pass at the front then typically the app just throws them away, so what you are seeing makes sense to me for a typical Immich install.

However I thought that you would be routing everything through a reverse-proxy and requiring basic auth on all requests - if so then I don't understand what is going on there as I don't believe that the basic auth headers would be sent.

rovo89:
I pushed an alternative implementation in #6840 which "simply" changes all headers from `Authorization: Bearer` to `x-immich-user-token` (in more than a dozen places...). Please take a look @jpg0 @jrasm91.

jpg0:
Looks like i missed a few places! I'm certainly happy with switching the token header entirely - I only left it as-is for non Basic to try to change as little as possible, but I agree a single approach is better.

Only one thing I have a question on which is the client sending basic auth headers - i'll comment on your PR though.

alextran1502:
superseded by #6840

PR #6632, opened 2024-01-24
refactor(server): e2e

jrasm91:
- Consolidate e2e clients
- Remove unused env variables for "run all tests"
- Remove unused ifit usages
- Update docs for split e2e tests
- Removed unused job options for test app init, since all these tests are for jobs


danieldietzler (commented):
Good stuff! Maybe it would be nice to have something like `@api-test` in server instead of always having to type the same relative path?

danieldietzler (approved):
LGTM!

PR #6634, opened 2024-01-25
feat(server): optimize partial facial recognition

mertalev:
## Description

Based on the discussion [here](https://github.com/immich-app/immich/discussions/6571), this PR changes facial recognition queuing:

* Face detection will not queue facial recognition if no faces are detected
* Only detected faces are queued, not all missing faces
* A nightly job is added to do more thorough recognition
* Manually running "missing" still does the more thorough recognition

Specifically, this is the new behavior for clustering:
* If the face is a core face (i.e. there are min faces - 1 faces around it) and a person exists for it, clustering works as normal
* If the face is a core face and there's no similar face with a person, a person is created, but the other faces aren't clustered unless they're in the same batch of new faces
* If the face is a non-core face and there's a similar face with a person, clustering works as normal
* If the face is a non-core face and there's no similar face with a person, then it won't be clustered

There is an edge case where this change can lead to a duplicate person: if a new person is created in the partial clustering and the faces similar to it existed before this batch of faces, then only the core face that created the person will be assigned this person until the nightly job runs. This means that if another image containing a face of this person is uploaded before the nightly job, then it will only be clustered if it can reach the face that created the person, not just any face of that person. If it can't and the face is a core face, a new person will be created.

Implementation notes:
* The queue facial recognition job is added before adding the individual recognition jobs as a gatekeeper to prevent the jobs from running until all queued detection jobs are finished
* The recognition jobs are queued within the detection handler instead of in `onDone` to support the more nuanced queueing
* Bulk upload should work the same, just with a different queueing method

## How Has This Been Tested?

Tested with debug logging to check that it queues the faces from Face Detection, not all missing faces.

### DIFF
--- a/server/src/domain/person/person.service.ts
+++ b/server/src/domain/person/person.service.ts
@@ -334,28 +334,26 @@ export class PersonService {
 
     if (faces.length) {
       await this.jobRepository.queue({ name: JobName.QUEUE_FACIAL_RECOGNITION, data: { force: false } });
+
+      const createFaces = [];
+      for (const face of faces) {
+        const mappedFace = {
+          assetId: asset.id,
+          embedding: face.embedding,
+          imageHeight: face.imageHeight,
+          imageWidth: face.imageWidth,
+          boundingBoxX1: face.boundingBox.x1,
+          boundingBoxX2: face.boundingBox.x2,
+          boundingBoxY1: face.boundingBox.y1,
+          boundingBoxY2: face.boundingBox.y2,
+        };
+
+        createFaces.push(this.repository.createFace(mappedFace));
+      }
+
+      const faceIds = await Promise.all(createFaces);
+      await this.jobRepository.queueAll(faceIds.map((id) => ({ name: JobName.FACIAL_RECOGNITION, data: { id } })));
     }
-
-    const promises = [];
-    for (const face of faces) {
-      const mappedFace = {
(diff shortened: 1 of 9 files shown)


mertalev:
No clue why those mobile tests are failing

server/src/domain/person/person.service.ts
```diff
@@ -332,19 +332,26 @@ export class PersonService {
...
+          embedding: face.embedding,
+          imageHeight: face.imageHeight,
+          imageWidth: face.imageWidth,
+          boundingBoxX1: face.boundingBox.x1,
+          boundingBoxX2: face.boundingBox.x2,
+          boundingBoxY1: face.boundingBox.y1,
+          boundingBoxY2: face.boundingBox.y2,
+        };
+
+        return this.repository.createFace(mappedFace);
```
jrasm91, line 350: Hmm, this is creating faces one at a time, but now you have changed it to run in parallel. It might be worth changing this to use something like `this.repository.createFaces` and pass in a list of faces instead. I know typeorm has some bulk insert methods (or save works in bulk as well).

jrasm91 (approved):

server/src/domain/person/person.service.ts
mertalev: Oh right, good point.

PR #6635, opened 2024-01-25
feat(server): sql access checks

jrasm91:
Refactor the access repository and sql generator to make it possible to generate sql for all the access checks.


mertalev (approved):
This is really nice. It should also let us use `@ChunkedSet` for these methods now (but this can be done in a follow-up PR)

PR #6636, opened 2024-01-25
refactor(server): move asset detail endpoint to new controller

jrasm91:
> [!NOTE]
> `/asset/assetById/:id` is deprecated and will be removed in a future release

- Rename api-v1/asset/asset-repository to `*V1` everywhere
- Add a new `/asset/:id` endpoint (in the new controller)
- Move the existing service method to the new service
- Point the old method at the new service
- Delete the old repository implementation and use the new one instead
- Added e2e tests for the asset detail endpoint

Note: A lot of relations where loaded for `serveFile`, but now it should just load the asset, so that endpoint should be faster.

### DIFF
--- a/server/src/domain/asset/asset.service.ts
+++ b/server/src/domain/asset/asset.service.ts
@@ -419,21 +419,21 @@ export class AssetService {
       throw new BadRequestException('Asset not found');
     }
 
-    if (!auth.sharedLink || auth.sharedLink?.showExif) {
-      const data = mapAsset(asset, { withStack: true });
-
-      if (data.ownerId !== auth.user.id) {
-        data.people = [];
-      }
-
-      if (auth.sharedLink) {
-        delete data.owner;
-      }
-
-      return data;
-    } else {
+    if (auth.sharedLink && !auth.sharedLink.showExif) {
       return mapAsset(asset, { stripMetadata: true, withStack: true });
     }
+
+    const data = mapAsset(asset, { withStack: true });
+
+    if (auth.sharedLink) {
+      delete data.owner;
+    }
+
+    if (data.ownerId !== auth.user.id) {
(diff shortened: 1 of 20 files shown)


web/src/lib/components/asset-viewer/detail-panel.svelte
```diff
@@ -234,7 +234,7 @@
       {/key}
     </section>
   {:else if description}
-    <p class="break-words whitespace-pre-line w-full text-black dark:text-white text-base">{description}</p>
+    <p class="px-4 break-words whitespace-pre-line w-full text-black dark:text-white text-base">{description}</p>
```
jrasm91, line 237: This seemed like a regression that I fixed here while testing the asset detail endpoint. cc: @martabal

server/src/domain/asset/asset.service.ts
```diff
@@ -400,6 +398,44 @@ export class AssetService {
...
+      }
+
+      if (auth.sharedLink) {
+        delete data.owner;
+      }
+
+      return data;
+    } else {
+      return mapAsset(asset, { stripMetadata: true, withStack: true });
+    }
```
danieldietzler, line 436: Can we flip this? So like `if (auth.SharedLink && !auth.sharedLink?.showExif) {return mapAsset(asset, { stripMetadata: true, withStack: true });}` and then all the other stuff below without an else. I prefer having as little deep code as possible

danieldietzler (approved):
Nice! Making progress in getting rid of api-v1 is always great :)

server/src/domain/asset/asset.service.ts
jrasm91: Yeah, that makes sense. I copied it as is from the other one, but I like that approach better.

PR #6646, opened 2024-01-25
fix(ml): error logging

mertalev:
## Description

The ML service's logging configuration is buggy. It only logs `log.error` calls, not actual exceptions. This led to misleading logs that suggested that it e.g. loads models several times, when in fact it has just been failing to load them and didn't print the exception.

I also tweaked the traceback output to suppress frames from FastAPI etc. The traceback ends up being very long and noisy with irrelevant frames. It still shows the path and line for those, but doesn't show the code.

### DIFF
--- a/machine-learning/app/config.py
+++ b/machine-learning/app/config.py
@@ -91,7 +91,7 @@ class CustomRichHandler(RichHandler):
                 if any(excluded in tb.tb_frame.f_code.co_filename for excluded in self.excluded):
                     tb.tb_frame.f_locals["_rich_traceback_omit"] = True
                 tb = tb.tb_next
-            
+
         return super().emit(record)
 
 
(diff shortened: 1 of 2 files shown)


jrasm91 (approved):
Were you able to produce a 500 and see it correctly log and pass the error the the immich-server/microservices?

mertalev:
I produced a 500 error by adding a runtime exception and requesting from Postman.

excluded-traceback

mertalev:
Also for reference, this is what it looks like without the traceback hack to exclude fastapi etc.
suppressed-traceback

jrasm91:
Yo that stack trace looks fancy haha

mertalev:
lol I can even get it to print local variables and stuff

PR #6649, opened 2024-01-26
chore: remove no-op patch

benmccann:
If I remove this patch, the same output is generated. The presence of this patch prevents an upgrade of the OpenAPI Generator. There are other issues preventing an upgrade as well, but this would address one of them at least


jrasm91 (commented):
That looks to be true, however, we can actually remove the templates/typescript directory, all apiInner* files, and update the bash script to not download the file in the first place (or use the template directory). Do you mind removing all of those changes as well?

benmccann:
Excellent! That really simplified things. I've made the suggested changes

jrasm91 (approved):

PR #6650, opened 2024-01-26
feat(ml): conditionally download .armnn models

mertalev:
## Description

Currently, the entire model repo is downloaded for a given model. This works well when the models are all  in ONNX format, but the addition of ARM NN makes this inefficient as an extra model is downloaded. This PR changes this to conditionally download based on whether ARM NN is available in the service. The model paths are also adapted to use the preferred model path, ensuring a service with ARM NN tries to download the ARM NN model even if an ONNX version already exists.

### DIFF
--- a/machine-learning/app/models/base.py
+++ b/machine-learning/app/models/base.py
@@ -240,7 +240,7 @@ class InferenceModel(ABC):
 
     @property
     def preferred_runtime_default(self) -> ModelRuntime:
-        return ModelRuntime.ONNX if ann.ann.is_available and settings.ann else ModelRuntime.ARMNN
+        return ModelRuntime.ARMNN if ann.ann.is_available and settings.ann else ModelRuntime.ONNX
 
 
 # HF deep copies configs, so we need to make session options picklable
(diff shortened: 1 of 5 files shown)


machine-learning/app/models/base.py
```diff
@@ -215,6 +229,19 @@ def sess_options_default(self) -> ort.SessionOptions:
...
+        return self._preferred_runtime
+
+    @preferred_runtime.setter
+    def preferred_runtime(self, preferred_runtime: ModelRuntime) -> None:
+        log.debug(f"Setting preferred runtime to {preferred_runtime}")
+        self._preferred_runtime = preferred_runtime
+
+    @property
+    def preferred_runtime_default(self) -> ModelRuntime:
+        return ModelRuntime.ONNX if ann.ann.is_available and settings.ann else ModelRuntime.ARMNN
```
fyfrey, line 243: This should be the other way around?

mertalev: Oops, good catch

fyfrey (approved):
LGTM

PR #6652, opened 2024-01-26
refactor(server): trash endpoints

jrasm91:
> [!Warning] 
> ### Deprecated endpoints
> The three asset trash and restore endpoints have been deprecated and will be removed in a future release

```
POST /asset/trash/empty   => POST /trash/empty
POST /asset/trash/restore => POST /trash/restore
POST /asset/restore       => POST /trash/restore/assets
```

Move trash endpoints out of the asset controller and into a new trash service.

### DIFF
--- a/server/src/domain/trash/trash.service.ts
+++ b/server/src/domain/trash/trash.service.ts
@@ -1,0 +1,65 @@
+import { Inject } from '@nestjs/common';
+import { DateTime } from 'luxon';
+import { AccessCore, Permission } from '../access';
+import { BulkIdsDto } from '../asset';
+import { AuthDto } from '../auth';
+import { usePagination } from '../domain.util';
+import { JOBS_ASSET_PAGINATION_SIZE, JobName } from '../job';
+import {
+  ClientEvent,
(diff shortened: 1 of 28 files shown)


server/src/domain/trash/trash.service.ts
```diff
@@ -0,0 +1,65 @@
...
+    );
+
+    for await (const assets of assetPagination) {
+      await this.jobRepository.queueAll(
+        assets.map((asset) => ({ name: JobName.ASSET_DELETION, data: { id: asset.id } })),
+      );
+    }
+  }
+
+  private async restoreAndSend(auth: AuthDto, ids: string[]) {
```
danieldietzler, line 57: I personally find "send" unintuitive. I'd prefer "notify" or "inform" but idk

danieldietzler (approved):
LGTM

server/src/domain/trash/trash.service.ts
jrasm91: Yeah, it is just what the other method was called already. Maybe we can change it to notify later.

PR #6655, opened 2024-01-26
deps: open-api generator

jrasm91:
Upgrade open-api-generator to 7.2


etnoy:
Have tested this and seems to work fine

etnoy (approved):

shenlong-tanwen:
We can remove the following line as well since the latest generator has proper intl version in their template file

https://github.com/immich-app/immich/blob/2d0db5d0f25ed7c75821e9536c6321f5d1e2d48f/open-api/bin/generate-open-api.sh#L17

PR #6678, opened 2024-01-27
chore: migrate CLI unit tests to vitest

benmccann:
I think it'd be nice to switch to vitest for a few reasons:
- I find it to be much simpler. It doesn't require understanding how multiple tools like jest, ts-node, babel, etc. work together

This mostly converts the CLI e2e tests as well, but I ran into an issue there where I learned the e2e tests depend on the server project pretty heavily, so you basically have to convert the e2e tests from the cli and server projects together. Since it was turning into a bigger job, I thought I should check for feedback on the approach before proceeding further.


jrasm91:
Personally, I have had a lot of annoyances with jest and ESM, so I'm not against using vitest instead. 

Regarding the e2e tests, I don't really have strong opinion on any of that stuff, except for the fact that I don't think the cli should depend on the server at all. We could maybe share some scripts to setup a test environment, but it shouldn't be importing or using anything else from the server codebase. 

I'm open to suggestions for how to re-work anything preventing a migration to vitest. It is nice to have some automatic tests that verify the cli does communicate with the server correctly in an end to end fashion though, so I would like to be able to re-implement those checks one way or another. I would prefer to refactor/fix that setup prior to migrating to vitest.

benmccann:
The e2e CLI tests get `testApp` by importing `@test/../e2e/jobs/utils`, which in turn depends on `app.service.ts`, which looks to be the whole app. I think the main issue here is that the CLI tests depend on all the server source code in this way. You can see that in this PR where the old jest code is setting up `moduleNameMapper` and the setup of `resolve.alias` in the new vitest code. I guess the way to address that would be to first compile the server code and then have the CLI tests depend on this package rather than using these aliases to depend on the source code.

benmccann:
closing in favor of https://github.com/immich-app/immich/pull/6740. this one only had the unit tests converted while that one converts everything

PR #6690, opened 2024-01-27
chore: use relative import to find package.json

benmccann:
This import line keep causing me headaches as I try to convert things over to use vitest. I don't understand the current import syntax. There's no `src` directory inside the `domain` folder, which is what you'd expect this to be referring to. And if you treat it relative from the `server` directory `src/../../package.json` isn't a file that exists. Hopefully this relative syntax would be equivalent to what the current syntax is.

### DIFF
--- a/server/src/domain/domain.constant.ts
+++ b/server/src/domain/domain.constant.ts
@@ -1,7 +1,7 @@
 import { AssetType } from '@app/infra/entities';
 import { Duration } from 'luxon';
 import { extname, join } from 'node:path';
-import pkg from 'src/../../package.json';
+import pkg from '../../package.json';
 
 export const AUDIT_LOG_MAX_DURATION = Duration.fromObject({ days: 100 });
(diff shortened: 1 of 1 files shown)


jrasm91:
I don't fully understand this, so it is a bit hard to explain, but for whatever reason, it works. The end goal is simple though - `src/` and `dist/` need to exist with the same folder structure (at the same depth).

| Current | PR |
| - | - |
| image | image |

Your change makes it "import from two directories above the current file", this actuallys pulls in package.json into the output, which changes the depth of the rest of the output. This is problematic because relative paths _resolve differently_ between the source files and the dist files.

Originally, the import says "starting at the project root, navigate into src, then back up and then find package.json". I have no idea why "../.." is needed instead of just "../" but with it it works.

jrasm91:
I would also be fine just using `fs.readFileSync` if that accomplishes the same thing while avoiding the import headaches.

benmccann:
Hmm. That didn't work. But thanks for sharing the image above. That gives me something to work off of

PR #6692, opened 2024-01-27
chore(server): use fs readsync instead of js import for reading package.json

danieldietzler:
Picks up #6690.


jrasm91 (commented):
Let's use __dirname instead of cwd

danieldietzler:
@martabal will this work with the AIO image? Is the package.json right under the workdir there as well?

martabal:
Yep, just like upstream :)

jrasm91 (approved):

PR #6695, opened 2024-01-27
fix(ml): handle missing `context_length` field

mertalev:
## Description

A few models seem not to have `context_length` in their `config.json` file, so search will fail for these models. This PR sets a default of 77 in these cases.

## How Has This Been Tested?

Tested by running Smart Search and searching with `nllb-clip-large-siglip__v1` and `XLM-Roberta-Large-ViT-H-14__frozen_laion5b_s13b_b90k`.

### DIFF
--- a/machine-learning/app/config.py
+++ b/machine-learning/app/config.py
@@ -70,6 +70,8 @@ LOG_LEVELS: dict[str, int] = {
 settings = Settings()
 log_settings = LogSettings()
 
+LOG_LEVEL = LOG_LEVELS.get(log_settings.log_level.lower(), logging.INFO)
+
 
 class CustomRichHandler(RichHandler):
     def __init__(self) -> None:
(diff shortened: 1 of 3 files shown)


machine-learning/app/config.py
```diff
@@ -81,6 +83,7 @@ def __init__(self) -> None:
             console=console,
             rich_tracebacks=True,
             tracebacks_suppress=[*self.excluded, concurrent.futures],
+            tracebacks_show_locals=LOG_LEVEL == logging.DEBUG,
```
jrasm91, line 86: Is this change intentional or only for debugging?

jrasm91 (approved):

machine-learning/app/config.py
mertalev: Intentional since it makes debugging this sort of thing easier

PR #6696, opened 2024-01-28
refactor(server): drop module aliases

jrasm91:
Using modules aliases can be convenient, but also make the build more complex and often frustrating to deal with. This PR replaces modules aliases with root path imports. This has several benefits, including:
- More transparency around where imports are coming from
- Less required configuration to integrate with other tools (like jest)
- Simpler migration path to ESM

Essentially the changeset comes down to removing `tsconfig.paths` settings and `jestConfig.moduleNameMapper` configuration by updating the following imports:

Before:
```typescript
import {} from '@app/domain';
import {} from '@app/infra';
import {} from '@app/immich';
import {} from '@test';
```
After:
```typescript
import {} from 'src/domain';
import {} from 'src/infra';
import {} from 'src/immich';
import {} from 'test';
```

TLDR is this is simpler and looks almost the same as before.

### DIFF
--- a/server/e2e/api/jest-e2e.json
+++ b/server/e2e/api/jest-e2e.json
@@ -14,11 +14,5 @@
     "!<rootDir>/src/**/*.spec.(t|s)s",
     "!<rootDir>/src/infra/migrations/**"
   ],
-  "coverageDirectory": "./coverage",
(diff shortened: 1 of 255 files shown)


jrasm91:
I'll come back to this another time.

PR #6710, opened 2024-01-28
fix(server): do not process faces of deleted assets

### LINKED ISSUE #6709: Issue with facial recognition of deleted images
### The bug

This but is closely related to #1988 - deleted images cause an error during facial recogniton.

Error message:
```
[Nest] 8  - 01/28/2024, 7:44:38 PM   ERROR [JobService] Unable to run job handler (facialRecognition/facial-recognition): TypeError: Cannot read properties of null (reading 'ownerId')
[Nest] 8  - 01/28/2024, 7:44:38 PM   ERROR [JobService] TypeError: Cannot read properties of null (reading 'ownerId')
    at PersonService.handleRecognizeFaces (/usr/src/app/dist/domain/person/person.service.js:297:34)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /usr/src/app/dist/domain/job/job.service.js:116:37
    at async Worker.processJob (/usr/src/app/node_modules/bullmq/dist/cjs/classes/worker.js:394:28)
    at async Worker.retryIfFailed (/usr/src/app/node_modules/bullmq/dist/cjs/classes/worker.js:581:24)
[Nest] 8  - 01/28/2024, 7:44:38 PM   ERROR [JobService] Object:
{
  "id": "d34ff9e7-c2a6-4544-929f-348a0c44729e",
  "deferred": false
}
```

Postgres:
```
immich=# select "id","assetId" from asset_faces where "id" = 'd34ff9e7-c2a6-4544-929f-348a0c44729e';
                  id                  |               assetId
--------------------------------------+--------------------------------------
 d34ff9e7-c2a6-4544-929f-348a0c44729e | b14ebe91-3420-45be-ac78-86b56c791296
(1 row)

immich=# select "id","deletedAt" from assets where "id" = 'b14ebe91-3420-45be-ac78-86b56c791296';
                  id                  |           deletedAt
--------------------------------------+-------------------------------
 b14ebe91-3420-45be-ac78-86b56c791296 | 2024-01-28 13:15:43.272135-05
(1 row)

immich=#
```

Similar to #1988 , I believe we need to check if the image is deleted when looking for facial recognition.

Debian 12 on docker

### Version of Immich Server

v1.93.3

n/a

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
n/a
```

### Your .env content

```Shell
n/a
```

### Reproduction steps

```bash
1. Delete an image
2. Run facial recognition
```

### Additional information

_No response_

mertalev:
## Description

When querying for faces during recognition, a face from an asset that was deleted after queueing will throw an error. While TypeORM has a default filter for deleted entities, it still returns an entity if it has deleted relations. The quick fix is to check the asset relation. Moving forward, it may be good to switch to inner joins for relations. This is beyond the scope of this fix as it involves switching to query builders, complicates custom selects, etc.

Fixes #6709

### DIFF
--- a/server/src/domain/person/person.service.spec.ts
+++ b/server/src/domain/person/person.service.spec.ts
@@ -788,7 +788,7 @@ describe(PersonService.name, () => {
     });
 
     it('should return false if face does not have asset', async () => {
-      const face = { ...faceStub.face1, asset: null } as AssetFaceEntity & { asset: null};
+      const face = { ...faceStub.face1, asset: null } as AssetFaceEntity & { asset: null };
       personMock.getFaceByIdWithAssets.mockResolvedValue(face);
 
       expect(await sut.handleRecognizeFaces({ id: faceStub.face1.id })).toBe(false);
--- a/server/src/domain/person/person.service.spec.ts
+++ b/server/src/domain/person/person.service.spec.ts
@@ -1,4 +1,4 @@
-import { AssetFaceEntity, Colorspace, SystemConfigKey } from '@app/infra/entities';
+import { AssetEntity, AssetFaceEntity, Colorspace, SystemConfigKey } from '@app/infra/entities';
 import { BadRequestException, NotFoundException } from '@nestjs/common';
 import {
   IAccessRepositoryMock,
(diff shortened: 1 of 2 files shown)


jrasm91 (commented):
Why is this required?

jrasm91:
It sounds like this is probably caused by trashed assets. So we should probably just make sure we ignore deleted assets instead.

mertalev:
That was the original PR, but then I noticed that there actually is a `AND ("AssetEntity"."deletedAt" IS NULL)` at the end of the query. So the issue is specific to `ownerId`

mertalev:
Although I'm not sure a row still exists if `ownerId` is null.

jrasm91:
Right, I think it would be good to reproduce the actual issue before making a change like this.

I have seen owner being null before, specifically when something references a relation that has been deleted.

jrasm91 (approved):
Looks great. Can you add a unit test for this still working with deleted assets?

PR #6711, opened 2024-01-29
chore: stop cli tests from reading server source

benmccann:
as suggested in https://github.com/immich-app/immich/pull/6678#issuecomment-1913007590. this should unblock the vitest migration for the CLI project


etnoy:
Looks good for the most part, but do we really want to move the api e2e to the tests folder?

jrasm91:
#6719 should fix the core dump issue btw. No need to specify the node version, etc.

benmccann:
> Looks good for the most part, but do we really want to move the api e2e to the tests folder?

We need the e2e api to be compiled because it's shared by the server and CLI projects. Currently the CLI project depends on the e2e api source code, which means the CLI project depends on the source code of the whole server project rather than the compiled output. We can put it in a different directory like `src/e2e` instead of `src/test` if you prefer, but it does have to move to the `src` folder to get compiled or else we can't have the CLI project use it.

> Node versions below 20.8 usually core dump when running tests, if you want a hint for debugging the failing test

Thanks! The tests are passing now by switching to Node 20

> https://github.com/immich-app/immich/pull/6719 should fix the core dump issue btw. No need to specify the node version, etc.

Sure, that works too. Though I was hoping to replace Jest with Vitest (https://github.com/immich-app/immich/pull/6678) after this PR is merged, which would let us run on Node 18 again without issue or needing to mock anything.

benmccann:
closing in favor of https://github.com/immich-app/immich/pull/6747

PR #6713, opened 2024-01-29
refactor: rename clip -> smart search

mertalev:
## Description

There's an inconsistency where some places use Smart Search and others use CLIP. This changes everything in the server and web code to use Smart Search with the exception of the config file. It also renames mobile usages, but I'm less confident about this change.

## How Has This Been Tested?

Tested that browsing and search works on web.


jrasm91 (approved):
LGTM

PR #6717, opened 2024-01-29
feat(mobile): Add people list to exit bottom sheet

l0nax:
This PR adds a list of recognized people on the exif bottom sheet of an asset. This is similar to the feature already available on the web.

The changes are an MVP approach because the information is fetched from the server every time an asset is opened. In a future PR this can be changed to work offline as well.

https://github.com/immich-app/immich/assets/29659953/695736e6-2d3f-4465-8711-22c1c471506f

No connectivity to the server/ no internet:

https://github.com/immich-app/immich/assets/29659953/554233ef-9138-4ba6-8567-8ed44035f227

### DIFF
--- a/mobile/lib/modules/asset_viewer/ui/exif_bottom_sheet.dart
+++ b/mobile/lib/modules/asset_viewer/ui/exif_bottom_sheet.dart
@@ -264,21 +264,24 @@ class ExifBottomSheet extends HookConsumerWidget {
               ),
               SizedBox(
                 height: imageSize,
-                child: CuratedPeopleRow(
-                  content: curatedPeople,
-                  onTap: (content, index) {
-                    context
-                        .pushRoute(
-                          PersonResultRoute(
-                            personId: content.id,
-                            personName: content.label,
-                          ),
-                        )
-                        .then((_) => peopleProvider.refresh());
-                  },
-                  onNameTap: (person, index) => {
-                    showPersonNameEditModel(person.id, person.label),
-                  },
+                child: Padding(
+                  padding: const EdgeInsets.only(top: 8.0),
+                  child: CuratedPeopleRow(
+                    content: curatedPeople,
+                    onTap: (content, index) {
+                      context
+                          .pushRoute(
+                            PersonResultRoute(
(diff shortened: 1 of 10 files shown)


jrasm91:
Looks awesome!

martyfuhry (approved):
This looks super cool! All the code looks fine and it runs great here.

I've only a few nit-picks with the package imports. I have a few issues with the layout, but nothing that can't be fixed later in a separate PR because this is awesome just as it is.

First, I'd like to see the edge of the circle line up with the titles:

image

Second, horizontal overscrolls shouldn't have edge padding:

image

The above should layout from edge to edge.

l0nax:
@martyfuhry thanks for the comments. I've rebased the branch against the latest `main` and changed the imports.

l0nax:
I rebased the branch against the upstream `main` and resolved the conflicts. The feature behaves and looks the same.

mobile/lib/modules/asset_viewer/ui/exif_bottom_sheet.dart
```diff
@@ -212,6 +224,72 @@ class ExifBottomSheet extends HookConsumerWidget {
...
+          return PersonNameEditForm(personId: personId, personName: personName);
+        },
+      ).then((_) {
+        // ensure the people list is up-to-date.
+        peopleProvider.refresh();
+      });
+    }
+
+    buildPeople() {
+      return people.widgetWhen(
```
martyfuhry, line 243: I just found in #7671 that this `widgetWhen` builder performs terribly in the sliding up panel for some reason. It's very jittery and doesn't let the panel slide up smoothly.

In that PR, I'm refactoring this bottom sheet to use widgets instead of using these `buildThing` functions. One of us should refactor this to use a widget (in this file is fine) instead of this function, and we should avoid using `widgetWhen` and opt to using

```dart
if (people.value == null) {
  return Container();
} else {
  return _PeopleWidget(people: people);
}
```

Or something like that. I'm also happy to just merge this in as-is and fix the performance issue in my own refactor, because this is not a regression.

Issue #6691, opened 2024-01-27
[Web] Storage quota is higher than possible


aviv926:
Currently, a storage quota can be set for the user that is higher than the system is capable of.

When a new storage quota is set and the quota exceeds the current drive size, show a warning message to the administrator.

mertalev:
The storage on the server can be changed at any time, so I'm not sure we should try to enforce a limit based on that. If you set a quota and remove a drive, should it now try to clamp the quota based on that? If yes, then the admin's quota isn't respected if a drive is re-added later. If not, then we're back to square 1 with a quota that's higher than what the storage allows.

aviv926:
> The storage on the server can be changed at any time, so I'm not sure we should try to enforce a limit based on that. If you set a quota and remove a drive, should it now try to clamp the quota based on that? If yes, then the admin's quota isn't respected if a drive is re-added later. If not, then we're back to square 1 with a quota that's higher than what the storage allows.

That's a good thought, I think that in general the system administrator should choose how much storage space there will be for the Immich only, from that he can divide/allocate/limit the amount between the users.

Another option would be to store the available storage space in the database, every time this space is added it could suggest to the system administrator to allocate new space to users.

mertalev:
Maybe we could validate the quota when it's set and from then on just display either the quota or the available storage, whichever is lower. So if the storage was decreased it would look normal, and if it was later increased it would still respect the original quota since the value in the DB never changed.

jrasm91:
I think this is all a bit confusing, but it involves a hypothetical situation where the admin incorrectly sets a limit above the drive capacity, which is odd. I don't think we should make any server changes for this. I would say show a popup warning in this case, but that's probably the only thing I would change.

aviv926:
I think the warning message would be helpful in this case. 
Although Mart's method can be good too.

mertalev:
The pop-up idea seems like the simplest solution.

jrasm91:
I've updated the title/description to indicate the current status. Seems like an easy issue for a new contributor to work on.

etnoy:
Closed via #6737

PR #6737, opened 2024-01-29
feat(web): add warning when setting a quota superior to the disk size

martabal:
fix for #6691

## Screenshots

Screenshot from 2024-01-30 14-35-29

### DIFF
--- a/web/src/lib/components/forms/create-user-form.svelte
+++ b/web/src/lib/components/forms/create-user-form.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
   import { api } from '@api';
-  import { createEventDispatcher, onMount } from 'svelte';
+  import { createEventDispatcher } from 'svelte';
   import ImmichLogo from '../shared-components/immich-logo.svelte';
   import { notificationController, NotificationType } from '../shared-components/notification/notification';
   import Button from '../elements/buttons/button.svelte';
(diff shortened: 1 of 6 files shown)


etnoy:
Just an idea, but can it display the warning already in the modal, when the number in the text field is too large? 

I.e. like "quota is too large..." next to the text field, shown before you click the button? Hope this makes sense

martabal:
> Just an idea, but can it display the warning already in the modal, when the number in the text field is too large?
> 
> I.e. like "quota is too large..." next to the text field, shown before you click the button? Hope this makes sense

That's even better ! thanks

web/src/lib/components/forms/create-user-form.svelte
```diff
@@ -13,9 +14,17 @@
...
   let isCreatingUser = false;
 
+  $: warning = quotaSize && convertToBytes(Number(quotaSize), 'GiB') > $serverInfo.diskSizeRaw;
+
+  onMount(async () => {
+    if (!$serverInfo) {
+      const { data } = await api.serverInfoApi.getServerInfo();
+      $serverInfo = data;
+    }
+  });
```
jrasm91, line 27: why is this needed? it seems like it should be initialized in the layout or when the user logs in, not in every component that uses it.

martabal: It feels a bit overkill to have this API request in a layout. I moved it to the page.ts, can you check my last commit ?

jrasm91: Why is it overkill to have it in the layout? That's essentially an entrypoint and means it will only be called once per _page load_.

jrasm91 (approved):
LGTM

web/src/lib/components/forms/create-user-form.svelte
jrasm91: It may be more complicated since it requires auth, so we can look into this again in the future.

PR #6740, opened 2024-01-30
chore: migrate CLI tests to vitest

benmccann:
This is https://github.com/immich-app/immich/pull/6711 with one additional commit on top. We can either merge that one and I can rebase this or we can close that one and just do it all in one go here.

It's nice to switch to vitest for a few reasons:
- I find it to be much simpler. It doesn't require understanding how multiple tools like jest, ts-node, babel, etc. work together. This PR removes thousands of lines of code from the lockfile, which demonstrates that


jrasm91:
I like the improved simplicity, but these changes also impact the dist output, which now includes `src/` and `test/` subdirectories.

benmccann:
Yes, that's true. Though I didn't know how much the matters. Since it's the server project the small additional size won't have much impact like shipping JS to the browser would. The cleanest way to do it would be to create a sibling project of `server` and `cli` called something like `test-utils`. If you prefer, I could do that.

jrasm91:
It's not so much about the size as it is about just being a huge pain to work with. For example, referencing files by relative paths outside of the src might work in development and break in production. If the directory structure changes again for whatever reason, all the start-up references will break (again).

For example, you did not update `server/start.sh`, so the production build is probably broken with the current changes. There are just a lot of things that are impacted when you change the folder structure in the dist folder and it is quite a headache.

I've started to take a look at this a bit and there are a few things that I am seeing:
1. The CLI is quite messy and in need of some basic clean up and refactoring
2. The `@api` import should not be needed at all, the cli should do everything with the generated `@immich/sdk`.
3. The only other code from the server directory that is actually needed is the code that starts the app and resets the database. It might be possible to move those two specific functions into the server/src directory. Or, just re-implement them in the cli itself, using references from the compiled output (AppModule, etc.).

I appreciate the motivation to migrate to vitest, remove complicated dependencies, and decouple the reliance on the server codebase. I think we might need to address some of these obstacles in intermediate steps before fully switching over to vitest.

benmccann:
closing for now pending https://github.com/immich-app/immich/pull/6747. I can make a new attempt after that

PR #6747, opened 2024-01-30
refactor(cli): organize files, simplify types, use @immich/sdk

jrasm91:
| Before | After |
| - | - |
| image | image |


cli/src/services/session.service.ts
```diff
@@ -1,31 +1,40 @@
-import fs from 'node:fs';
-import yaml from 'yaml';
+import { existsSync } from 'fs';
```
benmccann, line 1: I probably wouldn't remove the `node:` prefix here
```suggestion
import { existsSync } from 'node:fs';
```

cli/test/cli-test-utils.ts
```diff
@@ -1,13 +1,31 @@
...
+export const setup = async () => {
+  const api = new ImmichApi(process.env.IMMICH_INSTANCE_URL as string, '');
+  await api.authenticationApi.signUpAdmin({
+    signUpDto: { email: 'cli@immich.app', password: 'password', name: 'Administrator' },
+  });
+  const { data: admin } = await api.authenticationApi.login({
+    loginCredentialDto: { email: 'cli@immich.app', password: 'password' },
+  });
+  const { data: apiKey } = await api.keyApi.createApiKey(
+    { aPIKeyCreateDto: { name: 'CLI Test' } },
```
benmccann, line 21: the capitalization looks off here

benmccann (commented):
awesome! thank you!!

cli/src/services/session.service.ts
jrasm91: True, true. Do you know if there is a linting rule or some way to prefer the prefix? The code base is quite inconsistent.

cli/test/cli-test-utils.ts
jrasm91: Haha yeah. This is the auto generated code from open API generator. I think if we changed the dto on the server from APIKey to ApiKey it would generate more how you would expect it to.

mertalev: Based on [this](https://stackoverflow.com/questions/15526107/acronyms-in-camelcase) SO post, It seems style guidelines for acronyms lean toward using the normal casing instead of capitalizing. So `XmlHttpRequest`, not `XMLHTTPRequest`

cli/src/services/session.service.ts
etnoy: I created #6751 as a proof of concept for this (and most other unicorn rules)

etnoy (approved):

cli/src/commands/upload.command.ts
```diff
@@ -1,15 +1,112 @@
...
+    this.fileSize = stats.size;
+    this.albumName = this.extractAlbumName();
+  }
+
+  async getUploadFormData(): Promise<FormData> {
+    if (!this.deviceAssetId) throw new Error('Device asset id not set');
+    if (!this.fileCreatedAt) throw new Error('File created at not set');
+    if (!this.fileModifiedAt) throw new Error('File modified at not set');
+
+    // TODO: doesn't xmp replace the file extension? Will need investigation
```
danieldietzler, line 42: Normally you get a `.xmp` after the original extension, so `foo.jpg.xmp`

danieldietzler (approved):

etnoy (approved):

cli/src/commands/upload.command.ts
danieldietzler: Oh really? I hate this.

PR #6751, opened 2024-01-30
chore(server,cli,web): housekeeping and stricter code style

etnoy:
Inspiration came from https://github.com/immich-app/immich/pull/6747#discussion_r1470590345

This turned into a housekeeping and code style PR in general

* Add unicorn eslint rules which makes linting more strict across the whole codebase
* Bump es version
* Bump ts version

I've been able to fix a surpring number of bugs with this increased linting!


cli/src/commands/upload.ts
```diff
@@ -68,7 +68,8 @@ export class Upload extends BaseCommand {
       }
     }
 
-    const existingAlbums = (await this.immichApi.albumApi.getAllAlbums()).data;
+    const allAlbums = await this.immichApi.albumApi.getAllAlbums();
+    const existingAlbums = allAlbums.data;
```
jrasm91:
```suggestion
    const { data: existingAlbums } = await this.immichApi.albumApi.getAllAlbums();
```
Does this work instead for these types of variables?

cli/src/commands/upload.command.ts
```diff
@@ -276,12 +276,12 @@ export class UploadCommand extends BaseCommand {
...
-      maxBodyLength: Infinity,
+      maxContentLength: Number.POSITIVE_INFINITY,
+      maxBodyLength: Number.POSITIVE_INFINITY,
       data,
     };
 
-    const res = await axios(config);
-    return res;
+    const response = await axios(config);
+    return response;
```
benmccann, line 285:
```suggestion
    return await axios(config);
```

cli/src/commands/upload.command.ts
```diff
@@ -8,7 +8,7 @@ import { BaseCommand } from './base-command';
 import { basename } from 'node:path';
 import { access, constants, stat, unlink } from 'node:fs/promises';
 import { createHash } from 'node:crypto';
-import Os from 'os';
+import Os from 'node:os';
```
benmccann, line 11: I'd expect this to be `os`. I wasn't going to mention it since it's unrelated, but then I saw it imported as `os` in the next file in this PR

benmccann (commented):
personally, I'm not a fan of the `prevent-abbreviations` rule. things like `dir` are so common that everyone knows what they mean and all that typing out `directory` accomplishes is making things longer

cli/src/commands/upload.command.ts
etnoy: Thanks Ben. I agree with this, will try to find a decent balance with which unicorn rules we apply.

benmccann:
What's the reason for the `UUIDParamDto as UIDParameterDto` changes?

etnoy:
> What's the reason for the `UUIDParamDto as UIDParameterDto` changes?

That's likely an artifact from me running eslint --fix. Thanks for spotting!

cli/package.json
```diff
@@ -28,18 +28,18 @@
     "@types/cli-progress": "^3.11.0",
     "@types/mock-fs": "^4.13.1",
     "@types/node": "^20.3.1",
-    "@typescript-eslint/eslint-plugin": "^6.0.0",
-    "@typescript-eslint/parser": "^6.0.0",
+    "@typescript-eslint/eslint-plugin": "^6.4.1",
+    "@typescript-eslint/parser": "^6.4.1",
+    "eslint": "^8.56.0",
```
benmccann, line 33: personal preference, but I preferred them alphabetized

etnoy: me too, now that you mentioned it!

web/src/lib/components/forms/create-user-form.svelte
```diff
@@ -69,11 +69,11 @@
           error = 'Error create user account';
           isCreatingUser = false;
         }
-      } catch (e) {
+      } catch (error_) {
```
benmccann, line 72: there's quite a few places throughout with trailing underscore, which seems a bit funny to me. I thought I'd point this out in case it was done by the automatic fixer and we thought it was worth changing. i'm not sure how to really identify all the places it did it though. this PR almost crashes my browser as is :laughing:

etnoy: lol. Yeah, maybe this PR is a little much as-is. At least all the lints are fixed now. Thanks for noticing the underlines.

jrasm91 (approved):
Thanks for working on this! It looks like a good start and we can tweak and refine moving forward.

etnoy:
> Thanks for working on this! It looks like a good start and we can tweak and refine moving forward.

Thanks for not being too mad about the mess and size of this pr, lol

PR #6753, opened 2024-01-30
feat(web): add a setting to load the original file

martabal:
## Changes made in this PR

This PR adds a new section in the user setting called "Quality" and a toggle "Load original photos". Instead of displaying by default the JPEG thumbnail, the user has the choice to load the original file if it's web-compatible. This may result in slower photo display speeds but at least users have the choice to enable it or not.

### DIFF
--- a/web/src/lib/components/user-settings-page/quality-settings.svelte
+++ b/web/src/lib/components/user-settings-page/quality-settings.svelte
@@ -3,7 +3,7 @@
   import { alwaysLoadOriginalFile } from '../../stores/preferences.store';
   import SettingSwitch from '../admin-page/settings/setting-switch.svelte';
 
-  export const handleToggle = () => {
+  const handleToggle = () => {
     $alwaysLoadOriginalFile = !$alwaysLoadOriginalFile;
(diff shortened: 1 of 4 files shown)


web/src/lib/components/user-settings-page/quality-settings.svelte
```diff
@@ -0,0 +1,24 @@
+<script lang="ts">
+  import { fade } from 'svelte/transition';
+  import { alwaysLoadOriginalFile } from '../../stores/preferences.store';
+  import SettingSwitch from '../admin-page/settings/setting-switch.svelte';
+
+  export const handleToggle = () => {
```
danieldietzler, line 6: This should not be exported, right?

danieldietzler (approved):

web/src/lib/components/user-settings-page/quality-settings.svelte
martabal: Yep, thanks :smile:

jrasm91 (approved):
LGTM

PR #6764, opened 2024-01-30
Fix for Notifications cover the upload control panel #6102

### LINKED ISSUE #6102: [BUG] Uploading directly to album spams notification feed
### The bug

When uploading a lot of new photos to an **album** this causes many messages to appear one above the other. We should instead suppress these messages in this scenario, either adding a single notification once uploading is complete or just not having any notification at all.

image

Win 10 with Docker

### Version of Immich Server

last

null

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
default
```

### Your .env content

```Shell
default
```

### Reproduction steps

```bash
1. Create a new album
2. Upload +50 photos
3. Try accessing the upload control panel
```

### Additional information

**Tested on several browsers:**
- Google Chrome
- Edge
- Firefox

Menghini:
The notifications for adding assets covers the upload control panel. This PR moves it those notifications to the left of the screen, where there are no controls. This allows easier access to the upload control panel on the right side of the screen. This is a potential fix for #6102

### DIFF
--- a/web/src/lib/components/shared-components/notification/notification-list.svelte
+++ b/web/src/lib/components/shared-components/notification/notification-list.svelte
@@ -10,7 +10,7 @@
 </script>
 
 {#if $notificationList.length > 0}
-  <section transition:fade={{ duration: 250 }} id="notification-list" class="absolute left-4 top-[80px] z-[99999999]">
+  <section transition:fade={{ duration: 250 }} id="notification-list" class="absolute right-5 top-[80px] z-[99999999]">
     {#each $notificationList as notificationInfo (notificationInfo.id)}
       <div animate:flip={{ duration: 250, easing: quintOut }}>
         <NotificationCard {notificationInfo} />
--- a/web/src/lib/components/shared-components/notification/notification-list.svelte
+++ b/web/src/lib/components/shared-components/notification/notification-list.svelte
@@ -10,7 +10,7 @@
 </script>
 
 {#if $notificationList.length > 0}
-  <section transition:fade={{ duration: 250 }} id="notification-list" class="absolute right-5 top-[80px] z-[99999999]">
+  <section transition:fade={{ duration: 250 }} id="notification-list" class="absolute left-4 top-[80px] z-[99999999]">
     {#each $notificationList as notificationInfo (notificationInfo.id)}
       <div animate:flip={{ duration: 250, easing: quintOut }}>
         <NotificationCard {notificationInfo} />
--- a/web/src/lib/components/shared-components/notification/notification-list.svelte
+++ b/web/src/lib/components/shared-components/notification/notification-list.svelte
@@ -11,7 +11,7 @@
 
 {#if $notificationList.length > 0}
   <section transition:fade={{ duration: 250 }} id="notification-list" class="absolute right-5 top-[80px] z-[99999999]">
-    {#each $notificationList as notificationInfo (notificationInfo.id)}
+    {#each $notificationList.slice(0, 3) as notificationInfo (notificationInfo.id)}
       <div animate:flip={{ duration: 250, easing: quintOut }}>
         <NotificationCard {notificationInfo} />
       </div>


alextran1502:
Moving this to the left will block the sidebar; I'd rather move it down a little on the right-hand side. Can you please add a screenshot of before and after?

alextran1502:
You might not need to navigate away but I think it will be better to not blocking "primary" UI components

Menghini:
I agree with you. I think another approach would be better, such as limiting the number of notifications or similar approach.

Menghini:
I have switched the `notification-list` back to the right side.

I have two ideas:
1) Moving it behind the `upload-panel`
image

2) Only show three notifications as is currently done in my fork and in this PR.

Thoughts?

waclaw66:
Why there has to be any notification for each file at all when a progress bar is shown?
I would follow Google Photos way, it has pretty good and simple upload notification/progress. Popping toasts are quite annoying.
obrazek

Menghini:
I've looked into this issue further. I think the problem lies that if you are adding assets to an album from anywhere (such as one that has already been uploaded), there will be a popping toast notification. This normally doesn't occur when uploading pictures, except for when uploading them directly to an album, which is why there is an upload box and the redundant popping toast notification displaying the same information.

As far as I am concerned, you can do one of these things:
1. Limit the amount of toasts to limit the overflow into another UI element (and from going off screen).
2. If we are uploading assets directly to the album, don't show the "added x assets to album" toast notifications.

I would personally argue that either way, showing an overflow of notifications shouldn't happen and be limited to avoid a really long list that scrolls outside the view. This will not only make it a cleaner experience, but also allow access to the upload panel in an album. The 2nd one is up to grabs. I'm not quite sure who makes this decision, but that's my two cents.

hirenshah:
I don't believe there is any benefit in having all these notifications, especially since the upload panel provides a better view of the progress.  You also don't need the view album button a hundred times because a) you are already in the album and b) they often move too fast to even click on.

Can we not just have a summary at the end?

image

hirenshah:
I've noticed that the notification flood results in uploads taking longer.  Uploading media outside of albums is much faster.  So, until this is sorted out, I've added the following under Theme Settings > Custom CSS:

`#notification-list { display: none !important; }`

By hiding the notifications, I've not lost visibility of any useful information and it has definitely improved performance. 

**What was the rationale behind these album upload notifications? Assuming there was a reason, surely that reason should apply to both album and non-album uploads?**

Menghini:
I can't speak to the developers, @hirenshah, but to my understanding, it is doing the notifications because it 1) first uploads to the general photos and then 2) adds it to the album, which makes the notification (adding each asset, one at a time)

hirenshah:
@Menghini Yeah, I get what it is doing but I am wondering why.  This "feature" makes no sense to me and I think it is just ridiculous.

However, my day job is all about software development, so I recognise the fact that I may not have considered all scenarios / use cases, and am open to understanding the rationale. Hopefully the developers can provide some background information. 

In the meantime, @Menghini (or anyone else), do you benefit from these notifications?  If so, in what way?

jrasm91:
You can add photos to an album without uploading brand new ones. In that situation we provide a notification which says how many were added. This is helpful especially when there were some that already existed and the actual added count is different from the selected one. In that use case you get one notification for the one action with the totals. I'm guessing the upload and add action is implemented in a way that is trying to reuse existing code but results in this bad experience. I don't know why you think we would implement it like this on purpose.

hirenshah:
Thanks for the explanation.  In order to better understand the situation, the additional information is exactly what I was after.  

To date, I had never added existing images to albums (at least not via the UI) but following your comment, I've had a quick play.  Everything makes a lot more sense 👍 

My takeaway (high level):
- code has been reused (quite rightly so)
- unforeseen outcome for certain use cases
- doesn't break core functionality
- cosmetic/minor defect
- goes to the end of the backlog

Standard 😊

<br>
<br>

> I don't know why you think we would implement it like this on purpose.

@jrasm91 Was this comment directed at me? If so, maybe you missed the following:
<br>

> However, my day job is all about software development, so I recognise the fact that I may not have considered all scenarios / use cases, and am open to understanding the rationale. Hopefully the developers can provide some background information.

<br>
<br>

Obviously I didn't think it was done on purpose.  If I did, I would have deleted my immich containers by now  😝

zackpollard:
Hey, based on the feedback above, I am going to close this PR. I'll update the existing issue #6102 to instead fix the bug around uploading directly to albums spamming the notification feed. Cheers!

PR #6765, opened 2024-01-30
refactor: authentication on public routes

martabal:
This PR aims to refactor the authentication on public routes to avoid 400 requests on shared links


web/src/lib/utils/auth.ts
```diff
@@ -19,19 +21,19 @@ export const getAuthUser = async () => {
...
+    if (!user) {
+      redirect(302, AppRoute.AUTH_LOGIN);
+    }
+    if (options.admin && !user.isAdmin) {
+      redirect(302, AppRoute.PHOTOS);
+    }
   }
-
-  if (!savedUser) {
+  if (!savedUser && user) {
```
jrasm91, line 36: this is quite messy. can you remove nested if statements and clean up the logic so it is more readable.

web/src/lib/utils/cookies.ts
```diff
@@ -0,0 +1,30 @@
+import { browser } from '$app/environment';
+
+const IsAuthenticatedCookieName = 'immich_is_authenticated';
+
+export const setAuthCookie = (): void => {
```
jrasm91, line 5: set and remove should just be deleted as this should be implemented server-side.

web/src/lib/utils/cookies.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (browser) {
+    document.cookie = `${IsAuthenticatedCookieName}=${String(true)}; Path=/; Max-Age=34560000; SameSite=Lax;`;
+  }
+};
+
+export const getAuthCookie = (): string | null => {
```
jrasm91, line 11: this should probably just be a private method inside of the file that has `authenticate()` and it can use it to avoid calling getMyInfo when it isn't set.

jrasm91 (changes requested):

server/src/domain/auth/auth.service.ts
```diff
@@ -443,14 +444,17 @@ export class AuthService {
 
     let authTypeCookie = '';
     let accessTokenCookie = '';
+    let isAuthenticatedCookie = '';
 
     if (isSecure) {
       accessTokenCookie = `${IMMICH_ACCESS_COOKIE}=${loginResponse.accessToken}; HttpOnly; Secure; Path=/; Max-Age=${maxAge}; SameSite=Lax;`;
       authTypeCookie = `${IMMICH_AUTH_TYPE_COOKIE}=${authType}; HttpOnly; Secure; Path=/; Max-Age=${maxAge}; SameSite=Lax;`;
+      isAuthenticatedCookie = `${IMMICH_IS_AUTHENTICATED}=true; Path=/; Max-Age=${maxAge}; SameSite=Lax;`;
```
jrasm91, line 452: you didn't add secure here, should it have that attribute?

web/src/lib/utils/auth.ts
```diff
@@ -17,21 +19,34 @@ export const getAuthUser = async () => {
...
   options = options || {};
-
   const savedUser = getSavedUser();
-  const user = savedUser || (await getAuthUser());
-
-  if (!user) {
-    redirect(302, AppRoute.AUTH_LOGIN);
-  }
+  const user = savedUser || (isAuthenticated() ? await getAuthUser() : null);
+  if (!options.public) {
```
jrasm91: I think you might be able to just add one new condition at the top that is "if public && !user then return" and leave everything else the same.

jrasm91: Or, inside of !user either do the redirect or not, depending if the route is public. Either way you should not do anything else doing down if the user is null.

web/src/lib/components/user-settings-page/user-profile-settings.svelte
```diff
@@ -3,27 +3,28 @@
...
   import { fade } from 'svelte/transition';
   import { handleError } from '../../utils/handle-error';
   import SettingInputField, { SettingInputFieldType } from '../admin-page/settings/setting-input-field.svelte';
   import Button from '../elements/buttons/button.svelte';
-  import { setUser } from '$lib/stores/user.store';
+  import { user } from '$lib/stores/user.store';
 
-  export let user: UserResponseDto;
+  // deep clone user
+  let editedUser = JSON.parse(JSON.stringify($user)) as typeof $user;
```
danieldietzler, line 14: I mean... it works. But I'd prefer the lodash cloneDeep function for that :P

danieldietzler (approved):
LGTM

web/src/lib/utils/auth.ts
```diff
@@ -1,53 +1,61 @@
...
 export const authenticate = async (options?: AuthOptions) => {
   options = options || {};
+  const user = await loadUser();
 
-  const savedUser = getSavedUser();
-  const user = savedUser || (await getAuthUser());
-
-  if (!user) {
-    redirect(302, AppRoute.AUTH_LOGIN);
+  if (options.public && !user) {
```
jrasm91, line 43: why do you care if the user is defined or not if the route is public?

jrasm91: it seems like you want:
1. if public return we're done with all the checks
2. if not user, redirect to login, this is an auth required route
3. if admin route and not an admin, redirect to /photos

jrasm91 (approved):
LGTM

PR #6766, opened 2024-01-30
Provide instructions of how to specify a default timezone

mmomjian:
We should specify to the user how to enter their timezone as an env var.

### DIFF
--- a/docs/docs/install/environment-variables.md
+++ b/docs/docs/install/environment-variables.md
@@ -46,7 +46,7 @@ These environment variables are used by the `docker-compose.yml` file and do **N
 
 `exiftool` is only present in the microservices container.
 
-Timezones should be entered as the "TZ identifier" from [this list](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#List). For example, TZ="Etc/UTC".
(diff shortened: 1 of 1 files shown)


mmomjian:
Sort of a followup to this - how is `/etc/localtime` vs `TZ` handled? I had /etc/localtime mounted but I had some images that would not get the correct timezone until I added the `TZ` env var as well.

etnoy:
Having localtime mounted should be equivalent to the TZ env var. We had th TZ method documented before switching over to the localtime volume method which is more reliable. Are you sure there was a difference? If you remove the TZ method and rerun metadata extraction, does the problem come back, and on which assets?

mmomjian:
> Having localtime mounted should be equivalent to the TZ env var. We had th TZ method documented before switching over to the localtime volume method which is more reliable. Are you sure there was a difference? If you remove the TZ method and rerun metadata extraction, does the problem come back, and on which assets?

I don't want to play on my primary instance but I will be starting up a second instance in the next week or two for testing purposes, and I will run some tests there.

etnoy:
I'm closing this for now, feel free to reopen when you did your tests

PR #6777, opened 2024-01-30
chore: migrate CLI to ESM and vitest

benmccann:
Followup from the last attempt in https://github.com/immich-app/immich/pull/6740

It's nice to switch to vitest for a few reasons:
- I find it to be much simpler. It doesn't require understanding how multiple tools like jest, ts-node, babel, etc. work together. This PR removes thousands of lines of code from the lockfile, which demonstrates that

To support ESM, three changes were needed:
- migrate to vitest
- update `tsconfig.json` to output ESM
- update `package.json` to specify `"type": "module"`


jrasm91:
Can you summarize the required changes you had to make for esm to work in the description? It looks like you were able to address the directory structure change by moving the test utils to the src directory, is that right? That seems like a good compromise to me.

jrasm91 (commented):
I believe we only configured the typescript-sdk package as common js in the first place because of the CLI dependency on it. Would it be worth updating that as well? Or is it normal to publish packages as both cjs and esm?

benmccann:
> Can you summarize the required changes you had to make for esm to work in the description?

Done. Most of it was just getting rid of jest. There wasn't much left after that

> It looks like you were able to address the directory structure change by moving the test utils to the src directory, is that right? That seems like a good compromise to me.

Yep. It was previously in the src directory, but was in `src/test`, which maybe TypeScript interpreted as containing tests. I made it `src/test-util` in this one, which seemed to work better.

> I believe we only configured the typescript-sdk package as common js in the first place because of the CLI dependency on it. Would it be worth updating that as well? Or is it normal to publish packages as both cjs and esm?

Yes, we should make it ESM only as well. I planned to do that in a followup. I haven't investigated it to know what else is necessary. I'm hoping it shouldn't be much after this change, but sometimes surprises pop up :laughing:

jrasm91:
Perfect, sounds good. The typescript-sdk changes should be pretty minimal. Mostly updating the package.json exports/type:module, and then removing the cjs part of the build output.

jrasm91 (approved):
LGTM

PR #6780, opened 2024-01-30
feat(mobile): Add end page to the end to memories

martyfuhry:
image


martyfuhry:
There's a bug here where scrolling down and back up quickly will have the wrong index for an image, but it's in production, as well.

alextran1502 (approved):

PR #6793, opened 2024-01-31
feat(mobile): Memories activity is now full screen, better image fit, adds progress indicator

martyfuhry:
image

- Similar to the full-screen activity view of the gallery page, this will hide the system navigation bars when you enter the memories activity. 
- The navigation bars will return when you leave the activity.
- Adds a progress bar at the top which animates along with the page moving
- Computes aspect ratio of the photo and the available space; if the aspect ratios are nearly identical, then use much better looking `BoxFit.cover` image fit
- Removes the "1|12" text in the top right
- Refactors the widget to use more page controllers
- Changed the close icon to have a little more contrast
- Close icon now stays in the same place during page transitions


martyfuhry:
Ah, shoot I gotta dispose those `PageController` objects. But this is a `HookConsumerWidget` so I don't have a dispose method. Hold on while I fix that.

martyfuhry:
OK, I'm back to using Flutter Hooks once more. This is working fine and ready for review!

martyfuhry:
The latest commit adds ticks to the progress bar:

image

I'd love to get some kind of autoplay working, but it's proven a bit more challenging than I'd like for this PR. So I'll open a new one for that later.

alextran1502 (approved):

PR #6818, opened 2024-01-31
feat(server)!: oauth encryption algorithm setting

### LINKED ISSUE #6271: [BUG] Supporting multiple OAuth token algorithms
### The bug

Error 'unexpected JWT alg received, expected RS256, got: ES384' when using OAuth. Could you please add support to some other token algorithms? This can be fixed by simply modifying the configuration of the oidc library.
Thanks!

ArchLinux

### Version of Immich Server

v1.9.2

N/A

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
Irrelevant
```

### Your .env content

```Shell
Irrelevant
```

### Reproduction steps

```bash
Irrelevant
```

### Additional information

_No response_

danieldietzler:
Notes from @jrasm91:

> [!WARNING]
> ### Breaking Change
> OAuth setups using `HS256` (mainly Authentik) will need to either (1) update the signing algorithm in Immich or (2) specify a signing key in the provider settings (so that it uses `RS256` instead).

### Specify a signing key in Authentik:
[Screencast from 02-02-2024 12:05:04 AM.webm](https://github.com/immich-app/immich/assets/4334196/e51e00ca-7750-476c-982b-070806b7cbef)

### New Immich OAuth Setting
image

### Background

`RS256` is generally better than `HS256`. `RS256` is pretty much the most commonly used algorithm. The client library we use for open-id defaults to `RS256`. It's very easy to setup **Authentik** _without_ specifying a signing key, which will default to use `HS256`. The original implementation added a hack/fallback to `HS256` in some conditions to try to handle that situation. The current code removes the fallback, and adds a specific `Signing Algortithm` setting which can be explicitly set. Alternatively, the issue could be fixed by specifying a signing key in Authentik.

References:
- https://auth0.com/blog/rs256-vs-hs256-whats-the-difference/
- Fixes #6271


server/src/domain/auth/auth.service.ts
```diff
@@ -337,7 +337,7 @@ export class AuthService {
     try {
       const issuer = await Issuer.discover(issuerUrl);
       const algorithms = (issuer.id_token_signing_alg_values_supported || []) as string[];
-      if (algorithms[0] === 'HS256') {
+      if (algorithms[0] === config.oauth.signingAlgorithm) {
```
jrasm91, line 340: If it is set use it, otherwise don't set it. (No need to compare to [0] anymore

danieldietzler: Oh but what if it's not supported by the issuer?

jrasm91: Doesn't matter. If the user put it in just use it as is. The user is specifically saying that the auth server has been configured this way for this client.

jrasm91 (approved):

halkeye:
Well this was a fun redirect loop.

For anyone else surprised by this and disabled local login

```
insert into system_config VALUES ('oauth.signingAlgorithm', '"HS256"');
```

I'm still trying to figure out how to configure authentik with cert-manager to use rs instead

muava12:
i'm a noob in outh, how to do "(1) update the signing algorithm in Immich " as mentioned in release note??

jrasm91:
> i'm a noob in outh, how to do "(1) update the signing algorithm in Immich " as mentioned in release note??

Log in as the immich admin and click on "Administration" then "Settings" then "OAuth"

PR #6840, opened 2024-02-01
feat(mobile): Add support for Basic Authentication

rovo89:
Use the `x-immich-user-token` header instead of `Authentication: Bearer <token>` to allow for Basic Authentication credentials in the server URL (e.g. `https://user:pass@example.org`). Most of the codebase handles the latter automatically, however the `Authorization: Basic ...` header used to be overwritten.

Only exception: `socket_io_client` doesn't seem to support getting credentials this way, so we have to add them via manually crafted `Authorization` header.

Tested on Android against nginx reverse proxy. With these changes, the nginx log doesn't show any unauthenticated requests anymore, so I hope I found all places.

While I was at it, I changed all variable names to `accessToken`. Some were called `authToken`, `token` or even `jwtToken`.

### DIFF
--- a/mobile/lib/modules/album/ui/album_thumbnail_listtile.dart
+++ b/mobile/lib/modules/album/ui/album_thumbnail_listtile.dart
@@ -49,7 +49,7 @@ class AlbumThumbnailListTile extends StatelessWidget {
           type: ThumbnailFormat.WEBP,
         ),
         httpHeaders: {
-          "Authorization": "Bearer ${Store.get(StoreKey.accessToken)}",
+          "x-immich-user-token": Store.get(StoreKey.accessToken),
         },
         cacheKey: getAlbumThumbNailCacheKey(album, type: ThumbnailFormat.WEBP),
         errorWidget: (context, url, error) =>
--- a/mobile/lib/modules/asset_viewer/views/gallery_viewer.dart
+++ b/mobile/lib/modules/asset_viewer/views/gallery_viewer.dart
@@ -78,8 +78,7 @@ class GalleryViewerPage extends HookConsumerWidget {
     final isPlayingMotionVideo = useState(false);
     final isPlayingVideo = useState(false);
     Offset? localPosition;
-    final authToken = 'Bearer ${Store.get(StoreKey.accessToken)}';
-    final header = {"Authorization": authToken};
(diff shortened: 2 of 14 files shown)


danieldietzler:
Please look for other PRs before opening one :) (#6617)

danieldietzler:
Sorry just read the your comment on the other PR

rovo89:
> Sorry just read your comment on the other PR

Maybe I should have mentioned the other PR in the first comment. Didn't expect someone to react that fast. 😉 

By the way, I didn't touch `mobile/openapi/lib/auth/oauth.dart` because it seems to be auto-generated. Not sure if that file is relevant for mobile at all - if yes, it might mean that OAuth and Basic Authentication can't be used at the same time (unless the generator can be told to set the other header instead).

jpg0:
I see that you manually add the basic auth `Authorization` header in the websocket connection. Are you 100% sure that the header will be added for the normal http requests? I don't believe that they were when I tested, plus I cannot see any logic in the dart http lib which would do so.

Other than that it looks good to me.

rovo89:
Yes, because I tested it. Would be great if you could also test with your setup.

This might be the place where the credentials get added: https://github.com/dart-lang/sdk/blob/2a88162b841b5b93a36f698336404dc8f4e9b150/sdk/lib/_http/http_impl.dart#L2162

jpg0:
I just tested with my setup and indeed it does work!

The source that I was looking at is https://github.com/dart-lang/http/tree/master/pkgs/http - which I thought was being used as the import looked like a package rather than part of the SDK. I've never used Dart before so probably got this wrong. Regardless, as long as it works, and it certainly appears to, I'm happy.

alextran1502 (approved):

kocane:
Sorry for what might be a dumb question but is this supported in the lastest version of the app?

danieldietzler:
> Sorry for what might be a dumb question but is this supported in the lastest version of the app?

Yes, it should be.

kocane:
> Yes, it should be.

Hm. I'm not having any luck with it. When pressing Login it doesnt do anything at all but the log shows this:

312490729-9ac26006-125d-4fcd-b97e-bb2e5acaacca

Issue #6490, opened 2024-01-19
Errors appearing only in logcat, not in app log


rovo89:
### The bug

I'm currently experiencing some issues with uploading from the Android app. The root cause is probably in my server config (EDIT: `client_max_body_size` was missing), but that isn't why I open this issue.

While trying to debug the problem, I obviously looked into the "Logs" in the app and in the various Docker containers, but nothing was mentioned there. Even after incrasing the log level. Also no error message in the app, the photos simply didn't end up on the server and the remaining count stayed at 10.

Finally I set up adb to check logcat, and there I found more information:
```
01-19 11:31:13.362 10831 10866 I flutter : ERROR backupAsset: FormatException: Unexpected character (at character 1)
01-19 11:31:13.362 10831 10866 I flutter : <html>
01-19 11:31:13.362 10831 10866 I flutter : ^
01-19 11:31:13.362 10831 10866 I flutter :
```

Would it be possible to make that error message appear in the app log?

And by the way, there are usually hundreds of messages in the log, so maybe it would be a good idea to show the latest message at the top (or scroll down by default).

### The OS that Immich Server is running on

Debian 12 / Docker

### Version of Immich Server

v1.193.0

### Version of Immich Mobile App

v1.193.0

### Platform with the issue

- [ ] Server
- [ ] Web
- [X] Mobile

### Your docker-compose.yml content

```YAML
...
```

### Your .env content

```Shell
...
```

### Reproduction steps

```bash
1. Mess up your Nginx config.
2. Try to upload files.
```

jrasm91:
The mobile in-app logging has been improved. I believe this issue has been fixed.

PR #6866, opened 2024-02-02
feat(mobile): show more errors in the app log

rovo89:
Many helpful errors are sent to `debugPrint()` (i.e. logcat on Android) which requires additional efforts to read them.

Use the app's own logging mechanism instead to make them appear in the builtin log viewer, making them easily available to end-users.

### DIFF
--- a/mobile/lib/modules/login/providers/authentication.provider.dart
+++ b/mobile/lib/modules/login/providers/authentication.provider.dart
@@ -52,8 +52,8 @@ class AuthenticationNotifier extends StateNotifier<AuthenticationState> {
       // Resolve API server endpoint from user provided serverUrl
       await _apiService.resolveAndSetEndpoint(serverUrl);
       await _apiService.serverInfoApi.pingServer();
-    } catch (e) {
-      debugPrint('Invalid Server Endpoint Url $e');
+    } catch (error, stack) {
+      _log.severe("Error logging in: ${error.toString()}", error, stack);
       return false;
     }
 
@@ -91,9 +91,9 @@ class AuthenticationNotifier extends StateNotifier<AuthenticationState> {
         accessToken: loginResponse.accessToken,
         serverUrl: serverUrl,
       );
-    } catch (e) {
+    } catch (error, stack) {
       HapticFeedback.vibrate();
-      debugPrint("Error logging in $e");
+      _log.severe("Error logging in ${error.toString()}", error, stack);
       return false;
     }
   }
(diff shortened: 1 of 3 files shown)


rovo89:
I would like to get some feedback on this one before proceeding:

1. Is there any overall reason not to do this? I had a hard time when initially trying to set up my reverse proxy because I couldn't find logs in the app nor in the docker containers. Finally checked logcat and there they were, but I would have really liked to see them in the integrated log viewer (see #6490).
2. Is the error message text fine like this? To me it feels a bit strange to merge the error details (usually the raw server response) into the error description (e.g. "Error logging in"). I looked around a bit and noticed that `ImmichLogger` doesn't care about `record.error` - maybe it should save that attribute as well, so that the error description could be kept short and human-readable, while error details could be displayed in the details page?
3. Should the stack trace be included or not?

shenlong-tanwen:
Thanks a lot for working on this. I wanted to do this for a long time but never got around to do this.
 
>     1. Is there any overall reason not to do this?

There isn't but one issue that I can think of is that if the logs are left uncleared, we'll be keep on adding entries into the db but it shouldn't be that big of an issue though. I'd love more errors to be logged with more suitable logger levels though. Not all info logs we have are required by default and few can be moved to fine / finer level and stuff.

>     2. Is the error message text fine like this? To me it feels a bit strange to merge the error details (usually the raw server response) into the error description (e.g. "Error logging in"). I looked around a bit and noticed that `ImmichLogger` doesn't care about `record.error` - maybe it should save that attribute as well, so that the error description could be kept short and human-readable, while error details could be displayed in the details page?

This is something that has to be fixed as well. Can you fix this in a separate PR though? i.e, Update ImmichLogger to use the error from the record.
 
>     3. Should the stack trace be included or not?

It depends. For server side issues and http errors, we might not need the stack trace as long as it is clear where the error is from based on the error message. But since we rely on the openapi generated dart client for http calls as well, a stack trace can be used to catch errors in the auto generated code though

rovo89:
> Update ImmichLogger to use the error from the record.

I'll see what I can do, I think this should be the first step. I've never worked with Dart and all the frameworks before, so I'll have to see what is generated by what, how to change the database layout and add migrations (for the app database) etc. Any pointers to docs for these things would be helpful.

rovo89:
Ended up being easier than I thought (unless I missed something). 😄 Just had to add the new column to `mobile/lib/shared/models/logger_message.model.dart` and run `flutter pub run build_runner build` to regenerate the accompanying `.g.dart` file.

shenlong-tanwen:
@rovo89 Do you still plan to work on this or can we close this PR so you can open one later?

rovo89:
I'll close the PR for now as I'm busy with other things. When I have more time, I'll send a new one or reopen this one.

PR #6871, opened 2024-02-03
fix(ml): openvino not working with dynamic axes

mertalev:
## Description

There's an upstream bug with OpenVINO where it can't handle ONNX models with dynamic axes. This can cause errors for CLIP models as they all have dynamic axes. Facial recognition models are fixed at a batch size of 1, so they don't have this issue.

Fixes #6869


jrasm91 (approved):
LGTM

Issue #6647, opened 2024-01-26
PERF: switch from `axios` to `fetch`


benmccann:
### The bug

The typescript-axios generator pulls in an extra dependency on `axios`. The typescript-fetch generator could avoid this by using the browser's built-in `fetch` method.

Note that `fetch` is not supported in Internet Explorer (https://caniuse.com/fetch). I didn't see in the docs which browsers are supported, but I'm guessing IE is not a concern since it's a fairly recent app

### The OS that Immich Server is running on

any

### Version of Immich Server

any

### Version of Immich Mobile App

any

### Platform with the issue

- [X] Server
- [X] Web
- [X] Mobile

### Your docker-compose.yml content

```YAML
n/a
```

### Your .env content

```Shell
n/a
```

### Reproduction steps

```bash
https://github.com/immich-app/immich/blob/main/open-api/bin/generate-open-api.sh
https://github.com/immich-app/immich/blob/ca28e1e7a82b4467012e27934c2cddda39ea2c6c/.github/workflows/dispatch_sdk_update.yml#L23
```

jrasm91:
I'd be fine with this, just implies testing and validating the new client. Specifically, the file/attachment upload endpoints have always been tricky with the auto-generated clients.

benmccann:
I took a look at this and overall it's quite nice as it removes the need to do `.data` on every response.

The trickiest thing is figuring out how to re-implement `onUploadProgress` and `onDownloadProgress` from axios. Firefox/Safari don't have full support for the necessary APIs (one only supports upload and the other only supports download):
- https://github.com/axios/axios/issues/1219#issuecomment-1281827059
- https://developer.mozilla.org/en-US/docs/Web/API/Request/body
- https://developer.mozilla.org/en-US/docs/Web/API/Response/body

So the only way to monitor progress of these large file transfers appears to be to be to drop down and use XHR directly for that API call rather than using the `typescript-fetch` API for that specific call.

PR #6876, opened 2024-02-03
chore: remove `form-data` dependency

benmccann:
This uses the standard built-in class for `FormData` vs an external package. It will also let us remove `axios` as a dependency in a follow-up (https://github.com/immich-app/immich/issues/6647).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -103,6 +103,11 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
       - name: Run setup typescript-sdk
         run: npm ci && npm run build
         working-directory: ./open-api/typescript-sdk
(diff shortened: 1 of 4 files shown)


mertalev (approved):

jrasm91:
Does this still stream the file from disk? Also, ever since the CLI was converted to esm it no longer works. TSC doesn't modify the import paths and esm requires extensions. Can we use vite to bundle the CLI to fix this?

benmccann:
Good call outs. Let me check into these.

How about if we add the file extensions rather than bundling? I can update the tsconfig to fail the build if the file extensions are missing

benmccann:
Although maybe bundling is better as it will result in a smaller distributable. I'm not quite sure how the CLI is distributed currently

jrasm91:
> Good call outs. Let me check into these.
> 
> How about if we add the file extensions rather than bundling? I can update the tsconfig to fail the build if the file extensions are missing

I already looked into this, but the generated typescript-sdk is typescript files, which don't have extensions in the imports, so I'm not sure how to address the issue in that code. The rest of it was easy enough to change.

benmccann:
> Does this still stream the file from disk?

It seems we are very lucky to have to support only Node 20, which makes life here vastly easier. This PR uses `openAsBlob` which should allow for streaming the data: https://github.com/nodejs/undici/issues/2202#issuecomment-1666688797. We do wrap the result of that in `new File`, so I don't know if we'd need to remove that or not. If it's not streaming, it seems like it'd probably be a small change to make it do so. I could probably use some help in testing whether it does stream or not

PR #6858, opened 2024-02-02
feat(cli): dockerize

etnoy:
* Create docker image for v2 cli.
* bump npm commander to v12
* add documentation for docker usage of cli
* fix module not found and allow the cli to run with esm

To be done after this PR: 
* push immich-cli image to ghcr

PR #6888, opened 2024-02-03
chore: remove axios dependency from CLI

benmccann:
Ref https://github.com/immich-app/immich/issues/6647

This generates a second version of the API in the `open-api` project that uses the `fetch` API that is built into Node and the browser rather than `axios`. This migrates the CLI to the new API and leaves the `web` project on the old API. We can later migrate the `web` project as well and then delete the old `axios` API. It was too big of a change to figure out how to migrate both in one go, so this more incremental approach was easier.

The `fetch` API is nicer because it removes a bunch of unnecessary `.data` accesses and it removes a dependency resulting in faster project installs and less security risk due to a smaller attack area. It will make a bigger difference in the `web` project where removing a dependency will speed page load times

`main`: dist/index.js  1,817.04 kB
This PR: dist/index.js  1,088.98 kB


jrasm91 (commented):
I haven't tested it yet, but this looks like a good approach. Have you been able to run/test the cli locally yet? I know fetch and axios handle errors differently. Have you been able to validate how errors handling works with the new client?

benmccann:
I believe Axios will automatically throw an exception where as `fetch` does not and you have to manually check the status code. However, the generated OpenAPI client checks the status code and throws and exception if the response wasn't okay, so in practice they end up being the same since we use them via the OpenAPI client. You can see in this PR there was one place where the underlying API was being used directly and there I had to check the status code

etnoy:
I tried it out and it seems to work well. Verified server-info and uploading assets.

etnoy (approved):

etnoy:
This is great and helped me fix my outstanding issues with #6858. I suggest we merge this now and then I can merge my PR and all will be happy.

Thanks Ben!

PR #6892, opened 2024-02-03
fix: add file extensions to address CLI ESM compatibility

benmccann:
I can follow up with forcing these extensions or bundling the project, but wanted to at least get a fix out in the meantime to unbreak the project for folks and I think even if we did end up bundling the project this is probably still a good practice.


benmccann:
Hmm. I guess this doesn't work yet. Will have to go back to the drawing board and send a fix a bit later

PR #6895, opened 2024-02-03
chore(mobile):  Put the real delete button before other delete variants

lukaszwawrzyk:
Hi!
After recent update that split the trash action into 3 buttons (https://github.com/immich-app/immich/pull/4505) the most frequently used button - actual delete from the phone and remote - requires scrolling to access.
I moved it before its 2 cousins so that it is visible right away.
I believe that this is the primary delete action so it should be first.


shenlong-tanwen:
Thank you for the PR. But more than being primary, it is the most destructive of the three as well, right after the remove local file. So wouldn't it make sense that it takes a bit more effort to actually perform this operation?

alextran1502 (approved):

PR #6898, opened 2024-02-03
refactor(mobile): move error details to separate DB column

rovo89:
As discussed in #6866, `logRecord.error` previously wasn't stored in the DB. Instead, a string copy of it was merged into the `message` in most places. Separate the two, so the log overview shows a brief description of the error while the details (usually an exception, but could be any string) can be found in the details page.

I adjusted many places to split the log message into its headline and the details, but for sure there will be some that I've missed. It shouldn't be a problem to adjust those later. There might also be places where adding further details as string could be helpful, like a few statistics when the upload has finished.

### DIFF
--- a/mobile/lib/modules/asset_viewer/services/image_viewer.service.dart
+++ b/mobile/lib/modules/asset_viewer/services/image_viewer.service.dart
@@ -39,7 +39,8 @@ class ImageViewerService {
           final failedResponse =
               imageResponse.statusCode != 200 ? imageResponse : motionReponse;
           _log.severe(
-            "Motion asset download failed with status - ${failedResponse.statusCode} and response - ${failedResponse.body}",
+            "Motion asset download failed",
+            "Status: ${failedResponse.statusCode}\nResponse: ${failedResponse.body}",
           );
           return false;
         }
@@ -76,7 +77,8 @@ class ImageViewerService {
 
         if (res.statusCode != 200) {
           _log.severe(
-            "Asset download failed with status - ${res.statusCode} and response - ${res.body}",
+            "Asset download failed",
+            "Status: ${res.statusCode}\nResponse: ${res.body}",
           );
           return false;
         }
@@ -98,7 +100,7 @@ class ImageViewerService {
         return entity != null;
       }
     } catch (error, stack) {
-      _log.severe("Error saving file ${error.toString()}", error, stack);
+      _log.severe("Error saving downloaded asset", error, stack);
       return false;
     } finally {
       // Clear temp files
--- a/mobile/lib/modules/asset_viewer/services/image_viewer.service.dart
+++ b/mobile/lib/modules/asset_viewer/services/image_viewer.service.dart
@@ -1,6 +1,7 @@
 import 'dart:io';
 
 import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:immich_mobile/extensions/response_extensions.dart';
 import 'package:immich_mobile/shared/models/asset.dart';
 import 'package:immich_mobile/shared/providers/api.provider.dart';
 import 'package:immich_mobile/shared/services/api.service.dart';
@@ -38,10 +39,7 @@ class ImageViewerService {
             motionReponse.statusCode != 200) {
           final failedResponse =
               imageResponse.statusCode != 200 ? imageResponse : motionReponse;
-          _log.severe(
-            "Motion asset download failed",
-            "Status: ${failedResponse.statusCode}\nResponse: ${failedResponse.body}",
-          );
+          _log.severe("Motion asset download failed", failedResponse.toLoggerString());
           return false;
         }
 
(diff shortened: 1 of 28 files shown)


rovo89:
@shenlong-tanwen: PTAL

mobile/lib/modules/asset_viewer/services/image_viewer.service.dart
```diff
@@ -39,7 +39,8 @@ class ImageViewerService {
           final failedResponse =
               imageResponse.statusCode != 200 ? imageResponse : motionReponse;
           _log.severe(
-            "Motion asset download failed with status - ${failedResponse.statusCode} and response - ${failedResponse.body}",
+            "Motion asset download failed",
+            "Status: ${failedResponse.statusCode}\nResponse: ${failedResponse.body}",
```
shenlong-tanwen, line 43: Can the HTTP errors be refactored to a custom object for errors and the actual logging string be moved to the `toString` method of the custom class instead?

shenlong-tanwen (commented):
Thanks a lot for the PR!

mobile/lib/modules/asset_viewer/services/image_viewer.service.dart
rovo89: Makes sense. I'll probably need a few days though because a) I'm new to Dart and b) family keeps me busy 😉

rovo89: Sorry, still didn't get around to do this. I hope tomorrow or on the weekend.

rovo89: I went a slightly different route by adding an extension for the `Response` class. So whenever there's something to log for an HTTP call, the `toLoggerString()` method can be called. I hope that is fine as well. If not, I can add a wrapper class with a `toString()` instead, but I don't see advantages with that.

mobile/lib/shared/views/app_log_detail_page.dart
```diff
@@ -194,11 +127,13 @@ class AppLogDetailPage extends HookConsumerWidget {
...
           children: [
-            buildLogMessage(logMessage.message),
+            buildTextWithCopyButton("MESSAGE", logMessage.message),
+            if (logMessage.details != null)
+              buildTextWithCopyButton("DETAILS", logMessage.details.toString()),
             if (logMessage.context1 != null)
               buildLogContext1(logMessage.context1.toString()),
             if (logMessage.context2 != null)
-              buildStackMessage(logMessage.context2.toString()),
+              buildTextWithCopyButton("STACK TRACE", logMessage.context2.toString()),
```
shenlong-tanwen, line 136: Can we update this to separate private widgets instead of using functions to return them?

shenlong-tanwen (approved):
LGTM!

mobile/lib/shared/views/app_log_detail_page.dart
rovo89: I have to look up what that means 😉 But I would prefer to do that in a separate PR along with various other changes I have in mind, so that this one can get merged and I can follow up with migrating other places from debugPrint to logger.

shenlong-tanwen: It is not a blocking suggestion. We can do it in a separate PR, yes. Can you re-base your change over main so we can get this merged?

rovo89: > Can you re-base your change over main so we can get this merged?

Done.

rovo89:
That's funny... first it complains that
```dart
          final failedResponse =
              imageResponse.statusCode != 200 ? imageResponse : motionReponse;
          _log.severe("Motion asset download failed", failedResponse.toLoggerString());
          return false;
```
is bad and formats it to
```dart
          final failedResponse =
              imageResponse.statusCode != 200 ? imageResponse : motionReponse;
          _log.severe(
              "Motion asset download failed", failedResponse.toLoggerString());
          return false;
```
(which is less readable from my point of view).
But somehow the linter isn't friends with the formatter and wants to have a final comma:
```dart
          final failedResponse =
              imageResponse.statusCode != 200 ? imageResponse : motionReponse;
          _log.severe(
              "Motion asset download failed", failedResponse.toLoggerString(),);
          return false;
```
The end? No, the formatter wants some changes again:
```dart
          final failedResponse =
              imageResponse.statusCode != 200 ? imageResponse : motionReponse;
          _log.severe(
            "Motion asset download failed",
            failedResponse.toLoggerString(),
          );
          return false;
```
Looks better at least, but it pushed a simply log message from one to four lines and made it seem more important than it needs to be. Similar for the line above that. I thought the formatter is supposed to improve readability but, the 80 character limit achieves the opposite in my eyes (literally :grin:).

OK. Rant over, Dart seems to be happy now, so hopefully ready to merge. 😉

alextran1502 (approved):

PR #6926, opened 2024-02-05
chore: build API with esnext target

benmccann:
Removes a bunch of polyfill code from the API that's unnecessary. The CLI and web projects are built with Vite and support for any newer features will already be added at that time if needed.

This should make the web frontend load a bit faster. I didn't benchmark the web project, but for reference here's the output difference for the CLI

Size of the CLI on `main`:
dist/index.js  1,088.98 kB

Size of CLI with this PR:
dist/index.js  760.50 kB

### DIFF
--- a/open-api/typescript-sdk/tsconfig.axios.json
+++ b/open-api/typescript-sdk/tsconfig.axios.json
@@ -1,6 +1,7 @@
 {
   "include": ["axios.ts", "axios-client/**/*"],
   "compilerOptions": {
+    "target": "esnext",
     "strict": true,
     "skipLibCheck": true,
     "declaration": true,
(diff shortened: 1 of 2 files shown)


etnoy:
Is it preferable to use esnext or to fix it to something like es2022?

etnoy (approved):

jrasm91 (approved):

benmccann:
esnext is best for packages like the API that aren't used directly, but are included and built as part of an application (the web package and CLI package in this case). When we build the web and CLI those packages specify their own target and compile the API in a way that works to ensure support for older browsers or older versions of Node.

danieldietzler (approved):

PR #6927, opened 2024-02-05
chore(cli): use upload api and update documentation

etnoy:
* Don't manually call fetch for the upload file
* Streamline the upload code
* Fix the include hidden option
* Improve documentation
* Bump to version 2.0.7

### DIFF
--- a/cli/src/commands/upload.command.ts
+++ b/cli/src/commands/upload.command.ts
@@ -49,7 +49,7 @@ class Asset {
     let sidecarData: Blob | undefined = undefined;
     try {
       await access(sideCarPath, constants.R_OK);
-      sidecarData = await fs.openAsBlob(sideCarPath);
+      sidecarData = new File([await fs.openAsBlob(sideCarPath)], basename(sideCarPath));
     } catch {}
 
     return {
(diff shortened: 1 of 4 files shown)


cli/src/commands/upload.command.ts
```diff
@@ -48,28 +49,18 @@ class Asset {
     let sidecarData: Blob | undefined = undefined;
     try {
       await access(sideCarPath, constants.R_OK);
-      sidecarData = new File([await fs.openAsBlob(sideCarPath)], basename(sideCarPath));
+      sidecarData = await fs.openAsBlob(sideCarPath);
```
jrasm91, line 52: Why doesn't this need to be a new file, but the asset data does?

jrasm91 (approved):
LGTM. Did you test uploading a sidecar file?

cli/src/commands/upload.command.ts
etnoy: After testing I reverted this.

benmccann:
Cool. Thanks! This was something I noticed and was on my list too 😆  Something that would be good to confirm if you didn't already is that the axios and fetch clients use the same date serialization

etnoy:
> Cool. Thanks! This was something I noticed and was on my list too 😆 Something that would be good to confirm if you didn't already is that the axios and fetch clients use the same date serialization

You don't want to know how long I fought with the axios sdk to get uploads working. In the end I had to concede to do manual http calls instead. 

With fetch, it just worked. Thanks for helping me with it

PR #6950, opened 2024-02-06
fix(web): Handle duplicate library settings gracefully

### LINKED ISSUE #6694: Web UI handling of duplicate library import paths
### The bug

It's possible to add the same import path to a library twice (oops) and save it.

Once it's been saved, if you try and remove it via the edit modal -> Delete button the UI removes both copies of the path. I didn't click Save to find out what the UI/API actually does if you try and commit the delete.

debian stable

### Version of Immich Server

v1.93.3

n/a

### Platform with the issue

- [X] Server
- [X] Web
- [ ] Mobile

```YAML
n/a
```

### Your .env content

```Shell
n/a
```

### Reproduction steps

```bash
see description
```

### Additional information

_No response_

etnoy:
Don't allow adding duplicate import paths. Disable save button if input is empty or duplicate. Add error text if duplicate path.
image

Add text saying import paths are empty
image

Also, the same thing for exclusion patterns


web/src/lib/components/forms/library-import-path-form.svelte
```diff
@@ -5,15 +5,27 @@
   import FullScreenModal from '../shared-components/full-screen-modal.svelte';
   import { mdiFolderSync } from '@mdi/js';
 
-  export let importPath: string;
+  import { onMount } from 'svelte';
```
jrasm91, line 8: can you not put new lines between import statements please?

jrasm91 (approved):
LGTM

web/src/lib/components/forms/library-import-path-form.svelte
etnoy: Oh, my bad. Will fix

web/src/lib/components/forms/library-scan-settings-form.svelte
```diff
@@ -48,13 +48,16 @@
     }
 
     try {
-      library.exclusionPatterns.push(exclusionPatternToAdd);
+      // Check so that exclusion pattern isn't duplicated
+      if (!library.exclusionPatterns.includes(exclusionPatternToAdd)) {
+        library.exclusionPatterns.push(exclusionPatternToAdd);
+        exclusionPatterns = library.exclusionPatterns;
+      }
```
danieldietzler, line 55: We're doing that check in 3 different places or so. Not sure if I'm a huge fan of this...

danieldietzler (approved):
Nice

web/src/lib/components/forms/library-scan-settings-form.svelte
etnoy: We check both in the modal for add/edit and in the handle function. I think both are needed. I'll likely also add a check in the server to prevent duplicates submitted via API.

jrasm91: Another option is to not do anything client side and auto dedupe server side instead

etnoy: We still need to dedupe client side because of the client-side bugs in #6694. I'll address the server side in another PR

PR #6954, opened 2024-02-06
fix(server) don't print password to console in admin-cli

### LINKED ISSUE #6953: Fix code scanning alert - Clear-text logging of sensitive information
Tracking issue for:
- [ ] https://github.com/immich-app/immich/security/code-scanning/33

etnoy:
We should not print the password to the console when changed via the admin cli

### DIFF
--- a/server/src/immich-admin/commands/reset-admin-password.command.ts
+++ b/server/src/immich-admin/commands/reset-admin-password.command.ts
@@ -13,7 +13,7 @@ export class ResetAdminPasswordCommand extends CommandRunner {
     super();
   }
 
-  ask = (admin: UserResponseDto) => {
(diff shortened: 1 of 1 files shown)


jrasm91 (commented):
The user isn't required to enter a password. When they don't a random one is selected. How do you communicate the password to the user without printing it to the console.

Normally, printing passwords with `console.log` is bad, but in this case that is the exact purpose of this command.

etnoy:
> The user isn't required to enter a password. When they don't a random one is selected. How do you communicate the password to the user without printing it to the console.
> 
> Normally, printing passwords with `console.log` is bad, but in this case that is the exact purpose of this command.

Thanks, closing. It wasn't easy to see that it could auto-gen passwords

Issue #6843, opened 2024-02-02
[Tech-Debt] Use luxon for user deletion calculation


jrasm91:
`user.service.ts#isReadyForDeletion` uses Date millisecond parsing, but should use luxon, for a much cleaner implementation.

```typescript
    // TODO use luxon for date calculation
    const msInDay = 86_400_000;
    const msDeleteWait = msInDay * 7;
    const msSinceDelete = Date.now() - (Date.parse(user.deletedAt.toString()) || 0);

    return msSinceDelete >= msDeleteWait;
```

manasadepu:
Hey, first time contributor here. Can I take this issue?

manasadepu:
`import { DateTime } from 'luxon';
    const msDeleteWait = 7 * 24 * 60 * 60 * 1000;
    const deletedAt  = DateTime.fromJSDate(user.deletedAt || new Date(0));
    const msSinceDelete = DateTime.now().diff(deletedAt).as('milliseconds');

    return msSinceDelete >= msDeleteWait;`

manasadepu:
This is my idea for the solution, but should I strive for conciseness?

jrasm91:
Feel free to open a pull request to discuss it

danieldietzler:
Done in #6958

PR #6958, opened 2024-02-07
fix(server): use luxon for deleted date calculation

a-m-n-s:
Fix for #6843


server/src/domain/user/user.service.ts
```diff
@@ -189,9 +190,9 @@ export class UserService {
     }
 
     // TODO use luxon for date calculation
-    const msInDay = 86_400_000;
-    const msDeleteWait = msInDay * 7;
-    const msSinceDelete = Date.now() - (Date.parse(user.deletedAt.toString()) || 0);
+    const msDeleteWait = 7 * 24 * 60 * 60 * 1000; // milliseconds in a week
```
jrasm91, line 193: Can you use luxon.Duration here instead?

jrasm91 (approved):

PR #6964, opened 2024-02-07
chore(cli): publish docker image on ghcr

etnoy:
This should push the docker version of the v2 cli to ghcr.

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -3,10 +3,14 @@ on:
   workflow_dispatch:
   push:
     branches: [main]
-    paths: ["cli/**"]
+    paths: 
+      - 'cli/**'
+      - '.github/workflows/cli.yml'
   pull_request:
     branches: [main]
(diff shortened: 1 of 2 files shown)


danieldietzler (approved):

.github/workflows/cli.yml
```diff
@@ -0,0 +1,99 @@
+name: CLI Build
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+    paths: 
+      - 'cli/**'
+      - '.github/workflows/cli.yml'
+  pull_request:
```
jrasm91, line 9: Do we really need to run these jobs on anything except workflow dispatch? I think you could simply this quite a bit by only doing the "release" portions of the jobs.

bo0tzz: Doing that would mean the flow is never tested except for when we want to make a release. That seems like a recipe for problems.

jrasm91 (approved):

PR #6984, opened 2024-02-08
chore(server): remove deprecated endpoints

jrasm91:


server/e2e/api/specs/download.e2e-spec.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    await testApp.reset();
+
+    await api.authApi.adminSignUp(server);
+    const admin = await api.authApi.adminLogin(server);
+
+    await api.userApi.create(server, admin.accessToken, userDto.user1);
+    user1 = await api.authApi.login(server, userDto.user1);
+    libraries = await api.libraryApi.getAll(server, user1.accessToken);
+    asset1 = await createAsset(user1, new Date('1970-01-01'));
```
danieldietzler, line 43: Don't we have a stub for dates already?

danieldietzler (approved):

server/e2e/api/specs/download.e2e-spec.ts
jrasm91: I'm not sure. I just copy/pasted the code from asset e2e to download e2e. There are actually a few things that are done different ways throughout the e2e tests that need to be cleaned up.

PR #6985, opened 2024-02-08
refactor(server): asset core

jrasm91:
- Move the two asset core methods into the asset service (the only place they are used).
- Remove obsolete/unused asset repository v1 methods


danieldietzler (approved):
Removing api-v1 stuff; always nice! :)

PR #6986, opened 2024-02-08
fix(cli): Run npm ci in sdk folder before deploying to npm

etnoy:
* Run npm ci and build in sdk before publishing cli
* Bump cli version

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -39,6 +39,7 @@ jobs:
       - name: Build SDK
         run: npm run build --prefix ../open-api/typescript-sdk/
       - run: npm ci
+      - run: npm run build
       - run: npm publish --dry-run
         if: ${{ github.event_name == 'workflow_dispatch' }}
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/cli.yml
```diff
@@ -73,10 +78,13 @@ jobs:
         id: metadata
         uses: docker/metadata-action@v5
         with:
+          flavor: |
+            latest=false
```
jrasm91, line 82: It only pushes on workflow dispatch so why can't we just turn this on and delete the conditional one below?

bo0tzz: Because the default one is triggered only when the workflow event is `tag`. We tried that before and it didn't work, which is why I added it here.

jrasm91: Ah, I see. I didn't realize the defaults for the action don't apply it when it's a manual dispatch.

jrasm91 (approved):

Issue #3490, opened 2023-07-31
[BUG] Photospheres forcefully sitched together at poles


dominikheinz:
### The bug

When a photosphere is not fully spherical, but only partially (see the pictures below for illustration), then immich shouldn't automatically assume the image is a complete sphere.

| Complete sphere | Partial sphere (poles not included) |
|---|---|
| ![Screenshot from 2023-07-31 23-19-32](https://github.com/immich-app/immich/assets/32770080/317486f2-3292-4682-bb35-a78992bac38c) | ![Screenshot from 2023-07-31 23-22-14](https://github.com/immich-app/immich/assets/32770080/9dc8aa30-6eda-4ed8-a313-2fa020856e2e) |

If you have a partial photosphere (wich isn't uncommon, think of it like a 360 panorama), then immich stretches the image all the way to the poles resulting in a totally distorted image (See below).

Instead of stretching the photosphere to the poles, the free spots should have a neutral background color, e.g. gray or black, and not stretching should occur. This is how it is for example handled by [Aves](https://github.com/deckerst/aves) and many other mobile gallery apps.

| Current behaviour | Expected behaviour |
|---|---|
| ![Screenshot from 2023-07-31 23-23-04](https://github.com/immich-app/immich/assets/32770080/8e7e7677-e984-4a89-ac5e-1e1aef5b5a91) | ![signal-2023-07-31-232955_002](https://github.com/immich-app/immich/assets/32770080/74467a11-9d24-4efa-9e99-4254aab96fab) |

### The OS that Immich Server is running on

Arch Linux

### Version of Immich Server

v1.71.0

### Version of Immich Mobile App

v1.71.0 build.94

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
- Not relevant -
```

### Your .env content

```Shell
- Not relevant -
```

### Reproduction steps

```bash
1. Create a partial photosphere, e.g. with Google Camera on a Pixel, where the poles dont have any coverage
2. Upload the image to Immich.
3. View the photosphere in Immich and observe the distorted photosphere
```

EnochPrime:
Related to #3465

hoopahmadness:
Not only are they stretched together at the poles, but if the scene is not fleshed out in a full 360 degree circle then the incomplete sphere is yanked together in the back too. The solution should support photospheres with an arbitrary completion around a circle as well as up/down from the horizon.

FireBird4:
[...]

PR #3412, opened 2023-07-24
feat (web/server) 360 degrees Web panoramas [attempt 2]

dmitry-brazhenko:
Initial PR: https://github.com/immich-app/immich/pull/3098

In this PR I aded
1. Support for 360 Panoramas as it was discussed here: https://github.com/immich-app/immich/discussions/1630

**How it was implemented**
1. Once an images is uploaded, in [metadata-extraction.processor.ts](https://github.com/immich-app/immich/pull/3098/files#diff-ed7be838d2a88fd5de449b030f1fc0952987eba3b56897b7cb68a8b1442edbd0) it checks aspectRatio. We check exif (ProjextionType) property to determine if it is a panorama. 
2. If it is a panorama, we add a property projectionType as Enum (none/EQUIRECTANGULAR/...)
3. We use library @egjs/svelte-view360 to show the panorama
4. In the library we use a special symbol to display if a thumb is a panorama. 

**Potential further improvements:**
1. Probably we can switch library and use smth else. But this liibrary works OK both on desktop and mobile-web
2. Support panoramas in Flutter mobile app 
3. There 360 videos. Later we can support them as well

Here is thumb that shows a panorama:

Panorama after is opened:

PR #6992, opened 2024-02-09
fix (web): panorama / 360 view - partial panorama: use photo-sphere-viewer

dmitry-brazhenko:
Previous PR: https://github.com/immich-app/immich/pull/3412

This PR should fix 2 issues: https://github.com/immich-app/immich/issues/3490 and https://github.com/immich-app/immich/issues/3465

I changed 360 images viewer. Now I use photo-sphere-viewer. 

I checked "cropped/partial" panoramas and they work fine. 

image

### DIFF
--- a/web/src/lib/components/asset-viewer/photo-sphere-viewer-adapter.svelte
+++ b/web/src/lib/components/asset-viewer/photo-sphere-viewer-adapter.svelte
@@ -1,0 +1,35 @@
+<script lang="ts">
+  import { Viewer } from '@photo-sphere-viewer/core';
+  import '@photo-sphere-viewer/core/index.css';
+  import { onDestroy, onMount } from 'svelte';
+
+  export let panorama: string;
+  let container: HTMLDivElement;
+  let viewer: Viewer;
+
+  onMount(() => {
+    viewer = new Viewer({
+      container,
+      panorama,
+      touchmoveTwoFingers: true,
+      mousewheelCtrlKey: false,
+      navbar: false 
+    });
+  });
+
+  onDestroy(() => {
+    if (viewer) {
+      viewer.destroy();
+    }
+  });
+</script>
+
+<div bind:this={container} />
+
+<style>
+  div {
+    width: 100%;
+    height: 100%;
+    margin-bottom: 0;
+  }
+</style>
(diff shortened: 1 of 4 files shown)


web/src/lib/components/asset-viewer/photo-sphere-viewer-adapter.svelte
```diff
@@ -0,0 +1,35 @@
...
+
+<div bind:this={container} />
+
+<style>
+  div {
+    width: 100%;
+    height: 100%;
+    margin-bottom: 0;
+  }
+</style>
```
martabal, line 35: Can you use 
```html
<div class="h-full w-full mb-0" bind:this={container} />
``` 
instead ?

danieldietzler (approved):

dmitry-brazhenko:
@danieldietzler  could you help me to merge this PR?

danieldietzler:
> @danieldietzler could you help me to merge this PR?

I don't merge PRs, only some people do :)
Why is it so urgent? It's not like we'll do a release right after this PR

dmitry-brazhenko:
No urgency, just wanted not to forget about this :)

danieldietzler:
We usually don't "forget" pull requests, no worries.

mertalev:
The photo-sphere-viewer documentation says that it handles cropped/partial panoramas by looking at image metadata. We strip most metadata from preview images now, so can you confirm that a newly uploaded partial panorama still works?

alextran1502 (approved):

alextran1502:
@mertalev tested on panorama taken on iPhone and it works

dmitry-brazhenko:
Yep, I also tested on newly uploaded images. As far as I see, exif is not removed from image

rovo89:
@dmitry-brazhenko I think panorama-viewer.css is now unused, isn't it? Then I guess it should be deleted.

dmitry-brazhenko:
@rovo89  thanks a lot for your attentiveness!!

Here is a PR: https://github.com/immich-app/immich/pull/7012

waclaw66:
A few comments...
- panorama photos got attributes `InitialViewHeadingDegrees` and `InitialViewPitchDegrees`, ignoring them causes that image is rotated to the wrong side initially, they should be kept and set to `poseHeading` and `posePitch` panorama viewer properties
- there is a white vertical line between info panel and panorama, it's because viewport of panorama has partial dimensions, but the viewer uses integer (truncated) ones
 
obrazek

dmitry-brazhenko:
That's good point. I created an issue: https://github.com/immich-app/immich/issues/7243

PR #6996, opened 2024-02-09
fix: update lodash calls

benmccann:
The web is currently broken from my last PR. Guess I trusted TypeScript and the tests too much. This fixes it

### DIFF
--- a/web/src/lib/utils/timeline-util.ts
+++ b/web/src/lib/utils/timeline-util.ts
@@ -1,5 +1,5 @@
 import type { AssetResponseDto } from '@api';
-import { chain } from 'lodash-es';
+import { groupBy, sortBy } from 'lodash-es';
 import { DateTime, Interval } from 'luxon';
 
 export const fromLocalDateTime = (localDateTime: string) => DateTime.fromISO(localDateTime, { zone: 'UTC' });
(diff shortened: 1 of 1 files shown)


web/src/lib/utils/timeline-util.ts
```diff
@@ -45,8 +45,8 @@ export function splitBucketIntoDateGroups(
...
   locale: string | undefined,
 ): AssetResponseDto[][] {
-  return chain(assets)
-    .groupBy((asset) => fromLocalDateTime(asset.localDateTime).toLocaleString(groupDateFormat, { locale }))
-    .sortBy((group) => assets.indexOf(group[0]))
-    .value();
+  const grouped = groupBy(assets, (asset) =>
+    fromLocalDateTime(asset.localDateTime).toLocaleString(groupDateFormat, { locale }),
+  );
+  return sortBy(grouped, (group) => assets.indexOf(group[0]));
```
danieldietzler, line 51: Unrelated to this PR but isn't this `assets.indexOf` *really* slow given how often we call it?

danieldietzler (approved):

web/src/lib/utils/timeline-util.ts
mertalev: It does look slow 😅

PR #7005, opened 2024-02-09
fix: unnecessary .env file usage

jrasm91:


bo0tzz (approved):

gboudreau:
I was wondering why the `env_file` value was not needed.  
The lines that follow (`environment` and `volumes`) use variables that are defined in the `.env` file, eg. `POSTGRES_PASSWORD: ${DB_PASSWORD}`, so if you remove `env_file`, how is docker-compose loading those values?

Answering myself: variables defined in `env_file` are NOT available as variables in the compose file; they are injected as environment variables when running the container. Variables available in the compose file are defined in the `.env` file, in the same folder as the `compose.yml` file.  
Re-using `.env` as `env_file` makes sense, because that means Docker containers configurations will match values used by running containers.

PR #7009, opened 2024-02-09
chore: remove unneccessary async/awaits

benmccann:
Addressing @jrasm91's comment: https://github.com/immich-app/immich/pull/6973#discussion_r1484788267

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -67,6 +67,7 @@
       "license": "MIT",
       "devDependencies": {
         "@types/node": "^20.11.0",
+        "oazapfts": "^5.1.4",
         "typescript": "^5.3.3"
       },
       "peerDependencies": {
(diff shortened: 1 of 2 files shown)


jrasm91 (approved):
:pray:

web/package-lock.json
```diff
@@ -67,6 +67,7 @@
       "license": "MIT",
       "devDependencies": {
         "@types/node": "^20.11.0",
+        "oazapfts": "^5.1.4",
```
danieldietzler, line 70: Why is that change in here?

danieldietzler (approved):

web/package-lock.json
benmccann: guess it must have gotten missed somehow in my last PR. not sure why it didn't pick it up then

PR #7010, opened 2024-02-09
feat(server): restore modified at timestamp after upload, preserve when copying

jextrevor:
Created to address https://github.com/immich-app/immich/discussions/2581.

This change makes it so that the asset service restores the mtime of the asset file using the fileModifiedAt parameter.

Also, if renaming the file fails and the storage template service falls back to the copy and delete path instead, the filesystem provider now preserves the atime and mtime from the old file when copying.

This is my first time contributing - let me know if I missed something or what I can do to make this cleaner. Thanks!

(Also, not sure if I should be classifying this as a feature or a fix, I can change the commit message if needed)

### DIFF
--- a/server/src/immich/api-v1/asset/asset.service.ts
+++ b/server/src/immich/api-v1/asset/asset.service.ts
@@ -7,6 +7,7 @@ import {
   IAssetRepository,
   IJobRepository,
   ILibraryRepository,
+  IStorageRepository,
   IUserRepository,
   ImmichFileResponse,
   JobName,
@@ -55,6 +56,7 @@ export class AssetService {
     @Inject(IAssetRepository) private assetRepository: IAssetRepository,
     @Inject(IJobRepository) private jobRepository: IJobRepository,
     @Inject(ILibraryRepository) private libraryRepository: ILibraryRepository,
+    @Inject(IStorageRepository) private storageRepository: IStorageRepository,
     @Inject(IUserRepository) private userRepository: IUserRepository,
   ) {
     this.access = AccessCore.create(accessRepository);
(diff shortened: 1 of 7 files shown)


server/src/immich/api-v1/asset/asset.service.ts
```diff
@@ -80,10 +81,15 @@ export class AssetService {
       const libraryId = await this.getLibraryId(auth, dto.libraryId);
       await this.access.requirePermission(auth, Permission.ASSET_UPLOAD, libraryId);
       this.requireQuota(auth, file.size);
+      await this.fixModifiedAtTimestamp(file, dto);
       if (livePhotoFile) {
+        await this.fixModifiedAtTimestamp(livePhotoFile, dto);
```
jrasm91, line 86: These two can just go inside of the create method instead. Probably can move the sidecar one in there as well.

server/src/immich/api-v1/asset/asset.service.ts
```diff
@@ -369,4 +375,8 @@ export class AssetService {
       throw new BadRequestException('Quota has been exceeded!');
     }
   }
+
+  private async fixModifiedAtTimestamp(file: UploadFile, dto: CreateAssetDto) {
+    await utimes(file.originalPath, new Date(), new Date(dto.fileModifiedAt));
```
jrasm91, line 380: We don't call fs methods directly in services. Can you just add this to the storage repository and use that here instead? It makes the tests simpler.

jextrevor: Sounds good, I've added a `utimes` method to the storage repository/filesystem provider

server/src/immich/api-v1/asset/asset.service.ts
jextrevor: Sounds good, I've moved all timestamp updating into the `create` method

jrasm91 (approved):
LGTM

danieldietzler (approved):
LGTM

PR #7016, opened 2024-02-10
refactor(mobile): Immich image provider

martyfuhry:
- `ImmichImage` now uses its own image providers for much better image loading
- Big performance improvements profiling the Gallery Pager
- Uses a remote image provider with three part image loading
- Local image provider for original files
- Keeps using AssetEntityImageProvider for thumbnails for now
- May fix #6726 since Alex identified the multipart image provider issue
- No longer uses multiple widgets for multipart loading, instead it is all done in the image provider
- Room for improvements and more optimizations later on
- Still using Fynn's image cache for old original image provider, renamed to ImmichLocalImageProvider


mobile/lib/shared/cache/custom_image_cache.dart
```diff
@@ -1,6 +1,5 @@
 import 'package:flutter/painting.dart';
-
-import 'original_image_provider.dart';
+import 'package:immich_mobile/modules/asset_viewer/image_providers/original_image_provider.dart';
```
fyfrey, line 2: This file should use the new local image provider. Otherwise, the both large and small images will end up in the same cache again.

martyfuhry: Thank you! I've removed the reference.

mobile/lib/shared/ui/immich_image.dart
```diff
@@ -147,64 +109,68 @@ class ImmichImage extends StatelessWidget {
...
+        ImmichLocalImageProvider(asset: asset),
         context,
+        onError: onError,
       );
     } else {
-      final accessToken = Store.get(StoreKey.accessToken);
-      // Precache the remote image since we are not using local images
       return precacheImage(
-        remoteThumbnailProvider(asset, type, {"x-immich-user-token": accessToken}),
+        ImmichRemoteImageProvider(assetId: asset.remoteId!),
```
shenlong-tanwen, line 134: `isThumbnail` has to be set after checking if the thumbnail format is WEBP. Alternatively, should we remove the type `param` and replace it with `isThumbnail` as a bool param instead?

mobile/lib/shared/ui/immich_image.dart
```diff
@@ -147,64 +109,68 @@ class ImmichImage extends StatelessWidget {
...
+  /// whether the asset should be local or remote
   /// Precaches this asset for instant load the next time it is shown
-  static Future<void> precacheAsset(
+  static Future<void> precacheAssetImageProvider(
     Asset asset,
     BuildContext context, {
-    type = api.ThumbnailFormat.WEBP,
-    size = 250,
+    api.ThumbnailFormat type = api.ThumbnailFormat.WEBP,
+    int size = 250,
```
shenlong-tanwen, line 123: I cannot find the usage for size in the precache method? Is this for the future once we add in the custom thumbnail impl instead of the one from photo_manager?

mobile/lib/shared/ui/immich_image.dart
```diff
@@ -19,8 +18,7 @@ class ImmichImage extends StatelessWidget {
     this.fit = BoxFit.cover,
     this.useGrayBoxPlaceholder = false,
     this.useProgressIndicator = false,
-    this.type = api.ThumbnailFormat.WEBP,
-    this.preferredLocalAssetSize = 250,
+    this.isThumbnail = false,
```
shenlong-tanwen, line 21: Can we make `isThumbnail` true by default and use a factory `ImmichImage.original` instead to fetch the original image?`ImmichImage` widget is used in few more places for Album thumbnails. The previous implementation also had WEBP by default.

mobile/lib/modules/asset_viewer/image_providers/immich_remote_image_provider.dart
```diff
@@ -0,0 +1,138 @@
...
+      scale: 1.0,
+      chunkEvents: chunkEvents.stream,
+    );
+  }
+
+  /// Whether to show the original file or load a compressed version
+  bool get _useOriginal => AppSettingsEnum.loadOriginal.defaultValue;
+
+  /// Whether to load the preview thumbnail first or not
+  bool get _loadPreview => AppSettingsEnum.loadPreview.defaultValue;
```
shenlong-tanwen, line 54: This should not use the default values and has to be fetched from the store

mobile/lib/modules/home/ui/asset_grid/thumbnail_image.dart
```diff
@@ -136,10 +136,8 @@ class ThumbnailImage extends StatelessWidget {
           tag: isFromDto
               ? '${asset.remoteId}-$heroOffset'
               : asset.id + heroOffset,
-          child: ImmichImage(
+          child: ImmichImage.thumbnail(
             asset,
```
shenlong-tanwen, line 140: `useGrayBoxPlaceholder` has to be passed to the factory constructor and should be passed down here. Without this, the main timeline do not show any grey placeholder while loading the thumbnails.

mobile/lib/shared/ui/immich_image.dart
```diff
@@ -48,67 +57,22 @@ class ImmichImage extends StatelessWidget {
...
+      ),
       width: width,
       height: height,
-      // keeping memCacheWidth, memCacheHeight, maxWidthDiskCache and
-      // maxHeightDiskCache = null allows to simply store the webp thumbnail
-      // from the server and use it for all rendered thumbnail sizes
       fit: fit,
-      fadeInDuration: const Duration(milliseconds: 250),
-      progressIndicatorBuilder: (context, url, downloadProgress) {
+      frameBuilder: (context, child, frame, wasSynchronouslyLoaded) {
```
shenlong-tanwen, line 71: `CachedNetworkImage` handled fade in implicitly so when we scroll through the timeline, the assets used to fade in during loading. Since we switched to a custom handling, this is lost now. We have to handle this in `frameBuilder` or `loadingBuilder`

shenlong-tanwen (changes requested):
Thanks a lot for working on this. I've added few comments, Can you please look into them?

mobile/lib/utils/image_url_builder.dart
```diff
@@ -59,6 +59,10 @@ String getImageUrl(final Asset asset) {
   return '${Store.get(StoreKey.serverEndpoint)}/asset/file/${asset.remoteId}?isThumb=false';
 }
 
+String getImageUrlFromId(final String id) {
+  return '${Store.get(StoreKey.serverEndpoint)}/asset/file/$id?isThumb=false';
+}
+
```
fyfrey, line 65: please refactor `getImageUrl` to use this function, otherwise its duplicated

mobile/lib/shared/ui/immich_image.dart
fyfrey: yeah, we use the thumbnails more frequently.

mobile/lib/modules/activities/widgets/activity_tile.dart
```diff
@@ -106,7 +106,7 @@ class _ActivityAssetThumbnail extends StatelessWidget {
       decoration: BoxDecoration(
         borderRadius: const BorderRadius.all(Radius.circular(4)),
         image: DecorationImage(
-          image: ImmichImage.remoteThumbnailProviderForId(assetId),
+          image: ImmichRemoteImageProvider(assetId: assetId),
```
shenlong-tanwen, line 109: `isThumbail: true` is needed here since the previous implementation was only requesting a thumbnail

mobile/lib/shared/ui/immich_image.dart
```diff
@@ -1,143 +1,178 @@
...
-    this.type = api.ThumbnailFormat.WEBP,
-    this.preferredLocalAssetSize = 250,
+    this.isThumbnail = false,
+    this.thumbnailSize = 250,
     super.key,
   });
+
   final Asset? asset;
   final bool useGrayBoxPlaceholder;
   final bool useProgressIndicator;
```
shenlong-tanwen, line 29: `useProgressIndicator` can be removed since no usage

mobile/lib/shared/ui/immich_image.dart
```diff
@@ -1,143 +1,178 @@
...
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
-import 'package:flutter_cache_manager/flutter_cache_manager.dart';
 import 'package:immich_mobile/extensions/build_context_extensions.dart';
+import 'package:immich_mobile/modules/asset_viewer/image_providers/immich_local_image_provider.dart';
+import 'package:immich_mobile/modules/asset_viewer/image_providers/immich_remote_image_provider.dart';
 import 'package:immich_mobile/shared/models/asset.dart';
 import 'package:immich_mobile/shared/models/store.dart';
-import 'package:immich_mobile/utils/image_url_builder.dart';
+import 'package:immich_mobile/shared/ui/transparent_image.dart';
```
shenlong-tanwen, line 8: Unused import

mobile/lib/shared/ui/immich_image.dart
```diff
@@ -1,143 +1,178 @@
...
-              if (useProgressIndicator)
-                const Center(
-                  child: CircularProgressIndicator(),
-                ),
-            ],
+    final Asset asset = widget.asset!;
+
+    return OctoImage(
+      fadeInDuration: const Duration(milliseconds: 200),
+      fadeOutDuration: const Duration(milliseconds: 200),
```
shenlong-tanwen, line 140: This is more of a personal preference, but the following duration seems to create a more seamless fade out animation from the thumb-hash to the actual thumbnail. We don't want the image to fade-in, we just want the placeholder to fade-out instead

```suggestion
      fadeInDuration: const Duration(milliseconds: 0),
      fadeOutDuration: const Duration(milliseconds: 400)
```

shenlong-tanwen (changes requested):
Can you please check the outstanding comments? Also, we need to update `ImmichImage` to `ImmichImage.thumbnail` in the following two places:

- `SharedAlbumThumbnailImage`
- `SharingPage::buildAlbumList`

mobile/lib/modules/asset_viewer/image_providers/immich_remote_image_provider.dart
shenlong-tanwen:
```suggestion
  /// Whether to show the original file or load a compressed version
  bool get _useOriginal => Store.get(
        AppSettingsEnum.loadOriginal.storeKey,
        AppSettingsEnum.loadOriginal.defaultValue,
      );

  /// Whether to load the preview thumbnail first or not
  bool get _loadPreview => Store.get(
        AppSettingsEnum.loadPreview.storeKey,
        AppSettingsEnum.loadPreview.defaultValue,
      );
```

mobile/lib/modules/activities/widgets/activity_tile.dart
martyfuhry: Nice catch! Got it.

mobile/lib/shared/ui/immich_image.dart
martyfuhry: I like it. Let me try it locally and see how it works.

mobile/lib/shared/ui/immich_image.dart
martyfuhry: I would like to make a thumbnail widget separate from this in the future, so I'd prefer to keep the thumbnail constructor the factory and refactor from that later.

shenlong-tanwen (approved):
LGTM! Fantastic work <3

mobile/lib/shared/ui/immich_image.dart
martyfuhry: We use the `thumbnailSize` parameter now.

martyfuhry:
I left the thumbhash support out for now and we will reimplement it back after the image provider work gets merged for the sanity of our reviewers.

I've resolved everything once more and this is ready for review once more!

mobile/lib/modules/asset_viewer/image_providers/immich_local_image_provider.dart
```diff
@@ -0,0 +1,112 @@
...
+  //bool get _useOriginal => AppSettingsEnum.loadOriginal.defaultValue;
+  bool get _loadPreview => AppSettingsEnum.loadPreview.defaultValue;
+
+  // Streams in each stage of the image as we ask for it
+  Stream<ui.Codec> _codec(
+    Asset key,
+    ImageDecoderCallback decode,
+    StreamController<ImageChunkEvent> chunkEvents,
+  ) async* {
+    if (_loadPreview) {
```
fyfrey, line 51: this is technically not a preview in our sense; it's a thumbnail and we always load thumbnails.... so we could remove that check alltogether.

mobile/lib/modules/asset_viewer/image_providers/immich_local_image_provider.dart
```diff
@@ -0,0 +1,112 @@
...
+      codec: _codec(key, decode, chunkEvents),
+      scale: 1.0,
+      chunkEvents: chunkEvents.stream,
+      informationCollector: () sync* {
+        yield ErrorDescription(asset.fileName);
+      },
+    );
+  }
+
+  //bool get _useOriginal => AppSettingsEnum.loadOriginal.defaultValue;
```
fyfrey, line 42: not needed

mobile/lib/modules/asset_viewer/image_providers/immich_local_image_provider.dart
martyfuhry: Done, thanks!

mobile/lib/modules/asset_viewer/image_providers/immich_local_image_provider.dart
martyfuhry: Thank you!

fyfrey (approved):
LGTM! Can't wait to see it in action on my phone :)

PR #7028, opened 2024-02-11
feat(mobile): thumbhash support

shenlong-tanwen:
#### Changes made:

- Finally (Hopefully) a non-flickering thumb hash implementation in mobile

### DIFF
--- a/mobile/lib/shared/models/asset.dart
+++ b/mobile/lib/shared/models/asset.dart
@@ -36,7 +36,7 @@ class Asset {
         isOffline = remote.isOffline,
         stackParentId = remote.stackParentId,
         stackCount = remote.stackCount,
-        thumbhash = remote.thumbhash;
+        thumbhash = _decodeThumbhash(remote.thumbhash);
 
   Asset.local(AssetEntity local, List<int> hash)
       : localId = local.id,
@@ -118,7 +118,7 @@ class Asset {
   /// because Isar cannot sort lists of byte arrays
   String checksum;
 
-  String? thumbhash;
+  List<byte>? thumbhash;
 
   @Index(unique: false, replace: false, type: IndexType.hash)
   String? remoteId;
@@ -374,7 +374,7 @@ class Asset {
     ExifInfo? exifInfo,
     String? stackParentId,
     int? stackCount,
-    String? thumbhash,
+    List<byte>? thumbhash,
   }) =>
       Asset(
         id: id ?? this.id,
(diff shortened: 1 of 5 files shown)


mobile/lib/shared/models/asset.dart
```diff
@@ -116,6 +118,8 @@ class Asset {
   /// because Isar cannot sort lists of byte arrays
   String checksum;
 
+  String? thumbhash;
```
fyfrey, line 121: Please make this a `List<byte>` to store the raw bytes instead of the longer base64. This also avoids the wasteful decoding of base64 to bytes on every render.

shenlong-tanwen: Updated. Thank you 😄

mobile/lib/shared/models/asset.dart
```diff
@@ -504,3 +513,12 @@ extension AssetsHelper on IsarCollection<Asset> {
     return where().anyOf(ids, (q, String e) => q.localIdEqualTo(e));
   }
 }
+
+List<byte>? _decodeThumbhash(String? hash) {
+  if (hash == null) {
+    return null;
+  }
+  final hashBytes = base64.decode(base64.normalize(hash));
+  final rgbaImage = thumbhash.thumbHashToRGBA(hashBytes);
```
fyfrey, line 522: Sorry, maybe my previous comment was unclear.
This will now store the actual image in the database - for large collections this will blow up the database (it has a max size of 100s MB).
I only meant to decode the base64 string to bytes (so only use the first line of this function). The thumb hash calculation to make an image should still be performed when rendering.

shenlong-tanwen: My bad. Should be good now. Thank you :)

mobile/lib/shared/ui/immich_image.dart
```diff
@@ -174,3 +110,103 @@ class ImmichImage extends StatelessWidget {
...
+  // Creating the Uint8List from the List<bytes> during each build results in flickers during
+  // the fade transition. Calculate the hash in the initState and cache it for further builds
+  Uint8List? thumbHashBytes;
+  static const _placeholderDimension = 300.0;
+
+  @override
+  void initState() {
+    super.initState();
+    if (widget.asset?.thumbhash != null) {
+      final bytes = Uint8List.fromList(widget.asset!.thumbhash!);
```
fyfrey, line 124: it should be safe to cast the thumbhash to an Uint8List (after my other comment is addressed). Isar also returns its `List<byte>` as Uint8List. This removes 1 unnecessary array copy operation.

mobile/lib/shared/models/asset.dart
```diff
@@ -504,3 +512,10 @@ extension AssetsHelper on IsarCollection<Asset> {
     return where().anyOf(ids, (q, String e) => q.localIdEqualTo(e));
   }
 }
+
+List<byte>? _decodeThumbhash(String? hash) {
+  if (hash == null) {
+    return null;
+  }
+  return base64.decode(base64.normalize(hash)).toList();
```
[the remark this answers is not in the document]
martyfuhry: I'll merge this in to my branch and address this and the other comment there.

PR #7035, opened 2024-02-11
fix(web): remove warnings during the build

martabal:
Remove all warnings during the build.

### DIFF
--- a/web/src/routes/(user)/map/+page.svelte
+++ b/web/src/routes/(user)/map/+page.svelte
@@ -9,11 +9,12 @@
   import { mapSettings } from '$lib/stores/preferences.store';
   import { featureFlags } from '$lib/stores/server-config.store';
   import { type MapMarkerResponseDto, api } from '@api';
-  import { isEqual, omit } from 'lodash-es';
+  import { isEqual } from 'lodash-es';
   import { DateTime, Duration } from 'luxon';
   import { onDestroy, onMount } from 'svelte';
   import type { PageData } from './$types';
   import Map from '$lib/components/shared-components/map/map.svelte';
+  import type { MapSettings } from '$lib/stores/preferences.store';
 
   export let data: PageData;
 
@@ -35,6 +36,9 @@
   });
 
   $: $featureFlags.map || goto(AppRoute.PHOTOS);
+  const omit = (obj: MapSettings, key: string) => {
+    return Object.fromEntries(Object.entries(obj).filter(([k]) => k !== key));
+  };
 
   async function loadMapMarkers() {
     if (abortController) {
--- a/web/src/routes/(user)/map/+page.svelte
+++ b/web/src/routes/(user)/map/+page.svelte
@@ -14,6 +14,7 @@
   import { onDestroy, onMount } from 'svelte';
   import type { PageData } from './$types';
   import Map from '$lib/components/shared-components/map/map.svelte';
+  import type { MapSettings } from '$lib/stores/preferences.store';
 
   export let data: PageData;
 
--- a/web/src/routes/(user)/map/+page.svelte
+++ b/web/src/routes/(user)/map/+page.svelte
@@ -14,7 +14,6 @@
   import { onDestroy, onMount } from 'svelte';
   import type { PageData } from './$types';
   import Map from '$lib/components/shared-components/map/map.svelte';
-  import type { MapSettings } from '$lib/stores/preferences.store';
 
   export let data: PageData;
 
--- a/web/vite.config.js
+++ b/web/vite.config.js
@@ -12,20 +12,6 @@ const upstream = {
 };
 
 export default defineConfig({
-  build: {
-    // largest chunk size is maplibre-gl
-    chunkSizeWarningLimit: 800,
-    rollupOptions: {
-      output: {
-        // optimize chunks size https://github.com/vitejs/vite/discussions/9440
-        manualChunks(id) {
-          if (id.includes('node_modules')) {
-            return id.toString().split('node_modules/')[1].split('/')[0].toString();
-          }
-        },
-      },
-    },
-  },
   resolve: {
     alias: {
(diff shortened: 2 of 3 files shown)


web/vite.config.js
```diff
@@ -12,6 +12,9 @@ const upstream = {
 };
 
 export default defineConfig({
+  build: {
+    chunkSizeWarningLimit: 2_000_000,
```
benmccann, line 16: 2mb would be a really big chunk. It'd probably be nicer if we can figure out how to reduce chunk size instead of silencing this warning

web/src/routes/(user)/map/+page.svelte
```diff
@@ -35,6 +36,9 @@
   });
 
   $: $featureFlags.map || goto(AppRoute.PHOTOS);
+  const omit = (obj: MapSettings, key: string) => {
```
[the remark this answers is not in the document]
martabal: I don't know if it's a vite or a lodash issue, I haven't investigated further than that.
```ts
"omit" is imported from external module "lodash-es" but never used in "src/lib/components/elements/dropdown.svelte", "src/lib/components/admin-page/settings/admin-settings.svelte", "src/lib/stores/assets.store.ts", "src/lib/utils/timeline-util.ts", "src/lib/components/admin-page/settings/ffmpeg/ffmpeg-settings.svelte", "src/lib/components/admin-page/settings/machine-learning-settings/machine-learning-settings.svelte", "src/lib/components/admin-page/settings/map-settings/map-settings.svelte", "src/lib/components/admin-page/settings/thumbnail/thumbnail-settings.svelte", "src/lib/components/admin-page/settings/server/server-settings.svelte", "src/lib/components/admin-page/settings/trash-settings/trash-settings.svelte", "src/lib/components/admin-page/settings/new-version-check-settings/new-version-check-settings.svelte", "src/lib/components/admin-page/settings/library-settings/library-settings.svelte", "src/lib/components/admin-page/settings/logging-settings/logging-settings.svelte", "src/lib/components/admin-page/settings/job-settings/job-settings.svelte", "src/lib/components/admin-page/settings/password-login/password-login-settings.svelte", "src/lib/components/admin-page/settings/oauth/oauth-settings.svelte", "src/lib/components/admin-page/settings/storage-template/storage-template-settings.svelte", "src/lib/components/admin-page/settings/theme/theme-settings.svelte", "src/routes/(user)/albums/+page.svelte" and "src/routes/(user)/map/+page.svelte".
```
Generally speaking, I prefer to avoid using lodash

web/vite.config.js
martabal: That would be great, I think the problem comes from `Photosphere` and `Map`, without these modules the asset-viewer chunk size goes from `1,581.17 kB` to `178.95 kB` (uncompressed)

web/src/routes/(user)/map/+page.svelte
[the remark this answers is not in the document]
benmccann: Sounds good to me. I'll file an issue with Vite, but we might as well go with this in the meantime.

web/vite.config.js
```diff
@@ -12,6 +13,20 @@ const upstream = {
 };
 
 export default defineConfig({
+  build: {
+    // largest chunk size is maplibre-gl
+    chunkSizeWarningLimit: 800,
+    rollupOptions: {
+      output: {
+        // optimize chunks size https://github.com/vitejs/vite/discussions/9440
```
benmccann, line 21: I think this is going to create tons of small chunks which isn't really better either. Vite's defaults are pretty decent and so this feels unlikely to be an overall improvement

What I was thinking would be ideal would be if we could use a dynamic import to avoid putting some of the larger libraries in the main chunk. E.g. if I recall correctly, I think the map viewer is quite large and not loaded until you click into an image. Currently I believe it gets loaded straightaway and we pay the cost for it even if the user doesn't use it. If we do a dynamic import, we can wait to load it so that it doesn't slow down loading of the JavaScript needed for initial page load. E.g. maybe we could load it in `onMount` so that it doesn't block the initial JS loading, but is there in time for the user to use it

web/src/routes/(user)/map/+page.svelte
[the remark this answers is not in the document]
benmccann: I just filed an issue over in the Vite repo for this: https://github.com/vitejs/vite/issues/15890

web/src/routes/(user)/map/+page.svelte
```diff
@@ -9,11 +9,12 @@
   import { mapSettings } from '$lib/stores/preferences.store';
   import { featureFlags } from '$lib/stores/server-config.store';
   import { type MapMarkerResponseDto, api } from '@api';
-  import { isEqual, omit } from 'lodash-es';
+  import { isEqual } from 'lodash-es';
   import { DateTime, Duration } from 'luxon';
   import { onDestroy, onMount } from 'svelte';
   import type { PageData } from './$types';
   import Map from '$lib/components/shared-components/map/map.svelte';
+  import type { MapSettings } from '$lib/stores/preferences.store';
```
benmccann, line 17: I think this import is unused
```suggestion
```

web/vite.config.js
benmccann: Here's a PR to split the big chunk into two: https://github.com/immich-app/immich/pull/7057

They're both over the limit, so doesn't help in terms of removing the warning, but it should be a better user experience

I don't think there's a way to really shrink the photo sphere chunk further, so I guess we could set the limit at 600 kB to silence the warning for that one. We can probably do more with the asset viewer by splitting out the map stuff as well

martabal: Oh that's nice 😃

Yeah, I saw another project doing the same thing because of maplibre. Sounds reasonable.

benmccann (approved):

web/src/routes/(user)/map/+page.svelte
benmccann: ah, nevermind. I see now it was used by the new `omit` function

PR #7041, opened 2024-02-11
feat(ml): improve test coverage

mertalev:
### Description

This PR adds a bunch of tests, increasing unit test coverage from 80% to 84%, and increasing coverage with E2E tests included from 89% to 92%.

Other changes:
- Updates the E2E tests that were failing
  - In the case of the CLIP models, this was because we now normalize the embeddings from those models so they didn't match with the non-normalized embeddings
  - For facial recognition, the preprocessing is now done with NumPy so the numbers are *very* slightly different. 
  - These tests have been updated to use `allclose` for comparison to avoid floating point flakiness, with an absolute tolerance of 0.00000001.
- Tests no longer hardcode `/cache` and respect the `MACHINE_LEARNING_MODEL_CACHE` env for assertions
- Update test workflow to print lines with missing coverage
- Remove dead code that either only existed for reasons that are no longer relevant, or that were never used at all and only existed for the sake of completeness


jrasm91 (approved):
Lgtm

PR #7053, opened 2024-02-12
fix(ml): re-enable ARMNN integration

fyfrey:
ARMNN is not working on the current release. This PR re-enables ARMNN.

### DIFF
--- a/machine-learning/app/models/facial_recognition.py
+++ b/machine-learning/app/models/facial_recognition.py
@@ -28,7 +28,10 @@ class FaceRecognizer(InferenceModel):
 
     def _load(self) -> None:
         self.det_model = RetinaFace(session=self._make_session(self.det_file))
-        self.rec_model = ArcFaceONNX(self.rec_file.as_posix(), session=self._make_session(self.rec_file))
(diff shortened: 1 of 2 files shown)


machine-learning/app/models/facial_recognition.py
```diff
@@ -28,7 +28,10 @@ def __init__(
 
     def _load(self) -> None:
         self.det_model = RetinaFace(session=self._make_session(self.det_file))
-        self.rec_model = ArcFaceONNX(self.rec_file.as_posix(), session=self._make_session(self.rec_file))
+        self.rec_model = ArcFaceONNX(
+            self.rec_file.with_suffix(".onnx").as_posix(),
```
mertalev, line 32: Out of curiosity, what was the issue with passing the `.armnn` path here? Was `ArcFaceONNX` checking that it exists?

mertalev (approved):

machine-learning/app/models/facial_recognition.py
fyfrey: It analyzes the onnx model to get the input dimensions etc. This requires a file parsable by an onnx parser^^

PR #7055, opened 2024-02-12
feat(mobile): Adds blurhash

martyfuhry:
Adds blur hash to the thumbnails and animates them in using an AnimatedSwitcher from the asset grid.

We tried to implement thumbhash support in #7016 but it became a little unwieldy once we added in the Animated Switcher here.

This branch contains almost everything from that other PR, so it's completely dependent on that being merged in first. Once that merges in, this PR can be cleaned up significantly.


fyfrey (approved):
LGTM

shenlong-tanwen:
Closing in favor of #7279

PR #7064, opened 2024-02-12
fix(mobile): Multipart image loading for iOS double swipe

martyfuhry:


mobile/lib/modules/asset_viewer/image_providers/immich_local_image_provider.dart
```diff
@@ -49,18 +50,18 @@ class ImmichLocalImageProvider extends ImageProvider<Asset> {
     StreamController<ImageChunkEvent> chunkEvents,
   ) async* {
     if (_loadPreview) {
-      // TODO: Use local preview
+      // Load a small thumbnail
+      final thumbBytes = await asset.local?.thumbnailDataWithSize(
+        const ThumbnailSize.square(2000),
```
fyfrey, line 55: If we reuse the identical settings (default) add for thumbs on the grid, this will share the cached image because it's identical and super fast (already loaded for the grid view...)

Otherwise if this is too small: at least make the dimensions multiples of 16 (image codecs usually work in batches of 8x8 or 16x16) and set the quality to 80 (100 wastes a lot of space for almost identical visual quality as photo manager stores these thumbnails on the device)

mobile/lib/modules/asset_viewer/image_providers/immich_local_image_provider.dart
```diff
@@ -49,18 +50,18 @@ class ImmichLocalImageProvider extends ImageProvider<Asset> {
...
+        quality: 100,
+      );
+      if (thumbBytes == null) {
+        throw StateError("Loading thumb for ${asset.fileName} failed");
```
fyfrey, line 59: No need to error. We can simply try the original still..?

mobile/lib/modules/asset_viewer/image_providers/immich_local_image_provider.dart
martyfuhry: How does 256 with 80 quality sound?

mobile/lib/modules/asset_viewer/image_providers/immich_local_image_provider.dart
martyfuhry: Good point.

fyfrey (approved):

PR #7065, opened 2024-02-12
feat(server, web): Include partner's photos on map

ttyridal:
Include photos shared by partner to the map (map-marker endpoint) ~~if user has
chosen to include the shared photos on her timeline~~ if requested with mapOption / query parameter _withPartners_


server/src/domain/asset/asset.service.ts
```diff
@@ -187,8 +187,19 @@ export class AssetService {
     return folder;
   }
 
-  getMapMarkers(auth: AuthDto, options: MapMarkerDto): Promise<MapMarkerResponseDto[]> {
-    return this.assetRepository.getMapMarkers(auth.user.id, options);
+  async getMapMarkers(auth: AuthDto, options: MapMarkerDto): Promise<MapMarkerResponseDto[]> {
+    const userIds: string[] = [auth.user.id];
+    if (1) {
```
jrasm91, line 192: Is the idea to show partner assets under some condition? Like if the url contains `withPartner=true` or something?

ttyridal: for the `if(1)` you mean?   Yes,  I was thinking about it.. But kind of decided against it..  open for input! :)
I think it's "obvious" that you want the partners photos to behave like your own if `partner.inTimeline` is set.  The downside is that we'll always to the more complex DB-lookup.


will remove the if(1) or add the withPartner condition, depending on the input

ttyridal: Sure I can. Remove the dependency on `partner.inTimeline` then?

Arguably both convey the desire to include partner's photos in the response, and it's kind of illogical to require both.

jrasm91: I don't think we need to keep it or use it to filter the partner list, no.

web/src/lib/components/map-page/map-settings-modal.svelte
```diff
@@ -33,6 +33,7 @@
       <SettingSwitch title="Allow dark mode" bind:checked={settings.allowDarkMode} />
       <SettingSwitch title="Only favorites" bind:checked={settings.onlyFavorites} />
       <SettingSwitch title="Include archived" bind:checked={settings.includeArchived} />
+      <SettingSwitch title="Include shared with me" bind:checked={settings.withPartners} />
```
jrasm91, line 36: Shared with me is ambiguous here as assets can be shared via albums or partner sharing. Can we change this to just say include partner shared with me or something like that?

jrasm91 (approved):

web/src/lib/components/map-page/map-settings-modal.svelte
ttyridal: You are absolutely right! I didn't like it at all myself.   "include partner shared with me" will be very long and verbose compared to the other options.
English is not my naive language though, so I was out of other ideas 🤷‍♂️

Is it really a problem that it (could) include shared via albums?
I was playing with the idea of inversing the logic: "only my photos"...  Short, concise and would just exactly allow for such extensions.

True, it would let the user down at the moment, that it is _not_ including shared albums.

tbh I don't care much, I'm more interest in the end goal.   On your confirmation I'll change it to "include partner shared with me"

jrasm91: That's a good point. Let's leave it as is for now, knowing it only includes partner shared assets at the moment.

PR #7071, opened 2024-02-13
refactor: library watching

jrasm91:
Make testing the library service easier and decouple with chokidar types.


etnoy (approved):
I like this approach, makes things clearer and separates duties

server/src/domain/library/library.service.ts
```diff
@@ -116,63 +114,57 @@ export class LibraryService extends EventEmitter {
...
-      this.logger.debug(`Detected file change for ${path} in library ${library.id}`);
-
-      if (matcher(path)) {
-        // Note: if the changed file was not previously imported, it will be imported now.
-        await this.scanAssets(library.id, [path], library.ownerId, false);
-      }
-      this.emit('change', path);
-    });
+    let _resolve: () => void;
+    const ready$ = new Promise<void>((resolve) => (_resolve = resolve));
```
danieldietzler, line 118: Good idea!

danieldietzler (approved):
nice nice

PR #7076, opened 2024-02-13
feat(server,web) Semantic import path validation

etnoy:
Check if import paths are valid and warn user if they aren't.

This should help some users to prevent simple mistakes I've seen several times in discord

* Does the path exist?
* Is the path a directory?
* Do we have read access?

When scanning the library, import paths that do not pass validation are ignored

image

We also pop up a notification when the validation button is clicked
image


jrasm91:
Maybe the warning can go by the path itself not the action buttons

etnoy:
> Maybe the warning can go by the path itself not the action buttons

Yes, the front end stuff needs more work, I'm currently focusing on the backend first

open-api/immich-openapi-specs.json
```diff
@@ -3599,6 +3599,58 @@
...
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/ValidateLibraryDto"
+              }
+            }
+          },
+          "required": true
+        },
+        "responses": {
+          "201": {
```
jrasm91, line 3627: This should return a 200

jrasm91 (commented):
Looking at all of this, is there any reason these checks can't be run inside of the library create and update methods? It is a bit confusing if you can create an invalid external library and have to validate it independently.

bo0tzz:
IMO running the checks separately from changes makes sense. The library can break for external reasons (eg the mount goes offline), in which case having a check that immediately shows that would be nice.

jrasm91:
Sure, but that should be in addition to doing the check at time of creation imo.

open-api/immich-openapi-specs.json
etnoy: Done

jrasm91 (commented):
Looks pretty good!

server/src/domain/library/library.service.ts
```diff
@@ -509,6 +583,14 @@ export class LibraryService extends EventEmitter {
     return true;
   }
 
+  // Check if a given path is in a user's external path. Both arguments are assumed to be normalized
+  private isInExternalPath(filePath: string, externalPath: string | null): boolean {
+    if (externalPath === null) {
+      return false;
+    }
+    return filePath.match(new RegExp(`^${externalPath}`)) !== null;
```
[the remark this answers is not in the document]
etnoy: Done!

jrasm91 (approved):
LGTM

PR #7090, opened 2024-02-13
refactor(mobile): bring back backup selected local assets to timeline

shenlong-tanwen:
Follows #7000

#### Changes made:

- Local albums will not be rendered in the main timeline after 7000. This PR brings it back. After this change:
   - Assets that are a part of backup will be displayed in the main timeline
   - The remaining local assets can be accessed in Library -> Albums on Device even if they are not selected for backup. This would allow users to select and manually upload assets instead of relying on auto backup
- Moves the filtering asset to backup logic during album selection. (This would help us filter assets to be backed up during the actual backup process much easily. This will be added in a follow up PR)


mobile/lib/shared/models/asset.dart
```diff
@@ -35,7 +37,8 @@ class Asset {
         isReadOnly = remote.isReadOnly,
         isOffline = remote.isOffline,
         stackParentId = remote.stackParentId,
-        stackCount = remote.stackCount;
+        stackCount = remote.stackCount,
+        selectedForBackup = BackupSelection.none;
```
fyfrey, line 41: I'd really like to avoid adding this field to the asset class. Can we store this elsewhere?
E.g. a list of asset IDs with each selection state?

mobile/lib/shared/models/asset.dart
```diff
@@ -185,6 +190,12 @@ class Asset {
   @ignore
   bool get isImage => type == AssetType.image;
 
+  @enumerated
+  BackupSelection selectedForBackup;
+
+  @Backlink(to: Album.assetsLinkId)
+  final IsarLinks<LocalAlbum> localAlbums = IsarLinks<LocalAlbum>();
```
fyfrey, line 197: I'd like to avoid this as well as we can't move assets across isolates when they contain links

shenlong-tanwen:
@fyfrey Did the following changes to address all the comments:

- Moved backupSelection state to `DeviceAsset`
- Unified `AndroidDeviceAsset` and `IOSDeviceAsset` back into `DeviceAsset`. Changed the Isar ID to AutoIncrement. We might need to migrate the DB and recalculate all hash on device once. Will add this change if we are okay with proceeding with this change
- Removed the Isar links from asset collection and manually queries the localAlbums db when required

mobile/lib/shared/providers/asset.provider.dart
```diff
@@ -377,16 +384,27 @@ QueryBuilder<Asset, Asset, QAfterSortBy>? getRemoteAssetQuery(WidgetRef ref) {
...
+  final localIds =
+      ref.watch(deviceAssetsProvider).valueOrNull?.assetIdsForBackup ??
+          <String>[];
+
   return query
       .filter()
       .isArchivedEqualTo(false)
       .isTrashedEqualTo(false)
       .stackParentIdIsNull()
+      .group(
```
fyfrey, line 403: What's the purpose of group operation here?

fyfrey (approved):
Thanks for making these changes. Code looks good, didn't run it, though.

mobile/lib/shared/providers/asset.provider.dart
shenlong-tanwen: It is to make the timeline related filter for local assets more obvious.

PR #7097, opened 2024-02-13
refactor(web): use new open api client

jrasm91:
Migrate the follow apis to use the new fetch client:
- activity api
- album api
- library api
- audit api
- api key api
- authentication api
- face api
- job api
- oauth api
- partner api
- server info api
- system config api
- trash api
- user api

### axios-client/api.js

Before:
image

After:
image

PR #7103, opened 2024-02-13
refactor(web) open api client

jrasm91:
Continuation of #7097.

The only usages left are:
- Enums
- Requests that use response type blob
- Requests that use progress event
- Some types (identical) are imported from the axios package still I think.

image

### DIFF
--- a/web/src/lib/utils.ts
+++ b/web/src/lib/utils.ts
@@ -1,5 +1,6 @@
 import { goto } from '$app/navigation';
 import { page } from '$app/stores';
+import { JobName } from '@immich/sdk/axios';
 import { get } from 'svelte/store';
 
 interface UpdateParamAction {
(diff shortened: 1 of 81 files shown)


web/src/lib/utils.ts
```diff
@@ -31,3 +33,91 @@ export const updateParamList = async ({ param, value, add }: UpdateParamAction)
...
+    [JobName.Migration]: 'Migration',
+    [JobName.BackgroundTask]: 'Background Tasks',
+    [JobName.Search]: 'Search',
+    [JobName.Library]: 'Library',
+  };
+
+  return names[jobName];
+};
+
+let _key: string | undefined;
```
benmccann, line 56: This is fine for now, but having a single global key wouldn't work in the case that we had a cloud version of immich serving multiple users and doing ssr. It's not a change from the current code and not a blocker, but I thought I'd mention it anyway. Could potentially be worth a TODO. The typical way to avoid such a situation is with context: https://github.com/sveltejs/kit/discussions/4339#discussioncomment-2372710

benmccann (approved):

PR #7106, opened 2024-02-14
feat(server): only transcode streams that require it

mertalev:
## Description

Transcoding is currently "all-or-nothing": either both audio and video are transcoded, or neither are. This PR separates the two, meaning that transcoding is done only for the stream that requires it: the other can be copied as-is unless it also needs to be transcoded.

It also refactors the transcode policy code accordingly.

## How Has This Been Tested?

Tested that videos are transcoded successfully, that streams that should be transcoded based on a policy have a codec in the command instead of "copy" and that whether a stream is copied is responsive to the policy.


jrasm91 (approved):
LGTM

mingqlin:
"the other can be copied as-is unless it also needs to be transcoded."
why copied as-is? if no transcoding is required, should we not copy the original video into the encoded folder? it is wasted space.

mertalev:
If no transcoding is needed, then it still doesn't transcode. But if the audio stream needs to be transcoded, then the video has to either be transcoded or copied. The previous behavior was that it would always transcode the video, but now it will only transcode the video if it's required based on the transcode policy.

mertalev:
But I do see your point that this can increase space usage quite a bit, especially since the default transcode policy doesn't consider resolution. Hmm...

mingqlin:
I am ok with storage space usage if it is for a good cause. Think of YouTube, It always pre-transcode video into different qualities (fixed resolution and bitrate), this way its player could play different resolution base on users network speed. Another approach would be like Plex, which transcode real time, it will save a lot of space (no pre-transcoding)

mertalev:
> Another approach would be like Plex, which transcode real time, it will save a lot of space (no pre-transcoding, cons, it is slower response time when user click on play, it will take seconds to start if not longer)

We do want to add real-time transcoding at some point since it's a lot more efficient in terms of space and CPU usage. There are too many things to work on, not enough time haha

PR #7110, opened 2024-02-14
chore: enum support for new API

benmccann:
@jrasm91 has been doing a great job migrating the API over. One of the TODOs from https://github.com/immich-app/immich/pull/7103 was enums. I got that setup and working in this PR. I only migrated a single enum as a demonstration to avoid creating merge conflicts with his outstanding PR

Looks like this is causing a bunch of type errors, so there are a bunch of places we'll need to update imports from `@api` to `@immich/sdk`. It's probably best to wait for https://github.com/immich-app/immich/pull/7103 before doing that. Feel free to take this over as I have a busy day tomorrow or else I'll get back to it later in the week


jrasm91 (approved):
I've converted all imports and endpoints to use the new  enum types. The only things left are a few apis that use progress events and blob responses.

web/src/lib/components/admin-page/settings/job-settings/job-settings.svelte
```diff
@@ -29,7 +28,8 @@
     JobName.Migration,
   ];
 
-  function isSystemConfigJobDto(jobName: JobName): jobName is keyof SystemConfigJobDto {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
benmccann, line 31: I'm not much of a Typescript expert. Could this be `unkown` to avoid the `any`?

jrasm91: Doesn't look like it:
image

PR #7116, opened 2024-02-14
chore(cli): auto-sort imports

jrasm91:


cli/src/services/crawl.service.spec.ts
```diff
@@ -1,5 +1,5 @@
 import mockfs from 'mock-fs';
-import { CrawlService, CrawlOptions } from './crawl.service';
+import { CrawlOptions, CrawlService } from './crawl.service';
```
danieldietzler, line 2: Didn't know that even sorts within a destructed object, nice

danieldietzler (approved):

PR #7118, opened 2024-02-14
chore(web): auto sort imports

jrasm91:
Doesn't look like this plugin supports `.svelte` files yet, but it'll require and auto-fix `.ts` ones.

### DIFF
--- a/web/.prettierrc
+++ b/web/.prettierrc
@@ -3,6 +3,7 @@
   "trailingComma": "all",
   "printWidth": 120,
   "semi": true,
-  "plugins": ["prettier-plugin-svelte"],
+  "plugins": ["prettier-plugin-organize-imports", "prettier-plugin-svelte"],
+  "organizeImportsSkipDestructiveCodeActions": true,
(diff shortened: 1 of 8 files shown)


danieldietzler (approved):

benmccann (approved):

benmccann:
Looks like there's a feature request for Svelte support: https://github.com/simonhaenisch/prettier-plugin-organize-imports/issues/39

benmccann:
I'll note that this may make formatting quite a bit slower judging from their issue tracker: https://github.com/simonhaenisch/prettier-plugin-organize-imports/issues/114

benmccann:
I checked with the Svelte maintainers about Svelte support for the prettier plugin and one suggested that it'd be better to focus on getting something https://github.com/lydell/eslint-plugin-simple-import-sort working since it has more advanced sorting capabilities. E.g. it will put `node:`-prefixed imports into a separate group as opposed to intermixing with the other imports

jrasm91:
I looked at that briefly, but auto fixing eslinting errors seems quite a bit different than "format on save" functionality.

PR #7135, opened 2024-02-15
refactor(mobile): validate checksum before upload

shenlong-tanwen:
Follows #7090 

This is a WIP branch to test using checksum for validating assets before uploading them to the server.

#### Changes made in the PR

- Refactor BackupState into different classes specific to current backup state and backup settings
- Removeed `DuplicateAsset` collection
- Changes the background service to sync local albums first on asset change and use regular upload flow. This will help us further unify the upload flow

TODO:
- [ ] Verify background upload


mobile/lib/modules/backup/background_service/background.service.dart
```diff
@@ -344,35 +350,43 @@ class BackgroundService {
...
+    HashService hashService = HashService(db, this);
+    SyncService syncService = SyncService(db);
+    BackupAlbumService backupAlbumService = BackupAlbumService(db);
+    LocalAlbumService localAlbumService =
+        LocalAlbumService(db, hashService, syncService, backupAlbumService);
 
     await PhotoManager.setIgnorePermissionCheck(true);
 
     do {
+      await localAlbumService.refreshDeviceAlbums();
```
fyfrey, line 362: We need a new lock:

This function must not run concurrently with any sync operation in the foreground app.

So...we probably need to add the backup lock to the sync service (so it becomes a two-step lock, first either FG or BG gains lock, then individual tasks can run sequentially)

mobile/lib/modules/backup/models/backup_album.model.dart
```diff
@@ -14,7 +14,6 @@ enum BackupSelection {
 class BackupAlbum {
```
fyfrey, line 14: Can we get rid of this entire file and store all infos in the local album class?

mobile/lib/modules/backup/providers/backup.provider.dart
```diff
@@ -67,209 +53,76 @@ class BackupNotifier extends StateNotifier<BackUpState> {
...
+    }
 
-      Set<AssetEntity> assetsWillBeBackup = Set.from(state.allUniqueAssets);
-      // Remove item that has already been backed up
-      for (final assetId in state.allAssetsInDatabase) {
-        assetsWillBeBackup.removeWhere((e) => e.id == assetId);
-      }
+    await PhotoManager.clearFileCache();
+
+    final idsForBackup = await _db.deviceAssets
```
fyfrey, line 79: This is duplicated in the background service. Can we remove this duplication?

mobile/lib/shared/providers/asset.provider.dart
```diff
@@ -391,9 +422,13 @@ QueryBuilder<Asset, Asset, QAfterSortBy> _commonFilterAndSort(
...
-  final localIds =
-      ref.watch(deviceAssetsProvider).valueOrNull?.assetIdsForBackup ??
-          <String>[];
+  final localIds = ref
+      .read(dbProvider)
+      .deviceAssets
+      .filter()
+      .backupSelectionEqualTo(BackupSelection.select)
+      .idProperty()
+      .findAllSync();
```
fyfrey, line 431: Should this really be a synchronous operation? It can take quite some time..

fyfrey (commented):
Love it! Wanted to do this refactoring for a long time. Finally using the DB to get the assets to backup,.so much easier and faster.

Biggest issue is the required locking mechanism for the sync operation (must not run concurrently in foreground/background). It's certainly challenging. I can help with that once this PR is otherwise done :)

mobile/lib/modules/backup/models/backup_album.model.dart
shenlong-tanwen: We clear local albums on force refresh. Having this data stored in the local albums class results in the backup data getting removed during refresh

fyfrey: Good point.

PR #7139, opened 2024-02-15
feat(server, web): search location

martabal:
Search the location with a text box

## Screenshots

https://github.com/immich-app/immich/assets/74269598/9fd04fad-f245-4c10-b89e-4b2059f51109


server/src/infra/repositories/search.repository.ts
```diff
@@ -249,4 +260,20 @@ export class SearchRepository implements ISearchRepository {
...
+  @GenerateSql({ params: [DummyValue.STRING] })
+  async searchPlaces(placeName: string): Promise<GeodataPlacesEntity[]> {
+    return await this.geodataPlacesRepository
+      .createQueryBuilder('geoplaces')
+      .leftJoinAndSelect('geoplaces.admin1', 'admin1')
+      .leftJoinAndSelect('geoplaces.admin2', 'admin2')
+      .where(
+        '(LOWER(geoplaces.name) LIKE :nameStart OR LOWER(geoplaces.name) LIKE :nameAnywhere OR LOWER(admin1.name) LIKE :nameStart OR LOWER(admin1.name) LIKE :nameAnywhere OR LOWER(admin2.name) LIKE :nameStart OR LOWER(admin2.name) LIKE :nameAnywhere)',
+        { nameStart: `${placeName.toLowerCase()}%`, nameAnywhere: `% ${placeName.toLowerCase()}%` },
+      )
```
mertalev, line 273: I'm looking at optimizing this. We should add a [trigram indices](https://about.gitlab.com/blog/2016/03/18/fast-search-using-postgresql-trigram-indexes/) for the names and use `~* '\m<insert_text_here>'` for the comparison.

mertalev:
Actually, can't we use this for the search box as well? I think this is a sleek way to search for a location. @alextran1502

alextran1502:
@mertalev @martabal Wow, this is very cool. With the current search box, we know what is available; with this, we will need to perform an estimation. Am I understanding correctly?

mertalev:
The search box here uses `geodata_places`, but it should be possible to adapt this to the `exif` table too

server/src/infra/repositories/search.repository.ts
mertalev: Done

martabal: Nice, thank you !

The logic was a bit dumb. I want to improve it too by passing multiple input, so like if you type "France, Paris" this string will be split into "Paris" and "France". I want it to be accents and characters insensitive too.

Do you have any idea how to achieve this ?

martabal:
> @mertalev @martabal Wow, this is very cool. With the current search box, we know what is available; with this, we will need to perform an estimation. Am I understanding correctly?

Yes, but more like a suggestion

martabal:
@alextran1502 Yes, that probably can be used for your search

server/src/domain/search/dto/search.dto.ts
```diff
@@ -234,13 +234,33 @@ export class SearchDto {
...
+}
 
+export class SearchPeopleDto extends SearchPlacesDto {
   @IsBoolean()
   @Transform(toBoolean)
   @Optional()
   withHidden?: boolean;
 }
+
+export class PlacesReponseDto {
```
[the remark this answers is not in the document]
martabal: What's the typo ?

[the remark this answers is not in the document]
danieldietzler:
```suggestion
export class PlacesResponseDto {
```

server/src/infra/repositories/search.repository.ts
mertalev: The new query is accent and case insensitive, so we should be good to go there. "France, Paris" is a little trickier since the individual `name`, `admin1Name` and `admin2Name` fields are searched separately. I tweaked the query so it can handle that kind of input. It might need some more tweaking to get the ranking right.

server/src/domain/search/dto/search.dto.ts
```diff
@@ -234,13 +234,33 @@ export class SearchDto {
...
 }
 
-export class SearchPeopleDto {
+export class SearchPlacesDto {
   @IsString()
   @IsNotEmpty()
   name!: string;
+}
 
+export class SearchPeopleDto extends SearchPlacesDto {
```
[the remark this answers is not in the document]
martabal: I felt like copying the block just above ahah

server/src/infra/repositories/search.repository.ts
martabal: Hey, maybe we can extract the population to have a better sort. So when you search Washington, you prioritize the cities, with the most inhabitants

martabal: Oh we can even prioritize by features code (like capitals, populated places, airports ...)

server/src/domain/search/dto/search.dto.ts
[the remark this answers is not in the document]
martabal: Wow, I wasn't completely awake this morning, thank you

server/src/infra/repositories/search.repository.ts
```diff
@@ -172,6 +179,27 @@ export class SearchRepository implements ISearchRepository {
...
+      .where(`f_unaccent(name) %>> f_unaccent(:placeName)`)
+      .orWhere(`f_unaccent("admin2Name") %>> f_unaccent(:placeName)`)
+      .orWhere(`f_unaccent("admin1Name") %>> f_unaccent(:placeName)`)
+      .orWhere(`f_unaccent("alternateNames") %>> f_unaccent(:placeName)`)
+      .orderBy(
+        `
+        COALESCE(f_unaccent(name) <->>> f_unaccent(:placeName), 0) +
+        COALESCE(f_unaccent("admin2Name") <->>> f_unaccent(:placeName), 0) +
+        COALESCE(f_unaccent("admin1Name") <->>> f_unaccent(:placeName), 0) +
+        COALESCE(f_unaccent("alternateNames") <->>> f_unaccent(:placeName), 0)
```
danieldietzler, line 195: I trust this just does some magic that makes it work lol

web/src/lib/components/shared-components/change-location.svelte
```diff
@@ -38,8 +62,82 @@
       dispatch('cancel');
     }
   };
+
+  const showLocation = (name: string, admin1Name?: string, admin2Name?: string): string => {
```
danieldietzler, line 66: Nit: technically, `getLocation` makes more sense because it returns something and doesn't actually show it

danieldietzler (approved):
Nice work!

server/src/infra/repositories/search.repository.ts
mertalev: Haha let me explain:

`f_unaccent` removes accents from the words to make searching easier. The index uses this, so only `placeName` has to be un-accented at search time. But Postgres won't use the index if the columns don't also have `f_unaccent`, so we add this function to the columns too.

The `%>>`  operator checks if the two strings are similar using strict word similarity, meaning it checks against word boundaries instead of comparing the entire strings.

The `<->>>` operator, like `%>>`, compares two strings with strict word similarity, but the difference is that it returns a score instead of a boolean. Since it returns null if one of the strings is null, `COALESCE` prevents that from affecting the other comparisons.

We take the sum of all of those to make the ranking more robust. If the right location doesn't show up in the dropdown, the user can type more details about the state or country and the sum ranking will make the right location rank higher. If it was a strict ordering by city, state etc., then it might never rise to the top.

danieldietzler: Who tf figured `<->>>` is a good, precise operator for string comparison? lmao

server/src/immich/controllers/search.controller.ts
```diff
@@ -48,6 +49,11 @@ export class SearchController {
     return this.service.searchPerson(auth, dto);
   }
 
+  @Get('places')
+  searchPlaces(@Query() dto: SearchPeopleDto): Promise<PlacesResponseDto[]> {
```
alextran1502, line 53: why does `searchPlaces` use `SearchPeopleDto`?

alextran1502:
Amazing works! it looks and works very well, let's address my comment above and resolve the conflict then we can have this merged

waclaw66:
I'm getting

```
Error: 500 - 500
y@https://photos-dev.bolesiny.net/node_modules/.vite/deps/@oazapfts_runtime.js?v=80e2c728:152:5
l@https://photos-dev.bolesiny.net/node_modules/.vite/deps/@oazapfts_runtime.js?v=80e2c728:139:9
```

since this PR, is it possible to fix that something?
I did fresh `git clone`, remove everything from docker (including caches) and still occurs.

waclaw66:
I'm getting

```
Error: 500 - 500
y@https://photos-dev.server/node_modules/.vite/deps/@oazapfts_runtime.js?v=80e2c728:152:5
l@https://photos-dev.server/node_modules/.vite/deps/@oazapfts_runtime.js?v=80e2c728:139:9
```

since this PR, is it possible to fix that somehow?
I did fresh `git clone`, remove everything from docker (including caches) and still occurs.

martabal:
Does it work without your reverse proxy ?

waclaw66:
> Does it work without your reverse proxy ?

No, it's the same...
```
Error: 500 - 500
y@http://localhost:2284/node_modules/.vite/deps/@oazapfts_runtime.js?v=bc1795a2:152:5
l@http://localhost:2284/node_modules/.vite/deps/@oazapfts_runtime.js?v=bc1795a2:139:9
```

martabal:
Does it affect the whole webapp or only this modal ?

I'll take a look this evening

waclaw66:
It affects whole web app.
The root cause is probaly here in server...
```
Debugger listening on ws://0.0.0.0:9230/91b6385c-5e36-4d1f-8ad7-13cd3a97199e
For help, see: https://nodejs.org/en/docs/inspector
[Nest] 20  - 02/26/2024, 7:34:09 AM    WARN [AppService] Unable to open `www/index.html, skipping SSR.
[Nest] 20  - 02/26/2024, 7:34:09 AM     LOG [CommunicationRepository] Initialized websocket server
Migration "GeodataLocationSearch1708059341865" failed, error: permission denied for language c
/usr/src/app/src/driver/postgres/PostgresQueryRunner.ts:331
            throw new QueryFailedError(query, parameters, err)
                  ^
QueryFailedError: permission denied for language c
    at PostgresQueryRunner.query (/usr/src/app/src/driver/postgres/PostgresQueryRunner.ts:331:19)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at GeodataLocationSearch1708059341865.up (/usr/src/app/src/infra/migrations/1708059341865-GeodataLocationSearch.ts:5:5)
    at MigrationExecutor.executePendingMigrations (/usr/src/app/src/migration/MigrationExecutor.ts:336:17)
    at DataSource.runMigrations (/usr/src/app/src/data-source/DataSource.ts:403:13)
    at DatabaseRepository.runMigrations (/usr/src/app/src/infra/repositories/database.repository.ts:191:5)
    at /usr/src/app/src/domain/database/database.service.ts:50:7
    at /usr/src/app/src/infra/repositories/database.repository.ts:200:15
```

martabal:
Yes, it's a migration issue. Weird, I had no problems with this migration.

@mertalev Do you have any idea what could cause that ?

waclaw66:
I've found the cause, it was missing superuser privilige for postgres user, definitely my mistake, I'm sorry for confusion.

PR #7140, opened 2024-02-16
refactor(mobile): Use widgets in Immich asset grid

martyfuhry:
- Replaced all `buildThing` with a Widget
- This will ensure that the Flutter render engine can better cache and reuse existing Widgets
- Performance is about the same
- More boilerplate code, but good for future improvements and breaking the Widgets out a little more
- More readable
- Should have no regressions or changes
- I profiled it and it seemed about the same as `main`

### DIFF
--- a/mobile/lib/modules/home/ui/asset_grid/immich_asset_grid_view.dart
+++ b/mobile/lib/modules/home/ui/asset_grid/immich_asset_grid_view.dart
@@ -10,6 +10,7 @@ import 'package:immich_mobile/extensions/build_context_extensions.dart';
 import 'package:immich_mobile/extensions/collection_extensions.dart';
 import 'package:immich_mobile/modules/asset_viewer/providers/scroll_notifier.provider.dart';
 import 'package:immich_mobile/modules/home/ui/asset_grid/thumbnail_image.dart';
+import 'package:immich_mobile/modules/home/ui/asset_grid/thumbnail_placeholder.dart';
 import 'package:immich_mobile/shared/models/asset.dart';
 import 'package:scrollable_positioned_list/scrollable_positioned_list.dart';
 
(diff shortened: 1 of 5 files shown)


mobile/lib/modules/home/ui/asset_grid/thumbnail_placeholder.dart
```diff
@@ -0,0 +1,41 @@
...
+  Widget build(BuildContext context) {
+    final brightness = MediaQuery.platformBrightnessOf(context);
+    return Container(
+      width: width,
+      height: height,
+      margin: margin,
+      decoration: BoxDecoration(
+        gradient: LinearGradient(
+          colors: brightness == Brightness.light ? _brightColors : _darkColors,
+          begin: Alignment.topCenter,
```
shenlong-tanwen, line 35: Since we have a theme switcher from within the app, the apps theme need not be in sync with the platforms theme. Kindly use the isDarkTheme getter from the context itself, which checks the brightness from the current themedata of the context. Since it is already a boolean, we need not explicitly check for equality when assigning colors to the gradient as well

mobile/lib/modules/home/ui/asset_grid/immich_asset_grid_view.dart
```diff
@@ -485,3 +323,292 @@ class ImmichAssetGridViewState extends State<ImmichAssetGridView> {
...
+}
+
+/// A title row
+class _Title extends StatelessWidget {
+  final String title;
+  final List<Asset> assets;
+  final bool selectionActive;
+  final Function(List<Asset>) selectAssets;
+  final Function(List<Asset>) deselectAssets;
+  final Function(List<Asset>) allAssetsSelected;
```
shenlong-tanwen, line 499: NIT: Can we make the return type more explicit for the `allAssetsSelected` callback?

```suggestion
  final bool Function(List<Asset>) allAssetsSelected;
```

shenlong-tanwen (approved):
LGTM! The only thing that needs updating is the theme handling in placeholder

mobile/lib/modules/home/ui/asset_grid/thumbnail_placeholder.dart
martyfuhry: Thank you, I've addressed this in my revision.

mobile/lib/modules/home/ui/asset_grid/immich_asset_grid_view.dart
martyfuhry: I've fixed this in my revision.

mobile/lib/modules/home/ui/asset_grid/thumbnail_image.dart
```diff
@@ -138,6 +136,8 @@ class ThumbnailImage extends StatelessWidget {
               : asset.id + heroOffset,
           child: ImmichImage.thumbnail(
             asset,
+            height: 300,
```
fyfrey, line 139: Where does this 300 come from?
Thumbnails are 250px.

fyfrey (approved):

PR #7143, opened 2024-02-16
fix(docs): Add OpenVINO driver url

Rhevin:
Add OpenVINO driver url

If we don't install the driver, this path will not be found.

```
/dev/dri
```


mertalev:
I don't think this is the right link. You certainly don't need to install the whole OpenVINO runtime on the host for `/dev/dri` to appear.

Rhevin:
thanks for your correction
looks like you're right I removed this package and /dev/dri still exist

maybe because I install another intel packages

PR #7152, opened 2024-02-16
refactor(web): websocket events

michelheusschen:
Refactors the Websocket functionality by implementing a new event bus for real-time communication. The current implementation relies on stores for this purpose, but I don't think they're the right tool for the job. While stores are suitable for scenarios where a value needs to persist in memory and be updated, they are not the best choice for real-time events.

The `on_upload_success` event has been re-enabled to show uploaded assets without refreshing the page. Previously, this event was commented out but the reason for that isn't clear. Additionally, several small bugs are fixed related to using stores, including issues with unsubscribing and multiple updates not working as expected.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -29,6 +29,7 @@
       "devDependencies": {
         "@faker-js/faker": "^8.0.0",
         "@floating-ui/dom": "^1.5.1",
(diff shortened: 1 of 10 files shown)


danieldietzler (approved):
LGTM

jrasm91 (approved):
Welcome back!

PR #7156, opened 2024-02-16
chore: migrate to vitest

jrasm91:
* 20 seconds vs 40+ seconds for unit tests
* coverage doesn't seem to work for some reason
* run via swc
* remove dependency on jest-when


server/src/domain/trash/trash.service.spec.ts
```diff
@@ -8,16 +8,17 @@ import {
   newCommunicationRepositoryMock,
   newJobRepositoryMock,
 } from '@test';
+import { Mocked } from 'vitest';
 import { JobName } from '..';
```
danieldietzler, line 12: Not your fault but a bad import.

server/test/repositories/communication.repository.mock.ts
```diff
@@ -1,10 +1,11 @@
...
-export const newCommunicationRepositoryMock = (): jest.Mocked<ICommunicationRepository> => {
+export const newCommunicationRepositoryMock = (): Mocked<ICommunicationRepository> => {
   return {
-    send: jest.fn(),
-    broadcast: jest.fn(),
-    on: jest.fn(),
-    sendServerEvent: jest.fn(),
+    send: vi.fn(),
+    broadcast: vi.fn(),
+    on: vi.fn() as any,
```
danieldietzler, line 8: Why this `as any`?

danieldietzler (approved):
Nice!

server/test/repositories/communication.repository.mock.ts
jrasm91: this method has multiple signatures and it was throwing a type error.

server/package.json
```diff
@@ -123,42 +124,9 @@
     "ts-node": "^10.9.1",
```
benmccann, line 124: probably don't need this anymore

benmccann: also `@types/jest` above

jrasm91: It is still used in e2e, but I may try to migrate those here in this PR as well

PR #7177, opened 2024-02-18
chore(web): remove createEventDispatcher from LoginForm

benmccann:
Svelte 5 is encouraging callback props over custom events

### DIFF
--- a/web/src/lib/components/forms/login-form.svelte
+++ b/web/src/lib/components/forms/login-form.svelte
@@ -6,9 +6,13 @@
   import { oauth } from '$lib/utils';
   import { getServerErrorMessage, handleError } from '$lib/utils/handle-error';
   import { getServerConfig, login } from '@immich/sdk';
-  import { createEventDispatcher, onMount } from 'svelte';
+  import { onMount } from 'svelte';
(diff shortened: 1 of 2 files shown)


danieldietzler (approved):
That's an interesting transition. What's the reason behind that?

benmccann:
> That's an interesting transition. What's the reason behind that?

Events on elements (e.g. `on:click` on an `a` element) will be renamed to remove the `:` (e.g. `onclick`) and will be treated as regular props, which will enable event forwarding as has long been requested. As part of this, there was discussion around how to make events on components work in this new world and it was decided the best solution to the complications being introduced would be to remove them in favor of component props.

danieldietzler:
> > That's an interesting transition. What's the reason behind that?
> 
> Events on elements (e.g. `on:click` on an `a` element) will be renamed to remove the `:` (e.g. `onclick`) and will be treated as regular props, which will enable event forwarding as has long been requested. As part of this, there was discussion around how to make events on components work in this new world and it was decided the best solution to the complications being introduced would be to remove them in favor of component props.

Event forwarding! Yeeees!! :D

PR #7183, opened 2024-02-18
 feat(server): enable exiftool largefilesupport in metadata repo

fvdnabee:
Work in process

Fixes immich issue https://github.com/immich-app/immich/issues/4349
[Related exiftool-vendored.js comment](https://github.com/photostructure/exiftool-vendored.js/issues/109#issuecomment-1035595420)

### DIFF
--- a/server/src/infra/repositories/metadata.repository.ts
+++ b/server/src/infra/repositories/metadata.repository.ts
@@ -20,7 +20,7 @@ import {
 import { ImmichLogger } from '@app/infra/logger';
 import { Inject } from '@nestjs/common';
 import { InjectDataSource, InjectRepository } from '@nestjs/typeorm';
-import { DefaultReadTaskOptions, exiftool, Tags } from 'exiftool-vendored';
+import { DefaultExiftoolArgs, DefaultReadTaskOptions, ExifTool, Tags } from 'exiftool-vendored';
 import * as geotz from 'geo-tz';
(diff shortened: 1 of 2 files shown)


bo0tzz:
Thanks! Since this is an option that is not enabled in exiftool by default, presumably there are some considerations around setting it. I can't find much detail about what those are though, do you have any insight into that?

maxbraun91:
Hello @bo0tzz, I asked about that in the exiftool-Forum. Phil (the author of the tool) answered that this can cause a runtime error on older systems (which are not common any more). 

https://exiftool.org/forum/index.php?topic=15865.0

server/src/infra/repositories/metadata.repository.ts
```diff
@@ -20,31 +22,45 @@ import {
...
+  ) {
+    this.configCore = SystemConfigCore.create(configRepository);
+  }
 
   private logger = new ImmichLogger(MetadataRepository.name);
+  private exiftool: ExifTool;
 
   async init(): Promise<void> {
+    if (!this.subscription) {
+      this.subscription = this.configCore.config$.subscribe(() => this.init());
```
mertalev, line 59: This creates a new exiftool instance with the updated setting, but it wouldn't close the previous one.

fvdnabee:
See newer PR at https://github.com/immich-app/immich/pull/9894 which drops the system config.

PR #7186, opened 2024-02-18
fix: shell scripts

martabal:
Fix shell scripts which don't respect [shellcheck](https://github.com/koalaman/shellcheck) syntax.

Files found with this command : `find ./ -type f -name "*.sh" -not \( -path "**/open-api/**" -o -path "**/openapi/**" -o -path "**/node_modules/**" \) -exec shellcheck --format=gcc {} \;`

The `pump-version.sh` script bump the web version too.


martabal:
I don't know if it's worth adding a CI job for this

install.sh
```diff
@@ -1,15 +1,13 @@
+#!/bin/bash
```
mertalev, line 1: Change this to `#!/usr/bin/env bash` (also for the others, but keep `sh` as `sh`)

mertalev (approved):
Nice! I think a check would be fine if you'd like to add one

danieldietzler (approved):
So it isn't considered good practice to assign a value to an env variable directly? (without an intermediate variable)

martabal:
> So it isn't considered good practice to assign a value to an env variable directly? (without an intermediate variable)

Yes, if you use `export` https://www.shellcheck.net/wiki/SC2155

mertalev:
> In the original code, the return value of mycmd is ignored, and export will instead always return true. This may prevent conditionals, set -e and traps from working correctly.

Oh wow, I didn't know it ignored return values.

jrasm91 (approved):

PR #7187, opened 2024-02-18
chore: migrate more createEventDispatcher calls

benmccann:
Not ready yet... I think it might not be making some things async that should be. I need to check and see why `svelte-check` isn't catching these

### DIFF
--- a/web/src/lib/components/asset-viewer/asset-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/asset-viewer.svelte
@@ -756,9 +756,9 @@
   {#if isShowAlbumPicker}
     <AlbumSelectionModal
       shared={addToSharedAlbum}
-      on:newAlbum={({ detail }) => handleAddToNewAlbum(detail)}
-      on:album={({ detail }) => handleAddToAlbum(detail)}
-      on:close={() => (isShowAlbumPicker = false)}
+      onNewAlbum={handleAddToNewAlbum}
+      onAlbum={handleAddToAlbum}
+      onClose={() => (isShowAlbumPicker = false)}
     />
   {/if}
 
@@ -772,7 +772,7 @@
   {/if}
 
   {#if isShowProfileImageCrop}
-    <ProfileImageCropper {asset} on:close={() => (isShowProfileImageCrop = false)} />
+    <ProfileImageCropper {asset} onClose={() => (isShowProfileImageCrop = false)} />
   {/if}
 </section>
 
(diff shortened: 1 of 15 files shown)


web/src/lib/components/shared-components/base-modal.svelte
```diff
@@ -42,8 +40,8 @@
 >
   <div
     use:clickOutside
-    on:outclick={() => !ignoreClickOutside && dispatch('close')}
-    on:escape={() => dispatch('escape')}
+    on:outclick={() => !ignoreClickOutside && onClose && onClose()}
+    on:escape={() => onEscape && onEscape()}
```
jrasm91, line 44:
```suggestion
    on:escape={() => onEscape?.()}
```

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -756,9 +756,9 @@
   {#if isShowAlbumPicker}
     <AlbumSelectionModal
       shared={addToSharedAlbum}
-      on:newAlbum={({ detail }) => handleAddToNewAlbum(detail)}
-      on:album={({ detail }) => handleAddToAlbum(detail)}
-      on:close={() => (isShowAlbumPicker = false)}
+      onNewAlbum={handleAddToNewAlbum}
```
jrasm91, line 759: You handlers have a return signature of void, but the actual handler if often async. If svelte-check does catch that, it could be a problem.

Ideally, you would need to make the handler like this, and then you'd be forced to await it.
```ts
export let onNewAlbum: (search: string) => Promise<unknown>;
```

web/src/lib/components/shared-components/base-modal.svelte
jrasm91: FYI: You can use the `?.` null-ish chaining operator even when invoking functions

web/src/lib/components/asset-viewer/asset-viewer.svelte
benmccann: Yeah, I'm trying to figure out why `svelte-check`/`eslint` aren't catching it. That's why I sent the eslint PR (https://github.com/immich-app/immich/pull/7382), which I expected would catch this. Unfortunately, it's still not. Perhaps it's an issue with `eslint-plugin-svelte` since it's in the Svelte markup portion of the template. I'm leaving this broken and in draft mode until I track it down as I'm more concerned with having an automated check to avoid the error than fixing this particular instance of it. Next step on that journey is here: https://github.com/immich-app/immich/pull/7471

benmccann: I filed an issue in the `eslint-plugin-svelte` repo: https://github.com/sveltejs/eslint-plugin-svelte/issues/691

I'm going to close this PR until that's implemented as it's too hard to keep everything straight without the linter's help

PR #7189, opened 2024-02-18
fix: shell scripts

martabal:
optimize `start.sh`, format files and fix an incorrect variable introduced in #7186  (`DB_PASSWORD` is set twice)

### DIFF
--- a/server/start.sh
+++ b/server/start.sh
@@ -1,42 +1,17 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
-lib_path="/usr/lib/$(arch)-linux-gnu/libmimalloc.so.2"
-export LD_PRELOAD="$lib_path"
-
-if [ "$DB_URL_FILE" ]; then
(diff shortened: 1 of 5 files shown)


server/start.sh
```diff
@@ -1,42 +1,18 @@
...
+		unset "$1"
+	fi
+}
+
+read_file_and_export "DB_URL_FILE" "DB_URL"
+read_file_and_export "DB_HOSTNAME_FILE" "DB_HOSTNAME"
+read_file_and_export "DB_DATABASE_NAME_FILE" "DB_DATABASE_NAME"
+read_file_and_export "DB_USERNAME_FILE" "DB_USERNAME"
+read_file_and_export "DB_PASSWORD_FILE" "DB_PASSWORD"
+read_file_and_export "REDIS_PASSWORD_FILE" "REDIS_PASSWORD"
```
mertalev, line 16: Ooooh, fancy

mertalev (approved):

Issue #7052, opened 2024-02-12
Uploading large video file via immich cli fails


Froghut:
### The bug

When I try to upload a ~11GB video file via immich cli on windows I get the following error:
```
file:///C:/Users/user/scoop/persist/nodejs/bin/node_modules/@immich/cli/dist/index.js:9719
            throw new FetchError(e, "The request failed and the interceptors did not return an alternative response");
                  ^

FetchError: The request failed and the interceptors did not return an alternative response
    at AssetApi.fetchApi (file:///C:/Users/user/scoop/persist/nodejs/bin/node_modules/@immich/cli/dist/index.js:9719:19)
    at async AssetApi.request (file:///C:/Users/user/scoop/persist/nodejs/bin/node_modules/@immich/cli/dist/index.js:9771:22)
    ... 3 lines matching cause stack trace ...
    at async Command.<anonymous> (file:///C:/Users/user/scoop/persist/nodejs/bin/node_modules/@immich/cli/dist/index.js:22454:3) {
  cause: TypeError: fetch failed
      at node:internal/deps/undici/undici:12442:11
      at async AssetApi.fetchApi (file:///C:/Users/user/scoop/persist/nodejs/bin/node_modules/@immich/cli/dist/index.js:9704:20)
      at async AssetApi.request (file:///C:/Users/user/scoop/persist/nodejs/bin/node_modules/@immich/cli/dist/index.js:9771:22)
      at async AssetApi.uploadFileRaw (file:///C:/Users/user/scoop/persist/nodejs/bin/node_modules/@immich/cli/dist/index.js:13469:22)
      at async AssetApi.uploadFile (file:///C:/Users/user/scoop/persist/nodejs/bin/node_modules/@immich/cli/dist/index.js:13481:22)
      at async UploadCommand.run (file:///C:/Users/user/scoop/persist/nodejs/bin/node_modules/@immich/cli/dist/index.js:22389:30)
      at async Command.<anonymous> (file:///C:/Users/user/scoop/persist/nodejs/bin/node_modules/@immich/cli/dist/index.js:22454:3) {
    cause: RequestContentLengthMismatchError: Request body length does not match content-length header
        at AsyncWriter.write (node:internal/deps/undici/undici:9780:19)
        at writeIterable (node:internal/deps/undici/undici:9740:23) {
      code: 'UND_ERR_REQ_CONTENT_LENGTH_MISMATCH'
    }
  }
}
```

[...]

PR #7192, opened 2024-02-19
feat(cli): concurrent upload

mertalev:
### Description

When doing bulk uploads, I noticed that uploading through web was much faster than through the CLI (specifically when the upload component is minimized to avoid rendering slowdown). The former is able to upload files concurrently, while the latter currently does everything sequentially.

This PR addresses this limitation by refactoring the upload command to be "batch"-oriented - it runs a given number of promises at a time, finishing one step for all files before moving to the next step. Besides faster uploads, a benefit of this change is that it visualizes the progress of each step. In particular, this means users can now see how many of their assets were duplicates.

### DIFF
--- a/cli/src/commands/upload.command.ts
+++ b/cli/src/commands/upload.command.ts
@@ -134,7 +134,7 @@ export class UploadCommand extends BaseCommand {
 
     const assetsToCheck = files.map((path) => new Asset(path));
 
-    const { newAssets, duplicateAssets, rejectedAssets } = await this.checkAssets(
+    const { newAssets, duplicateAssets } = await this.checkAssets(
       assetsToCheck,
       options.concurrency ?? 4,
     );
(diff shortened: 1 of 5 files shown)


cli/src/commands/upload.command.ts
```diff
@@ -105,170 +115,333 @@ export class UploadOptionsDto {
...
+    } else {
+      console.log(`${messageStart} uploaded ${newAssets.length} assets (${byteSize(totalSizeUploaded)})`);
+    }
+
+    if (!options.album && !options.albumName) {
+      return;
+    }
+
+    const { createdAlbumCount, updatedAssetCount } = await this.updateAlbums(
+      [...newAssets, ...duplicateAssets, ...rejectedAssets],
```
mertalev, line 155: Side note that I think the distinction between duplicate and rejected assets is a bit murky. I was expecting existing assets to all show up under `duplicateAssets` and for `rejectedAssets` to only be for files that were an unsupported format, etc. But existing assets also show up as `reject`.

aviv926:
I wonder if this fixes the problem with large files as well?
https://github.com/immich-app/immich/issues/7052

cli/src/commands/upload.command.ts
```diff
@@ -9,15 +11,23 @@ import { ImmichApi } from 'src/services/api.service';
 import { CrawlService } from '../services/crawl.service';
 import { BaseCommand } from './base-command';
 
+const zipDefined = zip as <T, U>(a: T[], b: U[]) => [T, U][];
+
+enum CheckResponseStatus {
```
jrasm91, line 16: I think you should be able to just import this type from `@immich/sdk`

jrasm91 (approved):
LGTM

cli/src/commands/upload.command.ts
mertalev: Oops, this was just a bug in how I was checking the status

cli/src/commands/upload.command.ts
mertalev: The server divides the type into action (accept/reject) and reason (duplicate/invalid format), so there's no equivalent for this type in the sdk

PR #7219, opened 2024-02-20
feat(web): configure slideshow

martabal:
Add an option to configure the slideshow delay and show / hide the progress bar

## Screenshots

https://github.com/immich-app/immich/assets/74269598/4d1667a6-509c-4555-bcb5-865e2554fac3

### DIFF
--- a/web/src/lib/components/shared-components/progress-bar/progress-bar.svelte
+++ b/web/src/lib/components/shared-components/progress-bar/progress-bar.svelte
@@ -8,7 +8,6 @@
 <script lang="ts">
   import { createEventDispatcher, onMount } from 'svelte';
   import { tweened } from 'svelte/motion';
-  import { slideshowStore } from '$lib/stores/slideshow.store';
 
   /**
    * Autoplay on mount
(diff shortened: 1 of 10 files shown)


danieldietzler (approved):
Nice!

web/src/lib/components/shared-components/progress-bar/progress-bar.svelte
```diff
@@ -67,17 +78,13 @@
...
-  progress.subscribe((value) => {
-    if (value === 1) {
-      dispatch('done');
-    }
-  });
+  }
 </script>
 
-<span class="absolute left-0 h-[3px] bg-immich-primary shadow-2xl" style:width={`${$progress * 100}%`} />
+{#if $showProgressBar}
```
jrasm91, line 88: This means the component cannot be used anywhere else, since it's functionality is now coupled with the this slideshow setting.

jrasm91:
I've updated it to have a settings modal, similar to the map. It seems more consistent. I've also moved the shuffle setting into the same place now (less buttons on the screen all the time).

image

alextran1502 (approved):

PR #7232, opened 2024-02-20
feat(web): improve feedback for favorite and archive actions

michelheusschen:
When you change the favorite or archival status of an asset using the selection bar, changes aren't visible immediately (except when the asset gets removed from the view). Now the favorite and archive icons inside an asset are updated whenever they are changed.

Also some bug fixes for the `/albums/[id]` page:
- Archive icon inside assets never shows
- Selected assets that are archived cannot be unarchived
- Archive button shows when shared user doesn't own the asset


jrasm91 (approved):
LGTM

PR #7234, opened 2024-02-20
fix(web): websocket reconnect

michelheusschen:
The websocket connection doesn't properly re-establish after logging out and logging back in again. Fixed by using the `$user` store to detect whether the user is logged in.

### DIFF
--- a/web/src/lib/components/shared-components/navigation-bar/navigation-bar.svelte
+++ b/web/src/lib/components/shared-components/navigation-bar/navigation-bar.svelte
@@ -29,13 +29,13 @@
   }>();
 
   const logOut = async () => {
-    resetSavedUser();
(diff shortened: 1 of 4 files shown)


jrasm91 (commented):
Would it make more sense to set the store to null instead?

michelheusschen:
> Would it make more sense to set the store to null instead?

At first I didn't do that, because modifying the user store type requires lots of changes and would also cause errors because navigation hadn't completed yet. However, a workaround is possible by forcibly resetting the store to `undefined` and waiting until navigation has finished. Not the most elegant solution, but I've gone ahead and pushed the change.

jrasm91 (approved):
Nice, that looks better to me at least.

PR #7253, opened 2024-02-20
Update example.env add password note

aviv926:
Following the discord discussion [#](https://discord.com/channels/979116623879368755/1071165397228855327/1209286994090262528) [#](https://discord.com/channels/979116623879368755/1200197429518155907/1205215693516316682)
Add information on the correct way to change a password
You might want to change the wording of this.

### DIFF
--- a/docker/example.env
+++ b/docker/example.env
@@ -7,6 +7,9 @@ UPLOAD_LOCATION=./library
 IMMICH_VERSION=release
 
 # Connection secret for postgres. You should change it to a random password
+# NOTE: Password must be changed before docker compose up before the first time, after that do not change password via DB_PASSWORD.
+# To learn how to change the password later you can refer here: https://www.postgresql.org/docs/8.0/sql-alteruser.html
+
 DB_PASSWORD=postgres
 
 # The values below this line do not need to be changed
--- a/docker/example.env
+++ b/docker/example.env
@@ -7,7 +7,7 @@ UPLOAD_LOCATION=./library
 IMMICH_VERSION=release
 
 # Connection secret for postgres. You should change it to a random password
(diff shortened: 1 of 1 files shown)


docker/example.env
```diff
@@ -7,6 +7,9 @@ UPLOAD_LOCATION=./library
 IMMICH_VERSION=release
 
 # Connection secret for postgres. You should change it to a random password
+# NOTE: Password must be changed before docker compose up before the first time, after that do not change password via DB_PASSWORD.
```
danieldietzler, line 10:
```suggestion
# NOTE: This does not update the password for an already existing user.
```

bo0tzz:
In the spirit of keeping the file as minimal as we can - is this something we need to note here?

aviv926:
> In the spirit of keeping the file as minimal as we can - is this something we need to note here? 

I think so, it's required for those who don't know the database world closely, I think for most people this variable says "hey, you can change your database password here" that's half right and it can be confusing I think.

aviv926:
I am thinking of sending an update for the documentation soon, it will be possible to specify the information on a new page and give a link for this page. What do you think?

aviv926:
> Yeah, I'm not a fan of this change either. I saw just add a FAQ to the docs, that just links to the official postgres docker documentation where it clearly explains the purpose of this environment variable and its behavior.

Agreed, I believe I will send a PR soon.

jrasm91:
> > Yeah, I'm not a fan of this change either. I saw just add a FAQ to the docs, that just links to the official postgres docker documentation where it clearly explains the purpose of this environment variable and its behavior.
> 
> Agreed, I believe I will send a PR soon.

Awesome, thanks!

PR #7265, opened 2024-02-20
refactor(server): e2e

jrasm91:
Man, the person e2e tests are rough. There's not enough functionality exposed through the API at the moment.


e2e/src/api/specs/activity.e2e-spec.ts
```diff
@@ -245,11 +271,11 @@ describe(`${ActivityController.name} (e2e)`, () => {
     });
 
     it('should return a 200 for a duplicate like on the album', async () => {
-      const reaction = await api.activityApi.create(server, admin.accessToken, {
-        albumId: album.id,
-        type: ReactionType.LIKE,
-      });
-      const { status, body } = await request(server)
+      const [reaction] = await Promise.all([
```
danieldietzler, line 274: What's the reason for the promise.all here and a few lines below? Future-proofing?

e2e/src/api/specs/person.e2e-spec.ts
```diff
@@ -0,0 +1,176 @@
...
+    });
+
+    it('should return only visible people', async () => {
+      const { status, body } = await request(app)
+        .get('/person')
+        .set('Authorization', `Bearer ${admin.accessToken}`);
+
+      expect(status).toBe(200);
+      expect(body).toEqual({
+        total: 2,
```
danieldietzler, line 73: Why is that total 2?

danieldietzler (approved):
Nice! 🎉

e2e/src/api/specs/person.e2e-spec.ts
jrasm91: The response has total: number and items: Person[] or something like that. This returns both people.

danieldietzler: Shouldn't it only return the visible person?

jrasm91: Idk. Maybe the test case is mislabelled now. I just copied it is as.

Issue #6701, opened 2024-01-28
faces missing / rescanning does not help


lakemike:
### The bug

I am suspecting that my database has accumulated problems since my first docker installation some months ago: While some pictures were initially correctly scanned and faces were shown, these faces have disappeared over time, e.g. after updates and rescans. 

A complete redetection/rerecognition with latest version 1.93.3 does not fix the issue. This is actually weird: When I take an affected pictures (including faces), apply a crop to change checksum, I can reupload it and faces will then be recognized. I don't understand why the redetection/rerecognition fails on those other pictures in my database.

### The OS that Immich Server is running on

QNAP i386 Linux / Docker

### Version of Immich Server

v1.93.3

### Version of Immich Mobile App

v1.93.2 build 135

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
version: "3.8"

services:
  immich-server:
    container_name: immich_server
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    command: ["start.sh", "immich"]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - ${EXTPATH_LOCATION}:/mnt/media:ro
      - /etc/localtime:/etc/localtime:ro
    env_file:
      - .env
    ports:
      - 2283:3001
    depends_on:
      - redis
      - database
    labels:
      - traefik.enable=true
      - traefik.http.routers.immich.rule=Host(`myserver.domain.tld`)
      - traefik.http.services.immich.loadbalancer.server.port=3001
      - traefik.http.routers.immich.tls.certresolver=leresolver
      - traefik.http.routers.immich.entrypoints=websecure
    networks:
      - immichnet
    restart: always

  immich-microservices:
    container_name: immich_microservices
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    # extends:
    #   file: hwaccel.yml
    #   service: hwaccel
    command: ["start.sh", "microservices"]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - ${EXTPATH_LOCATION}:/mnt/media:ro
      - /etc/localtime:/etc/localtime:ro
    env_file:
      - .env
    depends_on:
      - redis
      - database
    networks:
      - immichnet
    restart: always

  immich-machine-learning:
    container_name: immich_machine_learning
[...]

PR #7278, opened 2024-02-21
fix(server): don't return archived assets by default

mertalev:
## Description

The search endpoint doesn't explicitly set `withArchived` right now, so it won't be part of the query unless the caller sets it to `false` themselves. This makes it default to false if not provided.

Fixes the behavior described in [this](https://github.com/immich-app/immich/issues/6701#issuecomment-1954813053) comment


jrasm91 (commented):
Would it be better to just default it in the dto instead? Also, why did you add another asset to the e2e? There was already an archived asset you should have been able to use.

mertalev:
> Would it be better to just default it in the dto instead? Also, why did you add another asset to the e2e? There was already an archived asset you should have been able to use.

There were too many tests failing with this change because they all reference asset3 and it's archived so it doesn't show by default now. I removed the `isArchived` flag from it and made a separate asset for this.

alextran1502 (approved):

PR #7319, opened 2024-02-21
fix(server, web): people page

martabal:
This PR moves the logic to hide people without thumbnail to the SQL query and fixes the count with total, visible and hidden people.

### DIFF
--- a/web/src/lib/components/faces-page/show-hide.svelte
+++ b/web/src/lib/components/faces-page/show-hide.svelte
@@ -6,6 +6,7 @@
   import { createEventDispatcher } from 'svelte';
   import LoadingSpinner from '$lib/components/shared-components/loading-spinner.svelte';
   import { mdiClose, mdiEye, mdiEyeOff, mdiRestart } from '@mdi/js';
+  import { locale } from '$lib/stores/preferences.store';
 
   const dispatch = createEventDispatcher<{
     close: void;
@@ -31,7 +32,7 @@
       <CircleIconButton icon={mdiClose} on:click={() => dispatch('close')} />
       <div class="flex gap-2 items-center">
         <p class="ml-2">Show & hide people</p>
-        <p class="text-sm text-gray-400 dark:text-gray-600">{`(${countTotalPeople.toString()})`}</p>
+        <p class="text-sm text-gray-400 dark:text-gray-600">{`(${countTotalPeople.toLocaleString($locale)})`}</p>
       </div>
     </div>
     <div class="flex items-center justify-end">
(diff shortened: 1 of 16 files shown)


web/src/lib/components/faces-page/show-hide.svelte
```diff
@@ -28,7 +30,10 @@
   >
     <div class="flex items-center">
       <CircleIconButton icon={mdiClose} on:click={() => dispatch('close')} />
-      <p class="ml-4 hidden sm:block">Show & hide people</p>
+      <div class="flex gap-2 items-center">
+        <p class="ml-2">Show & hide people</p>
+        <p class="text-sm text-gray-400 dark:text-gray-600">{`(${countTotalPeople.toLocaleString($locale)})`}</p>
```
jrasm91, line 35:
```suggestion
        <p class="text-sm text-gray-400 dark:text-gray-600">({countTotalPeople.toLocaleString($locale)})</p>
```

web/src/routes/(user)/people/+page.svelte
```diff
@@ -423,7 +430,10 @@
   </FullScreenModal>
 {/if}
 
-<UserPageLayout title="People" description={countTotalPeople === 0 ? undefined : `(${countTotalPeople.toString()})`}>
+<UserPageLayout
+  title="People"
+  description={countVisiblePeople === 0 ? undefined : `(${countVisiblePeople.toLocaleString($locale)})`}
```
jrasm91, line 435: Why are you doing the string building like this? Just put {count} inbetween the curly braces and put the parens outside of them to like ({count}). I think that would work fine. In general, I have found it rare that es6 template strings are needed inside of a svelte component html.

martabal: Sorry, I don't get how you can concatenate strings without using es6 template strings, do you prefer something like this ? ```ts
'(' + countVisiblePeople.toLocaleString($locale) + ')'

danieldietzler: Since it's svelte `{}` already escapes strings.
So what you can do is
```ts
description={countVisiblePeople === 0 && "({countVisiblePeople.toLocaleString($locale)})"}
```
I *think*.
Note that I've also changed `countVisiblePeople === 0 ? undefined : ...` to `countVisiblePeople === 0 && ...`

danieldietzler: Sorry you're right. This should:
```ts
description="{countVisiblePeople === 0 && `(${countVisiblePeople.toLocaleString($locale)})`}"
```

jrasm91: Ah, I didn't realize this was in an attribute value, so this should be fine. It seems you fixed the other case though.

jrasm91 (approved):

web/src/routes/(user)/people/+page.svelte
martabal: Sorry I was replying to @danieldietzler

PR #7320, opened 2024-02-21
refactor: album e2e

jrasm91:
Migrated to e2e and converted to not require full database resets between each tests. Test takes 1.5 seconds now, instead of 12.


danieldietzler (approved):
LGTM

danieldietzler (approved):

PR #7322, opened 2024-02-21
feat(web): search albums

martabal:
Search through albums

## Screenshots

https://github.com/immich-app/immich/assets/74269598/38d3676b-50e2-4f2f-8fd8-9990ea80b8f5

### DIFF
--- a/web/src/routes/(user)/albums/+page.svelte
+++ b/web/src/routes/(user)/albums/+page.svelte
@@ -290,7 +290,7 @@
     <!-- Album Card -->
     {#if $albumViewSettings.view === AlbumViewMode.Cover}
       <div class="grid grid-cols-[repeat(auto-fill,minmax(14rem,1fr))]">
-        {#each $albums.filter((album) => searchAlbum == '' || album.albumName
+        {#each $albums.filter((album) => searchAlbum === '' || album.albumName
               .toLowerCase()
               .includes(searchAlbum.toLowerCase())) as album, index (album.id)}
           <a data-sveltekit-preload-data="hover" href="{AppRoute.ALBUMS}/{album.id}" animate:flip={{ duration: 200 }}>
@@ -317,7 +317,7 @@
         <tbody
           class="block w-full overflow-y-auto rounded-md border dark:border-immich-dark-gray dark:text-immich-dark-fg"
         >
-          {#each $albums.filter((album) => searchAlbum == '' || album.albumName
+          {#each $albums.filter((album) => searchAlbum === '' || album.albumName
                 .toLowerCase()
                 .includes(searchAlbum.toLowerCase())) as album (album.id)}
             <tr
--- a/web/src/routes/(user)/albums/+page.svelte
+++ b/web/src/routes/(user)/albums/+page.svelte
@@ -182,6 +182,8 @@
     }
   }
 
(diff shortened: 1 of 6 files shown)


web/src/routes/(user)/albums/+page.svelte
```diff
@@ -285,7 +290,9 @@
     <!-- Album Card -->
     {#if $albumViewSettings.view === AlbumViewMode.Cover}
       <div class="grid grid-cols-[repeat(auto-fill,minmax(14rem,1fr))]">
-        {#each $albums as album, index (album.id)}
+        {#each $albums.filter((album) => searchAlbum == '' || album.albumName
+              .toLowerCase()
+              .startsWith(searchAlbum.toLowerCase())) as album, index (album.id)}
```
danieldietzler, line 295: Maybe `.includes` would be nicer?

danieldietzler (approved):

web/src/routes/(user)/albums/+page.svelte
martabal: By a lot !
Thanks

jrasm91 (approved):
LGTM

web/src/routes/(user)/albums/+page.svelte
```diff
@@ -285,7 +290,9 @@
     <!-- Album Card -->
     {#if $albumViewSettings.view === AlbumViewMode.Cover}
       <div class="grid grid-cols-[repeat(auto-fill,minmax(14rem,1fr))]">
-        {#each $albums as album, index (album.id)}
+        {#each $albums.filter((album) => searchAlbum === '' || album.albumName
```
mertalev, line 293: I don't think you need `searchAlbum === ''` here (or below)

mertalev (commented):
Awesome work!

andrewgdunn:
Apologies for commenting on a closed PR, but... would it be possible to do this on shared albums? (or maybe this applies there already)?

jrasm91:
This change only impacts the albums page, but in general it would be a good idea to consolidate the albums/shared albums pages to work the same.

martabal:
It's on my To do list
image

waclaw66:
I really appreciate this functionality, however there is a space for speed optimization. The page freezes for 3-4 second while searching among ~300 albums.

alextran1502:
@waclaw66 @martabal We probably need some debouncing

waclaw66:
The strange thing is that the first letter search is quite fast (<1s) after fresh page reload. When search input is cleared and first letter inserted again, it takes >3s. Even clearing the input is that slow. The search time depends on the letter, "e" is significantly slowed than "x", obviously because of frequency of occurrence.

https://github.com/immich-app/immich/assets/15554561/b8a1dddb-b094-4f32-b7bb-0c0b42b85bcf

martabal:
Yes it's still an issue for large collections

martabal:
One easy fix would be to disable the animation when you have more than 100 albums

waclaw66:
Another glitch I've observed with this PR is that resolution of album thumbnails is lowered when search is performed.

waclaw66:
> One easy fix would be to disable the animation when you have more than 100 albums

That's only workaround.

I just wonder why it's fast for the first search only. The flip animation will probably change the html code so that the next animation will be much more complex. The same occurs with album sorting as well.

martabal:
The best solution would be to have an infinite-scroll mechanism instead of loading all albums. But that would be a big and complex refactor

PR #7323, opened 2024-02-21
feat(mobile): Added "jump to date" functionality to the memory view

arnolicious:
Added a "Jump to date" Button in the bottom Info bar of the memory view.

It was somewhat mentioned [here](https://github.com/immich-app/immich/discussions/5858), but it is mainly a feature I've found myself wanting quite often.

As to the implementation, I don't have much Flutter experience, so my approach to communicate between the two components (the button in the memory view, and the Timeline list that needs to scroll) was with events, for which I've added the [event package](https://pub.dev/packages/event)

If there's a better/more "fluttery" way to do this, let me know!

And for the UI, at first I wanted to add an IconButton, however I couldn't really find a fitting Icon for this action (the closest I could come up with, was the `open_in_new` Icon), so I settled on a TextButton, and added the text to the languages I speak.

Here's a demo of what it looks like, on a very choppy emulator :D
https://github.com/immich-app/immich/assets/46051866/289750d1-d2b5-4619-8441-2c77295dd57b


mobile/assets/i18n/de-DE.json
```diff
@@ -208,6 +208,7 @@
   "image_viewer_page_state_provider_download_error": "Fehler beim Herunterladen",
   "image_viewer_page_state_provider_download_success": "Erfolgreich heruntergeladen",
   "image_viewer_page_state_provider_share_error": "Fehler beim Teilen",
+  "jump_to_date": "Zu Datum springen",
```
shenlong-tanwen, line 211: We don't need the keys in language files other than en-US. The other translations are handled through Localizely.

mobile/lib/modules/home/ui/asset_grid/immich_asset_grid_view.dart
```diff
@@ -138,6 +147,7 @@ class ImmichAssetGridViewState extends State<ImmichAssetGridView> {
     }
 
     final section = widget.renderList.elements[index];
+    // debugPrint("section: ${section.date} $index");
```
shenlong-tanwen, line 150: Can we removed

mobile/lib/shared/services/event.service.dart
```diff
@@ -0,0 +1,3 @@
+import 'package:event/event.dart';
+
+final scrollToDateEvent = Event<Value<DateTime>>();
```
shenlong-tanwen, line 3: We do not need a separate third party package for this functionality. The same can be achieved using `ValueListenable` or even `riverpod`. We can implement a event system over this to notify widgets about updates. Can you please remove the third party library and use a different implementation instead?

shenlong-tanwen (changes requested):
Thanks a lot for working on this PR. I'm sure a lot of users would find this useful. Can you please address the review comments. Also, Can we change the button to an `IconButton` instead? Choose any icon you find appropriate for now and we can change it later. `open_in_new`,`move_down` or `timeline` seems appropriate to me

mobile/lib/modules/home/ui/asset_grid/immich_asset_grid_view.dart
```diff
@@ -223,6 +233,22 @@ class ImmichAssetGridViewState extends State<ImmichAssetGridView> {
...
+    }
+    final index = widget.renderList.elements.indexWhere(
+      (e) =>
+          e.date.year == date.year &&
+          e.date.month == date.month &&
+          e.date.day == date.day,
+    );
+    if (index != -1) {
+      // Not sure why the index is shifted, but it works. :3
+      _scrollToIndex(index + 1);
```
martyfuhry, line 248: This is rather odd, but if it works, it works! I wonder if we shouldn't double check the position if it's the last asset in the list to avoid an overflow?

mobile/lib/modules/home/ui/asset_grid/immich_asset_grid_view.dart
```diff
@@ -237,6 +263,9 @@ class ImmichAssetGridViewState extends State<ImmichAssetGridView> {
   void initState() {
     super.initState();
     scrollToTopNotifierProvider.addListener(_scrollToTop);
+    scrollToDateNotifierProvider
```
martyfuhry, line 266: Do you need to remove this listener on dispose?

martyfuhry (commented):
This looks great and is super useful!

mobile/lib/modules/home/ui/asset_grid/immich_asset_grid_view.dart
arnolicious: After some testing, it seems that scrolling to an index too high just gets silently swallowed and the list scrolls to the last item anyways, but I added a check anyhow.

mobile/lib/modules/home/ui/asset_grid/immich_asset_grid_view.dart
arnolicious: Oh yeah, thanks!

alextran1502:
I tested the PR and when clicking on the button to jump to timeline, it doesn't do anything 🤔

arnolicious:
Well that's kind of embarassing, i had left an offset in the scroll index from when i was testing... 
I hope that was the issue, but I added an Error toast for when the scrolling fails due to the index being out of bounds (for debug purposes, I'd remove that before merging)

Did the memory view at least close when you clicked the button, @alextran1502 ?

alextran1502:
> Well that's kind of embarassing, i had left an offset in the scroll index from when i was testing... I hope that was the issue, but I added an Error toast for when the scrolling fails due to the index being out of bounds (for debug purposes, I'd remove that before merging)
> 
> Did the memory view at least close when you clicked the button, @alextran1502 ?

Hello, yes, when clicking on the button, it dismisses the memory view

arnolicious:
I'm kinda stumped, i added another toast, in case the date that the listener receives is somehow null...

Are you testing this on the demo instance by any chance?

DawidPietrykowski:
I tested it on my instance and it works, hides the memory view and jumps to the correct date

alextran1502:
Hey @arnolicious, I just tested again on my prod instance. When I clicked on the link, I got this error 

image

arnolicious:
Hm, an index of -1 would indicate no item in the list matching the date was found, which shouldn't happen, since if that memory exists, it should also exist in the timeline.

The check looks right to me, at first i suspected timezones to possibly be the cause of this, but I couldn't replicate it.
(I assume your server and your phone are in the same timezone, right?)

```dart
final index = widget.renderList.elements.indexWhere(
      (e) =>
          e.date.year == date.year &&
          e.date.month == date.month &&
          e.date.day == date.day,
);
```

Is the date in the error message correct, is that the date of the memory you clicked on?
I assume that somehow the date must be incorrect, maybe it's shifted by a day, and so I'd assume either the day before or after that memory does not have any assets... Is that the case? @alextran1502

martyfuhry:
I'm curious, what if you group the assets on the timeline by month instead of day? Does that change the expected index in the render list?

I'm thinking about this setting, in particular.

arnolicious:
Oh, didn't know that was a thing lol

Gonna look into it, thanks for pointing that out!

martyfuhry:
You should also gracefully handle the index not found case. In the ideal case, we'd hide the button entirely. But we can also show a nicely formatted error that we can't find the day in the timeline. Maybe we can do something even better?

arnolicious:
I'm not sure how feasible or necessary hiding the button would be, since it really shouldn't happen, once this bug is fixed. I will make the toast error more user friendly tho.

The question now is rather, what should happen when the timeline is grouped by month?
Scrolling to the month is fine, but I'm not sure how/if it's possible to then find the specific day in the month and scroll to it... 😅

alextran1502 (approved):

PR #7331, opened 2024-02-22
perf(server): optimize mapAsset

michelheusschen:
Noticed the `mapAsset` function was taking way too long. This seems to be caused by the object spread syntax, after removing it performance got much better.

Tested using `AssetService.getTimeBuckets`, specifically when the `mapAsset(asset, { withStack: true })` case gets returned. It would be useful to know if someone else can replicate these results (only includes calls to mapAsset function, nothing else).

| Assets     | Before   | After |
|------------|----------|-------|
| 100        | 3ms      | <1ms  |
| 1000       | 26ms     | 1ms   |
| 10000      | 397ms    | 6ms   |


jrasm91 (approved):
There have been reports of slowness in this method specifically.

danieldietzler (approved):

PR #7349, opened 2024-02-22
Web: Revamp message for Storage Template Engine in admin pannel

AndyPro720:
As discussed in ##7339 and subsequently on discord, this commit adds necessary context to the storage-template on admin panel, like so

image


jrasm91:
Can you add it at the same level of indentation as the inner settings? This will avoid a giant diff as well

Issue #7341, opened 2024-02-22
Edit people Done button shows wrong count


waclaw66:
### The bug

The toast "Edited XX people" shows wrong count (94) of detected people (46).

![obrazek](https://github.com/immich-app/immich/assets/15554561/f82dc166-99ae-424c-946e-a60115942f28)

### The OS that Immich Server is running on

Fedora 39

### Version of Immich Server

v1.95.1

### Version of Immich Mobile App

not relevant

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
not releavant
```

### Your .env content

```Shell
not releavant
```

### Reproduction steps

```bash
1. view asset view people detected
2. click to Edit people
3. click to Done button
4. wrong count of people is shown
```

aviv926:
@mertalev  Could it be because the rest of the face was not recognized as core?

mertalev:
It's probably related to counting hidden people. I think this should already be fixed with #7319.

mertalev:
Alternatively, if this issue was found on main, it could be *caused* by that issue.

waclaw66:
It occurs on main as well, checked now. There are no hidden people.

martabal:
It's cause by this PR #6751

aviv926:
fixed with #7352

PR #7352, opened 2024-02-22
fix(web): notification number of people when editing faces

martabal:
fix for #7341


jrasm91 (commented):
Why do we have null and undefined types in the list in the first place? Can you just add truthy values to begin with?

martabal:
That's what I was doing before https://github.com/immich-app/immich/pull/6751 

Do you prefer something like that ? `selectedPersonToReassign = Array.from({ length: peopleWithFaces.length }, () => null);`

danieldietzler (approved):

waclaw66:
May I ask why this fix was closed before merge?

martabal:
I had too many opened PRs. I will re-open it this evening

jrasm91 (approved):

PR #7359, opened 2024-02-23
Web: Revamp message for Storage Template Engine in admin pannel

AndyPro720:
Had to close last PR due to ssh verification issue. #7349 

As discussed in #https://github.com/immich-app/immich/pull/7339 and subsequently on discord, this commit adds necessary context to the storage-template on admin panel, like so 

image

### DIFF
--- a/web/src/lib/components/admin-page/settings/storage-template/storage-template-settings.svelte
+++ b/web/src/lib/components/admin-page/settings/storage-template/storage-template-settings.svelte
@@ -85,27 +85,25 @@
 </script>
 
 <section class="dark:text-immich-dark-fg mt-2">
-  <form autocomplete="off" on:submit|preventDefault class="mx-4 flex flex-col gap-4 py-4">
-    <div in:fade={{ duration: 500 }}>
-      <p class="text-sm dark:text-immich-dark-fg">
-        For more details about this feature, refer to the <a
-          href="https://immich.app/docs/administration/storage-template"
-          class="underline"
-          target="_blank"
-          rel="noreferrer"
-          >Storage Template
-        </a>
-        and its
-        <a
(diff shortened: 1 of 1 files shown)


web/src/lib/components/admin-page/settings/storage-template/storage-template-settings.svelte
```diff
@@ -84,7 +84,28 @@
   };
 </script>
 
-<section class="dark:text-immich-dark-fg">
+<section class="dark:text-immich-dark-fg mt-2">
+  <form autocomplete="off" on:submit|preventDefault class="mx-4 flex flex-col gap-4 py-4">
+    <div in:fade={{ duration: 500 }}>
```
michelheusschen, line 89: This form isn't needed, the classes can be moved to the div

michelheusschen: The stutter is caused by loading the storage template options from the server, it will wait for a response and then render the settings.

https://github.com/immich-app/immich/blob/3cd3411c1f5f5d9fd14218f31a391b3ed9351bae/web/src/lib/components/admin-page/settings/storage-template/storage-template-settings.svelte#L88

michelheusschen: And for the original comment, you can just push a new commit no need to create a new PR

AndyPro720: Ah thanks makes sense. About the PR I meant toss it in with another issue that I'll pick up. Anyways, I'll just commit it here

AndyPro720: @michelheusschen committed, have a look

alextran1502 (approved):

PR #7368, opened 2024-02-23
feat(web): Added password field visibility toggle

Jan108:
Added a toggle button to password fields, namely:
 - Login
 - Create User
 - Change Password
 - Change Password after first login
 - Admin register

First step for discussion: #5855

## Screenshots
Login with hidden password
Login with visible password

### DIFF
--- a/web/src/lib/components/shared-components/settings/setting-input-field.svelte
+++ b/web/src/lib/components/shared-components/settings/setting-input-field.svelte
@@ -10,8 +10,7 @@
 <script lang="ts">
   import { quintOut } from 'svelte/easing';
   import { fly } from 'svelte/transition';
-  import { mdiEyeOutline, mdiEyeClosed } from '@mdi/js';
-  import Icon from '../../elements/icon.svelte';
+  import PasswordField from '../password-field.svelte';
 
   export let inputType: SettingInputFieldType;
   export let value: string | number;
@@ -70,7 +69,7 @@
     <slot name="desc" />
   {/if}
 
-  {#if inputType != SettingInputFieldType.PASSWORD}
+  {#if inputType !== SettingInputFieldType.PASSWORD}
     <input
       class="immich-form-input w-full pb-2"
       aria-describedby={desc ? `${label}-desc` : undefined}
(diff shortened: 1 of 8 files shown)


waclaw66:
Wouldn't be better to use `eye-off-outline` instead of `eye-closed`?
obrazek

michelheusschen:
Thanks for the PR and nice convenience feature, however I have a few suggestions for improvement:
- Keep `<label>` out of the `PasswordField` component, for better reusability
- Use the `mdiEyeOffOutline` icon as already suggested
- Use Tailwind classes instead of custom CSS and inline styles
- Only show toggle button when password is entered
- Element IDs should be unique, now every password input has `id="password"`

As the list has gotten pretty long, I went ahead and pushed some changes. Feel free to make further adjustments as you see fit.

web/src/lib/components/shared-components/settings/setting-input-field.svelte
```diff
@@ -64,20 +69,36 @@
...
+    <PasswordField
+      aria-describedby={desc ? `${label}-desc` : undefined}
+      aria-labelledby="{label}-label"
+      id={label}
+      name={label}
+      type={fieldType}
+      autocomplete={passwordAutocomplete}
+      {required}
+      password={value.toString()}
+      onInput={(passwordValue) => (value = passwordValue)}
```
danieldietzler, line 99: We really should bind value.

danieldietzler (commented):
Unrelated but might be worth adding a password input component, since at the moment those "Passwords do not match" error messages + their handling is duplicated

web/src/lib/components/shared-components/settings/setting-input-field.svelte
michelheusschen: It's a bit tricky, just doing `bind:password={value}` won't work because `value` in this component also handles numbers

danieldietzler: Ugh. I missed that, sorry...

jrasm91 (approved):
LGTM

danieldietzler:
> Unrelated but might be worth adding a password input component, since at the moment those "Passwords do not match" error messages + their handling is duplicated

@michelheusschen did you take note of that comment as well? What's your opinion about this?

michelheusschen:
Validating forms in general is not done very uniformly and I think that improvements can certainly be made there. Once that's done, I'm not sure if a password confirmation component is necessary. I don't think the duplication is so bad as to be a problem, but if we decide to address it let's do it in a separate PR.

PR #7382, opened 2024-02-23
chore(web): cleanup promise handling

benmccann:
These are a subset of the errors pointed out by eslint when setting [recommended-type-checked](https://typescript-eslint.io/users/configs#recommended-type-checked).


web/src/lib/utils.ts
```diff
@@ -194,3 +194,9 @@ export const findLocale = (code: string | undefined) => {
     name: language?.name,
   };
 };
+
+export const asyncTimeout = (ms: number) => {
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms);
+  });
+};
```
mertalev, line 202: There's a promise version of `setTimeout` you can import from `timers/promises`

web/src/routes/admin/jobs-status/+page.svelte
```diff
@@ -4,28 +4,31 @@
...
     jobs = await getAllJobsStatus();
   };
 
+  let running = true;
+
   onMount(async () => {
-    await load();
-    timer = setInterval(load, 5000);
+    while (running) {
+      await load();
```
mertalev, line 25: There doesn't seem to be a point in having this `load` function over `jobs = await getAllJobsStatus();`

mertalev (commented):
Looks really good! I'm surprised there are so many places where we don't await

web/src/routes/admin/jobs-status/+page.svelte
benmccann: yeah, good catch. updated

web/src/lib/utils.ts
benmccann: Ah, thanks! I didn't know about that one. Can we import Node built-ins on the client? I'm not sure whether Vite will bundle Node built-ins or not. I'll have to test later when I get a minute

mertalev: Oops, I think not. I get `[plugin:vite:resolve] Module "timers/promises" has been externalized for browser compatibility` followed by `RollupError: [vite-plugin-sveltekit-compile] "setTimeout" is not exported by "__vite-browser-external"`.

mertalev (approved):

web/src/routes/admin/jobs-status/+page.svelte
```diff
@@ -4,28 +4,27 @@
...
-  const load = async () => {
-    jobs = await getAllJobsStatus();
-  };
+  let running = true;
 
   onMount(async () => {
-    await load();
-    timer = setInterval(load, 5000);
+    while (running) {
+      await getAllJobsStatus();
```
michelheusschen, line 21: Jobs don't show up on the page
```suggestion
      jobs = await getAllJobsStatus();
```

web/src/routes/(user)/albums/[albumId]/+page.svelte
```diff
@@ -220,12 +220,12 @@
 
   onMount(async () => {
     if (album.sharedUsers.length > 0) {
-      getFavorite();
-      getNumberOfComments();
+      await getFavorite();
+      await getNumberOfComments();
```
Comment (michelheusschen) [inline web/src/routes/(user)/albums/[albumId]/+page.svelte:224]: This now runs sequentially instead of in parallel, but it probably doesn't matter that much

web/src/lib/utils.ts
```diff
@@ -194,3 +195,13 @@ export const findLocale = (code: string | undefined) => {
...
+
+export const resolvePromise = <T>(promise: Promise<T>): void => {
+  Promise.resolve(promise).catch(async (error) => await handleError(error));
+};
```
michelheusschen, line 207: Why is `Promise.resolve` needed when only promises are accepted as input? Also I don't think the name `resolvePromise` properly communicates that the error will get handled.

michelheusschen (commented):
Will it still be possible to call an async function in `onMount` and (without awaiting) return a cleanup function? For example to abort a request.

web/src/lib/stores/assets.store.ts
```diff
@@ -175,7 +175,7 @@ export class AssetStore {
     for (const bucket of this.buckets) {
       if (height < viewport.height) {
         height += bucket.bucketHeight;
-        this.loadBucket(bucket.bucketDate, BucketPosition.Visible);
+        await this.loadBucket(bucket.bucketDate, BucketPosition.Visible);
```
michelheusschen, line 178: Will cause buckets to load sequentially and might slow down initial loading a bit

web/src/routes/(user)/albums/[albumId]/+page.svelte
Comment (benmccann) [reply web/src/routes/(user)/albums/[albumId]/+page.svelte]: updated

web/src/lib/utils.ts
jrasm91: All this does it automatically notify the user when an error happens. Maybe something like this would be better `notifyOnError(() => loadFavorites())`

benmccann: updated

web/src/lib/components/faces-page/assign-face-side-panel.svelte
```diff
@@ -128,7 +128,7 @@
       searchedPeopleCopy = data;
       searchWord = searchName;
     } catch (error) {
-      handleError(error, "Can't search people");
+      await handleError(error, "Can't search people");
```
jrasm91, line 131: Would it make sense to just not await all of these, but change `handleError` to return void instead?

web/src/routes/(user)/trash/+page.svelte
```diff
@@ -24,10 +24,15 @@
...
+  import { handlePromiseError } from '$lib/utils';
 
   export let data: PageData;
 
-  $: $featureFlags.trash || goto(AppRoute.PHOTOS);
+  featureFlags.subscribe(({ trash }) => {
+    if (!trash) {
+      handlePromiseError(goto(AppRoute.PHOTOS));
+    }
+  });
```
jrasm91, line 35: Does this create memory leaks? I don't see any unsubscribing happening here.

This begs a question though, in onMount hooks you can use async/await to be compliant with the rule, but that doesn't mean the implementation can't throw an promise rejection. It assume `onMount` (svelte) will handle it.

The same thing is true here. If you throw an error on `goto` svelte will just handle it. If the goal is to avoid throwing unhandled promise rejections to the runtime there are other things we should be changing, not just updating stuff to be "compliant" with the rule. 

Is there a reason you didn't just use `$featureFlags.trash || handlePromiseError(goto(AppRoute.Photos))`

web/src/routes/auth/change-password/+page.svelte
```diff
@@ -18,5 +19,12 @@
...
   </p>
 
-  <ChangePasswordForm user={$user} on:success={() => goto(AppRoute.AUTH_LOGIN)} />
+  <ChangePasswordForm
+    user={$user}
+    on:success={async () => {
+      await goto(AppRoute.AUTH_LOGIN);
+      resetSavedUser();
+      await logout();
+    }}
```
jrasm91, line 28: Can you move this on a handleSuccess method defined in the `<script>` section of the file instead?

web/src/routes/auth/login/+page.svelte
```diff
@@ -1,21 +1,12 @@
...
   import type { PageData } from './$types';
 
   export let data: PageData;
-
-  afterNavigate(async ({ from }) => {
-    if (from?.url?.pathname === AppRoute.AUTH_CHANGE_PASSWORD) {
-      resetSavedUser();
-      await logout();
-    }
-  });
```
jrasm91, line 18: Why did this section get removed?

web/svelte.config.js
```diff
@@ -4,12 +4,6 @@ import { vitePreprocess } from '@sveltejs/vite-plugin-svelte';
 /** @type {import('@sveltejs/kit').Config} */
 const config = {
   preprocess: vitePreprocess(),
-  onwarn: (warning, handler) => {
-    if (warning.code.includes('a11y')) {
-      return;
-    }
-    handler(warning);
-  },
```
jrasm91, line 12: Is this change related to the async linting?

benmccann: Yes, indirectly. I updated SvelteKit to start type-checking the `svelte.config.js` file, which was necessary to get the new eslint rules to work. It didn't like the code here because the types of `warning` and `handler` were not defined. At first I was going to fix it by adding the types here, but then I noticed that when I removed it that I didn't see any warnings printed, so I felt like that was the better fix since it's sort of bad practice to hide the warnings rather than fix them

web/src/routes/auth/change-password/+page.svelte
benmccann: done. I called it `onSuccess` though. That way when we migrate from event dispatchers to callback props for Svelte 5 we can use the shorthand syntax of `{onSuccess}` rather than having to write `onSuccess={handleSuccess}`

web/src/routes/auth/login/+page.svelte
benmccann: I moved it to `change-password/+page.svelte` since it was only applicable to that one page

web/src/lib/components/asset-viewer/photo-viewer.svelte
```diff
@@ -120,7 +120,7 @@
     if (state.currentZoom > 1 && isWebCompatibleImage(asset) && !hasZoomed && !$alwaysLoadOriginalFile) {
       hasZoomed = true;
 
-      loadAssetData({ loadOriginal: true });
+      handlePromiseError(loadAssetData({ loadOriginal: true }));
```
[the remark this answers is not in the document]
benmccann: I can't `await` it because `subscribe` requires a synchronous method

web/src/routes/(user)/trash/+page.svelte
benmccann: This was one of the very first ones I updated before `handlePromiseError` existed. I probably would have used it if it existed. I've updated it to use it now

web/svelte.config.js
jrasm91: Sounds good!

jrasm91 (approved):
Code looks good, but it would be good to do some basic testing as well.

benmccann:
I manually tested the app by navigating around and trying out various pieces of functionality. The only thing I found that looked like a bug also happens on `main`. When you click the heart to favorite or un-favorite a photo, that photo flashes and disappears for a second, which seemed unexpected to me.

Issue #4850, opened 2023-11-05
[BUG] Login screen shown if server is unreachable


opbod:
### The bug

Cannot browse pics using the app if Immich server is unreachable. I'm using Immich behind a VPN on my phone.

Expected that https://github.com/immich-app/immich/pull/3279 would fix, but it doesn't.

### The OS that Immich Server is running on

Ubuntu LTS

### Version of Immich Server

v1.84.0

### Version of Immich Mobile App

v1.84.0

### Platform with the issue

- [ ] Server
- [ ] Web
- [X] Mobile

### Your docker-compose.yml content

```YAML
version: "3.8"

services:
  immich-server:
    container_name: immich_server
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    command: ["start.sh", "immich"]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - /etc/localtime:/etc/localtime:ro
    env_file:
      - .env
    depends_on:
      - redis
      - database
      - typesense
    restart: always

  immich-microservices:
    container_name: immich_microservices
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    # extends:
    #   file: hwaccel.yml
    #   service: hwaccel
    command: ["start.sh", "microservices"]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - /etc/localtime:/etc/localtime:ro
    env_file:
      - .env
    depends_on:
      - redis
      - database
      - typesense
    restart: always

  immich-machine-learning:
    container_name: immich_machine_learning
    image: ghcr.io/immich-app/immich-machine-learning:${IMMICH_VERSION:-release}
    volumes:
      - model-cache:/cache
    env_file:
      - .env
    restart: always

  immich-web:
    container_name: immich_web
    image: ghcr.io/immich-app/immich-web:${IMMICH_VERSION:-release}
    env_file:
      - .env
    restart: always

  typesense:
    container_name: immich_typesense
    image: typesense/typesense:0.24.1@sha256:9bcff2b829f12074426ca044b56160ca9d777a0c488303469143dd9f8259d4dd
    environment:
      - TYPESENSE_API_KEY=${TYPESENSE_API_KEY}
      - TYPESENSE_DATA_DIR=/data
      # remove this to get debug messages
      - GLOG_minloglevel=1
    volumes:
      - tsdata:/data
    restart: always

  redis:
[...]

PR #7383, opened 2024-02-24
fix(mobile): Stop sending app to login page for unrelated auth errors

martyfuhry:
- Only validate access token with server if we have one in the `Store`
- Only return to the login page when the error exception has a 401 status.

### DIFF
--- a/mobile/lib/routing/auth_guard.dart
+++ b/mobile/lib/routing/auth_guard.dart
@@ -1,8 +1,8 @@
 import 'dart:io';
 
 import 'package:auto_route/auto_route.dart';
-import 'package:flutter/foundation.dart';
 import 'package:immich_mobile/routing/router.dart';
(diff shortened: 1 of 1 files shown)


mobile/lib/routing/auth_guard.dart
```diff
@@ -16,28 +17,31 @@ class AuthGuard extends AutoRouteGuard {
     resolver.next(true);
```
martyfuhry, line 17: This is the only section I'm not terribly sure about. I think we should block the resolver if `Store.get(StoreKey.accessToken)` excepts with a `StoreKeyNotFoundException` because we do that synchronously. But I didn't want to change the logic here, so I left this intact.

jrasm91 (approved):
LGTM 🎉

opbod:
Might fix https://github.com/immich-app/immich/issues/4850?

PR #7391, opened 2024-02-24
fix(web): fetch error reporting

michelheusschen:
Fetch errors don't show the error message returned from API responses, because it only expects axios errors. Fixed by handling both fetch and axios errors in the `handleError` function. The client hook has been changed to only handle fetch errors, because we don't use axios in any page load functions.


jrasm91 (approved):

danieldietzler (approved):
LGTM

Issue #7297, opened 2024-02-21
BUG(Web): Search allows broken/false place filters


JW-CH:
### The bug

When a country filter was set, the state filter shows only states in that country. If I set the state filter and change the country afterward, the search will never return a result because the filter allows a broken configuration. This is also possible with the city filter.

The place filter should work like the camera filter and not allow broken searches.

![image](https://github.com/immich-app/immich/assets/17313367/5796937c-4efe-4b62-bfd2-0d16245cf08a)

### The OS that Immich Server is running on

UnRaid

### Version of Immich Server

v1.95.1

### Version of Immich Mobile App

not needed

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
not needed
```

### Your .env content

```Shell
not needed
```

### Reproduction steps

```bash
1. Open search page
2. Filter city
3. Filter state (city not in state)
4. Filter country (state not in country)
===============
1. Open search page
2. Filter Country
3. Filter state (State filter shows only states in country)
4. Change country filter, state does not match with country anymore
```

aviv926:
fixed via https://github.com/immich-app/immich/pull/7397

PR #7397, opened 2024-02-24
refactor(web): search box

danieldietzler:
~The `GET /search/suggestions` endpoint has been updated and returns now an object of all suggestions. As a result, the `SearchSuggestionsType` has been removed.~


jrasm91 (dismissed):
I don't think we should return all of these results in a single request.

michelheusschen:
Thank you for picking this up. I agree with Jason, because it allow us to only fetch the data we need instead of having to update all suggestions. If we switch to managing filters and suggestions in each component instead of in `SearchFilterBox`, we could achieve this more easily.

I've pushed some changed that do this for the location filters and also fix #7297. Let me know what you think and feel free to make changes or revert.

web/src/lib/components/shared-components/search-bar/search-location-section.svelte
```diff
@@ -1,12 +1,62 @@
...
+  export let filter: SearchFilter['location'];
+
+  let countries: string[] = [];
+  let states: string[] = [];
+  let cities: string[] = [];
+
+  const toComboBoxOptions = (items: string[]) => items.map<ComboBoxOption>((item) => ({ label: item, value: item }));
+
+  $: countryFilter = filter.country;
+  $: stateFilter = filter.state;
```
danieldietzler, line 23: Why do we need those?

danieldietzler:
> Thank you for picking this up. I agree with Jason, because it allow us to only fetch the data we need instead of having to update all suggestions. If we switch to managing filters and suggestions in each component instead of in `SearchFilterBox`, we could achieve this more easily.
> 
> I've pushed some changed that do this for the location filters and also fix #7297. Let me know what you think and feel free to make changes or revert.

Thanks, I like this approach much more!

web/src/lib/components/shared-components/search-bar/search-location-section.svelte
michelheusschen: To prevent unneeded updates when `filter` is reassigned, but the values of `country` or `state` haven't actually changed

danieldietzler: And this prevents this? From my understanding `countryFilter` updates whenever `filter.country` updates, so `updateStates` will update whenever `filter.country` changes?

michelheusschen: Yes, this will prevent updating like shown in this [demo](https://svelte.dev/repl/1dde5d87b2e84090a3da4a33c126967a?version=4.2.12)

danieldietzler: Interesting. TIL!

web/src/lib/components/shared-components/search-bar/search-filter-box.svelte
```diff
@@ -282,11 +231,13 @@
 
       filter = {
         context: 'query' in searchQuery ? searchQuery.query : '',
-        people: orderBySelectedPeopleFirst(personIds.map((id) => ({ id }))),
+        people: orderBySelectedPeopleFirst(
+          personIds.map((id) => suggestions.people.find((person) => person.id === id)!),
+        ),
```
mertalev, line 236: I mentioned this in the [original PR](https://github.com/immich-app/immich/pull/6968#discussion_r1484959291) as well, but `filter.people`  doesn't need the person object, just the IDs. This scales quadratically for no reason.

mertalev (commented):
Loving these changes!

web/src/lib/components/shared-components/search-bar/search-filter-box.svelte
michelheusschen: Changed to using `Set<string>` with person ID

web/src/lib/components/shared-components/search-bar/search-camera-section.svelte
```diff
@@ -0,0 +1,63 @@
...
+
+  export let filters: SearchCameraFilter;
+
+  let makes: string[] = [];
+  let models: string[] = [];
+
+  $: makeFilter = filters.make;
+  $: modelFilter = filters.model;
+  $: updateMakes(modelFilter);
+  $: updateModels(makeFilter);
```
benmccann, line 20: This will have to be updated when https://github.com/immich-app/immich/pull/7382 is merged since it's calling an `async` function without error handling. I see the same in other files below

alextran1502 (approved):

Issue #4908, opened 2023-11-08
[BUG] web: "Appears in" album link doesn't work properly


waclaw66:
### The bug

"Appears in" link to different album works like url is changed, album title is also changed, but photos are shown from the original album.

### The OS that Immich Server is running on

Fedora 38

### Version of Immich Server

v.1.85.0

### Version of Immich Mobile App

v.1.85.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

```bash
1. go to album
2. click to asset that belongs to multiple albums
2. click to link to another album under "Appears in"
3. see original assets in the target album
4. hit F5 and see proper assets
```

PR #7403, opened 2024-02-25
fix(web): re-render albums

martabal:
## What's changed

This PR fixes 2 issues:
- Currently, when navigating from one album to another with the "Appears in" section in the detail-panel, the album infos and the timeline are not re-rendered. That's an issue we experienced with the `people/[personId]` route and this PR adds the same hacky solution to force re-rendering everything.
- the toggle event for the setting-switch / slider components

## Screenshots

| Before | After | 
| :----: | :---: |
| <video src="https://github.com/immich-app/immich/assets/74269598/e03c9e2b-1c01-41f0-b148-82c65ca6385e">  | <video src="https://github.com/immich-app/immich/assets/74269598/fb695df9-dba9-4b1c-b470-c93bb6801f41"> |


web/src/routes/(user)/albums/[albumId]/+page.svelte
```diff
@@ -82,6 +82,7 @@
 
   let album = data.album;
   let description = album.description;
```
Comment (michelheusschen) [inline web/src/routes/(user)/albums/[albumId]/+page.svelte:84]: Looke like these variables don't account for `data` to change and I think that's the root cause for the issues you're seeing with not re-rendering. Would be preferable if it can be fixed by making all these variables reactive, instead of the hacky solution in `afterNavigate`.

waclaw66:
Seems it could fix https://github.com/immich-app/immich/issues/4908

web/src/routes/(user)/albums/[albumId]/+page.svelte
Comment (martabal) [reply web/src/routes/(user)/albums/[albumId]/+page.svelte:84]: I don't think we should use reactive variables for description. When confirming the album description change, we compare the original album description with the new one. 

However, I forgot to re-initialize the description when navigating from one album to another (you can see on the video that the album description does not change, the description of the second album is "second album")

Comment (michelheusschen) [reply web/src/routes/(user)/albums/[albumId]/+page.svelte:84]: After taking another look, there are many more (reactivity) issues in this component. Just making these reactive variables won't work indeed, but I still believe it's better to fix these issues instead of coming up with a workaround. Maybe if we create new components for album title/description and do a little refactoring we can avoid the workaround?

Comment (jrasm91) [reply web/src/routes/(user)/albums/[albumId]/+page.svelte:84]: I think this would be a better approach as well.

Comment (martabal) [reply web/src/routes/(user)/albums/[albumId]/+page.svelte:84]: I don't know if using components would help. From my understanding, this is more of a sveltekit issue with variables not being reset when the slug changes.

Comment (jrasm91) [reply web/src/routes/(user)/albums/[albumId]/+page.svelte:84]: This isn't a svelte kit issue afaik. If you navigate to the same route with a different slug the component won't remount, which makes sense. To support these types of changes the data in the component needs to be reactive to the route param change.

web/src/lib/components/album-page/album-title.svelte
```diff
@@ -0,0 +1,45 @@
...
+
+    try {
+      await updateAlbumInfo({
+        id,
+        updateAlbumDto: {
+          albumName: newAlbumName,
+        },
+      });
+      albumName = newAlbumName;
+      notificationController.show({ type: NotificationType.Info, message: 'New album name has been saved' });
```
danieldietzler, line 26: Same thing here

web/src/lib/components/album-page/album-title.svelte
```diff
@@ -0,0 +1,45 @@
...
+    } catch (error) {
+      handleError(error, 'Unable to update album name');
+    }
+  };
+</script>
+
+<input
+  on:keydown={(e) => e.key === 'Enter' && titleInput.blur()}
```
danieldietzler, line 34: I think you can also get the titleInput with something like `e.target`, right? That would remove the extra variable

web/src/lib/components/album-page/album-description.svelte
```diff
@@ -0,0 +1,51 @@
...
+        id,
+        updateAlbumDto: {
+          description: newDescription,
+        },
+      });
+      notificationController.show({
+        type: NotificationType.Info,
+        message: 'Album description has been updated',
+      });
+      description = newDescription;
```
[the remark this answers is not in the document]
martabal: Why ? The goal to have that here is to send a notification only if the API request was successful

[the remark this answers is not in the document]
danieldietzler: Right, I would return in the catch case to resolve this.

```ts
try {
      await updateAlbumInfo({
        id,
        updateAlbumDto: {
          description: newDescription,
        },
      });
} catch (error) {
      handleError(error, 'Error updating album description');
      return;
}
// continue successful flow including showing the notification
```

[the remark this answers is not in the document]
martabal: Oh, that's an elegant solution. Thanks !

web/src/lib/components/album-page/album-description.svelte
```diff
@@ -0,0 +1,44 @@
...
+        id,
+        updateAlbumDto: {
+          description: newDescription,
+        },
+      });
+      description = newDescription;
+    } catch (error) {
+      handleError(error, 'Error updating album description');
+    }
+  };
```
danieldietzler, line 28: I think you missed adding the notification thing back in, right? You only deleted it but didn't move it below the try/catch

web/src/lib/components/album-page/album-title.svelte
```diff
@@ -0,0 +1,41 @@
...
+        id,
+        updateAlbumDto: {
+          albumName: newAlbumName,
+        },
+      });
+      albumName = newAlbumName;
+    } catch (error) {
+      handleError(error, 'Unable to update album name');
+    }
+  };
```
danieldietzler, line 27: Here as well. Or did you decide on removing it completely?

jrasm91: I think Alex deleted it in another pr

danieldietzler: Oh I see. Was that intentional or a bad merge?

jrasm91: On purpose because it seemed a little noisy

alextran1502 (approved):

PR #7439, opened 2024-02-26
feat(server, web): include pictures of shared albums on map

andreasgerstmayr:
This PR adds a new setting "Include shared albums" to the map settings, to includes pictures of albums shared with the current user on the map.

Related: #3176


server/src/infra/repositories/asset.repository.ts
```diff
@@ -511,19 +511,35 @@ export class AssetRepository implements IAssetRepository {
...
         },
       },
-      where: {
-        ownerId: In([...ownerIds]),
-        isVisible: true,
-        isArchived,
-        exifInfo: {
-          latitude: Not(IsNull()),
-          longitude: Not(IsNull()),
+      where: [
```
andreasgerstmayr, line 514: Is there a better way to write this OR expression?
`WHERE (ownerID IN ... OR albumId IN ...) AND isVisible = ... AND isArchived = ... AND ...`

server/src/domain/asset/asset.service.ts
```diff
@@ -159,14 +161,19 @@ export class AssetService {
...
     if (options.withPartners) {
       const partners = await this.partnerRepository.getAll(auth.user.id);
       const partnersIds = partners
         .filter((partner) => partner.sharedBy && partner.sharedWith && partner.sharedById != auth.user.id)
         .map((partner) => partner.sharedById);
       userIds.push(...partnersIds);
     }
-    return this.assetRepository.getMapMarkers(userIds, options);
+    if (options.withSharedAlbums) {
+      const sharedAlbums = await this.albumRepository.getShared(auth.user.id);
```
jrasm91, line 173: This doesn't include the asset relation does it?

server/src/domain/asset/dto/map-marker.dto.ts
```diff
@@ -31,4 +31,10 @@ export class MapMarkerDto {
   @IsBoolean()
   @Transform(toBoolean)
   withPartners?: boolean;
+
+  @ApiProperty()
```
jrasm91, line 35: This should not be needed

server/src/infra/repositories/asset.repository.ts
```diff
@@ -511,19 +515,35 @@ export class AssetRepository implements IAssetRepository {
...
+            id: In([...albumIds]),
+          },
+          isVisible: true,
+          isArchived,
+          exifInfo: {
+            latitude: Not(IsNull()),
+            longitude: Not(IsNull()),
+          },
+          isFavorite,
+          fileCreatedAt: OptionalBetween(fileCreatedAfter, fileCreatedBefore),
```
jrasm91, line 541: This section is repeated twice. Can you move it up to a variable?

web/src/lib/components/map-page/map-settings-modal.svelte
```diff
@@ -35,6 +35,7 @@
       <SettingSwitch title="Only favorites" bind:checked={settings.onlyFavorites} />
       <SettingSwitch title="Include archived" bind:checked={settings.includeArchived} />
       <SettingSwitch title="Include shared with me" bind:checked={settings.withPartners} />
+      <SettingSwitch title="Include shared albums" bind:checked={settings.withSharedAlbums} />
```
jrasm91, line 38: Should we remember this other option so it is more obvious what the difference is?

jrasm91 (approved):
This looks really good. Thanks!

danieldietzler:
@andreasgerstmayr You've requested a review from me but you haven't yet addressed @jrasm91 comments and there are also lots of merge conflicts right now...?

andreasgerstmayr:
> @andreasgerstmayr You've requested a review from me but you haven't yet addressed @jrasm91 comments and there are also lots of merge conflicts right now...?

hi @danieldietzler, sorry about that, must have been some automated process (it's showing up as 2:35 in the night), maybe because you're listed as a codeowner. No idea why it's happening now and not when I created the PR though.

Sorry for the late response, I was on holiday for a couple of weeks and then didn't get time so far.
Thanks for the review, I'll rebase and implement the changes soon.

danieldietzler:
Ah ok got it. No rush then :)

server/src/domain/asset/dto/map-marker.dto.ts
andreasgerstmayr: :+1:

The class got moved to `server/src/dtos/search.dto.ts` and doesn't have this annotation anymore.

server/src/infra/repositories/asset.repository.ts
andreasgerstmayr: Good idea, done.

web/src/lib/components/map-page/map-settings-modal.svelte
andreasgerstmayr: I renamed `Include shared with me` to `Include shared from partners`. Another option would be `Include assets from partners`.

What do you think, which option sounds better? Or do you have other suggestions?

server/src/domain/asset/asset.service.ts
andreasgerstmayr: I don't follow, where should I include the asset relation?

andreasgerstmayr:
I fixed the merge conflicts and implemented (most of) @jrasm91 suggestions now (except the one I didn't understand, see above).
Sorry for the delay on my side!

web/src/lib/components/map-page/map-settings-modal.svelte
NicholasFlamy: I'm not Jason but I would suggest `Include partner assets` or `Include shared partner assets`. Also, this rename is important for clarity.

andreasgerstmayr: I renamed it to `Include shared partner assets` now, thanks for the suggestion @NicholasFlamy

server/src/domain/asset/asset.service.ts
NicholasFlamy: Okay so Jason answered me on Discord about this:
https://discord.com/channels/979116623879368755/994044917355663450/1238973376349864108

Jason said: "Just want to make sure this is not loading the album AND all the assets for it"

I said: "If I understand correctly, you are asking if it is loading only the shared albums when figuring out what to show on the map, versus loding the shared albums and all the assets. And the problem with that is loading all the assets in the shared album takes more resources."

Jason responded: "Yup"

andreasgerstmayr: yes, this function does not query all assets, only the albums:
https://github.com/immich-app/immich/blob/f667c9597b9885059de00d9e8779147c4d08ddc2/server/src/repositories/album.repository.ts#L154-L166

zackpollard (approved):
LGTM, will let @jrasm91 take a look over it before we merge though

zackpollard:
I fixed a tsc error caused by changes in main since the branch was created as well as a prettier error in web. The PR should now be good to merge

jrasm91 (approved):
I updated this to include assets that are in shared albums where the user is the owner.

andreasgerstmayr:
> I updated this to include assets that are in shared albums where the user is the owner.

Good catch! I didn't think of that.

Thank you for merging! :)

PR #7456, opened 2024-02-27
fix(web): don't ask password for invalid shared link

michelheusschen:
Show `Invalid share key` message (passed through from server) instead of asking for password input for invalid shared links. Also adds e2e tests for this

invalid-shared-link-error

### DIFF
--- a/e2e/src/web/specs/shared-link.e2e-spec.ts
+++ b/e2e/src/web/specs/shared-link.e2e-spec.ts
@@ -31,25 +31,15 @@ test.describe('Shared Links', () => {
       },
       { headers: asBearerAuth(admin.accessToken) }
     );
-    sharedLink = await createSharedLink(
-      {
-        sharedLinkCreateDto: {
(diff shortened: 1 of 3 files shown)


e2e/src/web/specs/shared-link.e2e-spec.ts
```diff
@@ -40,6 +40,16 @@ test.describe('Shared Links', () => {
       },
       { headers: asBearerAuth(admin.accessToken) }
     );
+    sharedLinkPassword = await createSharedLink(
```
jrasm91, line 43: I think there is an apiUtils method that is more concise you could use.

jrasm91 (approved):
LGTM

e2e/src/web/specs/shared-link.e2e-spec.ts
michelheusschen: There is indeed, I'll update it

PR #7462, opened 2024-02-27
refactor(server): e2e

jrasm91:
* Turn on microservices in e2e tests
* Add support for awaiting webhook events for async tasks
* Migrate trash e2e
* Refactor cli.upload spec to not require a full database reset (runs much faster)
* Migrate a few asset endpoints to e2e


e2e/src/utils.ts
```diff
@@ -170,10 +177,42 @@ export interface AdminSetupOptions {
...
+    return new Promise<T>((resolve, reject) => {
+      const timeout = setTimeout(() => reject(new Error('Timeout')), 4000);
+      ws.once(event, (data: T) => {
+        clearTimeout(timeout);
+        resolve(data);
+      });
+    });
+  },
+};
+
```
jrasm91, line 210: This is the new websocket stuff

e2e/src/api/specs/trash.e2e-spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    it('should empty the trash', async () => {
+      const { id: assetId } = await apiUtils.createAsset(admin.accessToken);
+      await apiUtils.deleteAssets(admin.accessToken, [assetId]);
+
+      const before = await getAllAssets(
+        {},
+        { headers: asBearerAuth(admin.accessToken) },
+      );
+
+      expect(before.length).toBeGreaterThanOrEqual(1);
```
bo0tzz, line 40: Should this be an exact equals?

bo0tzz (approved):
Love it

e2e/src/api/specs/trash.e2e-spec.ts
jrasm91: The tests don't reset the database between test cases so there are some assets already there

e2e/src/utils.ts
```diff
@@ -59,7 +66,7 @@ let client: pg.Client | null = null;
 export const fileUtils = {
   reset: async () => {
     await execPromise(
-      `docker exec -i "${serverContainerName}" rm -R "${uploadMediaDir}"`
+      `docker exec -i "${serverContainerName}" /bin/bash -c "rm -rf ${dirs} && mkdir ${dirs}"`,
```
danieldietzler, line 69: When bun shell commands? :D

danieldietzler (commented):
Nice!

PR #7472, opened 2024-02-27
feat(server): rkmpp hardware decoding scaling

fyfrey:
enables full hardware transcoding via RKMPP on RK3588 platform: decode, scale filter, encode

transcode speed: 20x realtime for hevc, 27x realtime for h264 :rocket:

### DIFF
--- a/server/src/domain/media/media.util.ts
+++ b/server/src/domain/media/media.util.ts
@@ -14,7 +14,7 @@ class BaseConfig implements VideoCodecSWConfig {
 
   getOptions(target: TranscodeTarget, videoStream: VideoStreamInfo, audioStream?: AudioStreamInfo) {
     const options = {
-      inputOptions: this.getBaseInputOptions(),
+      inputOptions: this.getBaseInputOptions(videoStream),
       outputOptions: [...this.getBaseOutputOptions(target, videoStream, audioStream), '-v verbose'],
       twoPass: this.eligibleForTwoPass(),
     } as TranscodeOptions;
@@ -30,7 +30,8 @@ class BaseConfig implements VideoCodecSWConfig {
     return options;
   }
 
-  getBaseInputOptions(): string[] {
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  getBaseInputOptions(videoStream: VideoStreamInfo): string[] {
     return [];
   }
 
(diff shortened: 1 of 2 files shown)


server/src/domain/media/media.util.ts
```diff
@@ -615,7 +615,15 @@ export class RKMPPConfig extends BaseHWConfig {
...
       throw new Error('No RKMPP device found');
     }
-    return [];
+    return ['-hwaccel rkmpp', '-hwaccel_output_format drm_prime', '-afbc rga'];
+  }
+
+  getFilterOptions(videoStream: VideoStreamInfo) {
+    const options = this.shouldToneMap(videoStream) ? this.getToneMapping() : [];
+    if (this.shouldScale(videoStream)) {
+      options.push(`scale_rkrga=${this.getScaling(videoStream)}:format=nv12:afbc=1`);
```
mertalev, line 624: Would it be faster to put this before the tone-mapping? I'm thinking it'd let the decoding and scaling happen together and leave a lower resolution video to tone-map.

mertalev (commented):
Can you test that it works with an HDR video?

fyfrey:
Any freely available test video?
What should I look out for? I don't have any HDR monitor etc^^

mertalev:
It just needs to have the tone-mapping stuff in the command and transcode without any error logs. Any random YT video with HDR will work.

michelheusschen:
For test videos I use these [samples from Kodi](https://kodi.wiki/view/Samples), there are also links to external sources with even more videos

fyfrey:
I've disable hardware decoding/scaling for HDR videos (encoding still works).

I'll try to get hardware decoding and hardware tonemapping via OpenCL to work in another PR. the RK3588 SOC has hardware support for HDR tonemapping... I just don't know yet whether this ffmpeg build can use it and if so, how.

server/src/domain/media/media.util.ts
fyfrey: indeed. however, tonemapping does not work at all, so I've disabled it for now

mertalev:
Sounds good. Jellyfin uses libplacebo and we use their FFmpeg build, so it should be possible. I just haven't been able to get the right command to make it work.

mertalev (approved):

PR #7484, opened 2024-02-28
chore: build tweaks

jrasm91:
* ~~don't write to github action cache (with the base image there does not seem to be much benefit to caching as the build is faster without it)~~
* don't send test assets to the docker build context
* Use system tmp directory
* Only download chrome browser

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,30 +1,31 @@
 .vscode/
+.github/
+.git/
 
 design/
 docker/
 docs/
+e2e/
 fastlane/
 machine-learning/
 misc/
 mobile/
 
-server/node_modules/
+cli/coverage/
+cli/dist/
+cli/node_modules/
+
+open-api/typescript-sdk/build/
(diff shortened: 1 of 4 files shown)


alextran1502 (approved):

bo0tzz:
> (with the base image there does not seem to be much benefit to caching as the build is faster without it

How have you checked this? Scrolling through the log of successful docker builds (https://github.com/immich-app/immich/actions/workflows/docker.yml?query=is%3Asuccess), I see most jobs taking 6-8 minutes with some outliers at 14-15, and the jobs for this PR taking 11 minutes. That seems to me like the cache is usually giving a worthwhile speedup.

jrasm91:
I can't be bothered to look at this right now. I am under the impression that first runs take a long time because of the additional overhead to write out the cache, but also the slowest part of the whole process is definitely building for ARM. I will just omit these changes for now and merge the rest.

PR #7490, opened 2024-02-28
refactor(web): drop axios

### LINKED ISSUE #6647: PERF: switch from `axios` to `fetch`

jrasm91:
Closes #6647

* Drops axios in favor of fetch
* Migrates the remaining usages to `@immich/sdk`
* The 4 remaining endpoints (that rely of signals or progress events) have been temporarily migrated to use and xhr implementation.
* Tested by uploading and downloading files from albums, shared, links, and the main timeline

| Before | After | 
| - | - |
| image | image |


michelheusschen:
Nice work! You might even be able to use @immich/sdk for the AlbumCard, PanoramaViewer and PhotoViewer components because they don't use download progress.

jrasm91:
> Nice work! You might even be able to use @immich/sdk for the AlbumCard, PanoramaViewer and PhotoViewer components because they don't use download progress.

Oh I think you might be right, since it should support blob response type.

web/src/lib/components/album-page/__tests__/album-card.spec.ts
```diff
@@ -1,103 +1,99 @@
...
-        format: ThumbnailFormat.Jpeg,
-      },
-      { responseType: 'blob' },
-    );
-    expect(createObjectURLMock).toHaveBeenCalledWith(thumbnailFile);
-
-    expect(albumNameElement).toHaveTextContent('some album name');
-    expect(albumDetailsElement).toHaveTextContent('0 items');
-  });
+  // it.each([
```
benmccann, line 12: why are these commented out? should we add a TODO to re-enable or should we delete them?

benmccann (commented):
yesss!! this is great!

web/src/lib/components/album-page/__tests__/album-card.spec.ts
jrasm91: Yeah, we can add a TODO. The tests rely on mocking the `src/api`, which is now gone.I have no experience with these style of frontend tests, so I have commented them out for now since I don't know how to fix them and can't be bothered at the moment.

danieldietzler (approved):

martabal (approved):
Wow, that's great !

web/src/lib/components/album-page/__tests__/album-card.spec.ts
benmccann: Hmm, the old mocks were mocking `getAssetThumbnail` and that API endpoint doesn't seem to be called anywhere in the codebase now. I'm a bit surprised by that. I haven't really checked to see how it was being called before, but it makes it a little tricky to give a recommendation on this. I can say that Vitest needs a function to be on an object in order to be able to mock it. We don't want to put all the API methods in a single object as that will prevent tree shaking. So we probably want something more like an object for each tested API method or an object for each tested Svelte file.

benmccann:
Just FYI, it looks like it needs a rebase

web/src/lib/components/album-page/__tests__/album-card.spec.ts
jrasm91: Right. I brought this up originally when we did it that way in the CLI. We might just need to mock the whole `@immich/sdk` import or something like that.

jrasm91: This specific album card component uses the new downlladRequest method via a xhr request to download the image.

PR #7502, opened 2024-02-28
feat(web): ascending order for slideshow

martabal:
## What's changed

This PR adds an ascending / descending mode for the slideshow

## Screenshots

https://github.com/immich-app/immich/assets/74269598/fb743927-be1c-4504-b020-923c8fc24004


martabal:
> Looks good and works as expected. I'm only not so sure about how the order options are presented to the user. Maybe three buttons or a radio input with `Backward | Shuffle | Forward` would be more intuitive.

I agree. What do you think about using a dropdown menu (like the one in the album list to select the order) ?

michelheusschen:
A dropdown menu would be an improvement, but since there are only three options a radio input has my preference. But I'm fine with both.

martabal:
Yeah, I like dropdown more, it's easier to adapt if we had new navigation mechanisms in the future

michelheusschen:
The dropdown text disappears on small screens

image

web/src/lib/components/shared-components/settings/setting-dropdown.svelte
```diff
@@ -0,0 +1,48 @@
...
+  import { createEventDispatcher } from 'svelte';
+  import Dropdown, { type RenderedOption } from '$lib/components/elements/dropdown.svelte';
+
+  export let title: string;
+  export let subtitle = '';
+  export let options: RenderedOption[];
+  export let selectedOption: RenderedOption;
+  export let isEdited = false;
+
+  const dispatch = createEventDispatcher<{ toggle: RenderedOption }>();
```
danieldietzler, line 13: This should also export a toggle function instead

web/src/lib/stores/slideshow.store.ts
```diff
@@ -7,11 +7,20 @@ export enum SlideshowState {
   None = 'none',
 }
 
+export enum SlideShowNavigation {
```
danieldietzler, line 10: To stay persistent with the naming this should be `SlideshowNavigation`

michelheusschen:
Is hiding the dropdown text on small screens intentional? I'd say just show the text, there's enough space.

martabal:
> Is hiding the dropdown text on small screens intentional? I'd say just show the text, there's enough space.

Nope, it should be fixed

danieldietzler (approved):
LGTM

jinwater88:
Is the memory function in iPhone photo album the same as the slide show here? What technology is used to achieve it? Moviepy?

PR #7504, opened 2024-02-28
fix(web): settings accordion open state

### LINKED ISSUE #7499: Library settings folding is broken
https://github.com/immich-app/immich/assets/11602424/57ec5f75-1174-4dbd-a9bb-9563ffda09b6

michelheusschen:
Each `SettingAccordion` updates the search parameters shortly after another, causing a race condition where the state of accordions isn't updated correctly. This new solution avoids the race condition by keeping state in `SettingAccordionState` that is passed to each child using the context API.

Fixes #7499

### DIFF
--- a/web/src/lib/components/shared-components/settings/setting-accordion-state.svelte
+++ b/web/src/lib/components/shared-components/settings/setting-accordion-state.svelte
@@ -1,0 +1,35 @@
+<script lang="ts" context="module">
+  export type AccordionState = Set<string>;
+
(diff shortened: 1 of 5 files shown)


jrasm91 (approved):
LGTM

danieldietzler (approved):
Thanks, I've noticed this issue a while ago but didn't get around to fixing it. And in hindsight, I'm glad I didn't because your approach is better than everything I could've thought of lol

PR #7509, opened 2024-02-28
feat(server): Add emergency stop for library watcher

etnoy:
Fixes #6833 and related issues.

In some cases, the library watcher can hang, preventing immich from starting. Then the user has no way of disabling the watcher again.

This adds an env variable, IMMICH_STOP_LIBRARY_WATCHER, that causes the server container to change the configuration and then exit. After that, the env variable can be removed and Immich can start again.

### DIFF
--- a/docs/docs/features/libraries.md
+++ b/docs/docs/features/libraries.md
@@ -92,7 +92,7 @@ If your photos are on a network drive, automatic file watching likely won't work
 
 #### Troubleshooting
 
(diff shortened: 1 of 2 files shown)


martabal:
Wouldn't it be better if this was an admin CLI command?

etnoy:
> Wouldn't it be better if this was an admin CLI command?

No, since the server startup hangs and it will be hard to use the API.

jrasm91:
I would probably opt for moving library watching to microservices and use a postgres lock to prevent it from running on multiple instances. Then you could simple disable the microservices container, and the server/web is still responsiveness. These exact reasons are why we keep jobs off of the main api thread.

etnoy:
Closing in favor of #7533

Issue #3569, opened 2023-08-06
[BUG] asset dimensions info don't respect orientation


waclaw66:
### The bug

- asset details show dimensions always as `H x W`, it should be `W x H`
https://github.com/immich-app/immich/blob/c73832bd9cd81fe9739a463ca624b977b080c822/web/src/lib/components/asset-viewer/detail-panel.svelte#L189
https://github.com/immich-app/immich/blob/c73832bd9cd81fe9739a463ca624b977b080c822/mobile/lib/modules/asset_viewer/ui/exif_bottom_sheet.dart#L88

- asset details don't respect `orientation` column from `exif` table

### The OS that Immich Server is running on

Fedora 38

### Version of Immich Server

v1.71.0

### Version of Immich Mobile App

v1.71.0

### Platform with the issue

- [ ] Server
- [X] Web
- [X] Mobile

### Your docker-compose.yml content

```YAML
no relevant
```

### Your .env content

```Shell
no relevant
```

### Reproduction steps

```bash
no relevant
```

pamnanaimanish169:
Is this issue still open to work?

jrasm91:
It still exists for the mobile app, yes. Web has been fixed already.

PR #7513, opened 2024-02-29
feat(server): optimize person thumbnail generation

mertalev:
## Description

Person thumbnails are currently generated by cropping to a buffer and passing this buffer to sharp to resize. This is inefficient:
1. It ends up generating two thumbnails instead of one (the buffer is implicitly a new JPEG file)
2.  libvips is optimized for longer image processing pipelines, so separating the workload here is worse for performance
3. It lowers quality as person thumbnails are compressed at least three times: once due to the use of the compressed preview image, a second time with the buffer, and a third time with the resize

This PR combines cropping and resizing to the same pipeline for better performance and less compressed thumbnails. Additionally, it uses the original image to generate the thumbnail to further reduce compression artifacts, scaling the bounding box dimensions as needed.

### How Has This Been Tested?

I ran facial recognition on all assets and confirmed the person thumbnails are accurately cropped and look normal.

Before:

person-thumbnail-before

After:

person-thumbnail-after


alextran1502:
I tested with a new instance, and the facial thumbnail crop is off 
image

I also got some errors due to bad dimensions.

image

mertalev:
Hmm that's weird, it was working well when I tested it. I'll do more testing tomorrow

jrasm91 (approved):

server/src/domain/person/person.service.ts
```diff
@@ -488,19 +488,23 @@ export class PersonService {
...
+    const widthScale = asset.exifInfo.exifImageWidth / imageWidth;
+    const heightScale = asset.exifInfo.exifImageHeight / imageHeight;
+
+    const halfWidth = (widthScale * (x2 - x1)) / 2;
+    const halfHeight = (heightScale * (y2 - y1)) / 2;
 
-    const middleX = Math.round(x1 + halfWidth);
-    const middleY = Math.round(y1 + halfHeight);
+    const middleX = Math.round(widthScale * x1 + halfWidth);
+    const middleY = Math.round(heightScale * y1 + halfHeight);
```
jrasm91, line 507: I don't think if this will always work. We also use `fit: 'outside'`, which basically means there are no guarantees between the original size of the image and the resulting jpeg thumbnail, to which the coordinates for the face belong.

image

mertalev: > there are no guarantees between the original size of the image and the resulting jpeg thumbnail, to which the coordinates for the face belong

I don't think this is right. An outer fit just means the smaller dimension is set to 1440 and the larger dimension is scaled based on aspect ratio.

But I think you might be onto something here. All of the pictures in the test dataset I'm working with are horizontal, so this code might not be handling certain aspect ratios correctly.

jrasm91: You are trying to reverse map the face coordinates of the jpeg back to the original image, but this can only work if the thumbnail was not cropped at all. If it was I don't think you can reliable reverse-map the coordinates.

mertalev: Hmm I just tried cropping an image in a bunch of ways so it has extreme horizontal or vertical aspect ratios and it still works. Also tried cropping so the face is at the very most corner of the image and that works too. I'm kinda stumped.

mertalev:
@alextran1502 Can you confirm if you still have the issue as of the latest commit? I'm wondering if you maybe tried it with an older commit

alextran1502:
> @alextran1502 Can you confirm if you still have the issue as of the latest commit? I'm wondering if you maybe tried it with an older commit

I will take a look again and report back

server/src/interfaces/media.interface.ts
```diff
@@ -3,11 +3,19 @@ import { ImageFormat, TranscodeTarget, VideoCodec } from 'src/entities/system-co
...
+  top: number;
+  left: number;
+  width: number;
+  height: number;
+}
+
+export interface ThumbnailOptions {
   size: number;
-  format: ImageFormat;
+  format: 'webp' | 'jpeg';
```
danieldietzler, line 15: Why did you remove `ImageFormat`?

danieldietzler (approved):

server/src/interfaces/media.interface.ts
mertalev: Oops, bad rebase!

mertalev:
I tried on a slice of my library and reproduced the issue. I think it's because of #3569.

mertalev:
Can confirm that handling orientation fixed all the bad crops.

server/src/services/person.service.ts
```diff
@@ -614,4 +631,21 @@ export class PersonService {
...
+      case Orientation.MirrorHorizontal:
+      case Orientation.Rotate180:
+      case Orientation.MirrorVertical: {
+        return { width, height };
+      }
+      case Orientation.MirrorHorizontalRotate270CW:
+      case Orientation.Rotate90CW:
+      case Orientation.MirrorHorizontalRotate90CW:
+      case Orientation.Rotate270CW: {
+        return { width: height, height: width };
```
jrasm91, line 647: You don't have a fall through condition and the value of exif orientation is just stored as a string. It might be better to remove the top 4 cases and default to returning `{ width, height }` instead.

jrasm91 (approved):

server/src/services/person.service.ts
mertalev: Oh true, that's a good point.

mertalev:
@alextran1502 Could you test this PR when you get a chance? I want to be sure the cropping issues you noticed are fixed now.

alextran1502 (approved):

alextran1502:
Great work! Thank you for the discovery and the fix!

PR #7516, opened 2024-02-29
fix(server): only queue ml / transcoding jobs after thumbnail generation on upload

mertalev:
### Description

Re-running thumbnail generation has the side effect of queueing all ML tasks as well as transcoding. This is unnecessary and can be detrimental in the case of ML due to the indices being degraded. There's currently no way to *just* re-generate thumbnails (such as with a different quality setting, or after a [certain PR](https://github.com/immich-app/immich/pull/7513)).

This PR addresses this by selectively queueing ML and transcoding only on upload.

### DIFF
--- a/server/src/domain/job/job.service.spec.ts
+++ b/server/src/domain/job/job.service.spec.ts
@@ -286,12 +286,7 @@ describe(JobService.name, () => {
       },
       {
         item: { name: JobName.GENERATE_JPEG_THUMBNAIL, data: { id: 'asset-1' } },
-        jobs: [
-          JobName.GENERATE_WEBP_THUMBNAIL,
-          JobName.GENERATE_THUMBHASH_THUMBNAIL,
-          JobName.SMART_SEARCH,
-          JobName.FACE_DETECTION,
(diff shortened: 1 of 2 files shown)


jrasm91 (approved):

waclaw66:
I've been using that functionality for rescan faces after detection score change. Now there is no chance to rerun face detection selectively. It would be nice to add a separate function for face re-detection same as "Refresh thumbnails".

mertalev:
Face detection skips any assets that already have faces, so I don't think there's a difference between running thumbnail generation and running a "without" job for face detection.

waclaw66:
> Face detection skips any assets that already have faces, so I don't think there's a difference between running thumbnail generation and running a "without" job for face detection.

There is a discussion about it...
https://github.com/immich-app/immich/discussions/2472#discussioncomment-8505879

mertalev:
[Relevant XKCD](https://xkcd.com/1172/) haha

But I agree it'd be nice to have a face detection variant of Refresh Thumbnails

PR #7521, opened 2024-02-29
feat: filter people when using smart search

michelheusschen:
Implements filtering on people when using smart search. Ideally I would've added this to `searchAssetBuilder`, but then TypeORM creates an invalid query for metadata search.


server/src/infra/repositories/search.repository.ts
```diff
@@ -96,12 +104,21 @@ export class SearchRepository implements ISearchRepository {
...
 
     await this.assetRepository.manager.transaction(async (manager) => {
       let builder = manager.createQueryBuilder(AssetEntity, 'asset');
+
+      if (personIds?.length) {
+        const assetFaceBuilder = manager.createQueryBuilder(AssetFaceEntity, 'asset_face');
+        const cte = this.createPersonFilter(assetFaceBuilder, personIds);
+        builder
+          .addCommonTableExpression(cte, 'asset_face_ids')
+          .where(`asset.id IN (SELECT "assetId" FROM asset_face_ids)`);
```
mertalev, line 119: You should change this to an inner join

mertalev (approved):
Nice, I think grouping the faces first will work.

server/src/infra/repositories/search.repository.ts
michelheusschen: Done

PR #7525, opened 2024-02-29
Fix external weights loading during runtime in OpenVINO EP

agrawalsourav98:
Some models that are large e.g. [ViT-H-14-378-quickgelu__dfn5b](https://huggingface.co/immich-app/ViT-H-14-378-quickgelu__dfn5b/tree/main/visual) and have their attributes split into multiple files are loaded at runtime instead of the entire model being loaded during session creation.

This PR attempts to fix that behaviour.

### DIFF
--- a/machine-learning/app/main.py
+++ b/machine-learning/app/main.py
@@ -6,6 +6,7 @@ import threading
 import time
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import asynccontextmanager
+from pathlib import Path
 from typing import Any, AsyncGenerator, Callable, Iterator
 from zipfile import BadZipFile
 
@@ -105,7 +106,15 @@ async def predict(
 
     model = await load(await model_cache.get(model_name, model_type, **kwargs))
     model.configure(**kwargs)
-    outputs = await run(model.predict, inputs)
+    cwd = os.getcwd()
+    if model.mode == "vision":
+        os.chdir(Path(model.visual_path).parent)
+    elif model.mode == "text":
+        os.chdir(Path(model.textual_path).parent)
+    try:
+        outputs = await run(model.predict, inputs)
+    finally:
+        os.chdir(cwd)
     return ORJSONResponse(outputs)
 
 


machine-learning/app/main.py
```diff
@@ -105,7 +106,15 @@ async def predict(
...
-    outputs = await run(model.predict, inputs)
+    cwd = os.getcwd()
+    if model.mode == "vision":
+        os.chdir(Path(model.visual_path).parent)
+    elif model.mode == "text":
+        os.chdir(Path(model.textual_path).parent)
+    try:
+        outputs = await run(model.predict, inputs)
+    finally:
+        os.chdir(cwd)
```
mertalev, line 117: This doesn't make sense with multiple requests in flight.

mertalev (changes requested):
The concurrency issue aside, I also don't like that we're bolting on temporary code and adding overhead like this.

mertalev:
FWIW I think it's possible to solve the external data not being loaded eagerly by calling `run` with dummy data after the session is created. That should force all the weights to be loaded. Still not ideal but it at least keeps the temporary code in one place.

agrawalsourav98:
> FWIW I think it's possible to solve the external data not being loaded eagerly by calling `run` with dummy data after the session is created. That should force all the weights to be loaded. Still not ideal but it at least keeps the temporary code in one place.

Yeah, that is something we can do. I am not a fan of hacky stuff either it's just that I don't think intel cares enough that we would see newer packages anytime soon, so maybe we shouldn't rely on them.

I would try this out and update this PR if it is viable else let's just drop this approach.

mertalev:
I looked at their release cadence and they seem to like doing April releases. They probably scheduled it to be some time next month. Granted if we have a different issue with that release, it'll be half a year before they release the next one lol. I guess we can just see if this patch works and wait for April, then maybe look at building it ourselves if it comes to that.

agrawalsourav98:
> I looked at their release cadence and they seem to like doing April releases. They probably scheduled it to be some time next month. Granted if we have a different issue with that release, it'll be half a year before they release the next one lol. I guess we can just see if this patch works and wait for April, then maybe look at building it ourselves if it comes to that.

Yes, I building can be a better option in the future. They provide a ready to use DockerFile for OpenVINO [here](https://github.com/microsoft/onnxruntime/blob/main/dockerfiles/Dockerfile.openvino), so I don't think we should be too worried about building it ourselves.

agrawalsourav98:
Closing as it has been fixed by #7854 and #8018

PR #7533, opened 2024-02-29
chore(server): Move library watcher to microservices

etnoy:
This moves the library watcher to be running in the microservices.

Inspiration is from @jrasm91 who suggested this as a better option to #7509.

Here, if the watcher fails and stalls (as in #6833) we can stop the microservices container and disable the library watcher, unbricking the Immich install in a clean way

Added lock so that only one microservice takes care of the watcher in the case of multiple micros

Tests performed:
* App init works
* Library watcher automatically imports changes
* This workflow works: Disable microservices in docker compose, start app, disable library watcher
* Started multiple microservices and only one started the watcher


jrasm91 (commented):
This looks good so far, but you need to prevent library scanning from running on installs which run multiple instances of microservices. I think you can accomplish this by asking for a postgres lock, although you might need to get more info from @zackpollard or @mertalev on that.

server/src/infra/repositories/database.repository.ts
```diff
@@ -210,6 +210,25 @@ export class DatabaseRepository implements IDatabaseRepository {
...
+      }
+    } finally {
+      try {
+        await this.releaseLock(lock, queryRunner);
+      } finally {
+        await queryRunner.release();
+      }
+    }
+
+    return res as R;
```
jrasm91: This is not actually true in the case the lock was not acquired. Maybe we should change the return type to be something like this instead:

```ts
// : Promise<TypeLockResponse<R>> { ...
type TryLockResponse<R> = { response: R | null; acquired: boolean }
```

server/src/infra/repositories/database.repository.ts
```diff
@@ -210,6 +210,25 @@ export class DatabaseRepository implements IDatabaseRepository {
...
+    let res;
+    const queryRunner = this.dataSource.createQueryRunner();
+    try {
+      const lockAcquired = await this.tryLock(lock, queryRunner);
+      if (lockAcquired) {
+        res = await callback();
+      }
+    } finally {
+      try {
+        await this.releaseLock(lock, queryRunner);
```
jrasm91: will this line fail if the lock was not actually acquired?

server/src/infra/repositories/database.repository.ts
etnoy: Check the latest version, I already changed this

server/src/infra/repositories/database.repository.ts
etnoy: Fixed in latest version

server/src/domain/library/library.service.spec.ts
```diff
@@ -130,7 +137,16 @@ describe(LibraryService.name, () => {
       );
     });
 
-    it('should not initialize when watching is disabled', async () => {
+    it('should not initialize watcher when watching is disabled', async () => {
+      configMock.load.mockResolvedValue(systemConfigStub.libraryWatchEnabled);
+      databaseMock.tryLock.mockResolvedValue(false);
```
jrasm91, line 142: You probably don't need this and you probably should check that it is not even called if watching is turned off.

etnoy:
My test shows the current lock implementation did not work as expected, will have to do some debugging later:

```
immich_microservices-dev-2 | [Nest] 20  - 02/29/2024, 7:24:09 PM     LOG [SystemConfigService] LogLevel=verbose (set via LOG_LEVEL)
immich_microservices-dev-2 | [Nest] 20  - 02/29/2024, 7:24:09 PM     LOG [LibraryService] Starting to watch library df9f43b9-bf13-4b1a-a469-fbde3bda1f3f with import path(s) /data/import/bilder/2023/
immich_microservices-dev   | [Nest] 20  - 02/29/2024, 7:24:09 PM     LOG [SystemConfigService] LogLevel=verbose (set via LOG_LEVEL)
immich_microservices-dev   | [Nest] 20  - 02/29/2024, 7:24:09 PM     LOG [LibraryService] Starting to watch library df9f43b9-bf13-4b1a-a469-fbde3bda1f3f with import path(s) /data/import/bilder/2023/
```

only one microservice should have initiated the watch

server/src/domain/library/library.service.spec.ts
etnoy: Thanks for catching, I had put the lock = false statement in the wrong test

jrasm91:
The return type is an array:

```
[ { pg_try_advisory_lock: true } ]
```

etnoy:
> The return type is an array:
> 
> ```
> [ { pg_try_advisory_lock: true } ]
> ```

Yeah, I have to get used to ts not doing runtime type checking. Thanks :)

server/src/domain/library/library.service.spec.ts
etnoy: I'm not completely sure that I fixed this one, please let me know if I didn't.

server/src/infra/repositories/database.repository.ts
mertalev: I think Postgres will just give a warning but it won't throw

server/src/domain/library/library.service.ts
```diff
@@ -162,18 +175,30 @@ export class LibraryService extends EventEmitter {
 
   async unwatch(id: string) {
     if (this.watchers[id]) {
-      await this.watchers[id]();
+      this.watchers[id]();
```
mertalev, line 178: Shouldn't this be awaited?

etnoy: according to vscode: 'await' has no effect on the type of this expression.ts(80007)

It doesn't seem like await is needed after the refactor, and my testing makes me think I agree. Do you? It's no longer a record of async functions

mertalev: I looked at this and `watch` returns `return () => watcher.close();`, where `watcher.close()` returns `Promise<void>`. You aren't awaiting that promise if you remove the `await` keyword here.

etnoy: I dug deeper and changed the storage repository to actually return a promise, which it didn't do before. Now it's being awaited correctly!

mertalev: Nice!

server/src/infra/repositories/filesystem.provider.ts
```diff
@@ -148,6 +148,6 @@ export class FilesystemProvider implements IStorageRepository {
     watcher.on(StorageEventType.UNLINK, (path) => events.onUnlink?.(path));
     watcher.on(StorageEventType.ERROR, (error) => events.onError?.(error));
 
-    return () => watcher.close();
+    return async () => await watcher.close();
```
mertalev, line 151: You can actually keep this as-is or you might incur the wrath of @jrasm91 haha

etnoy: phew @mertalev , you really saved my bacon here ;)

mertalev (approved):

server/src/domain/library/library.service.ts
```diff
@@ -68,8 +71,16 @@ export class LibraryService extends EventEmitter {
 
   async init() {
     const config = await this.configCore.getConfig();
+
     const { watch, scan } = config.library;
-    this.watchLibraries = watch.enabled;
+
+    if (watch.enabled) {
+      // This ensures that library watching only occurs in one microservice
+      // TODO: we could make the lock be per-library instead of global
```
danieldietzler, line 79: I personally don't think we ever need this.

server/src/domain/library/library.service.ts
```diff
@@ -129,27 +141,28 @@ export class LibraryService extends EventEmitter {
...
           this.logger.debug(`Detected deleted file at ${path} in library ${library.id}`);
           const asset = await this.assetRepository.getByLibraryIdAndOriginalPath(library.id, path);
           if (asset && matcher(path)) {
             await this.assetRepository.save({ id: asset.id, isOffline: true });
           }
-          this.emit('unlink', path);
+          this.emit(StorageEventType.UNLINK, path);
         },
         onError: (error) => {
           // TODO: should we log, or throw an exception?
```
danieldietzler, line 163: Unrelated but I think this TODO is some artifact from an older PR that can be deleted

server/src/domain/repositories/database.repository.ts
```diff
@@ -19,6 +19,7 @@ export enum DatabaseLock {
   Migrations = 200,
   StorageTemplateMigration = 420,
   CLIPDimSize = 512,
+  LibraryWatch = 1337,
```
danieldietzler, line 22: 👀

danieldietzler (approved):
LGTM!

server/src/domain/library/library.service.ts
jrasm91: I agree as well

etnoy: You are correct, this would make it impossible to switch to watching at runtime

server/src/domain/repositories/database.repository.ts
etnoy: It was either that or 42069

server/src/domain/library/library.service.ts
etnoy: Let's just keep it as logging an error and emitting that event

PR #7535, opened 2024-02-29
chore: remove unused upload property

jrasm91:


etnoy (approved):

server/src/immich/api-v1/asset/dto/create-asset.dto.ts
```diff
@@ -3,7 +3,10 @@ import { ApiProperty } from '@nestjs/swagger';
 import { Transform, Type } from 'class-transformer';
 import { IsBoolean, IsDate, IsNotEmpty, IsString } from 'class-validator';
 
-export class CreateAssetBase {
+export class CreateAssetDto {
+  @ValidateUUID({ optional: true })
+  libraryId?: string;
+
```
danieldietzler, line 9: Could you sort them alphabetically? pls ;)

danieldietzler (approved):
LGTM!

server/src/immich/api-v1/asset/dto/create-asset.dto.ts
jrasm91: I want the boolean properties next to each other since that is why I think the decorators got missed.

danieldietzler: Oh I see. Got it

PR #7538, opened 2024-02-29
refactor: library e2e

jrasm91:
With jest `12.572 s`
With vitest `9332ms`
With vitest without the beforeEach `440ms`

### DIFF
--- a/e2e/src/api/specs/library.e2e-spec.ts
+++ b/e2e/src/api/specs/library.e2e-spec.ts
@@ -1,53 +1,40 @@
-import { LibraryResponseDto, LoginResponseDto } from '@app/domain';
-import { LibraryController } from '@app/immich';
-import { LibraryType } from '@app/infra/entities';
-import { errorStub, userDto, uuidStub } from '@test/fixtures';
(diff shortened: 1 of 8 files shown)


e2e/src/api/specs/library.e2e-spec.ts
```diff
@@ -467,7 +454,7 @@ describe(`${LibraryController.name} (e2e)`, () => {
         expect(pathResponse).toEqual({
           importPath: pathToTest,
           isValid: false,
-          message: `Path does not exist (ENOENT)`,
+          message: `Not a directory`,
```
jrasm91, line 457: It looks like this nest was never actually implemented so I've gone ahead and fixed it.

e2e/vitest.config.ts
```diff
@@ -1,9 +1,17 @@
 import { defineConfig } from 'vitest/config';
 
+// skip `docker compose up` if `make e2e` was already run
+const globalSetup: string[] = [];
```
jrasm91, line 4: Added this in here as well.

etnoy (approved):

etnoy:
Nice work!

danieldietzler (commented):
Dang you literally made it 28x faster! :O

PR #7541, opened 2024-02-29
fix(web): immich version

martabal:
With #7186, the `pump-version` script updates the web and the server at the same time, but they now need to be on the same version


jrasm91:
Do you want to have it update the typescript sdk too?

martabal:
Oh yes, I completely forget about it

alextran1502 (approved):

PR #7545, opened 2024-03-01
chore(server): eslint await-thenable

etnoy:
Looking for feedback on this before modifying other packages.

I had an unneeded await in my code and found an eslint rule to prevent it. It also found three other parts where it could have been used incorrectly


jrasm91 (approved):
LGTM. maybe we can look at making a parent config that these all extend so they can all have the same core set of rules.

etnoy:
> LGTM. maybe we can look at making a parent config that these all extend so they can all have the same core set of rules.

Seems like it shouldn't be too hard. https://turbo.build/repo/docs/handbook/linting/eslint

benmccann:
There are a few other rules to improve promise handling that I just turned on in the web project and which you might find to be useful here: https://github.com/immich-app/immich/pull/7382/files#diff-c1f89ef5e816f54437b1a48e36db79213d698329e9bbafd4c84a81506d8f197fR52

etnoy:
> There are a few other rules to improve promise handling that I just turned on in the web project and which you might find to be useful here: https://github.com/immich-app/immich/pull/7382/files#diff-c1f89ef5e816f54437b1a48e36db79213d698329e9bbafd4c84a81506d8f197fR52

Thanks, Ben! I added those rules and made the corresponding modifications :)

server/src/domain/download/download.service.ts
```diff
@@ -107,6 +107,7 @@ export class DownloadService {
       const assetIds = dto.assetIds;
       await this.access.requirePermission(auth, Permission.ASSET_DOWNLOAD, assetIds);
       const assets = await this.assetRepository.getByIds(assetIds);
+      // eslint-disable-next-line @typescript-eslint/require-await
```
jrasm91, line 110: you might be able to `yield Promise<assets>` and remove the async.

etnoy: Tried it, but 

 Property '[Symbol.asyncIterator]' is missing in type 'Generator<{ new (executor: (resolve: (value: AssetEntity[] | PromiseLike<AssetEntity[]>) => void, reject: (reason?: any) => void) => void): Promise<AssetEntity[]>; ... 7 more ...; readonly [Symbol.species]: PromiseConstructor; }, void, unknown>' but required in type 'AsyncGenerator<AssetEntity[], any, unknown>'.ts(2741)

jrasm91: Ahh, it might need async to be an actual async generator, opposed to a normal generator that yields a promise.

server/.eslintrc.js
```diff
@@ -25,6 +25,17 @@ module.exports = {
     'unicorn/prefer-top-level-await': 'off',
     'unicorn/prefer-event-target': 'off',
     'unicorn/no-thenable': 'off',
+    '@typescript-eslint/await-thenable': 'error',
+    '@typescript-eslint/no-floating-promises': 'error',
+    '@typescript-eslint/no-misused-promises': [
+      'error',
+      {
+        checksVoidReturn: false,
```
benmccann, line 33: we should probably try to avoid disabling this. You might find something like `handlePromiseError` from the web project's `$lib/utils` to be useful

etnoy: I wasn't able to get it working with that removed, but I did check with it enabled you'd like to take a stab at this. I'm going to bed for now

benmccann: I pushed a commit for it

etnoy: Thanks, but there's a strange error when running in github actions. Locally, everything compiles but not when the tests are running in github.

PR #7548, opened 2024-03-01
feat(server/web): add oauth defaultStorageQuota and storageQuotaClaim

samholton:
Fixes https://github.com/immich-app/immich/discussions/7399 and https://github.com/immich-app/immich/discussions/7300

### DIFF
--- a/server/src/domain/auth/auth.service.ts
+++ b/server/src/domain/auth/auth.service.ts
@@ -252,19 +252,11 @@ export class AuthService {
         storageLabel = null;
       }
 
-      let defaultQuota: number | null = config.oauth.defaultStorageQuota as keyof OAuthProfile as number;
-      if (typeof defaultQuota !== 'number') {
-        defaultQuota = null;
-      }
-
       let storageQuota: number | null = profile[config.oauth.storageQuotaClaim as keyof OAuthProfile] as number;
       if (typeof storageQuota !== 'number') {
-        storageQuota = defaultQuota;
+        storageQuota = config.oauth.defaultStorageQuota;
       }
-
-      if (typeof storageQuota == 'number' && storageQuota == 0) {
-        storageQuota = null;
-      }
+      storageQuota = storageQuota == 0 ? null : storageQuota;
 
       const userName = profile.name ?? `${profile.given_name || ''} ${profile.family_name || ''}`;
       user = await this.userCore.createUser({
--- a/server/src/domain/auth/auth.service.ts
+++ b/server/src/domain/auth/auth.service.ts
@@ -251,11 +251,26 @@ export class AuthService {
         storageLabel = null;
       }
 
+      let defaultQuota: number | null = config.oauth.defaultStorageQuota as keyof OAuthProfile as number;
+      if (typeof defaultQuota !== 'number') {
+        defaultQuota = null;
+      }
+
+      let storageQuota: number | null = profile[config.oauth.storageQuotaClaim as keyof OAuthProfile] as number;
+      if (typeof storageQuota !== 'number') {
+        storageQuota = defaultQuota;
+      }
+
+      if (typeof storageQuota == 'number' && storageQuota == 0) {
+        storageQuota = null;
+      }
+
       const userName = profile.name ?? `${profile.given_name || ''} ${profile.family_name || ''}`;
       user = await this.userCore.createUser({
         name: userName,
         email: profile.email,
         oauthId: profile.sub,
+        quotaSizeInBytes: storageQuota ? storageQuota * 1_073_741_824 : null,
         storageLabel,
       });
     }
--- a/server/src/domain/auth/auth.service.ts
+++ b/server/src/domain/auth/auth.service.ts
@@ -7,6 +7,7 @@ import {
   InternalServerErrorException,
   UnauthorizedException,
 } from '@nestjs/common';
+import { isNumber, isString } from 'class-validator';
 import cookieParser from 'cookie';
 import { DateTime } from 'luxon';
 import { IncomingHttpHeaders } from 'node:http';
@@ -63,6 +64,12 @@ interface LoginResponse {
 
 interface OAuthProfile extends UserinfoResponse {
   email: string;
+}
+
+interface ClaimOptions<T> {
+  key: string;
+  default: T;
+  isValid: (value: unknown) => boolean;
 }
 
 @Injectable()
@@ -235,9 +242,11 @@ export class AuthService {
       }
     }
 
+    const { autoRegister, defaultStorageQuota, storageLabelClaim, storageQuotaClaim } = config.oauth;
+
     // register new user
     if (!user) {
-      if (!config.oauth.autoRegister) {
+      if (!autoRegister) {
         this.logger.warn(
           `Unable to register ${profile.email}. To enable set OAuth Auto Register to true in admin settings.`,
         );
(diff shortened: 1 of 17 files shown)


server/src/domain/auth/auth.service.ts
```diff
@@ -251,11 +251,26 @@ export class AuthService {
...
+      if (typeof storageQuota == 'number' && storageQuota == 0) {
+        storageQuota = null;
+      }
+
       const userName = profile.name ?? `${profile.given_name || ''} ${profile.family_name || ''}`;
       user = await this.userCore.createUser({
         name: userName,
         email: profile.email,
         oauthId: profile.sub,
+        quotaSizeInBytes: storageQuota ? storageQuota * 1_073_741_824 : null,
```
samholton, line 273: I wasn't sure if/how I should use https://github.com/samholton/immich/blob/feature/add-quota-to-oauth-claim/web/src/lib/utils/byte-converter.ts

michelheusschen: You shouldn't import code from web in the server. We have these constants defined on the server side you can use:

https://github.com/immich-app/immich/blob/55ecfafa82e692f56bab40b7c559eface327158f/server/src/domain/domain.util.ts#L116

jrasm91: We should discuss how we want to accept input for the claim. Id be fine with just importing the bytes package and delegating all conversations to that.

https://www.npmjs.com/package/bytes

server/src/domain/system-config/dto/system-config-oauth.dto.ts
```diff
@@ -23,6 +23,11 @@ export class SystemConfigOAuthDto {
   @IsString()
   clientSecret!: string;
 
+  @ValidateIf(isEnabled)
+  @IsNotEmpty()
+  @IsNumber()
+  defaultStorageQuota!: number;
```
jrasm91, line 29: This probably doesn't belong in the oauth settings section.

Can we move it to it's own section? Maybe it can be under a new one called "user settings".

server/src/domain/auth/auth.service.ts
```diff
@@ -251,11 +251,26 @@ export class AuthService {
         storageLabel = null;
       }
 
+      let defaultQuota: number | null = config.oauth.defaultStorageQuota as keyof OAuthProfile as number;
+      if (typeof defaultQuota !== 'number') {
+        defaultQuota = null;
+      }
```
jrasm91, line 257: I don't think we need any additional type checking. It should already be valid.

server/src/domain/auth/auth.service.ts
```diff
@@ -251,11 +251,26 @@ export class AuthService {
...
+
+      let storageQuota: number | null = profile[config.oauth.storageQuotaClaim as keyof OAuthProfile] as number;
+      if (typeof storageQuota !== 'number') {
```
jrasm91, line 260: If the quota is a string it is ignored. I think you need to add some unit tests for this.

jrasm91 (commented):
Looks pretty good so far!

server/src/domain/auth/auth.service.ts
samholton: I'm open to suggestions. Originally I was thinking a string such as "50 GiB" but decided to keep it consistent with what you would enter on the edit user screen and used a number (in GiB). Are you suggesting I use a string and just pass it to `bytes`?

jrasm91: That is one option. I'm open to suggestions. Making it work the same way as the web seems like a good idea. So it would need to be a number and we'd assume it is in GiB?

samholton: Yeah, I think that makes sense. And if web is ever changed to be open ended as a string, this could be as well. I will update to bring in bytes, keep the type as number, and pass it to bytes with assumed GiB for conversion.

jrasm91: If we're assuming it is GiB we don't need to import the bytes package

server/src/domain/system-config/dto/system-config-oauth.dto.ts
samholton: Hmm, I was only attempting to set a default quota for oauth registered users if the did not have a claim. If this quota becomes global, should the create user box be populated with this default value and allow the user to manually change it?

server/src/domain/auth/auth.service.ts
samholton: A test that has a string in the claim and confirms it uses default quota?

server/src/domain/system-config/dto/system-config-oauth.dto.ts
jrasm91: Oh actually, this might be fine. When you create a new user in the web you have an option to specify a quota at that time, so the default probably would not ever be used. I am ok leaving this as an OAuth only setting.

server/src/domain/auth/auth.service.ts
jrasm91: I think this can just be `const defaultQuora = config.oauth.defaultStorageQuota`

danieldietzler (approved):
LGTM!

server/src/domain/system-config/dto/system-config-oauth.dto.ts
```diff
@@ -47,4 +52,9 @@ export class SystemConfigOAuthDto {
 
   @IsString()
   storageLabelClaim!: string;
+
+  @ValidateIf(isEnabled)
+  @IsNotEmpty()
```
jrasm91, line 57: I don't think we ever need to allow this being unspecified. It should just always be validated and always have a value.

server/src/domain/system-config/dto/system-config-oauth.dto.ts
```diff
@@ -23,6 +23,11 @@ export class SystemConfigOAuthDto {
   @IsString()
   clientSecret!: string;
 
+  @ValidateIf(isEnabled)
+  @IsNotEmpty()
```
jrasm91, line 27: This validation isn't needed imo. It should always be required and 0 should indicate unlimited quota.

web/src/lib/components/admin-page/settings/oauth/oauth-settings.svelte
```diff
@@ -130,6 +130,26 @@
...
+          bind:value={config.oauth.storageQuotaClaim}
+          required={true}
+          disabled={disabled || !config.oauth.enabled}
+          isEdited={!(config.oauth.storageQuotaClaim == savedConfig.oauth.storageQuotaClaim)}
+        />
+
+        <SettingInputField
+          inputType={SettingInputFieldType.NUMBER}
+          label="DEFAULT STORAGE QUOTA (GiB)"
+          desc="User's quota if not set in the claim (Note: Enter 0 for unlimited quota)."
```
jrasm91, line 146: Quota to be used when no claim is provided.

jrasm91 (commented):
Can you add some unit tests please?

samholton:
> Can you add some unit tests please?

Ok, I added some tests. I'm not familiar with this framework so let me know what I can do to improve/fix/change them. Is https://github.com/immich-app/immich/pull/7548/files#diff-b0539644bb10490d8fa4d18f4ee651ff6fe36b6e329f2390ceae1ecae772dbf0R26 the correct place to put this data?

jrasm91 (approved):
LGTM! Thanks :smile:

PR #7553, opened 2024-03-01
fix(mobile): bottom bar Upload translation

waclaw66:
This PR adds missing Upload button translation.


jrasm91 (approved):

JW-CH:
Why was the "Build Mobile" skipped?

JW-CH:
> > Why was the "Build Mobile" skipped?
> 
> It requires a signing key. The mobile is still built in the Test > Mobile action though. Are you looking for an APK or something?

No, I just thought it builds when there were changes in the mobile source. All good.

PR #7555, opened 2024-03-01
perf(web): asset delete

michelheusschen:
The server batches asset deletes, however on the web every delete action is done individually causing performance issues when deleting a large amount of assets. The main change is removing `AssetStore.removeAsset` in favor of a batched variant. Previously deleting a bucket with 10_000 assets took about 4 seconds for just updating the asset store, now it's down to 3-5 milliseconds.

### DIFF
--- a/web/src/lib/stores/assets.store.ts
+++ b/web/src/lib/stores/assets.store.ts
@@ -59,14 +59,14 @@ interface DeleteAsset {
   value: string;
 }
 
-interface TrashAsset {
+interface TrashAssets {
   type: 'trash';
-  value: string;
+  value: string[];
 }
 
 export const photoViewer = writable<HTMLImageElement | null>(null);
 
-type PendingChange = AddAsset | UpdateAsset | DeleteAsset | TrashAsset;
+type PendingChange = AddAsset | UpdateAsset | DeleteAsset | TrashAssets;
 
 export class AssetStore {
   private store$ = writable(this);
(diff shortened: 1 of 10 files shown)


web/src/lib/stores/assets.store.ts
```diff
@@ -372,35 +372,31 @@ export class AssetStore {
...
-    // TODO: this could probably be more efficient
-    for (const id of ids) {
-      this.removeAsset(id);
-    }
-  }
-
-  removeAsset(id: string) {
-    this.assets = this.assets.filter((asset) => asset.id !== id);
-    delete this.assetToBucket[id];
+  removeAssets = (ids: string[]) => {
```
jrasm91, line 375: Can we avoid the arrow syntax and update the usage to work accordingly?

michelheusschen: Forgot about that, I will revert to how it was before

jrasm91: Maybe add a handleDelete method in the components that calls the asset store that way you can use the simplified syntax there still.

michelheusschen: I think the syntax isn't too bad after all, might just leave it like this

danieldietzler:
Can't we just pass `assetStore.removeAssets` everywhere instead of `(assetIds) => assetStore.removeAssets(assetIds)`?

jrasm91 (approved):

PR #7564, opened 2024-03-01
dev(server): fun with bun

mertalev:
### Description

This is a test branch to compare performance characteristics.

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -46,6 +46,17 @@ WORKDIR /usr/src/app
 ENV NODE_ENV=production \
     NVIDIA_DRIVER_CAPABILITIES=all \
     NVIDIA_VISIBLE_DEVICES=all
+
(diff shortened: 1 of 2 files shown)


etnoy:
It was only a matter of time!

ttmx:
Well, how did the performance testing go?

mertalev:
It's a healthy uplift in some cases - for one case it took 13ms to do something that took 450-900ms on Node.js. It also finished importing geodata in 18s vs 26s.

In other cases it didn't change much. The CLI didn't benefit from Bun when I tested it. It also has some quirks - it didn't properly close streams in some cases so we had to make some changes there to avoid running out of file descriptors.

This branch was actually mainly because we were debugging a latency increase over time as more thumbnails are downloaded. Bun helped a bit here, but it only stopped that effect for the smaller WebP files and not the larger images.

PR #7569, opened 2024-03-02
feat(machine-learning): support cuda 12

martabal:
Add support for cuda 12:
- Onnxruntime [supports it since 1.17](https://onnxruntime.ai/docs/execution-providers/CUDA-ExecutionProvider.html) ([docs](https://onnxruntime.ai/docs/install/#install-onnx-runtime-gpu-cuda-12x))
- The default cuda version with ubuntu 23.10 is cuda 12.
- Keep the default version to cuda 11


machine-learning/pyproject.toml
```diff
@@ -59,6 +59,14 @@ optional = true
 [tool.poetry.group.armnn.dependencies]
 onnxruntime = "^1.15.0"
 
+[tool.poetry.group.cuda12.dependencies]
+onnxruntime-gpu = {version = "^1.17.0", source = "cuda12"}
+
+[[tool.poetry.source]]
+name = "cuda12"
+url = "https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/onnxruntime-cuda-12/pypi/simple/"
```
mertalev, line 67: What is this link lol

martabal: The official repo 🤷 https://onnxruntime.ai/docs/install/#install-onnx-runtime-gpu-cuda-12x

martabal: That's the microsoft's devops plateform

mertalev:
I'm not sure we actually need to keep 11. In practice, 525 is the minimum driver version for ONNX Runtime based on issues I've seen and the minimum compute capability for it is 5.2. I think anyone who's using CUDA right now is already in an environment where 12 will just work.

machine-learning/pyproject.toml
mertalev: Hmm fair enough. Weird that they publish it there and not on PyPI

martabal:
the original goal was only to support cuda 12 but if you want to move away from cuda 11 completely, I'm all in :smile:

machine-learning/pyproject.toml
```diff
@@ -59,6 +59,11 @@ optional = true
 [tool.poetry.group.armnn.dependencies]
 onnxruntime = "^1.15.0"
 
+[[tool.poetry.source]]
+name = "cuda12"
+url = "https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/onnxruntime-cuda-12/pypi/simple/"
+priority = "supplemental"
```
mertalev, line 65: You can make this explicit

mertalev (approved):

martabal:
What do you think about still supporting cuda 11 ?
It could be useful non-docker users so they still can install it with something like `poetry install ... --with cuda11`

mertalev:
I'm not sure if Poetry can handle different flavors of the same package since the optional dependencies are all squashed into the same lock file. It didn't work for cpu/gpu variants of PyTorch at least. Only one of them actually get resolved.

mertalev:
But that might have changed since I tried it. You can test and check if it can actually differentiate the cuda-11 and cuda-12 flavors

martabal:
Yeah, you're right, we can't https://github.com/python-poetry/poetry/issues/7748

mertalev (approved):

PR #7581, opened 2024-03-02
feat(server): allow oauth claim to set 0 for no quota

samholton:
@jrasm91 thanks for all the help on https://github.com/immich-app/immich/pull/7548 - Do we want to support the use case of having a default quota set but setting it to unlimited for specific users via the claim? I was sticking with `0` to match the UI and default quota language.

### DIFF
--- a/server/src/domain/auth/auth.service.spec.ts
+++ b/server/src/domain/auth/auth.service.spec.ts
@@ -15,7 +15,6 @@ import {
   newUserTokenRepositoryMock,
   sharedLinkStub,
   systemConfigStub,
-  userDto,
   userStub,
   userTokenStub,
 } from '@test';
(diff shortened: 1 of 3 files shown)


jrasm91 (approved):
Ahh, this allows 0 in the claim. Sounds good.

server/src/domain/auth/auth.service.spec.ts
```diff
@@ -543,7 +543,7 @@ describe('AuthService', () => {
         loginResponseStub.user1oauth,
       );
 
-      expect(userMock.create).toHaveBeenCalledWith(userDto.userWithDefaultStorageQuota);
+      expect(userMock.create).toHaveBeenCalledWith(userDto.userWithUnsetQuotaClaim);
```
jrasm91, line 546: It might be better to remove these user objects and just assert it matches an object containing the right quota value.

jrasm91 (approved):

server/src/domain/auth/auth.service.spec.ts
samholton: Yeah I can make that change

PR #7586, opened 2024-03-02
fix: bump version pipeline

martabal:
Currently, package-lock files still have incorrect open-api version mentions. This PR fixes it so now the `pump-version` script: 
1. Update the server version
2. Update the typescript SDK
3. Update all projects that use the TypeScript SDK

### DIFF
--- a/misc/release/pump-version.sh
+++ b/misc/release/pump-version.sh
@@ -66,8 +66,8 @@ if [ "$CURRENT_SERVER" != "$NEXT_SERVER" ]; then
   npm --prefix open-api/typescript-sdk version "$SERVER_PUMP"
   npm --prefix web version "$SERVER_PUMP"
   npm --prefix web i --package-lock-only
-  npm --prefix e2e i --package-lock-only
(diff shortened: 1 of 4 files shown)


danieldietzler (approved):

misc/release/pump-version.sh
```diff
@@ -62,9 +62,12 @@ fi
...
   echo "Pumping Server: $CURRENT_SERVER => $NEXT_SERVER"
   npm --prefix server version "$SERVER_PUMP"
-  npm --prefix web version "$SERVER_PUMP"
-  npm --prefix open-api/typescript-sdk version "$SERVER_PUMP"
   make open-api
+  npm --prefix open-api/typescript-sdk version "$SERVER_PUMP"
+  npm --prefix web version "$SERVER_PUMP"
+  npm --prefix web i --package-lock-only
+  npm --prefix e2e i --package-lock-only
+  npm --prefix cli i --package-lock-only
```
jrasm91, line 70: I think the e2e might depend on the CLI so it should be before e2e

PR #7589, opened 2024-03-02
fix(server, web): prevent reload when liking an asset

martabal:
This PR prevents photo viewer from reloading when you like an asset

### DIFF
--- a/web/src/lib/components/asset-viewer/photo-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/photo-viewer.svelte
@@ -25,6 +25,8 @@
   let copyImageToClipboard: (source: string) => Promise<Blob>;
   let canCopyImagesToClipboard: () => boolean;
 
+  const loadOriginalByDefault = $alwaysLoadOriginalFile && isWebCompatibleImage(asset);
+
   $: if (imgElement) {
     createZoomImageWheel(imgElement, {
       maxZoom: 10,
@@ -125,7 +127,7 @@
   transition:fade={{ duration: haveFadeTransition ? 150 : 0 }}
   class="flex h-full select-none place-content-center place-items-center"
 >
-  {#await loadAssetData({ loadOriginal: $alwaysLoadOriginalFile ? isWebCompatibleImage(asset) : false })}
+  {#await loadAssetData({ loadOriginal: loadOriginalByDefault })}
     <LoadingSpinner />
   {:then}
     <div bind:this={imgElement} class="h-full w-full">


jrasm91:
I think we should always send them since they will be sent eventually when a user _can_ do those actions.

martabal:
> I think we should always send them since they will be sent eventually when a user _can_ do those actions.

Currently, only owners can archive and favorite assets, but the day when we'll implement a mechanism so all users can like and/or archive assets, we'll have to update the backend and the frontend logic, so it's more maintenance.

And I don't really like the fact we're sending incorrect information.

e2e/vitest.config.ts
```diff
@@ -10,7 +10,7 @@ try {
 
 export default defineConfig({
   test: {
-    include: ['src/{api,cli}/specs/*.e2e-spec.ts'],
+    include: ['src/{api,cli}/specs/album.e2e-spec.ts'],
```
jrasm91, line 13: Can you revert this? You can run `npm test -- album` to only run one spec

martabal: Oops. Thanks for the tip !

jrasm91:
Sending false is not incorrect. The asset is not favorited by the user. It may change to true in the future in which case I don't think it makes sense to have deleted it from the response payload.

web/src/lib/components/asset-viewer/photo-viewer.svelte
```diff
@@ -7,14 +7,14 @@
...
-  import { type AssetResponseDto } from '@immich/sdk';
   import { useZoomImageWheel } from '@zoom-image/svelte';
   import { onDestroy, onMount } from 'svelte';
   import { fade } from 'svelte/transition';
   import LoadingSpinner from '../shared-components/loading-spinner.svelte';
   import { NotificationType, notificationController } from '../shared-components/notification/notification';
 
-  export let asset: AssetResponseDto;
+  export let assetId: string;
+  export let originalPath: string;
```
michelheusschen, line 17: It might be better to leave `export let asset: AssetResponseDto` and just use reactive declarations in this component. That allows you to keep the export the same with changes like in #7596

web/src/lib/components/asset-viewer/photo-viewer.svelte
```diff
@@ -125,7 +127,7 @@
   transition:fade={{ duration: haveFadeTransition ? 150 : 0 }}
   class="flex h-full select-none place-content-center place-items-center"
 >
-  {#await loadAssetData({ loadOriginal: $alwaysLoadOriginalFile ? isWebCompatibleImage(asset) : false })}
```
michelheusschen, line 128: The issue now occurs again. I would just do `$: isWebCompatible = isWebCompatibleImage(asset);` and use that here

martabal: Really ? I can't reproduce this issue on firefox and chrome.

>  I would just do $: isWebCompatible = isWebCompatibleImage(asset);

You mean without `$alwaysLoadOriginalFile` ? That would be a regression, it was implemented because some users want to always load the original file like on mobile.

michelheusschen: It is fixed. My bad, I still had an old version

michelheusschen (approved):

web/src/lib/components/asset-viewer/photo-viewer.svelte
martabal: Ahah no worries, thank you for your feedback, it's much better now

PR #7591, opened 2024-03-03
feat(web): show user quota on server stats page

samholton:
Addresses https://github.com/immich-app/immich/discussions/7582

**before**
before

**after**
after


jrasm91:
Can you add a screenshot to the PR description of before and after?

jrasm91 (approved):

PR #7593, opened 2024-03-03
perf(web): optimize date groups

mertalev:
## Description

The algorithm for grouping by date is inefficient and doesn't scale well. There's also no caching, so revisiting buckets re-groups again. This PR optimizes this with a more efficient algorithm and ensures that a given bucket will only ever be grouped into dates once.

## How Has This Been Tested?

Tested with a console log during grouping to check that a given bucket is only grouped once, even after scrubbing away and back to the bucket.

### DIFF
--- a/web/src/lib/components/photos-page/actions/select-all-assets.svelte
+++ b/web/src/lib/components/photos-page/actions/select-all-assets.svelte
@@ -21,7 +21,7 @@
         if ($isSelectAllCancelled) {
           break;
         }
-        await assetStore.loadBucket(bucket.bucketDate, BucketPosition.Unknown);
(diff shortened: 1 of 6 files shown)


waclaw66:
Does this solve issue https://github.com/immich-app/immich/issues/5030 ?

danieldietzler (approved):
Great stuff! :)

mertalev:
> Does this solve issue https://github.com/immich-app/immich/issues/5030 ?

No, that one is because we only do a rough estimate of height initially and update as we fetch the assets. The justified layout makes it hard to get it exact. There's probably room for improvement there though.

PR #7594, opened 2024-03-03
perf(web): optimize response sizes for initial page load

mertalev:
## Description

This PR tries to optimize the initial page load and generally reduce response sizes.
* Making the AssetViewer a dynamic import since it's bulky and doesn't need to be loaded immediately
* Enabling compression at the fastest setting (even the fastest dramatically reduces response size, and slower settings see diminishing returns)
* Using webp instead of jpeg for Memory Lane


server/src/immich/main.ts
```diff
@@ -19,6 +20,7 @@ export async function bootstrap() {
   app.set('trust proxy', ['loopback', 'linklocal', 'uniquelocal']);
   app.set('etag', 'strong');
   app.use(cookieParser());
+  app.use(compression({ level: 1 }));
```
benmccann, line 23: Looks like Cloudflare found level 4 to be best for brotli and level 8 for gzip

https://blog.cloudflare.com/this-is-brotli-from-origin

mertalev: Ooh interesting, let me compare

mertalev: GZip level 8:

gzip8

GZip level 4:

gzip4

GZip level 1:

gzip1

No compression:

no-compression

mertalev: Looking at the time bucket response, it goes from 328kb to 46.5kb with level 1, but level 8 only lowers that by 3kb down to 43.6kb and ends up taking longer. Level 4 seems reasonable, but the sizes all look pretty similar to level 1 so hard to tell if it's better overall.

web/src/lib/components/photos-page/asset-grid.svelte
```diff
@@ -451,17 +450,19 @@
...
-      {assetStore}
-      asset={$viewingAsset}
-      {isShared}
-      {album}
-      on:previous={handlePrevious}
-      on:next={handleNext}
-      on:close={handleClose}
-      on:action={({ detail: action }) => handleAction(action.type, action.asset)}
-    />
+    {#await import('../asset-viewer/asset-viewer.svelte') then AssetViewer}
```
martabal, line 453: Nice !

martabal (approved):

jrasm91 (approved):
LGTM

benmccann:
I just found this old SvelteKit issue: https://github.com/sveltejs/kit/issues/5431. We used to have `compression` built-in, but had to remove it because it was regularly causing the server to hang. Unfortunately the fix still hasn't been merged: https://github.com/expressjs/compression/pull/183. And you can't provide the answer to the question that the PR is blocked on as the repository has blocked all outside contributors from commenting :confused: 

The `precompress` option might be safer. Ideally we'd be able to use `precompress` to aggressively precompress the assets in brotli and then do a quicker gzip compression with `compress` here for dynamic responses. But I don't know how much we want to chance the server randomly hanging. Maybe we should read through that old issue and see if there were certain circumstances that would cause it to trigger. I don't remember all the details. I'm guessing `precompress` would cover most requests though and should get us an even larger gain.

mertalev:
Oof, that sucks. I'll remove the compression for now until we take a closer look

aviv926:
> I just found this old SvelteKit issue: [sveltejs/kit#5431](https://github.com/sveltejs/kit/issues/5431). We used to have `compression` built-in, but had to remove it because it was regularly causing the server to hang. Unfortunately the fix still hasn't been merged: [expressjs/compression#183](https://github.com/expressjs/compression/pull/183). And you can't provide the answer to the question that the PR is blocked on as the repository has blocked all outside contributors from commenting 😕
> 
> The `precompress` option might be safer. Ideally we'd be able to use `precompress` to aggressively precompress the assets in brotli and then do a quicker gzip compression with `compress` here for dynamic responses. But I don't know how much we want to chance the server randomly hanging. Maybe we should read through that old issue and see if there were certain circumstances that would cause it to trigger. I don't remember all the details. I'm guessing `precompress` would cover most requests though and should get us an even larger gain.

It seems that this project is not going to receive a commitment soon, there is this project that received a commitment 3 days ago and a year ago and he is the one who offered the PR
which is not merged into main

Maybe you should keep an eye on it?
https://github.com/fernandolguevara/compression/tree/master

benmccann:
That's the fork that the fix (https://github.com/expressjs/compression/pull/183) is coming from. It was called too hacky to merge as it relies on brittle Node internals, so I don't think we want to use it. The author of the PR wants to keep support for ancient versions of Node rather than address that, which means it will probably never be merged.

Unfortunately, all outside users are blocked from sending fixes, so it is likely to remain broken forever.

Screenshot from 2024-03-06 15-06-08

Perhaps we should build our own compression middleware into SvelteKit or polka or something.

benmccann:
There was a PR to polka that Vite copied and has been using the past couple of years. I'll see if we can't get that merged to use. See here: https://github.com/sveltejs/kit/issues/9593#issuecomment-1982019157

mertalev:
That'd be awesome!

PR #7595, opened 2024-03-03
feat(web): add share link on asset-viewer

samholton:
Adds a sharing link on the individual asset viewer. This matches the functionality on mobile to share a single image rather than going back to grid view and selecting it. Mentioned in https://github.com/immich-app/immich/discussions/7561 as well.

**before**
before

**after**
after

after with menu

### DIFF
--- a/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
+++ b/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
@@ -9,7 +9,6 @@
   import {
     mdiAlertOutline,
     mdiArrowLeft,
-    mdiCloudDownloadOutline,
     mdiContentCopy,
     mdiDeleteOutline,
     mdiDotsVertical,
@@ -39,7 +38,14 @@
 
   $: isOwner = asset.ownerId === $user?.id;
 
-  type MenuItemEvent = 'addToAlbum' | 'addToSharedAlbum' | 'asProfileImage' | 'runJob' | 'playSlideShow' | 'unstack';
+  type MenuItemEvent =
+    | 'addToAlbum'
+    | 'addToSharedAlbum'
+    | 'asProfileImage'
+    | 'download'
+    | 'playSlideShow'
+    | 'runJob'
+    | 'unstack';
 
   const dispatch = createEventDispatcher<{
     back: void;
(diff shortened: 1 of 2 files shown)


web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
```diff
@@ -139,6 +142,14 @@
         title="Download"
       />
     {/if}
+    {#if showShareButton}
```
martabal, line 145: Can you restrict it to asset owners ?

samholton: Initially I included restriction on owner but tested selecting multiple images from the grid view including those that were from partner share. The partner photos showed up in the share link as well. Should a user be able to share partner share photos?

jrasm91: I think owner or partner can both share it

jrasm91 (approved):
This looks fine to me, although I think shared should be the first icon. Also there are a lot of actions now. Maybe we should move download into the overflow menu now.

samholton:
> This looks fine to me, although I think shared should be the first icon. Also there are a lot of actions now. Maybe we should move download into the overflow menu now.

Do we think users would more often download a single image or share a single image?

jrasm91:
It is how Google does it I believe.

danieldietzler (approved):
LGTM!

web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
martabal: Yep, but that's an issue for assets in shared albums
image

samholton: I'll take a look at assets in shared albums

samholton: I was able to reproduce by inviting a user to the shared album that was also _not_ a partner share. The same error is displayed if selecting the image from the main album page and selecting share as well. So seems like an existing but not necessarily related to this change.

Created https://github.com/immich-app/immich/issues/7604 for tracking

PR #7596, opened 2024-03-03
feat(web): improve alt text

mertalev:
This PR aims to make the alt text on web more helpful and descriptive. It changes all alt texts that are an unreadable ID to either a static text, an asset/album description or a custom alt text based on date, location and people.


danieldietzler (approved):
Thanks for working on this! This is really something we should focus more on

PR #7620, opened 2024-03-04
fix(server): regular version check

rovo89:
`dt.diffNow()` equals `dt.diff(DateTime.now())`, so it returns a negative number when `dt` is in the past (which it always is in this case).

Therefore we could only get over the condition during startup (when `this.releaseVersionCheckedAt` isn't set yet), effectively breaking update notifications while the server is running.

There are different ways to write this, e.g. 
```javascript
1: DateTime.now().diff(this.releaseVersionCheckedAt).as('minutes') < 60
2: -this.releaseVersionCheckedAt.diffNow().as('minutes') < 60
3: Math.abs(this.releaseVersionCheckedAt.diffNow().as('minutes')) < 60
4: this.releaseVersionCheckedAt.diffNow().as('minutes') > -60
```

I prefer the first one because it feels most readable. Second is basically reversing the timestamps, third does the same (assuming that time only moves forward). I don't like the fourth way, hard to read from my point of view.

### DIFF
--- a/server/src/domain/server-info/server-info.service.ts
+++ b/server/src/domain/server-info/server-info.service.ts
@@ -149,7 +149,7 @@ export class ServerInfoService {
       }
 
       // check once per hour (max)
-      if (this.releaseVersionCheckedAt && this.releaseVersionCheckedAt.diffNow().as('minutes') < 60) {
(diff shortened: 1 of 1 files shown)


danieldietzler (approved):
I agree. I think this is the best solution

jrasm91:
Thanks for fixing this, but I noticed you didn't add any unit tests. Would that be something you would be able to do in a follow up PR?

PR #7637, opened 2024-03-05
fix(web): prettify album download filename

waclaw66:
This PR changes album download zip filename

from: `album-2024-03-05-09-42-43.zip`
to: `album-20240305_094243.zip`

### DIFF
--- a/web/src/lib/utils/asset-utils.ts
+++ b/web/src/lib/utils/asset-utils.ts
@@ -59,8 +59,8 @@ export const downloadArchive = async (fileName: string, options: DownloadInfoDto
 
   for (let index = 0; index < downloadInfo.archives.length; index++) {
     const archive = downloadInfo.archives[index];
-    const suffix = downloadInfo.archives.length === 1 ? '' : `+${index + 1}`;
-    const archiveName = fileName.replace('.zip', `${suffix}-${DateTime.now().toFormat('yyyy-LL-dd-HH-mm-ss')}.zip`);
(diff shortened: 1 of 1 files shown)


web/src/lib/utils/asset-utils.ts
```diff
@@ -59,8 +59,8 @@ export const downloadArchive = async (fileName: string, options: DownloadInfoDto
 
   for (let index = 0; index < downloadInfo.archives.length; index++) {
     const archive = downloadInfo.archives[index];
-    const suffix = downloadInfo.archives.length === 1 ? '' : `+${index + 1}`;
-    const archiveName = fileName.replace('.zip', `${suffix}-${DateTime.now().toFormat('yyyy-LL-dd-HH-mm-ss')}.zip`);
+    const suffix = downloadInfo.archives.length > 1 ? `+${index + 1}` : '';
+    const archiveName = fileName.replace('.zip', `${suffix}-${DateTime.now().toFormat('yyyyLLdd_HHmmss')}.zip`);
```
danieldietzler, line 63: I do prefer the underscore to separate day and hour, but I'm not sure if removing all hyphens makes it better readable.
I would probably vote for
```suggestion
    const archiveName = fileName.replace('.zip', `${suffix}-${DateTime.now().toFormat('yyyy-LL-dd_HH-mm-ss')}.zip`);
```

waclaw66: Seems it's matter of taste. I do prefer more compact format, however your suggestion is definetelly better then the original one.

jrasm91 (approved):

PR #7655, opened 2024-03-05
feat(server): hardware HDR tonemapping for RKMPP

fyfrey:
**enables full hardware transcoding for HDR videos on Rockchip SoCs**

- pipeline: hardware decode -> scaling -> opencl tonemapping -> hardware encoding
- only RK3588 supports tonemapping, other Rockchips SoCs still use software decoding/scaling/tonemapping but retain hardware encoding
- users need to provide `libmali.so` for tonemapping
- added docs for RKMPP transcoding
- enables 4K HDR -> FHD SDR transcode at 8x realtime speed :rocket:  (in software it's 0.25x realtime...)

### DIFF
--- a/docker/hwaccel.transcoding.yml
+++ b/docker/hwaccel.transcoding.yml
@@ -38,9 +38,10 @@ services:
       - /dev/dri:/dev/dri
       - /dev/dma_heap:/dev/dma_heap
       - /dev/mpp_service:/dev/mpp_service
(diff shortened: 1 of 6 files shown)


mertalev (approved):
Great stuff! This will also make it easier to add hardware tone-mapping for the other backends.

mertalev (approved):

PR #7663, opened 2024-03-05
feat(server,web): make user deletion delay configurable

samholton:
First part to add a new User Settings section to add configuration option for user deletion delay. Next up is to add a force option to queue up user for immediate deletion.

delete delay

### DIFF
--- a/web/src/lib/components/admin-page/settings/admin-settings.svelte
+++ b/web/src/lib/components/admin-page/settings/admin-settings.svelte
@@ -7,6 +7,7 @@
   } from '$lib/components/shared-components/notification/notification';
   import { handleError } from '$lib/utils/handle-error';
   import { getConfig, getConfigDefaults, updateConfig, type SystemConfigDto } from '@immich/sdk';
+  import { loadConfig } from '$lib/stores/server-config.store';
   import { cloneDeep } from 'lodash-es';
   import { createEventDispatcher, onMount } from 'svelte';
   import type { SettingsEventType } from './admin-settings';
@@ -34,6 +35,8 @@
       config = cloneDeep(newConfig);
       savedConfig = cloneDeep(newConfig);
       notificationController.show({ message: 'Settings saved', type: NotificationType.Info });
+
+      await loadConfig();
 
       dispatch('save');
     } catch (error) {
--- a/web/src/lib/components/admin-page/settings/user-settings/user-settings.svelte
+++ b/web/src/lib/components/admin-page/settings/user-settings/user-settings.svelte
@@ -26,7 +26,7 @@
           inputType={SettingInputFieldType.NUMBER}
           min={1}
           label="DELETE DELAY"
-          desc="Number of days after removal to permanently delete an user's account and assets. Changing this will impact users already queued for deletion."
+          desc="Number of days after removal to permanently delete a user's account and assets. The user deletion job runs at midnight to check for users that are ready for deletion. Changes to this setting will be evaluated at the next execution."
           bind:value={config.user.deleteDelay}
           isEdited={config.user.deleteDelay !== savedConfig.user.deleteDelay}
         />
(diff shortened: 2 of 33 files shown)


web/src/lib/components/admin-page/settings/admin-settings.svelte
```diff
@@ -35,6 +36,8 @@
       savedConfig = cloneDeep(newConfig);
       notificationController.show({ message: 'Settings saved', type: NotificationType.Info });
 
+      await loadConfig();
```
samholton, line 39: Somewhat outside the scope of this PR, but added this to ensure the system settings are reloaded after they are saved. This is used in the confirmation screen when deleting a user. Without this change, updating the setting and then removing a user without reloading the page will still show the old value.

jrasm91: newConfig should be the latest config. Maybe we just need to update the store with the new value.

web/src/lib/components/admin-page/settings/user-settings/user-settings.svelte
```diff
@@ -0,0 +1,44 @@
...
+</script>
+
+<div>
+  <div in:fade={{ duration: 500 }}>
+    <form autocomplete="off" on:submit|preventDefault>
+      <div class="ml-4 mt-4 flex flex-col gap-4">
+        <SettingInputField
+          inputType={SettingInputFieldType.NUMBER}
+          label="DELETE DELAY"
+          desc="Number of days after removal to permanently delete an user's account and assets. Changing this will impact users already queued for deletion."
```
jrasm91, line 28: I think you can add min=1 here

jrasm91 (commented):
Looks good.

jrasm91 (approved):

web/src/lib/components/admin-page/settings/user-settings/user-settings.svelte
samholton: Thanks, just noticed I was missing that

jrasm91:
I mean it might need to be min={1}

web/src/lib/components/admin-page/settings/admin-settings.svelte
samholton: Any examples of that being done somewhere? I tried replacing this with 
```
serverConfig.update(() => ({ ...savedConfig, loaded: true }));
``` 
Also tried newConfig but the delete confirm modal shows undefined after making a settings update and not refreshing page.

samholton: Ah, its different, one is `SystemConfigDto` and the other is `ServerConfigDto`

samholton: Ok, this works for my specific field and saves the network call. But does not refresh any other fields. What do you think, network call and ensure everything is up to date, or pick and choose what is updated?

```
serverConfig.update(() => ({ userDeleteDelay: newConfig.user.deleteDelay, loaded: true }));
```

samholton:
> I mean it might need to be min={1}

`min=1` worked but updated it to be consistent with the others

alextran1502:
Hello, thank you for the PR! What is the impact of changing the number of day for already queued for delete user?

samholton:
> Hello, thank you for the PR! What is the impact of changing the number of day for already queued for delete user?

The number of days is not stored with the user record, so each time the batch job runs, it will evaluate the queued users against the current setting value. 

I attempted to indicate this in the text: https://github.com/immich-app/immich/pull/7663/files#diff-2a84dfc76e5b23ae7a77e3ba11ca749128214aff7a18e5e207acfe551df563fbR29

web/src/lib/components/admin-page/settings/admin-settings.svelte
jrasm91: Oh right, this is just the public config. You can leave the load config in for now. I think the server can emit an event through the websocket to push out the update though, but it doesn't really matter.

jrasm91:
We should maybe clarify somewhere that a job runs at midnight to check for users that are ready to be deleted and then we can add that this change will take effect the next time the job runs.

samholton:
updated

updated text

jrasm91:
Yeah that looks good to me.

PR #7671, opened 2024-03-06
fix(mobile): Refactors exif bottom sheet to use widgets and fixes slow sliding up exif bottom sheet

martyfuhry:
Simple if / else statement to replace the problem line:

```dart
assetWithExif.when(
  data: (data) => DescriptionInput(asset: data),
  error: (error, stackTrace) => Icon(
    Icons.image_not_supported_outlined,
    color: context.primaryColor,
  ),
  loading: () => const SizedBox(
    width: 75,
    height: 75,
    child: CircularProgressIndicator.adaptive(),
  ),
),
```

Otherwise, the functionality should be left entirely intact. Alls I did was replace the `buildFoo` functions with Stateless Widgets to hopefully take more advantage of widget cache.


alextran1502 (approved):
Amazing refactor! Thank you so much

waclaw66:
Could you please show only non-hidden people as it is in the web client? Thanks.

PR #7677, opened 2024-03-06
fix(web): consistent modal escape behavior

michelheusschen:
The escape key inconsistently closes modals, sometimes doing nothing and other times closing the modal and navigating to the previous page.

To address this the FullscreenModal exposes a callback that combines the `clickOutside` and `onEscape` events. The clickOutside action has been modified (in a backwards compatible way) to accept callback functions. Using a callback function for `onEscape` prevents the escape key from propagating and navigating to previous pages.

Not all changes are related to this issue, some refactoring is included. I've manually tested all instances where `FullScreenModal` was used and some of `ConfirmDialogue`

### DIFF
--- a/web/src/lib/components/admin-page/delete-confirm-dialoge.svelte
+++ b/web/src/lib/components/admin-page/delete-confirm-dialoge.svelte
@@ -10,6 +10,7 @@
   const dispatch = createEventDispatcher<{
     success: void;
     fail: void;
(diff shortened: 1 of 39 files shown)


jrasm91 (approved):
LGTM.

danieldietzler (approved):
LGTM

PR #7681, opened 2024-03-06
feat(server,web): add force delete to immediately remove user

samholton:
Addresses https://github.com/immich-app/immich/discussions/1616

regular delete
force delete confirmation
force delete ready


server/src/domain/user/user.service.ts
```diff
@@ -75,14 +75,21 @@ export class UserService {
     return this.userCore.updateUser(auth.user, dto.id, dto).then(mapUser);
   }
 
-  async delete(auth: AuthDto, id: string): Promise<UserResponseDto> {
+  async delete(auth: AuthDto, id: string, options?: DeleteUserOptionsDto): Promise<UserResponseDto> {
```
jrasm91, line 78: Just make options required.

server/src/immich/controllers/user.controller.ts
```diff
@@ -3,6 +3,7 @@ import {
   CreateUserDto as CreateDto,
   CreateProfileImageDto,
   CreateProfileImageResponseDto,
+  DeleteUserOptionsDto,
```
jrasm91, line 6: Just call it DeleteUserDto

server/src/immich/controllers/user.controller.ts
```diff
@@ -66,8 +67,12 @@ export class UserController {
 
   @AdminRoute()
   @Delete(':id')
-  deleteUser(@Auth() auth: AuthDto, @Param() { id }: UUIDParamDto): Promise<UserResponseDto> {
-    return this.service.delete(auth, id);
+  deleteUser(
+    @Auth() auth: AuthDto,
+    @Body() dto: DeleteUserOptionsDto,
```
jrasm91, line 72: Nit: I usually do param and then body

web/src/lib/components/admin-page/delete-confirm-dialoge.svelte
```diff
@@ -30,9 +36,26 @@
...
-      <p>
-        <b>{user.name}</b>'s account and assets will be permanently deleted after {$serverConfig.userDeleteDelay} days.
-      </p>
+      {#if !forceDelete}
+        <p>
+          <b>{user.name}</b>'s account and assets will be scheduled for permanent deletion in {$serverConfig.userDeleteDelay}
+          days.
+        </p>
+      {/if}
+      {#if forceDelete}
```
jrasm91, line 45: Just use :else

web/src/lib/components/admin-page/restore-dialoge.svelte
```diff
@@ -11,10 +12,15 @@
   }>();
 
   const handleRestoreUser = async () => {
-    const { deletedAt } = await restoreUser({ id: user.id });
-    if (deletedAt == undefined) {
-      dispatch('success');
-    } else {
+    try {
```
samholton, line 15: Depending on how long it takes for the background job to actually clean up the user, it will most likely be shown on the ui as pending deletion (get user list happens before job runs). Noticed that the ui wasn't catching the exception and just hanging when trying to restore a user that no longer existed.

We could add some sort of force delete column to the user object to handle this edge case (filter them out from the list call).

jrasm91: Maybe we can just compare the deleted at timestamp with the delete duration to know if it is pending immediate deletion or not.

jrasm91: Although I guess force delete can be used independently of that... Hmm... I don't really want to add a column just for this. I would be down to send a web socket event and have the page wait up to a few seconds to receive it.

server/src/domain/user/user.service.ts
```diff
@@ -75,14 +75,21 @@ export class UserService {
...
-  async delete(auth: AuthDto, id: string): Promise<UserResponseDto> {
+  async delete(auth: AuthDto, id: string, options: DeleteUserDto): Promise<UserResponseDto> {
     const user = await this.findOrFail(id, {});
     if (user.isAdmin) {
       throw new ForbiddenException('Cannot delete admin user');
     }
 
     await this.albumRepository.softDeleteAll(id);
 
+    if (options && options.force) {
```
jrasm91, line 86: Options can't be null anymore so I would just add `const { force } = dto` at the top

jrasm91 (approved):
LGTM but I agree we can clean up the delete experience. Id recommend sending an event when a user is deleted over a web socket

web/src/lib/components/admin-page/restore-dialoge.svelte
samholton: Yeah, I didn't want to add a column either. If you have any examples of the websocket being used, I can take a look at that. But even with the websocket, the user will see the forced deleted user in the table with the restore button. Once the job is on the queue, even if they restore, the user will eventually get deleted.

jrasm91: Maybe we should add something to the message that says the action is irreversible.

samholton: Yeah, I was pointing out that the restore option could be misleading/confusing. The user can see it and click it, but at that point, the job is already in the queue and the delete is going to happen (even if they quickly click restore).

The ideal behavior is the user doesn't show up in the row after the confirmation modal is closed, but I'm not sure how we can do that without adding/setting something on the user table.

jrasm91: Maybe we can change the job to still require the user be soft delete either way. If the restore happens before the job is processed then you are all good. As soon as it is processed the user disappears from the table. Yeah I will have to think about it some more. Feel free to ping me on discord if it is easier to discuss it over there.

samholton: Calling it a night, will think more on it as well.

Although I like that idea, and actually I think that is already the behavior with the extra call out to `isReadyForDeletion`. I'll test it tomorrow, in that case would just need to add the webhook logic to update the user table once delete is complete.

web/src/lib/components/admin-page/delete-confirm-dialoge.svelte
```diff
@@ -30,9 +36,25 @@
...
+          <b>{user.name}</b>'s account and assets will be queued for permanent deletion <b>immediately</b>.
+        </p>
+      {/if}
+
+      <div class="flex justify-center m-4 gap-2">
+        <label class="text-sm dark:text-immich-dark-fg" for="forceDelete">
+          Queue user and assets for immediate deletion
+        </label>
+
+        <input type="checkbox" class="form-checkbox h-5 w-5 color" bind:checked={forceDelete} />
```
michelheusschen, line 55: You use `<label for="forceDelete">` above, so you should add that ID to the input field to link them together

```suggestion
        <input id="forceDelete" type="checkbox" class="form-checkbox h-5 w-5 color" bind:checked={forceDelete} />
```

web/src/lib/components/admin-page/restore-dialoge.svelte
jrasm91: Maybe we can add a new column called `status` and it defaults to active. Then we can have a "delete in progress state" and maybe others in the future like inactive, disabled, locked, etc.

web/src/lib/components/admin-page/delete-confirm-dialoge.svelte
samholton: I switched from slide to checkbox and didn't add, thanks!

web/src/routes/admin/user-management/+page.svelte
```diff
@@ -220,7 +221,7 @@
                       </button>
                     {/if}
                   {/if}
-                  {#if isDeleted(immichUser)}
+                  {#if isDeleted(immichUser) && immichUser.status != "removing"}
```
samholton, line 224: Looking to see if there is a way to reference the enum at https://github.com/immich-app/immich/pull/7681/files#diff-5769839dae446ca9c68221d6d1ec556da0810925a90ffc10096e8061d2ff3a3bR26 or if I need to recreate it in web

jrasm91:
I added some websocket stuff, and cleaned up a few things. Let me know what you think and I'll probably merge it tomorrow if it looks good to you.

web/src/lib/components/admin-page/delete-confirm-dialoge.svelte
```diff
@@ -30,15 +36,31 @@
...
+          days.
+        </p>
+      {/if}
+
+      <div class="flex justify-center m-4 gap-2">
+        <label class="text-sm dark:text-immich-dark-fg" for="forceDelete">
+          Queue user and assets for immediate deletion
+        </label>
+
+        <input id="forceDelete" type="checkbox" class="form-checkbox h-5 w-5 color" bind:checked={forceDelete} />
```
michelheusschen, line 61: Accidentally deleting a user is still a little too easy in my opinion, even with this extra checkbox. How about requiring an additional email input for permanent deletions, kinda like github does when deleting a repository?

Also the `color` class doesn't exist

server/src/domain/user/user.service.ts
```diff
@@ -73,22 +73,29 @@ export class UserService {
...
+    return mapUser(user);
   }
 
   async restore(auth: AuthDto, id: string): Promise<UserResponseDto> {
-    let user = await this.findOrFail(id, { withDeleted: true });
-    user = await this.userRepository.restore(user);
+    await this.findOrFail(id, { withDeleted: true });
     await this.albumRepository.restoreAll(id);
-    return mapUser(user);
+    return this.userRepository.update(id, { deletedAt: null, status: UserStatus.ACTIVE }).then(mapUser);
```
michelheusschen, line 98: `userRepository.restore` becomes unused

michelheusschen (approved):
No major issues, just some smaller things that could also be addressed later on

web/src/lib/components/admin-page/delete-confirm-dialoge.svelte
jrasm91: Alex had this same idea and I think it is a good one. I don't care if it is done now or in a separate pull request.

web/src/routes/admin/user-management/+page.svelte
```diff
@@ -211,11 +218,11 @@
                       </button>
                     {/if}
                   {/if}
-                  {#if isDeleted(immichUser)}
+                  {#if immichUser.deletedAt && immichUser.status === UserStatus.Deleted}
```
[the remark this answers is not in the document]
jrasm91: It is a little tricky because your can't actually restore the user besides in that error situation. Showing the button right away might be confusing.

server/src/domain/user/user.service.ts
jrasm91: Sweet. We should delete that code then.

web/src/lib/components/admin-page/delete-confirm-dialoge.svelte
samholton: I should have time this weekend to work on that - can do it here or in a new PR, doesn't matter to me either. You thinking an input box that appears when you check the box that disables the button until the user id is entered? Or a separate screen like github that shows up after you click delete that then asks you input?

server/src/domain/user/user.service.ts
samholton: removed and pushed

web/src/lib/components/admin-page/delete-confirm-dialoge.svelte
samholton: Added the extra input logic for force delete

samholton: Updated screenshots, the force checkbox jumping when enabling it may be a bit disorienting

jrasm91: Perfect. Looks great

aviv926:
Maybe a password for confirmation could be a more secure idea?

samholton:
> Maybe a admin password for confirmation could be a more secure idea?

The original intention was more about confirmation than authorization. I believe the GitHub workflow has the confirmation followed by MFA. 

I haven't actually tried disabling password with my oauth enabled, but could potentially be an issue if you have passwords disabled

PR #7689, opened 2024-03-07
chore(server): lower default max recognition distance for facial recognition

mertalev:
## Description

The feedback we've been receiving is more often that multiple people are recognized as the same person as opposed to duplicating the same person. The current algorithm is more effective at connecting faces than the previous one, so it seems appropriate to lower this default a touch.


samholton:
Agree, especially on younger faces. And I'd much rather merge a few extra faces than trying to separate them.

jrasm91 (approved):

PR #7706, opened 2024-03-07
feat(server): create a person with optional values

jrasm91:
* speed up person e2e (remove beforeEach)
* add option to set `birthDate`, `isHidden`, and `name` when creating a person
* sort person controller methods
* rename person service methods
* refactor person service update method to have less branching

### DIFF
--- a/e2e/src/api/specs/person.e2e-spec.ts
+++ b/e2e/src/api/specs/person.e2e-spec.ts
@@ -5,7 +5,7 @@ import { app, utils } from 'src/utils';
 import request from 'supertest';
 import { beforeAll, beforeEach, describe, expect, it } from 'vitest';
 
-describe('/activity', () => {
+describe('/person', () => {
   let admin: LoginResponseDto;
   let visiblePerson: PersonResponseDto;
   let hiddenPerson: PersonResponseDto;
--- a/e2e/src/api/specs/person.e2e-spec.ts
+++ b/e2e/src/api/specs/person.e2e-spec.ts
@@ -5,6 +5,14 @@ import { app, utils } from 'src/utils';
 import request from 'supertest';
 import { beforeAll, beforeEach, describe, expect, it } from 'vitest';
 
+const invalidBirthday = [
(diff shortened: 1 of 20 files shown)


e2e/src/api/specs/person.e2e-spec.ts
```diff
@@ -141,6 +145,41 @@ describe('/activity', () => {
...
 
+  describe('POST /person', () => {
+    it('should require authentication', async () => {
+      const { status, body } = await request(app).post(`/person`);
+      expect(status).toBe(401);
+      expect(body).toEqual(errorDto.unauthorized);
+    });
+
+    it('should not accept invalid birth dates', async () => {
+      for (const { birthDate, response } of invalidBirthday) {
```
danieldietzler, line 156: Maybe we should move the for each out of the `it` for a more helpful output if the test fails?

server/src/domain/person/person.service.ts
```diff
@@ -199,36 +196,41 @@ export class PersonService {
...
 
-      person = await this.repository.update({ id, faceAssetId: face.id });
+      faceId = face.id;
+    }
+
+    const person = await this.repository.update({ id, faceAssetId: faceId, name, birthDate, isHidden });
+
+    if (assetId) {
       await this.jobRepository.queue({ name: JobName.GENERATE_PERSON_THUMBNAIL, data: { id } });
     }
```
danieldietzler, line 228: Can't we merge this if with the if starting in l. 214?

danieldietzler (approved):
LGTM

server/src/domain/person/person.service.ts
jrasm91: No, the faceId needs to be saved before the job can be queued. I don't want there to be a race condition if the job runs before the update happens.

e2e/src/api/specs/person.e2e-spec.ts
jrasm91: I will do it in a follow up. There is only one test that runs either way right now.

PR #7709, opened 2024-03-07
refactor: api validators (boolean and date)

jrasm91:
* no longer allow so many "date" options, like `false`, 0, etc.
* standardize date validation (using isDateString)
* standardize boolean validation (correctly converts `'false'` and `'true'` in query params, etc.

The e2e tests for the cli failed without `fileCreatedAt.toISOString()` after these changes, which  now require valid date strings. I'm not sure what it was doing before but it is possible it was not working correctly.


danieldietzler (approved):
Nice!

PR #7710, opened 2024-03-07
fix: update e2e

martabal:


jrasm91 (approved):

danieldietzler (approved):

jrasm91:
I don't think e2e is at the correct version, can you update that?

martabal:
> I don't think e2e is at the correct version, can you update that?

Done

jrasm91 (approved):
you need to run `npm i` to update `package-lock.json` as well fyi.

PR #7732, opened 2024-03-08
refactor: search e2e

jrasm91:


e2e/src/api/specs/search.e2e-spec.ts
```diff
@@ -0,0 +1,224 @@
...
+        await utils.createAsset(admin.accessToken, {
+          deviceAssetId: `test-${filename}`,
+          assetData: { bytes, filename },
+        }),
+      );
+    }
+
+    for (const asset of assets) {
+      await utils.waitForWebsocketEvent({ event: 'upload', assetId: asset.id });
+    }
```
danieldietzler, line 56: Don't you want something like
```ts
await Promise.all(assets.map((asset) => utils.waitForWebsocketEvent({ event: 'upload', assetId: asset.id })));
```
instead? With the current approach this may wait for an event that already happened, no?

danieldietzler (approved):
LGTM!

e2e/src/api/specs/search.e2e-spec.ts
jrasm91: It works even if the event has already happened. I find mine more readable and there is no performance benefit with running them in parallel since they're just waiting either way.

PR #7734, opened 2024-03-08
docs: Update backup script example

mmomjian:
When using a deduplicating backup program (Borg, Restic, etc), using gzip is counter-productive because small changes in the input data stream can lead to much larger changes in the compressed data to backup.

As an alternative, some programs advise that you can use `gzip --rsyncable`, but uncompressed data is typically still preferred as it has a better chance of matching up with the existing chunks from prior backups.

I propose that we update the docs to make Borg more space efficient as Immich will often have small daily changes to the database.

Also fixed some unquoted vars

### DIFF
--- a/docs/docs/guides/template-backup-script.md
+++ b/docs/docs/guides/template-backup-script.md
@@ -47,6 +47,8 @@ REMOTE_BACKUP_PATH="/path/to/remote/backup/directory"
 
 # Backup Immich database
 docker exec -t immich_postgres pg_dumpall -c -U postgres > "$UPLOAD_LOCATION"/database-backup/immich-database.sql
+# For deduplicating backup programs such as Borg or Restic, compressing the content can increase backup size by making it harder to deduplicate. If you are using a different program or still prefer to compress, you can use the following command instead:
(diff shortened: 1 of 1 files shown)


docs/docs/guides/template-backup-script.md
```diff
@@ -46,18 +46,18 @@ REMOTE_BACKUP_PATH="/path/to/remote/backup/directory"
 ### Local
 
 # Backup Immich database
-docker exec -t immich_postgres pg_dumpall -c -U postgres | /usr/bin/gzip > $UPLOAD_LOCATION/database-backup/immich-database.sql.gz
+docker exec -t immich_postgres pg_dumpall -c -U postgres > "$UPLOAD_LOCATION"/database-backup/immich-database.sql
```
jrasm91, line 49: I think it is a good idea to leave gzip out, a lot of people run this on windows and such and it is simpler for that reason as well. It could still be a good idea to leave this original example un, perhaps commented out. What do you think?

mmomjian: Added a comment with the original example, added `--rsyncable` as this does not significantly change backup size and makes it much easier to transport and chunk.

danieldietzler (approved):
LGTM!

alextran1502 (approved):

PR #7744, opened 2024-03-08
chore: rename "Library" to "External Library" in system settings

declan8010:
This is intended to assist with any confusion regarding standard libraries.

### DIFF
--- a/web/src/routes/admin/system-settings/+page.svelte
+++ b/web/src/routes/admin/system-settings/+page.svelte
@@ -72,8 +72,8 @@
     },
     {
       item: LibrarySettings,
-      title: 'Library',
-      subtitle: 'Manage library settings',
+      title: 'External Library',
+      subtitle: 'Manage external library settings',
       key: 'library',
     },
     {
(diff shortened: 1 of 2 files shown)


web/src/routes/admin/system-settings/+page.svelte
```diff
@@ -72,8 +72,8 @@
     },
     {
       item: LibrarySettings,
-      title: 'Library',
-      subtitle: 'Manage library settings',
+      title: 'External Library',
+      subtitle: 'Manage external library settings',
       key: 'library',
```
jrasm91, line 77: might as well change this to `external-library` as well.

declan8010: Updated 👍

aviv926:
Can you change that as well?
Assuming there is no problem that it will be together with this PR @jrasm91.

image

declan8010:
Actioned 👍

jrasm91 (approved):

PR #7745, opened 2024-03-08
fix(web): album state after removing assets

michelheusschen:
The album summary (date range + asset count) isn't updated after removing assets and when removing all assets from an album, the page doesn't return to the initial 'empty state'. Fixed by making asset count reactive and creating a new `AlbumSummary` component, which also eliminates code duplication.

### DIFF
--- a/web/src/lib/components/album-page/album-viewer.svelte
+++ b/web/src/lib/components/album-page/album-viewer.svelte
@@ -3,11 +3,9 @@
   import SelectAllAssets from '$lib/components/photos-page/actions/select-all-assets.svelte';
   import { assetViewingStore } from '$lib/stores/asset-viewing.store';
   import { dragAndDropFilesStore } from '$lib/stores/drag-and-drop-files.store';
-  import { locale } from '$lib/stores/preferences.store';
   import { fileUploadHandler, openFileUploadDialog } from '$lib/utils/file-uploader';
   import type { AlbumResponseDto, SharedLinkResponseDto, UserResponseDto } from '@immich/sdk';
   import { onDestroy, onMount } from 'svelte';
-  import { dateFormats } from '../../constants';
   import { createAssetInteractionStore } from '../../stores/asset-interaction.store';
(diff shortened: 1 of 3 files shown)


web/src/lib/components/album-page/album-viewer.svelte
```diff
@@ -148,14 +122,7 @@
...
 
-      <!-- ALBUM SUMMARY -->
-      {#if album.assetCount > 0}
-        <span class="my-4 flex gap-2 text-sm font-medium text-gray-500" data-testid="album-details">
-          <p class="">{getDateRange()}</p>
-          <p>·</p>
-          <p>{album.assetCount} items</p>
-        </span>
-      {/if}
+      <AlbumSummary assetCount={album.assetCount} {album} />
```
jrasm91, line 125: it looks like you don't pass in assets here so does this mean it won't show ranges anymore? it seems like this doesn't work the way it used to.

web/src/routes/(user)/albums/[albumId]/+page.svelte
```diff
@@ -557,14 +532,12 @@
...
-                    <p class="">{getDateRange()}</p>
-                    <p>·</p>
-                    <p>{album.assetCount} items</p>
-                  </span>
-                {/if}
+                <AlbumSummary
+                  {assetCount}
+                  {album}
+                  oldestAsset={$assetStore.assets.at(-1)}
+                  newestAsset={$assetStore.assets.at(0)}
```
Comment (jrasm91) [inline web/src/routes/(user)/albums/[albumId]/+page.svelte:539]: This is the list of assets from the loaded time buckets for the album, not every asset in the album. Unless I am mistaken this will only show the date range for loaded assets.

web/src/lib/components/album-page/album-viewer.svelte
michelheusschen: It uses the range from the album and should behave exactly as before

jrasm91 (approved):
Perfect, thanks!

PR #7757, opened 2024-03-08
perf: precompress and cache assets

benmccann:
Rather than using `adapter-node` as I originally planned, I thought it would require fewer changes to keep using `adapter-static` and swap out Nest's static serving for the static serving that `adapter-node` does. Using `adapter-node` might not be a bad idea still, but this just seemed easier

I tested this locally with `make prod` and you can see assets being served with `br` content encoding now

Screenshot from 2024-03-08 16-17-36

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -51,6 +51,7 @@
         "rxjs": "^7.8.1",
         "sanitize-filename": "^1.6.3",
         "sharp": "^0.33.0",
+        "sirv": "^2.0.4",
         "thumbhash": "^0.1.1",
         "typeorm": "^0.3.17",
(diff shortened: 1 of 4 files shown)


mertalev:
Lighthouse likes the change - the score went from 61-62 to 65. But it's also saying the caching is inefficient for static assets:

efficient-cache-policy

mertalev:
On main it only shows `custom.css` there

mertalev:
It had to check for a path prefix of `/_app/immutable` for it to work

mertalev (approved):

benmccann:
Thanks for catching and fixing that!

PR #7685, opened 2024-03-07
perf(server): use queries to refresh library assets

mertalev:
## Description

The current implementation fetches all assets for a library, which is slow and uses a substantial amount of memory. This PR moves this logic to optimized queries instead.

Fixes #7373

PR #7759, opened 2024-03-09
fix(server): external library sync not working for large libraries

mertalev:
The earlier improvement in #7685 only works correctly up to 65k assets. This PR changes the implementation to receive paginated assets and keep track of which are online / offline / new. At the end, online and offline assets have their state updated and new assets are scanned. This is a bit slower than #7685, but works with large libraries and should still use little memory.

### DIFF
--- a/server/src/domain/library/library.service.ts
+++ b/server/src/domain/library/library.service.ts
@@ -640,9 +640,15 @@ export class LibraryService extends EventEmitter {
       .filter((validation) => validation.isValid)
       .map((validation) => validation.importPath);
 
-    const shouldRefresh = job.refreshAllFiles || job.refreshModifiedFiles;
-    const crawledAssetPaths = await this.crawlLibrary(library, validImportPaths);
-    let pathsToScan: string[] = shouldRefresh ? [...crawledAssetPaths] : [];
+    let rawPaths = await this.storageRepository.crawl({
+      pathsToCrawl: validImportPaths,
+      exclusionPatterns: library.exclusionPatterns,
+    });
+    const crawledAssetPaths = new Set<string>(rawPaths);
+
+    const shouldScanAll = job.refreshAllFiles || job.refreshModifiedFiles;
+    let pathsToScan: string[] = shouldScanAll ? rawPaths : [];
+    rawPaths = [];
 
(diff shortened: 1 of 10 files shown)


server/src/domain/library/library.service.ts
```diff
@@ -640,27 +640,56 @@ export class LibraryService extends EventEmitter {
...
       pathsToCrawl: validImportPaths,
       exclusionPatterns: library.exclusionPatterns,
     });
-    const crawledAssetPaths = rawPaths.map((filePath) => path.normalize(filePath));
+    const crawledAssetPaths = new Set<string>(rawPaths);
 
-    this.logger.debug(`Found ${crawledAssetPaths.length} asset(s) when crawling import paths ${library.importPaths}`);
+    const shouldScanAll = job.refreshAllFiles || job.refreshModifiedFiles;
+    let pathsToScan: string[] = shouldScanAll ? rawPaths : [];
+    rawPaths = [];
```
jrasm91, line 651: are you trying to get gc to free up the memory, or what is the purpose of setting this to an empty list?

server/src/infra/repositories/asset.repository.ts
```diff
@@ -184,10 +184,10 @@ export class AssetRepository implements IAssetRepository {
   }
 
   @GenerateSql({ params: [[DummyValue.UUID]] })
-  @ChunkedArray()
-  getByLibraryId(libraryIds: string[]): Promise<AssetEntity[]> {
-    return this.repository.find({
-      where: { library: { id: In(libraryIds) } },
+  getLibraryAssetPaths(pagination: PaginationOptions, libraryId: string): Paginated<AssetEntity> {
```
jrasm91, line 187: Do you think it is worth defining a specific type so we know in the service there are only 3 fields on the "asset"?

jrasm91 (approved):
LGTM

server/src/domain/library/library.service.ts
mertalev: It's just to help with GC

server/src/infra/repositories/asset.repository.ts
mertalev: Yeah, that's a good idea

etnoy (approved):

PR #7760, opened 2024-03-09
feat(mobile): Adds file upload progress stats

othyn:
Closes #7379

See screenshot below, specifically its the new "File progress: ..." row, as the new first row in the table at the bottom of the view containing all the current uploading file information.

Adds file upload progress stats:
- Current file size uploaded (formatted to nearest unit)
- Current file size to be uploaded (formatted to nearest unit)
- X [Y]bytes per second upload speed (formatted to nearest unit)

Tested:
- Uploading files to the server in the iOS emulator and watching for the file stats and upload speed to ensure they align
- Not sure what in the automated test suite we could add for checking this, as it is just an information readout

~~NOTE: This does require localisation for other languages, and I'm happy to be pointed in the right direction for how that's done (e.g. just a Google translate?), as it adds a new language key.~~

This was the reference I found and subsequently why I removed the non-'en-US' translations: https://github.com/immich-app/immich/pull/7323/files/25ca7303acf0bbcfd739d04dc318c6bfb2042867#r1500831992

Hope that was the right thing to do!

Screenshot of the new feature, its the new "File progress: ..." row, as the new first row in the table at the bottom of the view containing all the current uploading file information:

simulator_screenshot_FA6759D6-21F9-4E10-8ED9-EA194FC2AF9E


othyn:
This was the reference I found and subsequently why I removed the non-'en-US' translations: https://github.com/immich-app/immich/pull/7323/files/25ca7303acf0bbcfd739d04dc318c6bfb2042867#r1500831992

alextran1502:
Hello, thank you for the PR. I tested and found some bugs below regarding the stats of the upload speed

image

image

image

othyn:
> Hello, thank you for the PR. I tested and found some bugs below regarding the stats of the upload speed
> 
> 
> 
> image
> 
> 
> 
> image
> 
> 
> 
> image


Weird! I didn't have any such issues when testing on mine, but that tale is as old as time when it comes to software dev.

Firstly, glad that the file size and uploaded stats appear to work fine.

The first two images look like overflows, so I can get some boundary checks added to curtail those.

The third one - that's odd. Was it always/consistently displaying '0.0 B/s' or was it just showing it periodically? For example just showing 0 for one second then correct thereafter?

The upload speed is polled every second to stop calculating on every data state update (as it just looked a bit silly).

othyn:
So I've added a boundary check to the value before its stored into the data object to hopefully stop downstream overflow values.

However, I'm still unable to re-produce either issue in my local simulator and testing. Is there any re-producible scenario in which the `0.0 B/s` string occurs for more than one update cycle (one second)? As the upload speed is re-calculated and I'd expect at worst it would only occur for one cycle, a gut feeling at the moment without more information to go on.

It seems to be working pretty robustly for me, is all. I'd like to fix the bug, but cannot at this time re-produce it locally.

Here's an example of an upload for me:

https://github.com/immich-app/immich/assets/7256684/883582c1-2094-4bbc-bedf-88012d710a04

othyn:
Just under some purely heuristic debugging (as I can't get a re-producible scenario) I've added some more fixes just to ensure things are reset to a known state at the beginning of each file upload as a common strategy to reduce issues. Not sure if this will be a direct fix, but it will certainly just help keep things to be in an expected state, clean and tidy.

othyn:
Seems to be a bit more robust now, resetting correctly between cycles/file uploads.

https://github.com/immich-app/immich/assets/7256684/0f998289-0544-4528-a510-c9a7616ed6aa

othyn:
@alextran1502 ah, thanks for the test case! Can confirm, grabbing a load of test images and running an upload, I can see that behaviour now.

As the uploads occur so quickly (in under a second), it never has chance to hit the first 'cycle' to calculate the upload speed over the given period.

Although I'm not sure on a 'fix', as the only thing will be to lower the polling rate to sub-one-second to try and squeeze a calculation in before the upload ends. I can understand your push to get the functionality to be consistent, but given its on screen for less than a second anyway, is it a _big_ issue? I'd imagine upload speeds are more desirable to know on longer running transfers. If its done before you can see it then you're kinda winning anyway haha

All I can think is to drop the polling rate down, but then I ran into issues with the maths being wonky.

The other option is to dynamically hide the upload speed until _after_ the first poll. So items that upload in under a second just display the file size progress without the upload speed. But that may seem like inconsistent UX behaviour.

Here's a working re-production of the '0.0 B/s' issue using 10 x 500KB-1.5MB JPEG's:

https://github.com/immich-app/immich/assets/7256684/358caaad-8933-460a-bdad-39ee66c1b031

leleogere:
A potential fix (but it would probably require to change the implementation a lot) would be to compute the upload speed across multiple files, with like an exponential moving average maybe?

othyn:
@leleogere 

> A potential fix (but it would probably require to change the implementation a lot) would be to compute the upload speed across multiple files, with like an exponential moving average maybe?

That would be a good way to do it, just persist the upload speed. Hmmm, will have a think on how to do that. It shouldn't be too complex, its mainly just not fully resetting the upload speed, only showing zero when something isn't uploading then restoring the prior value.

alextran1502:
@leleogere I was thinking the same thing. If it doesn't work, I am fine to leave the upload speed out, just leave the file's size progress

leleogere:
Not completely in the scope of this PR, but as we gather the number of files to upload in any case, it might be interesting to also get their size, and estimate the remaining time before the end of the upload using the upload speed that is computed.

othyn:
Okay, so I've implemented an average upload speed and now I see what you guys were wanting as it is much better.

I capped it to the latest 10 upload speed calculations, in order to keep things somewhat recent _and_ memory sensitive.

I've also done a small tweak to the UI to place the stats directly underneath the progress bar, as its far more relevant being there than it is in the file table. The file table is contextual to the file, so it didn't make sense to have the upload progress stats placed there.

Instead the upload progress stats are now in smaller font directly underneath the progress bar, much closer to my initial mockup on #7379. They look far more natural and at home in their new location.

I've also swapped to using Immich's monospace font for the stats display, as to ensure that the numbers don't cause the UI to bounce around as the upload stats values change, which looks far better IMO. Its a common pet peeve of mine, and this change makes it far more readable.

See what you guys think (@alextran1502 & @leleogere) (apologies for the lag in places, thats iOS simulator for you):

https://github.com/immich-app/immich/assets/7256684/5f06adc6-bd7f-4d82-9750-ba8a14628821

leleogere:
This feels way better! Nice touch to use a monospace font.

alextran1502:
Amazing! I will look at the code later today and merge

alextran1502 (approved):

PR #7763, opened 2024-03-09
feat: Remove play button in memories if only 1 photo

declan8010:
This commit introduces a change to hide the play bar at the top of the memory viewer when there's only one image in the memory. Previously, the play bar was displayed regardless of the number of images, leading to unnecessary UI elements for single-image memories.

The Gif below shows single view and then switching between the two modes.

firefox_CWquOsB2YH

### DIFF
--- a/web/src/lib/components/memory-page/memory-viewer.svelte
+++ b/web/src/lib/components/memory-page/memory-viewer.svelte
@@ -112,7 +112,7 @@
         </p>
       </svelte:fragment>
 
-      {#if !galleryInView}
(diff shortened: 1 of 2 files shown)


declan8010:
No idea what has happened here it seems to have closed my [other PR](https://github.com/immich-app/immich/pull/7763/commits/bacb45b6f7eb1c53d6741a5ba7f6231b253a44ac) which is now merged into this one not sure if someone can split it or remove it for a seperate submission

PR #7764, opened 2024-03-09
fix: Hide play bar when there's only one image in memory

declan8010:
This commit introduces a change to hide the play bar at the top of the memory viewer when there's only one image in the memory. Previously, the play bar was displayed regardless of the number of images, leading to unnecessary UI elements for single-image memories.

The Gif below shows single view and then switching between the two modes.

firefox_CWquOsB2YH


declan8010:
Fixes #7572

web/src/lib/components/memory-page/memory-viewer.svelte
```diff
@@ -112,7 +112,7 @@
         </p>
       </svelte:fragment>
 
-      {#if !galleryInView}
+      {#if currentMemory.assets.length > 1} <!-- Display play bar only if there's more than 1 image -->
```
michelheusschen, line 115: We should still show this when there are more memories next. The comment adds no value in my opinion and is better left out.

```suggestion
      {#if canGoForward}
```

alextran1502 (approved):

PR #7769, opened 2024-03-09
refactor: asset e2e

jrasm91:
* refactor asset e2e tests
* remove unused files


server/e2e/api/specs/asset.e2e-spec.ts
```diff
@@ -1,1151 +0,0 @@
...
-    let assets: AssetEntity[] = [];
-    for await (const assetsPage of assetPagination) {
-      assets = [...assets, ...assetsPage];
-    }
-    return assets.map((a) => a.id);
-  };
-
-  describe(AssetRepository.name, () => {
-    describe('getWithout', () => {
-      describe('WithoutProperty.FACES', () => {
```
jrasm91, line 1108: FYI, I have removed this set of tests. There is no way to test them in e2e and I'm not sure I can about adding them back anywhere else.

danieldietzler (approved):
Nice one! I really love how the e2e are structured now

PR #7777, opened 2024-03-09
feat(mobile): ios - do not auto lock on foreground backup setting

shenlong-tanwen:
#### Changes made

- A new iOS specific app setting to not put the device to sleep when an active foreground backup is in progress. This is useful for initial bulk upload without the user needing to change the device level setting to keep the device alive during backup


dvbthien176:
Hi @shenlong-tanwen, thanks for your pull request. I have an ideal: when turn on disable auto lock screen, the device should automatically change to a black display to save battery, similar Amazon Photos work to disable auto lock screen. Below is an example screen from Amazon Photos.
Ảnh màn hình 2024-03-09 lúc 21 01 37Ảnh màn hình 2024-03-09 lúc 21 39 49

shenlong-tanwen:
> Hi @shenlong-tanwen, thanks for your pull request. I have an ideal: when turn on disable auto lock screen, the device should automatically change to a black display to save battery, similar Amazon Photos work to disable auto lock screen.

This seems like an entire page dedicated to over night uploads. We can even replace my current implementation with a specific route such as from your example to handle over night uploads, yes. This should also make it trivial on which page the app would not sleep. I'll wait for the opinion of others as well before switching the implementation. Thanks a lot for the input!

aviv926:
> > Hi @shenlong-tanwen, thanks for your pull request. I have an ideal: when turn on disable auto lock screen, the device should automatically change to a black display to save battery, similar Amazon Photos work to disable auto lock screen.
> 
> This seems like an entire page dedicated to over night uploads. We can even replace my current implementation with a specific route such as from your example to handle over night uploads, yes. This should also make it trivial on which page the app would not sleep. I'll wait for the opinion of others as well before switching the implementation. Thanks a lot for the input!

If in the end you decide to choose this design, you should pay attention to the thought of moving the elements on this page to avoid the possibility of screen burn.

DaanHend:
> Hi @shenlong-tanwen, thanks for your pull request. I have an ideal: when turn on disable auto lock screen, the device should automatically change to a black display after 5 minutes of idle state to save battery, similar Amazon Photos work to disable auto lock screen. Below is an example screen from Amazon Photos. What do you think of it? Thanks Ảnh màn hình 2024-03-09 lúc 21 01 37Ảnh màn hình 2024-03-09 lúc 21 39 49

This is a similar as Synology Photos does it. With this app it will count down an turn of the screen. I think after 10 seconds. 
Screenshot_20240311-151101

ddshd:
My opinion/input: maybe instead of a toggle change it to a button (something like "Overnight Upload") which routes to a mostly blank black screen with just a progress bar and keeping the screen from going to sleep while the upload is running.

Do include something for screen burn though (moving the element at 1 Hz)

alextran1502:
I am fine with adding this mechanism first in this PR and additional sleep mode in the future PRs

alextran1502:
@shenlong-tanwen Tested, it works well and looks good. Please feel free to merge to the other PR

alextran1502 (approved):

shenlong-tanwen:
Superseded by #8385

idominiki:
Is it possible to make this setting also working while manually uploading from share sheet

PR #7812, opened 2024-03-10
feat(web): dark color-scheme css for dark mode

ZTzTopia:
Ensure form controls and scroll bars follow the theme when toggled dark or light


web/src/app.css
```diff
@@ -136,4 +137,12 @@ input:focus-visible {
   .scrollbar-stable {
     scrollbar-gutter: stable both-edges;
   }
-}
+
+  [data-color-scheme="dark"] {
```
michelheusschen, line 141: These aren't really utilities, so I would move them out of this layer. The `dark` class gets already applied and we might as well use that instead of adding a data property:

```css
:root.dark {
  color-scheme: dark;
}

.root:not(.dark) {
  color-scheme: light;
}
```

michelheusschen: Sorry made a typo here

```diff
- .root:not(.dark) {
+ :root:not(.dark) {
```

jrasm91:
Can you fix the formatting? `npm run format:fix`

web/src/app.css
ZTzTopia: i don't see it either but it looks like it just needs `:root.dark`?

jrasm91 (approved):
Can you add screenshots of before and after to the pull request description please?

alextran1502:
| Before | After |
| -----  | ----- |
| image |  image |

PR #7819, opened 2024-03-10
feat(server, web): album orders

martabal:
## What's changed

This PR adds the ability to see album assets in ascending/descending order

## Screenshots

https://github.com/immich-app/immich/assets/74269598/ae144f2d-7293-4ff4-bcc0-9aebe15faad6


aviv926:
Fixing:
https://github.com/immich-app/immich/discussions/7031
https://github.com/immich-app/immich/discussions/6781
https://github.com/immich-app/immich/discussions/6624
https://github.com/immich-app/immich/discussions/1689


Maybe:
https://github.com/immich-app/immich/discussions/7489

server/src/infra/repositories/asset.repository.ts
```diff
@@ -594,12 +594,17 @@ export class AssetRepository implements IAssetRepository {
   @GenerateSql({ params: [{ size: TimeBucketSize.MONTH }] })
   getTimeBuckets(options: TimeBucketOptions): Promise<TimeBucketItem[]> {
     const truncated = dateTrunc(options);
-    return this.getBuilder(options)
+    const timeBuckets = this.getBuilder(options)
       .select(`COUNT(asset.id)::int`, 'count')
       .addSelect(truncated, 'timeBucket')
-      .groupBy(truncated)
-      .orderBy(truncated, 'DESC')
```
jrasm91, line 601: Why not just use a ternary here instead?

web/src/lib/components/album-page/album-options.svelte
```diff
@@ -1,22 +1,51 @@
...
+    if (selectedOption == returnedOption) {
+      return;
+    }
+    const ascendingOrder = options[0] == returnedOption;
+
+    try {
+      await updateAlbumInfo({
+        id: album.id,
+        updateAlbumDto: {
+          ascendingOrder: ascendingOrder,
```
jrasm91, line 41: Just use the simpler form of just the key

web/src/lib/components/photos-page/asset-grid.svelte
```diff
@@ -54,7 +55,8 @@
     showSkeleton = false;
     document.addEventListener('keydown', onKeydown);
     assetStore.connect();
-    await assetStore.init(viewport);
+    console.log(ascendingOrder);
```
jrasm91, line 58: Remove this

web/src/lib/stores/assets.store.ts
```diff
@@ -153,15 +153,15 @@ export class AssetStore {
     this.emit(true);
   }, 10_000);
 
-  async init(viewport: Viewport) {
+  async init(viewport: Viewport, ascendingOrder?: boolean) {
```
jrasm91, line 156: This is the wrong place to pass the option. You should pass it when the class is created like everything else.

web/src/lib/components/album-page/album-options.svelte
```diff
@@ -1,22 +1,51 @@
...
+  $: selectedOption = album.ascendingOrder ? options[0] : options[1];
+
   const dispatch = createEventDispatcher<{
     close: void;
     toggleEnableActivity: void;
     showSelectSharedUser: void;
   }>();
+
+  const handleToggle = async (returnedOption: RenderedOption) => {
+    if (selectedOption == returnedOption) {
```
jrasm91, line 32: You should use ===

jrasm91 (commented):
We have other existing query params, etc. We should use consistent naming here as well, not ascendingOrder. I think an enum would make more sense probably.

martabal:
> I think an enum would make more sense probably

Something like this ?

```ts
enum Order {
  ASCENDING = 'ascending',
  DESCENDING = 'descending',
}
```

web/src/lib/stores/assets.store.ts
```diff
@@ -153,15 +153,19 @@ export class AssetStore {
...
     this.buckets = [];
     this.assets = [];
     this.assetToBucket = {};
     this.albumAssets = new Set();
 
-    const buckets = await getTimeBuckets({ ...this.options, key: getKey(), ascendingOrder });
+    const buckets = await getTimeBuckets({
+      ...this.options,
+      key: getKey(),
+      ascendingOrder: this.options.ascendingOrder,
```
jrasm91, line 167: this is already included in line 165. What are you adding it again here?

jrasm91:
I was referring to this, which we already have

image

web/src/lib/stores/assets.store.ts
martabal: oops I forgot to remove it

server/src/domain/album/album-response.dto.ts
```diff
@@ -23,6 +23,8 @@ export class AlbumResponseDto {
   startDate?: Date;
   endDate?: Date;
   isActivityEnabled!: boolean;
+  @ApiProperty({ enumName: 'AssetOrder', enum: AssetOrder })
+  order!: AssetOrder;
```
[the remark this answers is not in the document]
alextran1502: I am pretty sure it will break the mobile app because the SDK will check for this property and it will be null from previous version, so the users will not see any albums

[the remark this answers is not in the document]
jrasm91: Let's make it optional then to avoid the breaking change. We can change it back in the future after a few releases.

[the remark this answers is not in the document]
shenlong-tanwen: > @shenlong-tanwen will this new property break the mobile app? Do we need to make it `order?: AssetOrder`?

Yes. The mobile app will not sync albums with older server versions. Making it optional would be best.

server/src/infra/entities/album.entity.ts
```diff
@@ -14,6 +14,11 @@ import { AssetEntity } from './asset.entity';
 import { SharedLinkEntity } from './shared-link.entity';
 import { UserEntity } from './user.entity';
 
+export enum AssetOrder {
+  ASC = 'asc',
+  DESC = 'desc',
+}
+
```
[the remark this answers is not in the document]
martabal: Yes exactly

package-lock.json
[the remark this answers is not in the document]
martabal: Oops

mobile/pubspec.lock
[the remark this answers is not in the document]
martabal: Mmmh, I think the dart extension changed it automatically

server/src/infra/entities/album.entity.ts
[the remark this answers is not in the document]
jrasm91: Can you add a comment to it then indicating that is why it is there?

server/src/infra/repositories/asset.repository.ts
```diff
@@ -598,7 +598,7 @@ export class AssetRepository implements IAssetRepository {
       .select(`COUNT(asset.id)::int`, 'count')
       .addSelect(truncated, 'timeBucket')
       .groupBy(truncated)
-      .orderBy(truncated, 'DESC')
+      .orderBy(truncated, options.order === AssetOrder.ASC ? 'ASC' : 'DESC')
```
danieldietzler, line 601: Can't we just do `options.order.toUpperCase()` instead?

server/src/infra/repositories/asset.repository.ts
```diff
@@ -611,7 +611,7 @@ export class AssetRepository implements IAssetRepository {
         // First sort by the day in localtime (put it in the right bucket)
         .orderBy(truncated, 'DESC')
         // and then sort by the actual time
-        .addOrderBy('asset.fileCreatedAt', 'DESC')
+        .addOrderBy('asset.fileCreatedAt', options.order === AssetOrder.ASC ? 'ASC' : 'DESC')
```
danieldietzler, line 614: Same here

web/src/lib/components/album-page/album-options.svelte
```diff
@@ -1,22 +1,58 @@
...
+    if (selectedOption === returnedOption) {
+      return;
+    }
+    let order = AssetOrder.Desc;
+    for (const [key, option] of Object.entries(options)) {
+      if (option === returnedOption) {
+        order = key as AssetOrder;
+        break;
+      }
+    }
```
danieldietzler, line 42: You can use `findKey(options, (option)  => option === returnedOption)` from lodash-es.

web/src/lib/components/album-page/album-options.svelte
```diff
@@ -34,8 +70,16 @@
...
-            <h2 class="text-gray text-sm mb-3">SHARING</h2>
-            <div class="p-2">
+            <h2 class="text-gray text-sm mb-2">SETTINGS</h2>
+            <div class="grid p-2 gap-y-2">
+              {#if album.order}
+                <SettingDropdown
+                  title="Direction"
+                  options={Object.values(options)}
+                  selectedOption={options[album.order]}
+                  onToggle={(option) => handleToggle(option)}
```
danieldietzler, line 80: If you just pass the option through you could do `onToggle={handleToggle}` here

server/src/infra/repositories/asset.repository.ts
martabal: I don't know about that. I don't like the idea of mixing SQL syntax and enum values

danieldietzler: Hm ok that's fair. Coupling those would result in the query breaking in case we were to ever rename the enum value. That's actually bad, true!

martabal: As soon as we add new values to this enum, we'll have to update that part too and I don't know if typescript / typeorm can alert it. It will definitely be a disaster :laughing:

alextran1502:
Hello, thank you for the PR!

I just ran some tests, and the order of the assets in the album doesn't seem to change. Can you double-check?

martabal:
> Hello, thank you for the PR!
> 
> I just ran some tests, and the order of the assets in the album doesn't seem to change. Can you double-check?

Can you try with `make open-api && make dev-update` ?

alextran1502:
@martabal It works now, sorry about my glitchy system

alextran1502:
@martabal actually, it doesn't work again, so probably there is a bug somewhere

martabal:
Ok, it should be fixed

alextran1502 (approved):

waclaw66:
It would be nice to have Immich-wide configuration for default album sorting. I personally prefer "oldest first" order for most of story-like albums, however timelapse albums looks better with "newest first" order.

martabal:
> It would be nice to have Immich-wide or user-wide configuration for default album sorting. I personally prefer "oldest first" order for most of story-like albums, however timelapse albums looks better with "newest first" order.

User-wide makes sense. You probably can open a FR for that

waclaw66:
> User-wide makes sense. You probably can open a FR for that

Created... https://github.com/immich-app/immich/discussions/8068

mondeg0:
This feature is awesome, and I was waiting for it for a long time! 
However, the options to sort seems erased when the album is shared via a link. In this scenario, the album presentation is still recent to oldest, even if set to the inverse in the albums settings. 

Is this a bug or desired behavior ?

martabal:
This is definitely a missing feature. We prefer to add features piece by piece, it's easier to correct bugs and get user feedback.

PR #7824, opened 2024-03-10
feat(web): add sticky date headers for asset-date-group

samholton:
Causes the date header element to be sticky as you scroll

addresses web part of https://github.com/immich-app/immich/discussions/1667

[sticky dates.webm](https://github.com/immich-app/immich/assets/3507151/776432cd-6819-48f5-b097-2a6592974efe)


jrasm91 (approved):
LGTM

Issue #7680, opened 2024-03-06
Duplicate file shows uploaded message


aviv926:
### The bug

When uploading a duplicate file, instead of showing only a duplicate message, a message that is uploaded + a duplicate message is displayed

![image](https://github.com/immich-app/immich/assets/51673860/cd43e5fe-3758-4063-b976-e842e023443c)

![image](https://github.com/immich-app/immich/assets/51673860/0ba4b039-2b79-4026-b123-c53cd6ed2a1b)

### The OS that Immich Server is running on

Main Demo

### Version of Immich Server

Main Demo

### Version of Immich Mobile App

Main Demo

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
Main Demo
```

### Your .env content

```Shell
Main Demo
```

### Reproduction steps

```bash
1. Upload a file
2. Upload the same file
```

aviv926:
fixed via #7858

PR #7825, opened 2024-03-10
fix: Separate duplicate asset notification from upload success

declan8010:
This commit separates the notification messages for duplicate assets and upload success in the upload panel component. Previously, if a single duplicated image was uploaded, it erroneously displayed both the "Upload success" and "Duplicate" messages, leading to the invalidation of the successful message. Now, the "Upload success" message is displayed only when there are successful uploads other than duplicates. Subsequently, if there are any skipped duplicate assets, their count will be displayed in a separate warning message. This organization ensures that the notifications are displayed as desired.

In each Gif below I am initially uploading 3 images and then proceeding to upload the same 3 again.

Before:
upload-before

After
upload-after

Fixes https://github.com/immich-app/immich/issues/7680


declan8010:
Fixes #7680

aviv926:
> Fixes #7680

Thanks for the PR, note the problem with the small slider still showing 3 uploaded.

image

web/src/lib/components/shared-components/upload-panel.svelte
```diff
@@ -34,13 +34,12 @@
...
     out:fade={{ duration: 250 }}
     on:outroend={() => {
-      notificationController.show({
-        message:
-          ($errorCounter > 0
-            ? `Upload completed with ${$errorCounter} error${$errorCounter > 1 ? 's' : ''}`
-            : 'Upload success') + ', refresh the page to see new upload assets.',
-        type: $errorCounter > 0 ? NotificationType.Warning : NotificationType.Info,
-      });
+      if ($successCounter > 0 && $successCounter !== $duplicateCounter) {
```
michelheusschen, line 37: You can push more commits to a PR, you don't have to create a new one. The root cause appears to be incrementing `$successCounter` for duplicate assets, we should fix that instead of creating a workaround. Also the error count is no longer reported, is that an intentional change?

declan8010:
> > Fixes #7680
> 
> Thanks for the PR, note the problem with the little slider still showing 3 uploaded but not 3 duplicates
> 
> image

Just sorted that in my latest commit.

@michelheusschen Could you drop me a screenshot on that one, Not 100% sure which error section you are reffering to, If its the one from the panel bar at the bottom the current production build does not report duplicates as errors.

Gif for current version:
firefox_INZPPDSS7R

Thanks both 👍

michelheusschen:
I'm referring to the `$errorCounter` you've removed from the notification: https://github.com/immich-app/immich/pull/7825/files#diff-b35c09fa6eb31538c2438c4b21451b27eeca9a16125b2328a4e6a7e5909d6d16L40

Also I still think it's better to only increment `$successCounter` for non-duplicates so we don't need this weird logic. Currently the success counter is always incremented: https://github.com/immich-app/immich/blob/5bd597f14bf920d83e6193fae47ab841d49c2680/web/src/lib/utils/file-uploader.ts#L113

declan8010:
@michelheusschen Any suggestions on sorting out this counter logic, I have used the increment method for the successcounter but am now seeing this:

image

michelheusschen:
Remove this line
https://github.com/immich-app/immich/blob/5bd597f14bf920d83e6193fae47ab841d49c2680/web/src/lib/utils/file-uploader.ts#L113

Add that to the if statement like:
```ts
if (res.duplicate) {
  uploadAssetsStore.duplicateCounter.update((count) => count + 1);
} else {
  uploadAssetsStore.successCounter.update((c) => c + 1);
}
```
https://github.com/immich-app/immich/blob/5bd597f14bf920d83e6193fae47ab841d49c2680/web/src/lib/utils/file-uploader.ts#L100-L102

Revert your changes to this line
https://github.com/immich-app/immich/blob/4eca2b0f34ab5897ab060da5b831b85c651d4bd1/web/src/lib/components/shared-components/upload-panel.svelte#L67

declan8010:
@michelheusschen Thank you for the help I really appreciate it, Ignore the two failed files above Github was throwing some errors while I was submitting my branch. The two which have passed are the ones we want

Issue #7724, opened 2024-03-08
Set profile picture from video


aviv926:
### The bug

I think by default this shouldn't be possible, but currently when a user tries to select a profile picture from a video it shows it as broken.

![image](https://github.com/immich-app/immich/assets/51673860/e1b7a0c0-6bf2-48ce-8794-5cea96ed5b54)

**Solution:**
Using a thumb image instead of the video itself.

### The OS that Immich Server is running on

Main Demo

### Version of Immich Server

Main Demo

### Version of Immich Mobile App

Main Demo

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
Main Demo
```

### Your .env content

```Shell
Main Demo
```

### Reproduction steps

```bash
1. Enter the video
2. Click on the three dots on the top right
3. Select `As profile picture`
```

PR #7830, opened 2024-03-10
Fix: Disable 'As profile picture' option for videos in context menu a…

declan8010:
This commit modifies the context menu behavior to disable the "As profile picture" option when interacting with video assets. Previously, the option was available for all asset types, including videos, which could lead to confusion when this displayed an error.

With this change, the "As profile picture" option is conditionally rendered based on the asset type. If the asset is a video, the option is not displayed in the context menu.

This adjustment enhances the web experience by preventing users from attempting to set a video as their profile picture, which is not supported by the system.

Suggested Changes Gif:

firefox_UTeLNKhfeK(1)

Fixes: #7724

### DIFF
--- a/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
+++ b/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
@@ -186,7 +186,9 @@
                 on:click={() => dispatch('toggleArchive')}
                 text={asset.isArchived ? 'Unarchive' : 'Archive'}
               />
-              <MenuOption on:click={() => onMenuClick('asProfileImage')} text="As profile picture" />
+              {#if asset.type !== AssetTypeEnum.Video}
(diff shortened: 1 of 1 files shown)


web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
```diff
@@ -186,7 +186,9 @@
                 on:click={() => dispatch('toggleArchive')}
                 text={asset.isArchived ? 'Unarchive' : 'Archive'}
               />
-              <MenuOption on:click={() => onMenuClick('asProfileImage')} text="As profile picture" />
+              {#if asset.type !== AssetTypeEnum.Video}
```
mertalev, line 189: Wouldn't it make more sense to check if it's an image?

declan8010: Great shout @mertalev I was too focused on the video aspect 👍

mertalev (approved):
Thanks for fixing this!

PR #7840, opened 2024-03-11
chore(cli): clarify use of concurrency option

mertalev:
## Description

The concurrency option takes a number, but this isn't reflected in the help message.


jrasm91 (approved):
Do we want to add any e2e tests for this? At least that it correctly accepts numbers and disallows other options?

mertalev:
Sure, I can add a test for that

PR #7841, opened 2024-03-11
feat(server): Add OAuth auto-redirect enable/disable CLI commands

ddshd:
Cannot use password login if OAuth auto-redirect is enabled and there is an issue with OAuth


ddshd:
Lmk if a main oauth enable/disable would be bettter

jrasm91 (commented):
I'm ok with adding this, although you can usually hit the back button and it will take you back to the login page. Also it should be OAuth everywhere instead of Oauth

jrasm91:
And I think just disabling it entirely might be better.

ddshd:
> And I think just disabling it entirely might be better.

Done

> I'm ok with adding this, although you can usually hit the back button and it will take you back to the login page. Also it should be OAuth everywhere instead of Oauth

Weird doesn't work for me, it just redirects back - maybe it's my oauth

jrasm91:
There is a url param autoLaunch that you can set to 0 I think to disable the auto launching.

jrasm91 (approved):
I think everything looks good, but could you rename the file to just oauth login?

PR #7857, opened 2024-03-11
fix(server): serve static directory only if it exists

benmccann:
alternative to https://github.com/immich-app/immich/pull/7796

### DIFF
--- a/server/src/immich/main.ts
+++ b/server/src/immich/main.ts
@@ -1,4 +1,4 @@
-import { envName, isDev, serverVersion } from '@app/domain';
+import { WEB_ROOT, envName, isDev, serverVersion } from '@app/domain';
 import { WebSocketAdapter } from '@app/infra';
 import { ImmichLogger } from '@app/infra/logger';
 import { NestFactory } from '@nestjs/core';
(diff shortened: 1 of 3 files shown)


server/src/immich/main.ts
```diff
@@ -29,20 +30,23 @@ export async function bootstrap() {
...
-      gzip: true,
-      brotli: true,
-      setHeaders: (res, pathname) => {
-        if (pathname.startsWith(`/_app/immutable`) && res.statusCode === 200) {
-          res.setHeader('cache-control', 'public,max-age=31536000,immutable');
-        }
-      },
-    }),
-  );
+  const staticDir = 'www';
```
jrasm91, line 33: this is a relative path, can you use an absolute path instead? Maybe you can just rework this (remove the index.html part from the env)

```ts
export const WEB_ROOT_PATH = join(process.env.IMMICH_WEB_ROOT || '/usr/src/app/www', 'index.html');
```

benmccann: done

jrasm91 (approved):

PR #7859, opened 2024-03-11
fix(web): small fixes for empty placeholder

michelheusschen:
fix(web): small fixes for empty placeholder

- Use button when placeholder is clickable, otherwise use div
- Remove alt text, because image is only decorative and text is already provided
- Some general cleanup and rename `actionHandler` to `onClick`

There are no visual changes


danieldietzler (approved):
Nice!

jrasm91 (approved):

PR #7863, opened 2024-03-11
feat(web): Implemented device last seen date and time with user locale support

declan8010:
This pull request introduces enhancements to the last seen functionality in our application. Previously, the last seen information displayed only the time elapsed since the device was last seen. However, with this update, I have implemented support for displaying the date and time of the last seen event, enhancing the user experience.

Before:
image

After:
image

Totally open to suggestions for this in terms of how its displayed

### DIFF
--- a/web/src/lib/components/user-settings-page/device-card.svelte
+++ b/web/src/lib/components/user-settings-page/device-card.svelte
@@ -12,7 +12,7 @@
     mdiMicrosoftWindows,
     mdiTrashCanOutline,
   } from '@mdi/js';
-  import { DateTime, type ToRelativeCalendarOptions } from 'luxon';
+  import { DateTime } from 'luxon';
   import { createEventDispatcher } from 'svelte';
 
   export let device: AuthDeviceResponseDto;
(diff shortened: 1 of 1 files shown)


declan8010:
No idea why prettier is failing, Its all showing good in my VSCode could someone take a look please?

alextran1502:
> No idea why prettier is failing, Its all showing good in my VSCode could someone take a look please?

Try run `npm i` again in the `web` directory

alextran1502:
I think if it is last seen over 7 days then we display the time, otherwise, keep the day count

jrasm91:
I like the relative date format myself. 90 seems more useful than some random date. I don't really care if it was October 11th or November 1st. I would be fine adding a way to get the actual date though. Something like text-underline on 90 days ago and when it is clicked or hovered shows the actual date.

declan8010:
Hey both what do you think of this change?

firefox_Tx91Zmj1Sh

alextran1502:
From UX point of view, I think it adds unnecessary interaction where you can just present the information on screen

declan8010:
Sounds good, Do you think just displaying it under the current values in a sub text style?

alextran1502:
Yeah, you can try that or also display it on the same line after a dash

declan8010:
What do we think of this layout? 

image

alextran1502:
> What do we think of this layout?
> 
> image

I like it!, thank you

declan8010:
All pushed and good to go 👍

alextran1502 (approved):

jrasm91:
Nice work, looks great.

PR #7865, opened 2024-03-11
feat(mobile): add labels to app bar buttons

ben-basten:
## Description

Adds labels to some of the buttons in the top app bar for screen reader users, to provide context about where the buttons go.

**Note:** I did not modify the rightmost profile circle button, because I think new internationalization text will need to be added for this. Something like "Profile and settings" maybe, and modify the header of the resulting popup to match for predictability?

## How Has This Been Tested?

- [x] VoiceOver for iOS
- [ ] TalkBack for Android - untested, I don't have a physical Android to test with yet...

No visible changes

## Checklist:

### DIFF
--- a/mobile/lib/shared/ui/immich_app_bar.dart
+++ b/mobile/lib/shared/ui/immich_app_bar.dart
@@ -1,4 +1,5 @@
 import 'package:auto_route/auto_route.dart';
+import 'package:easy_localization/easy_localization.dart';
 import 'package:flutter/material.dart';
 import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:immich_mobile/extensions/build_context_extensions.dart';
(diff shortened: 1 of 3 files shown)


mobile/lib/shared/ui/immich_app_bar.dart
```diff
@@ -80,6 +81,7 @@ class ImmichAppBar extends ConsumerWidget implements PreferredSizeWidget {
               strokeWidth: 2,
               strokeCap: StrokeCap.round,
               valueColor: AlwaysStoppedAnimation<Color>(iconColor),
+              semanticsLabel: 'backup_controller_page_backup'.tr(),
```
ben-basten, line 84: In the future, it would be cool if these labels included the status of the backup, since that's a cue for sighted users that screen readers miss. Something like:

* Backup (complete)
* Backup (in progress)
* Backup (off)

alextran1502 (approved):

PR #7888, opened 2024-03-12
fix(web): asset viewer navigation buttons

michelheusschen:
- Show cursor as pointer when hovering over clickable navigation area
- Improve accessibility
- Remove unused keydown event


jrasm91 (commented):
Why do we need to wrap the button in a div and have the click handler be in the div and not the button?

michelheusschen:
That allows focusing the button and not the whole clickable area (white border in image for illustration). The click event from the button bubbles up to the `div` where it's handled, that's also why we can safely ignore these a11y warnings.

asset-viewer-button-focus

jrasm91 (approved):

PR #7894, opened 2024-03-12
feat(server): YAML config file support

bo0tzz:


server/src/domain/system-config/system-config.core.ts
```diff
@@ -341,19 +342,19 @@ export class SystemConfigCore {
     if (force || !this.configCache) {
       try {
         const file = await this.repository.readFile(filepath);
-        const json = JSON.parse(file.toString());
+        const config = loadYaml(file.toString()) as any;
```
bo0tzz, line 345: js-yaml has this option:

> `json` (default: false) - compatibility with JSON.parse behaviour. If true, then duplicate keys in a mapping will override values rather than throwing an error.

I don't think we need it, right?

server/src/domain/system-config/system-config.core.ts
```diff
@@ -341,19 +342,19 @@ export class SystemConfigCore {
...
+          this.unsetDeep(config, key);
           if (value !== undefined) {
             overrides.push({ key, value });
           }
         }
 
-        if (!_.isEmpty(json)) {
-          this.logger.warn(`Unknown keys found: ${JSON.stringify(json, null, 2)}`);
+        if (!_.isEmpty(config)) {
+          this.logger.warn(`Unknown keys found: ${JSON.stringify(config, null, 2)}`);
```
bo0tzz, line 357: I left this as just JSON output rather than trying to format based on what the input was, to keep the code simple.

jrasm91: I think that's fine, it is just a warning anyways and the user should be able to map it back themselves.

server/src/domain/system-config/system-config.core.ts
jrasm91: Oh I didn't realize you could unit test it so easily, but that works too.

jrasm91 (approved):
Noice.

alextran1502 (approved):

PR #7907, opened 2024-03-12
fix(web): correctly use button and link elements

michelheusschen:
- Use native `button` element instead of emulating them
- Add blur event to people thumbnail to remove bounding box when focus is lost
- Remove button element from icons that have no action
- Converted div to link in `SharedLinkCard`


jrasm91 (approved):
I love these types of changes. Ty

PR #7910, opened 2024-03-12
fix(web): Remove excess spacing on #asset-grid and search bar

natedawg:
The timeline / asset-grid has a `1rem` left-margin that results in the sidebar looking larger than it actually is. Removing the margin makes the sidebar have even spacing and as bonus adds a little more room for photos. With this change photos are no longer offset from the main page titles like `Archive` and `Favorites` but those are stickied to the top anyway.  

Update: Also removed the left padding on the search bar so it's even with the timline. Pics below do not reflect this change. 

Original:
orig

Left margin removed:
fixed

### DIFF
--- a/web/src/lib/components/photos-page/asset-grid.svelte
+++ b/web/src/lib/components/photos-page/asset-grid.svelte
@@ -387,7 +387,7 @@
 <!-- Right margin MUST be equal to the width of immich-scrubbable-scrollbar -->
 <section
   id="asset-grid"
-  class="scrollbar-hidden h-full overflow-y-auto pb-[60px] {isEmpty ? 'm-0' : 'ml-4 mr-[60px]'}"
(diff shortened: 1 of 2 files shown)


alextran1502:
Can you also move the search bar over :D

natedawg:
> Can you also move the search bar over :D?

Sure. It's moved now.

jrasm91 (approved):
Seems like a good change!

waclaw66:
I would prefer as it was before, now it's too tight on Photos page, too close to scrollbar...

obrazek
obrazek

arpanghosh8453:
I would prefer the early look also because otherwise, it looks too compact ( negative space is very important ), and also the layout shift while hovering over dates causes issues with navigation ( For elderly people ).

alextran1502:
@arpanghosh8453 it has been fixed and will be available in the next release

arpanghosh8453:
Hey there! It's Alex (Our CEO) again saving my day! Thank you so much! You are so fast to reply dude! It's been 2 minutes since I made this comment.

PR #7913, opened 2024-03-12
fix: Prevent clear-text logging of sensitive information admin password

declan8010:
### Description
This PR addresses the clear-text logging of sensitive information in the `ResetAdminPasswordCommand` class to mitigate the risk of exposing sensitive user data to potential attackers. 

```typescript
      } else {
        console.log(`The admin password has been updated to:\n${password}`);
``` 

### Changes Made
- Modified the code to prevent clear-text logging of the new admin password.
- Updated the log message to provide a generic notification that the admin password has been updated, without revealing the actual password.

### Motivation and Context
Previously, the admin password was logged directly in the console output, which could lead to the exposure of confidential information if the logs were accessed by unauthorized parties. By implementing this fix, we reduce the risk of exposing sensitive user data and system information, thereby enhancing the overall security posture of the application.

### DIFF
--- a/server/src/immich-admin/commands/reset-admin-password.command.ts
+++ b/server/src/immich-admin/commands/reset-admin-password.command.ts
@@ -24,19 +24,15 @@ export class ResetAdminPasswordCommand extends CommandRunner {
     return this.inquirer.ask<{ password: string }>('prompt-password', {}).then(({ password }) => password);
   };
 
-  async run(): Promise<void> {
-    try {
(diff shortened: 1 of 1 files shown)


bo0tzz:
The password is logged only if it was randomly generated, in which case it is necessary to do that.

declan8010:
@bo0tzz Thank you for the clarification lets close this one off 👍

jrasm91 (commented):
The password is only returned if you don't provide one and we generate a random one instead. This is working as designed. If you don't want the password logged just specify one in the reset command.

PR #7920, opened 2024-03-13
feat(web): preload assets in photo-viewer

samholton:
Adding `preloadAssets = writable<AssetResponseDto[]>([]);` to `asset-viewing.store` to allow one or more assets to be preloaded after the primary asset has been loaded.

Design choices:
1. Only preload assets if they are of type `AssetTypeEnum.Image`, didn't want to interfere with video streaming
2. Only preload assets if primary asset is image - didn't want to interfere with video streaming
3. For stacks, preload the next image as well as the second image in the stack
4. Does not preload when initially selecting an asset - although we could preload both the previous and next until we know which direction they are going

The actual loading is currently in `photo-viewer` because it has the information when the primary is finished loading as well as the helper `downloadRequest`. Although maybe it makes sense to add an event when primary loading is done and do pre-loading in `asset-grid` or `asset-viewer`?

Tested by clearing cache, not enabling disable cache, and setting throttling to a slower connection. Tested both navigating around as well as slideshow.

Does not address the issue in Firefox and the XHR abort not happening.

**before without preload**
[before.webm](https://github.com/immich-app/immich/assets/3507151/de5b80ed-3973-4e90-861c-c7924d518000)

**after with preload**
[after with preload.webm](https://github.com/immich-app/immich/assets/3507151/2b02a925-e9b9-4678-80fa-ddd5d019d857)


web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -103,6 +104,8 @@
       $stackAssetsStore = [...$stackAssetsStore, asset].sort(
         (a, b) => new Date(b.fileCreatedAt).getTime() - new Date(a.fileCreatedAt).getTime(),
       );
+      // if its a stack, add the next stack image in addition to the next asset
+      preloadAssets.push($stackAssetsStore[1]);
```
michelheusschen, line 108: It's technically possible to create a stack with only one asset, opening such a stack will show the spinner forever

web/src/lib/components/asset-viewer/photo-viewer.svelte
```diff
@@ -60,8 +65,20 @@
       });
 
       assetData = URL.createObjectURL(data);
+      imageLoaded = true;
+
+      if (preloadAssets) {
```
danieldietzler, line 70: Could you return if `!preloadAssets` and otherwise continue to avoid such a huge `if`?

web/src/lib/components/asset-viewer/asset-viewer.svelte
samholton: Good to know, I don't have any native stacked photos and was only able to create a stack with 2+ photos in the UI

web/src/lib/components/asset-viewer/photo-viewer.svelte
```diff
@@ -60,8 +65,20 @@
...
+        for (const preloadAsset of preloadAssets) {
+          if (preloadAsset.type === AssetTypeEnum.Image) {
+            await downloadRequest({
+              url: getAssetFileUrl(preloadAsset.id, !loadOriginal, false),
```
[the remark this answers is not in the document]
samholton: You are suggesting to always preload the original asset? As it is now, I load whichever preference the user has selected

alextran1502 (approved):

jvgoch:
Great feature! Unfortunately it doesn't work when the slideshow is set to shuffle. Should I create an issue on the github for this?

samholton:
Will see if I can take a look this weekend at preloading when in shuffle mode

PR #7928, opened 2024-03-13
fix(web): url state of nested accordions

danieldietzler:


web/src/lib/components/shared-components/settings/setting-accordion.svelte
```diff
@@ -19,6 +20,11 @@
       $accordionState = $accordionState;
     }
   };
+
+  onDestroy(() => {
+    $accordionState.delete(key);
+    $accordionState = $accordionState;
```
michelheusschen, line 26: Looks good, you could also do

```suggestion
    setIsOpen(false);
```

danieldietzler: True, thanks!

jrasm91 (approved):

PR #7932, opened 2024-03-13
chore(server): introduce proper job status

danieldietzler:
In some cases one can certainly argue whether it is `SKIPPED` or `SUCCESS`, so I'm open for suggestions :)

### DIFF
--- a/server/src/domain/job/job.service.ts
+++ b/server/src/domain/job/job.service.ts
@@ -11,6 +11,7 @@ import {
   ISystemConfigRepository,
   JobHandler,
   JobItem,
+  JobStatus,
   QueueCleanType,
 } from '../repositories';
 import { FeatureFlag, SystemConfigCore } from '../system-config/system-config.core';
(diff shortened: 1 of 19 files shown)


server/src/domain/job/job.service.ts
```diff
@@ -155,8 +156,8 @@ export class JobService {
 
         try {
           const handler = jobHandlers[name];
-          const success = await handler(data);
-          if (success) {
+          const status = await handler(data);
+          if (status === JobStatus.SUCCESS || status == JobStatus.SKIPPED) {
```
jrasm91, line 160: Did you change any `false` => `SKIPPED`?

danieldietzler: I most certainly did, ugh.

jrasm91: Did you change any `true` => `SKIPPED`? Otherwise just remove skipped so we have the same behavior for now.

danieldietzler: I made sure that there aren't any `false`  -> `SKIPPED` among the jobs that cause something in onDone. So I'd argue that's fine as is now

jrasm91 (approved):

PR #7934, opened 2024-03-13
fix(server): don't crash when refreshing large libraries

etnoy:
There have recently been improvements that help refreshing large external libraries. However, it is still capped at some limit, depending on available memory and the size of the library. Specifically, handleQueueAssetRefresh has a high memory complexity that causes OOM for large libraries when queuing the individual asset refreshes.

This PR fixes this specific OOM crash. We previously kept a list of *all* assets in a library being refreshed because we would track which assets have been removed. This is the list that kept eating all memory. We no longer build a huge list from the filesystem crawl. Instead, we request them 100k at a time and then discard when the corresponding refresh jobs have been queued.

This, however, means we can't immediately infer when an asset is offline since that operation used to depend on us loading all online assets into memory. Instead, we do a second pass to check all currently-online assets. 

Quick testing indicated I was able to refresh millions of assets without a sweat on a VM with only 10G of ram. There is an upper limit to the things we can scan, but that seems to be above 5M assets which ought to be enough for everybody (TM)

image

We add the same pagination to other library jobs as well.

How this has been tested
* Queue a large library containing 5M+ assets. Queueing took 20 minutes, but I did not test how long it would take to complete the full import
* Remove assets from an import path and run scan library files. The correct files are marked as offline, and can be removed with the corresponding job
* More e2e tests

NOTE: Trying to clear a bullmq queue with millions of items in it locks up redis and it appears the lua script used for drain() doesn't finish ever


server/src/domain/library/library.service.ts
```diff
@@ -305,9 +304,8 @@ export class LibraryService extends EventEmitter {
     this.logger.verbose(`Queuing refresh of ${assetPaths.length} asset(s)`);
 
     // We perform this in batches to save on memory when performing large refreshes (greater than 1M assets)
-    const batchSize = 5000;
-    for (let i = 0; i < assetPaths.length; i += batchSize) {
-      const batch = assetPaths.slice(i, i + batchSize);
+    for (let i = 0; i < assetPaths.length; i += LIBRARY_SCAN_BATCH_SIZE) {
+      const batch = assetPaths.slice(i, i + LIBRARY_SCAN_BATCH_SIZE);
```
mertalev, line 308: The call to `scanAssets` is already batched now so this is redundant.

server/e2e/jobs/specs/library.e2e-spec.ts
```diff
@@ -50,20 +50,96 @@ describe(`${LibraryController.name} (e2e)`, () => {
...
           expect.objectContaining({
-            isOffline: true,
+            isOffline: false,
+            originalFileName: 'tanners_ridge',
+          }),
+        ]),
+      );
+    });
+
+    it('should not offline a missing file when performing regular scan', async () => {
```
mertalev, line 73: I'm not sure it makes sense to have a test for this. Is it undesirable for it to offline a missing file?

server/src/domain/library/library.service.ts
etnoy: Fixed

server/e2e/jobs/specs/library.e2e-spec.ts
etnoy: Thanks for spotting

server/src/domain/library/library.service.ts
```diff
@@ -568,14 +566,21 @@ export class LibraryService extends EventEmitter {
...
-      data: {
-        id,
-        refreshModifiedFiles: dto.refreshModifiedFiles ?? false,
-        refreshAllFiles: dto.refreshAllFiles ?? false,
-      },
-    });
+    if (dto.checkForOffline) {
+      if (dto.refreshAllFiles || dto.refreshModifiedFiles) {
+        throw new BadRequestException('Cannot use checkForOffline with refreshAllFiles or refreshModifiedFiles');
+      }
```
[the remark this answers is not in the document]
etnoy: thanks, yes, we can now do both

jrasm91 (commented):
I think we should not be adding new endpoints for each type of scan. Can you describe all the different scan situations and we can come up with a unified API for them?

etnoy:
> I think we should not be adding new endpoints for each type of scan. Can you describe all the different scan situations and we can come up with a unified API for them?

* Scan. Performs a crawl of the fs and imports any new files not in the db
* Scan modified files.  As above, but also re-imports any files with modified timestamp
* Force scan all. As above, but re-imports all files

And then the new scan:
* Scan deleted. No fs crawl. Instead, it checks each file in the db to see if it still exists on disk. If not, it marks is as offline

mertalev:
What I'm wondering is if we actually need to expose this as a new option - having so many variations of a scan is confusing. This can maintain the current behavior and just be a backend change in how the jobs are executed.

etnoy:
> What I'm wondering is if we actually need to expose this as a new option - having so many variations of a scan is confusing. This can maintain the current behavior and just be a backend change in how the jobs are executed.

as in a scan first looking for new files and then scanning for deleted files? Could work

mertalev:
Yup, that's what I had in mind

etnoy:
> I think we should not be adding new endpoints for each type of scan. Can you describe all the different scan situations and we can come up with a unified API for them?

Done. 

The main scan interweaves 5000 imports with 5000 checks for existing assets being offline. This should make it feel more responsive when using large libraries

etnoy:
> Yup, that's what I had in mind

Done

etnoy:
I reduced the batch size to 1000 in order to use less memory. I also fixed an edge case bug where a changed import path caused assets outside of it not to be marked offline

mertalev:
Would it be better to still have the offline and online stuff as separate jobs, but just queue both of them? That approach was a lot faster and used less memory when I tested it.

etnoy:
> Would it be better to still have the offline and online stuff as separate jobs, but just queue both of them? That approach was a lot faster and used less memory when I tested it.

No, because for a large refresh it can potentially take a long time before either offline or online starts running. By interweaving the queue you make it feel snappier to the user. In my opinion.

server/e2e/jobs/specs/library.e2e-spec.ts
danieldietzler, line 1: Could you move those to the new e2e tests https://github.com/immich-app/immich/blob/main/e2e/src/api/specs/library.e2e-spec.ts?

server/src/domain/library/library.service.ts
```diff
@@ -572,6 +571,36 @@ export class LibraryService extends EventEmitter {
     return JobStatus.SUCCESS;
   }
 
+  // Check if an asset is has no file or is outside of import paths, marking it as offline
```
danieldietzler, line 574:
```suggestion
  // Check if an asset has no file or is outside of import paths, marking it as offline
```

server/src/domain/library/library.service.ts
```diff
@@ -572,6 +571,36 @@ export class LibraryService extends EventEmitter {
...
+    const exists = await this.storageRepository.checkFileExists(asset.originalPath, R_OK);
+
+    let existsInImportPath = false;
+
+    for (const importPath of job.importPaths) {
+      if (asset.originalPath.startsWith(importPath)) {
+        existsInImportPath = true;
+        break;
+      }
+    }
```
danieldietzler, line 592:
```suggestion
    const path = job.importPaths.find((path) => asset.originalPath.startsWith(path));
```
And then down below check if `path && exists`

server/src/domain/library/library.service.ts
```diff
@@ -590,106 +619,96 @@ export class LibraryService extends EventEmitter {
...
 
-    const crawledAssetPaths = await this.getPathTrie(library);
-    this.logger.debug(`Found ${crawledAssetPaths.size} asset(s) when crawling import paths ${library.importPaths}`);
-
-    const assetIdsToMarkOffline = [];
-    const assetIdsToMarkOnline = [];
-    const pagination = usePagination(LIBRARY_SCAN_BATCH_SIZE, (pagination) =>
-      this.assetRepository.getLibraryAssetPaths(pagination, library.id),
+    const pathValidation = await Promise.all(
+      library.importPaths.map(async (importPath) => await this.validateImportPath(importPath)),
```
danieldietzler, line 629:
```suggestion
    const pathValidation = await Promise.all(
      library.importPaths.map((importPath) => this.validateImportPath(importPath)),
```

server/src/domain/library/library.service.ts
```diff
@@ -590,106 +619,96 @@ export class LibraryService extends EventEmitter {
...
-        if (isOffline && !asset.isOffline) {
-          assetIdsToMarkOffline.push(asset.id);
-          this.logger.verbose(`Added to mark-offline list: ${asset.originalPath}`);
+    const validImportPaths = pathValidation
+      .map((validation) => {
+        if (!validation.isValid) {
+          this.logger.error(`Skipping invalid import path: ${validation.importPath}. Reason: ${validation.message}`);
         }
+        return validation;
+      })
```
danieldietzler, line 638: You probably don't want a `map` here but only a `forEach`. In general this I'm not sure if it's worth it to loop over that array only for the sake of good error messages. Idk.

server/src/domain/library/library.service.ts
```diff
@@ -590,106 +619,96 @@ export class LibraryService extends EventEmitter {
...
+      .filter((validation) => validation.isValid)
+      .map((validation) => validation.importPath)
+      .map((importPath) => path.normalize(importPath));
```
danieldietzler, line 641:
```suggestion
      .map((validation) => path.normalize(validation.importPath));
```

server/src/domain/library/library.service.ts
```diff
@@ -590,106 +619,96 @@ export class LibraryService extends EventEmitter {
   async handleQueueAssetRefresh(job: ILibraryRefreshJob): Promise<JobStatus> {
```
danieldietzler, line 619: This function is currently hard to read. Maybe you could split it up into smaller functions to make it easier to read

server/src/domain/library/library.service.ts
```diff
@@ -590,106 +619,96 @@ export class LibraryService extends EventEmitter {
...
+        );
+        await this.jobRepository.queueAll(
+          existingAssetPage.value.map((asset: AssetEntity) => ({
+            name: JobName.LIBRARY_CHECK_OFFLINE,
+            data: { id: asset.id, importPaths: validImportPaths },
+          })),
+        );
       }
-    }
+    };
```
danieldietzler, line 673: If you already put it in its own function, why keep it in there and not move it out into its own private function?

server/src/domain/library/library.service.ts
```diff
@@ -590,106 +619,96 @@ export class LibraryService extends EventEmitter {
...
 
-    this.logger.verbose(`Crawled assets have been checked for online/offline status`);
+    let crawledAssetPaths: string[] = [];
 
-    if (assetIdsToMarkOffline.length > 0) {
-      this.logger.debug(`Found ${assetIdsToMarkOffline.length} offline asset(s) previously marked as online`);
-      await this.assetRepository.updateAll(assetIdsToMarkOffline, { isOffline: true });
-    }
+    while (!crawlDone) {
```
danieldietzler, line 677: TBH I can't really follow what comes now. There's lots of branching and nesting which makes it really hard to read imo. Could you maybe try to work with early returns and private functions to try to simplify the logic here?

server/src/interfaces/asset.repository.ts
```diff
@@ -48,6 +48,7 @@ export enum WithoutProperty {
 
 export enum WithProperty {
   SIDECAR = 'sidecar',
+  IS_ONLINE = 'isOnline',
```
danieldietzler, line 51: Isn't an asset always online when `!IS_OFFLINE`?

server/e2e/jobs/specs/library.e2e-spec.ts
danieldietzler: This is not a dependency to get it merged though. There are still some tests to be migrated anyways, so moving those as well isn't too much trouble

server/src/services/library.service.ts
```diff
@@ -598,106 +622,96 @@ export class LibraryService extends EventEmitter {
...
+    const validImportPaths = pathValidation
+      .map((validation) => {
+        if (!validation.isValid) {
+          this.logger.error(`Skipping invalid import path: ${validation.importPath}. Reason: ${validation.message}`);
         }
+        return validation;
+      })
+      .filter((validation) => validation.isValid)
+      .map((validation) => validation.importPath)
+      .map((importPath) => path.normalize(importPath));
```
jrasm91, line 644: This isn't a time sensitive operation so I would probably prefer to just have a normal for loop and add valid paths to a list.

server/src/services/library.service.ts
```diff
@@ -598,106 +622,96 @@ export class LibraryService extends EventEmitter {
...
+    };
 
-    this.logger.verbose(`Crawled assets have been checked for online/offline status`);
+    let crawledAssetPaths: string[] = [];
 
-    if (assetIdsToMarkOffline.length > 0) {
-      this.logger.debug(`Found ${assetIdsToMarkOffline.length} offline asset(s) previously marked as online`);
-      await this.assetRepository.updateAll(assetIdsToMarkOffline, { isOffline: true });
-    }
+    while (!crawlDone) {
```
jrasm91, line 680: This is a bit weird. We should be able to use a normal (async) for loop here instead.

server/src/services/library.service.ts
```diff
@@ -598,106 +622,96 @@ export class LibraryService extends EventEmitter {
...
+        this.logger.log(`Queueing scan of ${crawledAssetPaths.length} asset path(s) in library ${library.id}...`);
+        // We have reached the batch size or the end of the generator, scan the assets
+        await this.scanAssets(job.id, crawledAssetPaths, library.ownerId, job.refreshAllFiles ?? false);
+        crawledAssetPaths = [];
 
-        if (batch.length >= LIBRARY_SCAN_BATCH_SIZE) {
-          await this.scanAssets(job.id, batch, library.ownerId, job.refreshAllFiles ?? false);
-          batch = [];
+        if (!existingAssetsDone) {
+          // Interweave the queuing of offline checks with the asset scanning (if any)
```
jrasm91, line 697: What is the benefit of this? If anything it seems like it would be less efficient.

server/src/services/library.service.ts
```diff
@@ -598,106 +622,96 @@ export class LibraryService extends EventEmitter {
...
     }
 
-    await this.repository.update({ id: job.id, refreshedAt: new Date() });
-
-    return JobStatus.SUCCESS;
-  }
+    // If there are any remaining assets to check for offline status, do so
+    while (!existingAssetsDone) {
+      await checkIfOnlineAssetsAreOffline();
+    }
```
jrasm91, line 706: This can easily become an infinite loop. Let's use a for loop instead.

server/src/domain/library/library.service.ts
etnoy: Thanks!

server/src/domain/library/library.service.ts
etnoy: Done, thanks

server/src/domain/library/library.service.ts
etnoy: Nicely spotted

server/src/services/library.service.ts
```diff
@@ -580,6 +574,29 @@ export class LibraryService extends EventEmitter {
     return JobStatus.SUCCESS;
   }
 
+  // Check if an asset has no file or is outside of import paths, marking it as offline
```
jrasm91, line 577: Nice, this looks simple and easy to read.

server/e2e/jobs/specs/library.e2e-spec.ts
etnoy: the remaining e2e tests seem to be non-trivial to merge, will look at it in a separate PR

server/src/services/library.service.ts
etnoy: See previous comment

server/src/domain/library/library.service.ts
etnoy: I do agree that this function is growing more and more complex, but this PR does not greatly increase that complexity IMO. Perhaps I can refactor the function further in a separate PR?

server/src/domain/library/library.service.ts
etnoy: I think it's better now, do you agree?

server/src/domain/library/library.service.ts
danieldietzler: Can't you just pass the values? IMO arrow functions should be (as well as possible) independent and not have any side effects.

server/e2e/jobs/specs/library.e2e-spec.ts
```diff
@@ -236,6 +165,185 @@ describe(`${LibraryController.name} (e2e)`, () => {
...
+      await api.libraryApi.scanLibrary(server, admin.accessToken, library.id);
+
+      const onlineAssets = await api.assetApi.getAllAssets(server, admin.accessToken);
+      expect(onlineAssets.length).toBeGreaterThan(1);
+
+      await fs.promises.rm(`${IMMICH_TEST_ASSET_TEMP_PATH}/albums/nature/silver_fir.jpg`);
+
+      await api.libraryApi.scanLibrary(server, admin.accessToken, library.id);
+
+      const assets = await api.assetApi.getAllAssets(server, admin.accessToken);
```
danieldietzler, line 189: This is bound to break, correct? You have a race condition between the library scanning and the fetch for all assets.

server/e2e/jobs/specs/library.e2e-spec.ts
```diff
@@ -236,6 +165,185 @@ describe(`${LibraryController.name} (e2e)`, () => {
...
+      const onlineAssets = await api.assetApi.getAllAssets(server, admin.accessToken);
+      expect(onlineAssets.length).toBeGreaterThan(1);
+
+      await api.libraryApi.setImportPaths(server, admin.accessToken, library.id, [
+        `${IMMICH_TEST_ASSET_TEMP_PATH}/albums/other`,
+      ]);
+
+      await api.libraryApi.scanLibrary(server, admin.accessToken, library.id);
+
+      const assets = await api.assetApi.getAllAssets(server, admin.accessToken);
```
danieldietzler, line 229: Same here. As long as we don't have any events for this you could at least re-try it a couple of times or poll until the library queue is empty and then continue

server/e2e/jobs/specs/library.e2e-spec.ts
```diff
@@ -236,6 +165,185 @@ describe(`${LibraryController.name} (e2e)`, () => {
...
+
+      await fs.promises.rm(`${IMMICH_TEST_ASSET_TEMP_PATH}/albums/nature/silver_fir.jpg`);
+
+      await api.libraryApi.scanLibrary(server, admin.accessToken, library.id);
+
+      await fs.promises.cp(`${IMMICH_TEST_ASSET_PATH}/albums/nature`, `${IMMICH_TEST_ASSET_TEMP_PATH}/albums/nature`, {
+        recursive: true,
+      });
+
+      await api.libraryApi.scanLibrary(server, admin.accessToken, library.id);
```
danieldietzler, line 269: This is even more "racey" as the file copy could happen before the prior scan is done

server/e2e/jobs/specs/library.e2e-spec.ts
```diff
@@ -236,6 +165,185 @@ describe(`${LibraryController.name} (e2e)`, () => {
...
+      const initialAssets = await api.assetApi.getAllAssets(server, admin.accessToken);
+      expect(initialAssets.length).toBeGreaterThan(1);
+
+      await api.libraryApi.setImportPaths(server, admin.accessToken, library.id, [
+        `${IMMICH_TEST_ASSET_TEMP_PATH}/albums/other`,
+      ]);
+
+      await api.libraryApi.scanLibrary(server, admin.accessToken, library.id);
+
+      const offlineAssets = await api.assetApi.getAllAssets(server, admin.accessToken);
```
danieldietzler, line 311: Same here and a few lines below. I'm aware that it's currently hard to test these things as there aren't any websocket events for job/queue completion atm. But still, you shouldn't just write these tests as if they'd happen linearly because that (1) doesn't really test anything and (b) can result in flaky tests. In my opinion these tests don't add any value currently since they're simply not reliable.

server/src/services/library.service.spec.ts
```diff
@@ -247,49 +259,85 @@ describe(LibraryService.name, () => {
...
 
-      await sut.handleQueueAssetRefresh(mockLibraryJob);
+      storageMock.checkFileExists.mockResolvedValue(false);
 
-      expect(assetMock.updateAll).toHaveBeenCalledWith([assetStub.image.id], { isOffline: true });
-      expect(assetMock.updateAll).not.toHaveBeenCalledWith(expect.anything(), { isOffline: false });
-      expect(jobMock.queueAll).not.toHaveBeenCalled();
+      await sut.handleOfflineCheck(mockAssetJob);
+
+      expect(assetMock.update).toHaveBeenCalledWith({ id: assetStub.external.id, isOffline: true });
```
danieldietzler, line 277: I think it would be nice to also expect the response here.

server/src/services/library.service.spec.ts
```diff
@@ -247,49 +259,85 @@ describe(LibraryService.name, () => {
...
 
-      await sut.handleQueueAssetRefresh(mockLibraryJob);
+      storageMock.checkFileExists.mockResolvedValue(true);
 
-      expect(assetMock.updateAll).toHaveBeenCalledWith([assetStub.offline.id], { isOffline: false });
-      expect(assetMock.updateAll).not.toHaveBeenCalledWith(expect.anything(), { isOffline: true });
-      expect(jobMock.queueAll).not.toHaveBeenCalled();
+      await sut.handleOfflineCheck(mockAssetJob);
+
+      expect(assetMock.update).toHaveBeenCalledWith({ id: assetStub.external.id, isOffline: true });
```
danieldietzler, line 292: Same here

server/src/services/library.service.ts
danieldietzler: I don't understand. Why can't you do
```ts
for await (const crawlResult of crawledAssets) {
...
}
```
?

server/src/services/library.service.ts
```diff
@@ -607,106 +624,92 @@ export class LibraryService extends EventEmitter {
...
-        this.logger.debug(`Will import ${crawledAssetPaths.size} new asset(s)`);
+      if (!crawlDone) {
+        crawledAssetPaths.push(crawlResult.value);
+        crawlCounter++;
       }
 
-      let batch = [];
-      for (const assetPath of crawledAssetPaths) {
-        batch.push(assetPath);
+      if (crawledAssetPaths.length % LIBRARY_SCAN_BATCH_SIZE === 0 || crawlDone) {
```
danieldietzler, line 688: Can't we move that whole if outside the loop? Once we quit that loop we inherently are done with the scanning, no?

danieldietzler (dismissed):
This has already gotten much more readable, thanks!
And sorry that PR got stale lol

server/e2e/jobs/specs/library.e2e-spec.ts
etnoy: I don't see how this would race. In these e2e jobs, we mock the job queue so all queued jobs finish with the await

danieldietzler: Oh I see. Sorry I mixed that up then... Then it'll only become an issue once we move them to e2e/

etnoy: I'm not looking forward to that 😂

server/src/services/library.service.ts
etnoy: The code needs to do two tasks, and to make the system feel more responsive I don't want to wait for the first (potentially very long) job to be worked on before the next jobs start. We therefore interleave the two jobs in between

server/src/services/library.service.spec.ts
etnoy: Done

server/src/services/library.service.spec.ts
etnoy: Done

server/src/services/library.service.ts
etnoy: No, we aren't done with the scanning after the loop, it also needs to check if any existing assets need to be checked.

server/src/services/library.service.ts
danieldietzler: This reply does not make any sense to me. Doesn't this semantically do the exact same thing?

server/src/services/library.service.ts
danieldietzler: I don't understand this. You're implying that batching does not give you all results?

danieldietzler: Maybe I am also just very confused... This whole interleaved thing of those two jobs may make it more performant but it definitely adds a lot of complexity my brain apparently can't comprehend.

etnoy: I changed my mind and did it your way which is much simpler. We can revisit if we need more performance

server/src/services/library.service.ts
etnoy: I decided to simplify things and take the straightforward approach, otherwise I might be falling into the premature optimization trap. We can revisit later if needed

server/src/services/library.service.ts

server/src/domain/library/library.service.ts
etnoy: Done in the latest refactor

server/src/services/library.service.ts
```diff
@@ -540,6 +533,34 @@ export class LibraryService implements OnEvents {
...
+      }
+    }
+
+    // Either asset path does not exist, or it is outside of any import path
+    this.logger.debug(`Marking asset as offline: ${asset.originalPath}`);
+    await this.assetRepository.update({ id: asset.id, isOffline: true });
+
+    return JobStatus.SUCCESS;
+  }
+
```
jrasm91, line 563: I think a lot of these comments are unnecessary. If anything some could be communicated with better variable names.

server/src/services/library.service.ts
```diff
@@ -540,6 +533,34 @@ export class LibraryService implements OnEvents {
...
+    if (!asset || asset.isOffline) {
+      // We only care about online assets, we exit here if offline
+      return JobStatus.SKIPPED;
+    }
+
+    const exists = await this.storageRepository.checkFileExists(asset.originalPath, R_OK);
+
+    if (exists) {
+      const isInPath = job.importPaths.find((path) => asset.originalPath.startsWith(path));
```
jrasm91, line 548: this condition can actually be checked outside of this if statement. I think it would be more clear if the code was like this:
```ts
if(!fileExists || !validPath) { 
    this.logger.debug(`Marking asset as offline: ${asset.originalPath}`);
    await this.assetRepository.update({ id: asset.id, isOffline: true });
}

return JobStatus.SUCCESS

File: server/src/services/library.service.ts [hunk 27]
```diff
@@ -570,100 +591,78 @@ export class LibraryService implements OnEvents {
...
-      .map((validation) => {
-        if (!validation.isValid) {
-          this.logger.error(`Skipping invalid import path: ${validation.importPath}. Reason: ${validation.message}`);
-        }
-        return validation;
-      })
-      .filter((validation) => validation.isValid)
-      .map((validation) => validation.importPath);
+    // Check all existing online assets to see if they are still online
+    while (!existingAssetsDone) {
```
jrasm91, line 641: change this for a for loop. for examples, just look at other places in the code `usePagination` is used.

server/src/services/library.service.ts
```diff
@@ -570,100 +591,78 @@ export class LibraryService implements OnEvents {
...
-    }
+    // (Re-)import all assets found on disk
+    while (!crawlDone) {
+      const assetGenerator = await crawledAssets.next();
+      crawlDone = assetGenerator.done ?? true;
 
-    if (crawledAssetPaths.size > 0) {
-      if (!shouldScanAll) {
-        this.logger.debug(`Will import ${crawledAssetPaths.size} new asset(s)`);
+      if (!crawlDone) {
```
jrasm91, line 619: Use a for loop here

server/src/services/library.service.ts
danieldietzler: I think what you actually want is what I already commented [here](https://github.com/immich-app/immich/pull/7934#discussion_r1553707433), so no need to even look for examples :P

server/src/services/library.service.ts
```diff
@@ -252,24 +251,17 @@ export class LibraryService implements OnEvents {
...
-      );
-    }
-
-    this.logger.debug('Asset refresh queue completed');
+    await this.jobRepository.queueAll(
+      assetPaths.map((assetPath) => ({
+        name: JobName.LIBRARY_SCAN_ASSET,
+        data: {
+          id: libraryId,
+          assetPath: assetPath,
```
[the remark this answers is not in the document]

server/src/services/library.service.ts

server/src/services/library.service.ts
etnoy: @jrasm91 I've looked into using pagination instead of the current approach which would be idea. However, in this case we have a generator (which internally uses a ReadableStream). Would you be fine with me modifying usePagination to consume generators, too?

server/src/services/library.service.ts

server/src/services/library.service.ts
etnoy: Done. I implemented the batching in the walk* function directly

server/src/services/library.service.ts
etnoy: I've reworked the logic of this method and it should be much clearer

jrasm91 (approved):
LGTM

PR #7942, opened 2024-03-14
feat(web): add millisecond options to storage template settings

sevtdy:
Proposing to add an option for SSS (milliseconds) to the storage template.

Currently, my storage template setting looks like this:
`{{y}}/{{y}}-{{MM}}/{{y}}{{MM}}{{dd}}_{{HH}}{{mm}}{{ss}}`
`UPLOAD_LOCATION/admin/2022/2022-02/20220203_045605.jpg`

Although immich append +1, +2, etc. when the filename is duplicated, I would prefer to avoid this. 
I believe that adding an SSS (millisecond) option to the template would make more sense and be more suitable for most situations.

take my as example, ideally would be like this:
`{{y}}/{{y}}-{{MM}}/{{y}}{{MM}}{{dd}}_{{HH}}{{mm}}{{ss}}_{{SSS}}`
`UPLOAD_LOCATION/admin/2022/2022-02/20220203_045605_055.jpg`
`UPLOAD_LOCATION/admin/2022/2022-02/20220203_045605_171.jpg`

related: https://github.com/immich-app/immich/discussions/6545


jrasm91 (approved):
Nice. Do you want to add a preset with it?

PR #7943, opened 2024-03-14
feat(web): randomize password on reest

samholton:
Generate a random password when an admin resets a user's password.

[password reset.webm](https://github.com/immich-app/immich/assets/3507151/7cfdf004-3a20-49d0-86bc-18356cec4ede)

### DIFF
--- a/web/src/lib/components/forms/edit-user-form.svelte
+++ b/web/src/lib/components/forms/edit-user-form.svelte
@@ -72,6 +72,7 @@
     }
   };
 
+  // TODO move password reset server-side
   function generatePassword(length: number = 16) {
     let generatedPassword = '';
 
--- a/web/src/lib/components/forms/edit-user-form.svelte
+++ b/web/src/lib/components/forms/edit-user-form.svelte
@@ -13,6 +13,7 @@
 
   export let user: UserResponseDto;
   export let canResetPassword = true;
+  export let newPassword: string;
 
   let error: string;
   let success: string;
@@ -53,12 +54,12 @@
 
   const resetPassword = async () => {
     try {
-      const defaultPassword = 'password';
+      newPassword = generatePassword();
 
       await updateUser({
         updateUserDto: {
           id: user.id,
-          password: defaultPassword,
+          password: newPassword,
           shouldChangePassword: true,
         },
       });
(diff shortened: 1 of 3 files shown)


web/src/lib/components/forms/edit-user-form.svelte
```diff
@@ -70,6 +71,22 @@
...
+
+    for (let i = 0; i < length; i++) {
+      let randomNumber = crypto.getRandomValues(new Uint32Array(1))[0];
+      randomNumber = randomNumber / 2 ** 32;
+      randomNumber = Math.floor(randomNumber * characterSet.length);
+
+      generatedPassword += characterSet[randomNumber];
+    }
+
+    return generatedPassword;
```
michelheusschen, line 88: A bit lengthy for my liking, what do you think of this?

```suggestion
    const characterSet = '0123456789' + 'abcdefghijklmnopqrstuvwxyz' + 'ABCDEFGHIJKLMNOPQRSTUVWXYZ' + ',.-{}+!#$%/()=?';

    return Array.from(crypto.getRandomValues(new Uint32Array(length)))
      .map((x) => characterSet[x % characterSet.length])
      .join('');
````

File: web/src/routes/admin/user-management/+page.svelte [hunk 2]
```diff
@@ -147,29 +152,40 @@
...
+              <div class="flex flex-col gap-4">
+                <p>The user's password has been reset:</p>
+
+                <div class="flex justify-center gap-2">
+                  <code
+                    class="rounded-md bg-gray-200 px-2 py-1 font-bold text-immich-primary dark:text-immich-dark-primary dark:bg-gray-700"
+                  >
+                    {newPassword}
+                  </code>
+                  <LinkButton on:click={() => copyToClipboard(newPassword)}>
```
michelheusschen, line 175: This button should have `Copy password` somewhere. You could add it as `title=""` or regular text, but we shouldn't only rely on the icon.

There's also no feedback to the user when clicking the copy button

web/src/lib/components/forms/edit-user-form.svelte
oddlama: Both of these generators are biased because of using modulo / floor. The lower `N % len(characterSet)` characters will be chosen more often than others because the wrap around is most likely not aligned.

To make a secure one, you should either use a uniform random number distribution to pick characters or just make a loop and generate random bytes while discarding any that are not a valid characters. Looping is probably the easiest to get right, cryptographically speaking.

jrasm91: The server just uses random bytes and then stips out non alpha-numeric characters, which seems more than sufficient imo.

jrasm91: Generating stuff like this client-side has its complexities. I added a note that we should move this to the server in a follow-up PR.

jrasm91 (approved):

PR #7946, opened 2024-03-14
fix(web): keyboard shortcut handling

michelheusschen:
Keyboard shortcuts are currently handled pretty inconsistently and maybe in an unexpected way. Let's look at an example listening for the escape key using `on:keydown={(e) => e.key === 'Escape'}`. This not only triggers when pressing escape, but also with any combination of modifier key (ctrl, alt, shift or meta). Another issue may arise when listening for a key that is entered in an input field, where the event handler will still execute.

In some cases these issues are accounted for, but not everywhere. A new action `use:shortcut` and function `executeShortcuts` are added that manage these pitfalls. However, they aren't yet used everywhere as it would make this PR a bit large.


web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -551,7 +505,7 @@
   };
 </script>
 
-<svelte:window on:keydown={handleKeypress} />
+<svelte:window on:keydown={(event) => executeShortcuts(event, shortcuts)} />
```
jrasm91, line 508: Why can't you use the other form here?

jrasm91 (approved):
Awesome work. Eventually would it be possible to automatically show a list of registered shortcuts when you press `?`? Like make that modal, which is currently static, dynamic?

web/src/lib/components/asset-viewer/asset-viewer.svelte
michelheusschen: Do you mean just a bunch of `use:shortcut` with other form? That's certainly possible, but I think using the list looks better with a lot of shortcuts.

jrasm91: Oh, I think I misunderstood the interface for `use:shortcut`. I think it might be nicer if it basically took in a list of shortcuts (shortcut + callback) instead and then it looked the same everywhere. What do you think? Basically, why can't the default implementation for `use:shortcut` by `executeShortcuts`?

jrasm91:
> It can be done, but it's a little tricky. You need to register all shortcuts and add some logic for filtering out duplicate and irrelevant shortcuts such as enter or escape. Hoewever, this PR should make it a lot easier to implement that.

Yeah that makes sense. I assumed we'd have to add something like `title: string` and/or `description: string` to the ones that would be included. I also think you'd need to register them all directly, instead of using the deferred `executeShortcuts` implementation, which looks just like shorthand for shortcut aliases.

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -250,68 +250,22 @@
...
-        $isShowDetail = !$isShowDetail;
-        return;
-      }
-    }
+  let shortcuts: ShortcutList;
+  $: shortcuts = [
+    [{ key: 'a', shift: true }, () => toggleArchive()],
+    [{ key: 'ArrowLeft' }, () => navigateAsset('previous')],
+    [{ key: 'ArrowRight' }, () => navigateAsset('next')],
+    [{ key: 'd', shift: true }, () => downloadFile(asset)],
```
martabal, line 258: Can some shortcuts be case insensitive like before ?

web/src/lib/components/asset-viewer/asset-viewer.svelte
michelheusschen: I've added a new action called `use:shortcuts` which accept multiple shortcuts and replaces the `executeShortcuts` function. This means all shortcuts are registered via actions, which would make implementing the dynamic shortcuts modal easier

web/src/lib/components/asset-viewer/asset-viewer.svelte
michelheusschen: Matching shortcuts is case insensitive, this is handled with `shortcut.key.toLowerCase() === event.key.toLowerCase()`

web/src/lib/components/asset-viewer/asset-viewer.svelte
jrasm91: Awesome, thanks.

web/src/lib/components/asset-viewer/asset-viewer.svelte
martabal: Oh nice !

alextran1502 (approved):
Amazing work! Thank you

danieldietzler (approved):
Dang this is *amazing*! Really awesome! ❤️

PR #7952, opened 2024-03-14
chore(web): Sharing -> Partner Sharing

aviv926:
Makes the title clearer.


alextran1502 (approved):

jrasm91:
I think the idea was to put all "sharing" related features here, even though it is only partner sharing right now. We might have to change this back if we ever add additional "sharing"-related settings.

PR #7960, opened 2024-03-14
feat(server): added support for SVG files

declan8010:
This PR enhances the Immich open-source project by adding support for Scalable Vector Graphics (SVG) files. With this feature, users can now upload SVG files as assets and view them within the application. SVG files offer several advantages, including scalability and resolution independence, making them a valuable addition to the supported asset formats.

While I have thoroughly tested this implementation on the web interface, the integration with mobile clients remains untested. I encourage our contributors to review, expand, and refine this feature as needed.

There is going to be some rough edges so I invite any of our contributers to add or remove anything that will help 👍 

image


server/src/immich/api-v1/asset/asset.service.ts
```diff
@@ -338,7 +340,7 @@ export class AssetService {
       fileModifiedAt: dto.fileModifiedAt,
       localDateTime: dto.fileCreatedAt,
 
-      type: mimeTypes.assetType(file.originalPath),
```
alextran1502, line 341: We can leave this as is, it is self-explanatory so you don't need to add a comment

alextran1502 (approved):

server/src/immich/api-v1/asset/asset.service.ts
declan8010: Actioned

alextran1502:
I mean you can keep this `type: mimeTypes.assetType(file.originalPath),` instead of creating a local variable for the method :)

declan8010:
@alextran1502 Would you be able to take a look at why the server test is failing if you have a moment?

alextran1502:
You can read about why it failed here https://github.com/immich-app/immich/actions/runs/8286816641/job/22677716034?pr=7960

declan8010:
> You can read about why it failed here https://github.com/immich-app/immich/actions/runs/8286816641/job/22677716034?pr=7960

Its strange that its still failing even after I moved the mimetype to be in the correct order

image

jrasm91:
I think there are some security issues related to svgs that we previously discussed. cc: @mertalev

alextran1502:
@jrasm91 I remember now, like add code to SVG file?

jrasm91:
SVGs can container javascript and they'll be executed, etc. by the browser. There are other security considerations as well I believe, like SVGs can have references to external scripts, etc.

mertalev:
Yes, there are some [security issues](https://medium.com/sse-blog/beware-svg-its-not-just-a-scalable-graphic-9a2e9f3df6ef) with user-provided SVG. They're safe if used in `img` blocks, but if the user right-clicks the image and opens it in another tab they can have arbitrary code executed.

alextran1502:
Maybe we can make an option for the admin to disable uploading SVG files to the server later. So this change can be accepted now

PR #7962, opened 2024-03-14
refactor(cli): simplify

jrasm91:


e2e/src/cli/specs/login.e2e-spec.ts
```diff
@@ -21,7 +21,9 @@ describe(`immich login-key`, () => {
 
   it('should require a valid key', async () => {
     const { stderr, exitCode } = await immichCli(['login-key', app, 'immich-is-so-cool']);
-    expect(stderr).toContain('Failed to connect to server http://127.0.0.1:2283/api: Error: 401');
+    expect(stderr).toContain('Failed to connect to server');
+    expect(stderr).toContain('Invalid API key');
+    expect(stderr).toContain('401');
```
jrasm91, line 26: For API errors, it'll log the response body now.

cli/src/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+  instanceUrl: string;
+  apiKey: string;
+}
+
+export const authenticate = async (options: BaseOptions): Promise<void> => {
+  const { configDirectory: configDir, instanceUrl, apiKey } = options;
+
+  // provided in command
+  if (instanceUrl && apiKey) {
+    await connect(instanceUrl, apiKey);
```
danieldietzler, line 22: I believe you miss a return here, right?

danieldietzler (approved):
Very easy to read, awesome! :)

cli/src/utils.ts
jrasm91: Yuuuuup.

PR #7969, opened 2024-03-15
refactor: library e2e

jrasm91:
With help from @danieldietzler, moving some of the library e2e tests over.

### DIFF
--- a/e2e/src/utils.ts
+++ b/e2e/src/utils.ts
@@ -5,6 +5,7 @@ import {
   CreateAssetDto,
   CreateLibraryDto,
   CreateUserDto,
+  MetadataSearchDto,
   PersonCreateDto,
   SharedLinkCreateDto,
   ValidateLibraryDto,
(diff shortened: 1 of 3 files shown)


e2e/src/utils.ts
```diff
@@ -83,16 +86,30 @@ const events: Record<EventType, Set<string>> = {
...
 
 const execPromise = promisify(exec);
 
 const onEvent = ({ event, id }: { event: EventType; id: string }) => {
-  events[event].add(id);
-  const callback = callbacks[id];
-  if (callback) {
-    callback();
-    delete callbacks[id];
+  // console.log(`Received event: ${event} [id=${id}]`);
```
danieldietzler, line 95: Is that intentional?

danieldietzler (approved):
Thanks for finishing this up! :)

e2e/src/utils.ts
jrasm91: Yeah it is good for debugging, but it's quite noisy

PR #7974, opened 2024-03-15
perf(web): batch asset store changes

michelheusschen:
The asset store is updated for every add and update event, which causes noticable performance issues with >100 events. By batching these events we can reduce the amount of times `emit()` is called and minimize sorting. Events are still executed by insertion order to avoid breaking things.

Adding 1000 assets to the store (tested via uploading) now takes less than 50ms, compared to 3-5 seconds previously. Updating assets shows a similar decrease. The wait time between processing changes is also reduced to 2500ms (from 10_000ms).


jrasm91 (approved):
LGTM.

PR #7985, opened 2024-03-15
refactor: global validation pipe

jrasm91:
* remove validation pipe usage in asset v1 controller
* remove `UseValidation`
* add validation pipe globally

### DIFF
--- a/server/src/immich/app.module.ts
+++ b/server/src/immich/app.module.ts
@@ -70,7 +70,7 @@ import { ErrorInterceptor, FileUploadInterceptor } from './interceptors';
     PersonController,
   ],
   providers: [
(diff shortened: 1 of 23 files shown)


danieldietzler (approved):
LGTM

server/src/immich/app.module.ts
```diff
@@ -70,6 +70,7 @@ import { ErrorInterceptor, FileUploadInterceptor } from './interceptors';
     PersonController,
   ],
   providers: [
+    { provide: APP_PIPE, useValue: new ValidationPipe({ transform:true, whitelist: true}) },
```
jrasm91, line 73: This will apply it automatically to all controllers (query, params, body, etc.)

PR #7987, opened 2024-03-15
refactor(web): list navigation with keyboard

michelheusschen:
Adds the `use:listNavigation` action that can be used to navigate a list with the arrow up/down keys. This action replaces the current implementations to navigate through location and people suggestions.

The current implementation manually tracks the focused element, but this breaks when using `Tab` or `Shift+Tab` to navigate. `use:listNavigation` uses document.activeElement for tracking to avoid this issue. Other than that, behavior should be the same.


alextran1502 (approved):
Great improvement, thank you

PR #7997, opened 2024-03-16
feat(web) add filter when viewing all people in search box

samholton:
Addresses an issue when an install has many faces matched and you want to select specific faces. Before I had to scroll through all the faces looking for individuals I wanted to select. This can be tedious when there are many faces present. When clicking `See all people` on the enhanced search, a name filter appears.

[search people filter 2.webm](https://github.com/immich-app/immich/assets/3507151/70435872-4a7a-4541-94de-712b98c16fd6)


samholton:
I also played with moving the selected people to the front immediately. If you filtered and selected someone outside the range of the initial slice, they would move to the front when collapsing the all people. However, this may be disorienting when scrolling through the list and clicking the face - it will disappear out of view and move to the top.

[search filter - move to front.webm](https://github.com/immich-app/immich/assets/3507151/9d0e7885-5521-4c77-bb02-dee4fb57c5b9)

mertalev:
The spacing above and below PEOPLE changes when the search bar appears / disappears, making it a bit jarring. Can you make it so the search bar doesn't move anything?

samholton:
> The spacing above and below PEOPLE changes when the search bar appears / disappears, making it a bit jarring. Can you make it so the search bar doesn't move anything?

Fixed that and updated first screenshot, didn't re-record the second

web/src/lib/components/shared-components/search-bar/search-people-section.svelte
```diff
@@ -38,15 +40,23 @@
     }
     selectedPeople = selectedPeople;
   }
+
+  const filterPeople = (list: PersonResponseDto[], name: string) => {
+    return name == '' ? list : list.filter((p) => p.name.toLowerCase().startsWith(name.toLowerCase()));
```
mertalev, line 45: You can move `name.toLowerCase()` to a variable to only evaluate it once.

mertalev (approved):
This is really nice!

web/src/lib/components/shared-components/search-bar/search-people-section.svelte
```diff
@@ -13,6 +14,7 @@
   let peoplePromise = getPeople();
   let showAllPeople = false;
   $: numberOfPeople = (width - 80) / 85;
+  $: name = '';
```
michelheusschen, line 17: This should be a regular prop

web/src/lib/components/shared-components/search-bar/search-people-section.svelte
michelheusschen: Try to use strict equality by default, in this case you could also do `name ? list : ...`

web/src/lib/components/shared-components/search-bar/search-people-section.svelte
```diff
@@ -38,15 +40,23 @@
...
+    {@const peopleList = showAllPeople ? filterPeople(people, name) : people.slice(0, numberOfPeople)}
 
     <div id="people-selection" class="-mb-4">
-      <div class="flex items-center gap-6">
-        <p class="immich-form-label">PEOPLE</p>
+      <div class="flex items-center w-full justify-between gap-6">
+        <p class="immich-form-label py-3">PEOPLE</p>
+
+        {#if showAllPeople}
+          <SearchBar bind:name placeholder="Filter people" isSearching={false} />
```
michelheusschen, line 58: The searchbar contains a `<button>` for the search icon and clicking it will submit the form. Maybe add `type="button"` to prevent that

samholton: I see that behavior now. Adding `type="button"` didn't prevent it. Also tried adding `on:search={(event) => event.stopPropagation()})`. Will keep looking

michelheusschen: Strange, for me it does work when I add type="button" here. Maybe you added it to `<SearchBar>` directly?

https://github.com/immich-app/immich/blob/321525ead593e8b270741378cde912a5bb55bb27/web/src/lib/components/elements/search-bar.svelte#L26

samholton: I did indeed try adding it to `<SearchBar>` :)

michelheusschen: The dispatch isn't relevant actually, because we don't listen for it and custom events don't automatically bubble up. The behavior you're seeing is caused by the default type=submit that buttons have and by putting it inside `<form>` it will submit the form when you click it. Adding type=button disables the default behavior.

samholton: Ah, makes sense, thanks for the explanation.

web/src/lib/components/shared-components/search-bar/search-people-section.svelte
```diff
@@ -38,15 +40,24 @@
     }
     selectedPeople = selectedPeople;
   }
+
+  const filterPeople = (list: PersonResponseDto[], name: string) => {
+    const nameLower = name.toLowerCase();
+    return name === '' ? list : list.filter((p) => p.name.toLowerCase().startsWith(nameLower));
```
danieldietzler, line 46: You don't need `name === ''` as `''` already is falsy.
```suggestion
    return name ? list : list.filter((p) => p.name.toLowerCase().startsWith(nameLower));
```

danieldietzler: Also, do we want to use `includes` instead for a more fuzzy-ish search?

danieldietzler (approved):

web/src/lib/components/shared-components/search-bar/search-people-section.svelte
samholton: I actually originally had include but noticed the filter on people page does starts with. Starts with probably makes sense for names unless you want to search for all people with last name

samholton: I'll play with the if again tonight. I thought I tried that first and it didn't work. But it was late last night

PR #8007, opened 2024-03-16
feat(web): combobox accessibility improvements

ben-basten:
## Description

This change makes the combobox more accessible for keyboard use. It visibly should look the same. Design was modeled after the WCAG documentation here:

https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/

**Changes include:**

* Arrow key navigation of the completions
* Arrow key navigation of completions loops at the beginning or the end of the list
* Screen reader announces completion options
* Completions are removed from the tab order, so it's easier to navigate the form with a single press of the "Tab" key
* Integrate labels into the component, since they're a required part of any combobox
* Allow labels to be hidden for screen reader only
* Labeling icons within the component
* Show/hide the completions on focus and blur, rather than just click
* Auto scrolling to guarantee that the selected option is always visible
* opt + down arrow opens the completion list without selecting anything
* Maintain search magnifying glass icon while the component is active

**See also:** Bumped the z-index on the skip link, because it is showing underneath the search box in narrow screen widths.

Open to any thoughts, comments, or feedback if you'd like to see any changes in how this works!

## How Has This Been Tested?

- [x] MacOS + VoiceOver
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
    - One issue to note: VoiceOver doesn't announce autocomplete options on the initial load without filtering. This same issue exists in the WCAG documentation. I haven't found a fix for this yet.
- [x] Windows + NVDA - tested by @michelheusschen
  - [x] Chrome
  - [x] Firefox

Dropdown list with some example data:

Screenshot 2024-03-16 at 11 36 40

## Checklist:

* [x] npm run lint (linting via ESLint)
* [x] npm run format (formatting via Prettier)
* [x] npm run check:svelte (Type checking via SvelteKit)
* [x] npm test (Tests via Jest)


ben-basten:
The keyboard shortcut handling changes in #7946 were super helpful! #:heart:

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -11,81 +11,204 @@
...
-    // In rare cases it's possible for the input to still have focus and
-    // outclick to fire.
-    if (!inputFocused) {
-      isOpen = false;
+  const closeDropdown = () => {
+    isOpen = false;
+    selectedIndex = undefined;
+  };
+
+  const moveFocus = (event: KeyboardEvent) => {
```
michelheusschen, line 78: What does this function do exactly? List elements aren't focusable because that's managed with `aria-activedescendant` and moving focus to the next/previous element should be default browser behavior.

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -11,81 +11,204 @@
...
+          shortcut: { key: 'ArrowDown' },
+          onShortcut: () => {
+            openDropdown();
+            void incrementSelectedIndex(1);
+          },
+        },
+        {
+          shortcut: { key: 'ArrowDown', alt: true },
+          onShortcut: () => {
+            openDropdown();
```
michelheusschen, line 193: Is this shortcut even needed? The dropdown should always be opened when input has focus

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -11,81 +11,204 @@
...
+          shortcut: { key: 'Enter' },
+          onShortcut: () => {
+            if (selectedIndex !== undefined && filteredOptions.length > 0) {
+              onSelect(filteredOptions[selectedIndex]);
+            }
+            closeDropdown();
+          },
+        },
+        {
+          shortcut: { key: 'Escape' },
```
michelheusschen, line 206: Pressing escape closes the whole search bar and not the dropdown

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -11,81 +11,204 @@
...
+  /**
+   * Keeps track of whether the combobox is actively being used.
+   */
+  let isActive = false;
   let searchQuery = selectedOption?.label || '';
+  let selectedIndex: number | undefined;
+  let comboboxRef: HTMLElement;
+  let optionRefs: HTMLElement[] = [];
+  const inputId = `combobox-${crypto.randomUUID()}`;
+  const listboxId = `listbox-${crypto.randomUUID()}`;
```
michelheusschen, line 41: `randomUUID()` is only available in secure contexts, this will probably cause some issues for people. Maybe we can use an alternative?

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -95,37 +218,52 @@
...
-    <div
-      role="listbox"
-      transition:fly={{ duration: 250 }}
-      class="absolute text-left text-sm w-full max-h-64 overflow-y-auto bg-white dark:bg-gray-800 rounded-b-lg border border-t-0 border-gray-300 dark:border-gray-900 z-10"
-    >
+  <ul
+    role="listbox"
+    id={listboxId}
+    transition:fly={{ duration: 250 }}
+    class="absolute text-left text-sm w-full max-h-64 overflow-y-auto bg-white dark:bg-gray-800 r unded-b-lg border border-t-0 border-gray-300 dark:border-gray-900 rounded-b-xl z-10"
```
michelheusschen, line 233: There are some invalid classes here. I also see a horizontal line below the combobox resulting in this with multiple comboboxes:

combobox-horizontal-line

michelheusschen (commented):
Nice work! Have you considered using the [listNavigation](https://github.com/immich-app/immich/blob/321525ead593e8b270741378cde912a5bb55bb27/web/src/lib/utils/list-navigation.ts) action? That would cut down on a lot of code if that can be used, maybe with some changes to it.

ben-basten:
> Nice work! Have you considered using the [listNavigation](https://github.com/immich-app/immich/blob/321525ead593e8b270741378cde912a5bb55bb27/web/src/lib/utils/list-navigation.ts) action? That would cut down on a lot of code if that can be used, maybe with some changes to it.

I haven't seen that before! I'll look into this. I'm all for reusing code, especially for other comboboxes in the app 😄

web/src/lib/components/shared-components/combobox.svelte
ben-basten: I saw this too... will troubleshoot and find a way to prevent the escape key from bubbling and closing the search bar.

I was hesitant to fix this, because in my opinion I don't think escape should close the whole search bar. It could be a frustrating experience for users to have some fields where pressing Escape closes just the dropdown, and other fields it closes the whole search box and clears out all the input fields they've been working on.

web/src/lib/components/shared-components/combobox.svelte
ben-basten: The suggestion to add this shortcut came from the WCAG spec:

https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/#kbd_label_textbox

There are scenarios where the input field is focused but the dropdown isn't visible, such as when the user presses Escape, or presses Enter without a dropdown completion highlighted. This just gives a quick way to open the dropdown again.

web/src/lib/components/shared-components/combobox.svelte
ben-basten: Found a solution for this - making the `preventDefault()` in the shortcuts action configurable with a flag.

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -11,81 +11,198 @@
...
-      class:cursor-pointer={!isOpen}
-      value={isOpen ? '' : selectedOption?.label || ''}
-      on:input={(e) => (searchQuery = e.currentTarget.value)}
-      on:focus={handleClick}
-      on:blur={() => (inputFocused = false)}
+      class:cursor-pointer={!isActive}
+      class="immich-form-input text-sm text-left w-full !pr-12 transition-all"
+      id={inputId}
+      on:click={activate}
+      on:focus={activate}
```
michelheusschen, line 150: `on:click` is probably not needed, because there's already `on:focus`

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -11,81 +11,198 @@
...
+
+  const incrementSelectedIndex = async (increment: number) => {
+    if (filteredOptions.length === 0) {
+      selectedIndex = 0;
+    } else if (selectedIndex === undefined) {
+      selectedIndex = increment === 1 ? 0 : filteredOptions.length - 1;
+    } else {
+      selectedIndex = (selectedIndex + increment + filteredOptions.length) % filteredOptions.length;
     }
+    await tick();
```
michelheusschen, line 94: Why do we need to wait for DOM changes?

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -11,81 +11,198 @@
...
+      id={inputId}
+      on:click={activate}
+      on:focus={activate}
+      on:input={onInput}
+      role="combobox"
+      type="text"
+      value={searchQuery}
+      use:shortcuts={[
+        {
+          shortcut: { key: 'Tab' },
```
michelheusschen, line 157: Don't know how reliable this detection is, because macos safari has an alternative shortcut for moving focus if I recall correctly. Maybe we could use something like this on the outer div and also prevent duplicating outclick logic:

```svelte
<div
  use:clickOutside={{ onOutclick: deactivate }}
  on:focusout={(e) => {
    if (e.relatedTarget instanceof Node && !e.currentTarget.contains(e.relatedTarget)) {
      deactivate();
    }
  }}
```

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -95,37 +212,53 @@
...
-    <div
-      role="listbox"
-      transition:fly={{ duration: 250 }}
-      class="absolute text-left text-sm w-full max-h-64 overflow-y-auto bg-white dark:bg-gray-800 rounded-b-lg border border-t-0 border-gray-300 dark:border-gray-900 z-10"
-    >
+  <ul
+    role="listbox"
+    id={listboxId}
+    transition:fly={{ duration: 250 }}
+    class="absolute text-left text-sm w-full max-h-64 overflow-y-auto bg-white dark:bg-gray-800 r unded-b-lg border-t-0 border-gray-300 dark:border-gray-900 rounded-b-xl z-10"
```
michelheusschen, line 227:
```suggestion
    class="absolute text-left text-sm w-full max-h-64 overflow-y-auto bg-white dark:bg-gray-800 border-t-0 border-gray-300 dark:border-gray-900 rounded-b-xl z-10"
```

web/src/lib/components/shared-components/search-bar/search-bar.svelte
```diff
@@ -84,7 +85,18 @@
...
-<div class="w-full relative" use:clickOutside on:outclick={onFocusOut} on:escape={onFocusOut}>
+<div
+  class="w-full relative"
+  use:clickOutside
+  on:outclick={onFocusOut}
+  tabindex="-1"
+  use:shortcut={{
+    shortcut: { key: 'Escape' },
+    onShortcut: () => {
+      onFocusOut();
```
michelheusschen, line 96: This doesn't work if you clear focus by clicking on an empty area in the search bar. Moving the shortcut to `window` should solve that

web/src/lib/components/shared-components/combobox.svelte
ben-basten: Scenario: input box is focused, dropdown is closed, and up arrow is pressed to re-open the dropdown and select the last item in the list.

I was seeing an issue where the last item on the list would be selected, but not visible because `scrollIntoView` was called before the dropdown gets fully rendered. Waiting for DOM changes guarantees that the dropdown is rendered before doing any scrolling on it.

web/src/lib/components/shared-components/combobox.svelte
ben-basten: I added this to cover for the scenario where the input is focused, but the dropdown is closed after pressing "Escape" or "Enter". This gives mouse users a chance to re-open the dropdown by clicking on the input field again.

web/src/lib/components/shared-components/combobox.svelte
michelheusschen: Got it, that makes sense 👍

web/src/lib/components/shared-components/combobox.svelte
ben-basten: This approach is even simpler, nice!

web/src/lib/components/shared-components/search-bar/search-bar.svelte
ben-basten: Pushed out a fix - I noticed that the search field also needed an "Escape" key handler, to close the recent searches dropdown.

ben-basten:
> Nice work! Have you considered using the [listNavigation](https://github.com/immich-app/immich/blob/321525ead593e8b270741378cde912a5bb55bb27/web/src/lib/utils/list-navigation.ts) action? That would cut down on a lot of code if that can be used, maybe with some changes to it.

I've been thinking more about this - the approach in this PR uses aria tags and CSS for item selection in the dropdown, which is different from the other dropdowns that use focus movement.

I'm wondering if we could consider a different approach. What if we refactored to use this same reusable combobox for both location search and merging people? This could provide consistency in animations, behavior, and accessibility across the app. 

I understand this would require more refactoring to support these use cases, but it could be beneficial to have a single source of truth for comboboxes. What are your thoughts on this? 

I appreciate your time and detailed feedback on this pull request!

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -11,48 +11,93 @@
...
+    } else {
+      selectedIndex = (selectedIndex + increment + filteredOptions.length) % filteredOptions.length;
     }
+    await tick();
+    optionRefs[selectedIndex]?.scrollIntoView({ block: 'nearest' });
+  };
+
+  const onInput: FormEventHandler<HTMLInputElement> = (event) => {
+    openDropdown();
+    searchQuery = (event.target as HTMLInputElement).value;
```
michelheusschen, line 91: Should be the same in this case and `currentTarget` already has correct type

```suggestion
    searchQuery = event.currentTarget.value;
```

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -95,37 +190,53 @@
     >
       {#if selectedOption}
         <IconButton color="transparent-gray" on:click={onClear} title="Clear value">
-          <Icon path={mdiClose} />
+          <Icon path={mdiClose} ariaLabel="Clear value" />
```
michelheusschen, line 193: Should this also have `ariaHidden`?

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -95,37 +190,53 @@
...
-        {@const selected = option.label === selectedOption?.label}
-        <button
-          type="button"
+      {#each filteredOptions as option, index (option.label)}
+        <!-- svelte-ignore a11y-click-events-have-key-events -->
+        <li
+          aria-selected={index === selectedIndex}
+          bind:this={optionRefs[index]}
+          class:bg-gray-100={index === selectedIndex}
+          class:dark:bg-gray-700={index === selectedIndex}
```
michelheusschen, line 231: It's also possible to directly add these to class with `aria-selected:*`

michelheusschen (approved):
> I'm wondering if we could consider a different approach. What if we refactored to use this same reusable combobox for both location search and merging people? This could provide consistency in animations, behavior, and accessibility across the app. What are your thoughts on this?

Reusing the combobox in other places sounds good to me. Then we might not even need `listNavigation` anymore. We could still consider separating logic to improve maintainability, but it's not a requirement.

I noted some small things below, but having tested the PR using Windows+NVDA in both Chrome and Firefox without issues, this should be fine to merge. Thank you for working on this ❤️

ben-basten:
> Reusing the combobox in other places sounds good to me. Then we might not even need `listNavigation` anymore. We could still consider separating logic to improve maintainability, but it's not a requirement.
> 
> I noted some small things below, but having tested the PR using Windows+NVDA in both Chrome and Firefox without issues, this should be fine to merge. Thank you for working on this ❤️

Sounds good, appreciate your input. Thanks for the screen reader testing on the Windows side!

web/src/lib/components/shared-components/combobox.svelte
ben-basten: Can you share a snippet to clarify what you mean here? Happy to include this change once I have a better understanding.

michelheusschen: Basically just add `aria-selected:bg-gray-100 aria-selected:dark:bg-gray-700` to the class attribute of this element to replace these two lines

ben-basten: Oh cool, applied that change!

PR #8013, opened 2024-03-17
feat(docs): Database information

aviv926:
This PR adds a brief explanation of what each table in the Immich database contains, plus I've also included an image that can illustrate the table's hierarchy.

There are parts I'm not sure about and parts I don't know what their purpose is.
If anyone has any knowledge about this I would be happy to receive suggestions for fixes.


docs/docs/administration/img/database-structure.png
danieldietzler, line 1: TBH I'm not a huge fan of providing such an image. It is prone to be outdated the moment we release the docs. Additionally, I personally don't see the benefit of such a detailed graph. If people would like to dive this deep into how Immich persists data they are most likely able to just find that information by themselves.

docs/docs/administration/database-info.md
```diff
@@ -0,0 +1,49 @@
...
+| users                     | Contains information about the users of the system, email, password (encrypted), creation and update date, path to profile picture (optional), whether system administrator?, need to change password?, deletion date, oauthId (optional), storage label name (optional), whether Memories enabled?, username, avatar color (optional), available storage quota (optional), storage quota in use.                                                                                                                                                                                                                                                                                                                                                                                                  |
```
danieldietzler, line 49: I probably wouldn't try to give an exhaustive list here. If you want to include something like this I think it would be nice to briefly explain the *most important* tables. Something like `typeorm_metadata` or others however don't matter at all to get a quick overview.

danieldietzler (commented):
In general I think this page is way to exhaustive. If people are at a point where they need this kind of information, they are either able to figure it out themselves or they'll come on the discord and start something like a focus-discussion first anyways.
IMO - if anything - this should be nothing more than a brief summary of the most important tables, not a complete DB plan.

aviv926:
> In general I think this page is way to exhaustive. If people are at a point where they need this kind of information, they are either able to figure it out themselves or they'll come on the discord and start something like a focus-discussion first anyways. IMO - if anything - this should be nothing more than a brief summary of the most important tables, not a complete DB plan.

Got it, do you think it will merge? If so I will try to reduce the tables to what is most important.
If not I should spend the time working on the documentation in other aspects.

danieldietzler:
> > In general I think this page is way to exhaustive. If people are at a point where they need this kind of information, they are either able to figure it out themselves or they'll come on the discord and start something like a focus-discussion first anyways. IMO - if anything - this should be nothing more than a brief summary of the most important tables, not a complete DB plan.
> 
> Got it, do you think it will merge? If so I will try to reduce the tables to what is most important. If not I should spend the time working on the documentation in other aspects.

I can't really tell. I personally wouldn't merge it I think, but ultimately that's the admins' call.

alextran1502:
Hello, thank you for the PR. First, I am pro-documentation, so your contribution is very appreciated and we should capture information that must avoid repeatedly explaining the questions.

However, we don't need to capture this information in the official documentation site since this changes quite frequently due to our rapid development. I value your time and want to maintain a manageable amount of information that can be updated easily. This information should also belong to the development section instead, and as a developer, I would know how to find information about database relationships. Putting myself into an average user's shoes, I don't think you would care how the data structure would look; I would care more about explaining how a particular feature functions. For example, the difference between `Album share/partner share/public share` is more valuable to add and maintain.

With that said, I don't think I will merge this PR.

aviv926:
> Hello, thank you for the PR. First, I am pro-documentation, so your contribution is very appreciated and we should capture information that must avoid repeatedly explaining the questions.
> 
> However, we don't need to capture this information in the official documentation site since this changes quite frequently due to our rapid development. I value your time and want to maintain a manageable amount of information that can be updated easily. This information should also belong to the development section instead, and as a developer, I would know how to find information about database relationships. Putting myself into an average user's shoes, I don't think you would care how the data structure would look; I would care more about explaining how a particular feature functions. For example, the difference between `Album share/partner share/public share` is more valuable to add and maintain.
> 
> With that said, I don't think I will merge this PR.

Absolutely understandable, thanks for your feedback Alex and Daniel.

alextran1502:
Thank you very much for your understanding Aviv!

PR #8018, opened 2024-03-17
feat(ml): add cache_dir option to OpenVINO EP

dvdblg:
## Description

This PR enables caching for the OpenVINO EP by adding a `cache_dir` option to the provider config as documented [here](https://onnxruntime.ai/docs/execution-providers/OpenVINO-ExecutionProvider.html#summary-of-options) and suggested by @mertalev in https://github.com/immich-app/immich/pull/7854#issuecomment-2002035826

This greatly reduced memory consumption during smart search in my tests: using the model `XLM-Roberta-Large-Vit-B-32`, I saw a memory usage of ~2.5 GB during search instead of the ~7-8 GB without caching.

### DIFF
--- a/machine-learning/app/models/base.py
+++ b/machine-learning/app/models/base.py
@@ -186,7 +186,7 @@ class InferenceModel(ABC):
                 case "CPUExecutionProvider" | "CUDAExecutionProvider":
                     option = {"arena_extend_strategy": "kSameAsRequested"}
                 case "OpenVINOExecutionProvider":
-                    option = {"device_type": "GPU_FP32"}
+                    option = {"device_type": "GPU_FP32", "cache_dir": f"{self.cache_dir}/openvino"}
                 case _:
                     option = {}
             options.append(option)
--- a/machine-learning/app/models/base.py
+++ b/machine-learning/app/models/base.py
@@ -186,7 +186,7 @@ class InferenceModel(ABC):
                 case "CPUExecutionProvider" | "CUDAExecutionProvider":
                     option = {"arena_extend_strategy": "kSameAsRequested"}
                 case "OpenVINOExecutionProvider":
-                    option = {"device_type": "GPU_FP32", "cache_dir": f"{self.cache_dir}/openvino"}
+                    option = {"device_type": "GPU_FP32", "cache_dir": (self.cache_dir / "openvino").as_posix()}
                 case _:
                     option = {}
             options.append(option)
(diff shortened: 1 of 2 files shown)


dvdblg:
Maybe with caching enabled it may be worth to re-enable OpenVINO EP also for CPU (it was disabled in https://github.com/immich-app/immich/pull/7450) once the issue with dynamic axes is fixed.

machine-learning/app/models/base.py
```diff
@@ -186,7 +186,7 @@ def provider_options_default(self) -> list[dict[str, Any]]:
                 case "CPUExecutionProvider" | "CUDAExecutionProvider":
                     option = {"arena_extend_strategy": "kSameAsRequested"}
                 case "OpenVINOExecutionProvider":
-                    option = {"device_type": "GPU_FP32"}
+                    option = {"device_type": "GPU_FP32", "cache_dir": f"{self.cache_dir}/openvino"}
```
mertalev, line 189: Since `self.cache_dir` is a `Path`, you should do forward slashes instead of string concatenation:
```suggestion
                    option = {"device_type": "GPU_FP32", "cache_dir": (self.cache_dir / "openvino").as_posix()}
```

machine-learning/app/test_main.py
```diff
@@ -90,6 +90,7 @@ def test_sets_default_provider_options(self, mocker: MockerFixture) -> None:
         assert encoder.provider_options == [
             {"device_type": "GPU_FP32"},
             {"arena_extend_strategy": "kSameAsRequested"},
+            {"cache_dir": f"{encoder.cache_dir}/openvino"},
```
mertalev, line 93: This should go in the first dictionary instead of being a new entry

mertalev:
> Maybe with caching enabled it may be worth to re-enable OpenVINO EP also for CPU (it was disabled in #7450) once the issue with dynamic axes is fixed.

This is a good idea, but we should check if it's actually better than the normal CPU provider. That can be a separate PR.

machine-learning/app/test_main.py
dvdblg: right, I completely missed that

mertalev (approved):

agrawalsourav98:
> > Maybe with caching enabled it may be worth to re-enable OpenVINO EP also for CPU (it was disabled in #7450) once the issue with dynamic axes is fixed.
> 
> This is a good idea, but we should check if it's actually better than the normal CPU provider. That can be a separate PR.

There is no benefit to using OpenVINO EP for CPU as onnx runtime already optimizes for all the possible processor flags.

PR #8033, opened 2024-03-17
feat(server): use prisma

mertalev:
## Description

This is a work-in-progress PR that sets up the configuration for Prisma and switches the asset and search repositories to use it.

## Impressions
- Query-building is *very* nice: it feels powerful and expressive, and obviates the need for raw queries or SQL snippets in many cases.
- The typing is very good and caught some existing issues: `save` cannot actually create a new entity for most of its uses because the calls are missing several non-nullable properties, queueing assets with a null `assetId` or `embedding` is nonsensical beacuse they're non-nullable, etc.
- Performance is better than TypeORM: `getByIds` is sub-millisecond on average.
- Client extensions are awesome and make it easy to add custom functionality.
- Outside of client extensions, it isn't as hackable: you can't just add a raw SQL expression to parts of the query the way you can with TypeORM.
  - You can't group by truncated dates for instance, so I handled this query with Kysely.
  - In these cases, you can define a view in the schema file as well.
- Its support for extensions is pretty limited, hope they improve on this.
- The Kysely extension is a must and mitigates the above limitations.

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -10,12 +10,16 @@ RUN npm ci && \
   rm -rf node_modules/@img/sharp-libvips* && \
   rm -rf node_modules/@img/sharp-linuxmusl-x64
 COPY server .
+
+WORKDIR /usr/src/app/server
+RUN npm run prisma:generate
+
+WORKDIR /usr/src/app
 ENV PATH="${PATH}:/usr/src/app/bin" \
   IMMICH_ENV=development \
   NVIDIA_DRIVER_CAPABILITIES=all \
   NVIDIA_VISIBLE_DEVICES=all
 ENTRYPOINT ["tini", "--", "/bin/sh"]
-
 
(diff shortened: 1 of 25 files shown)


jrasm91 (commented):
The migration from save to update all seems quite independent of prisma. Can You pull those changes into a separate PR?

mertalev:
> The migration from save to update all seems quite independent of prisma. Can You pull those changes into a separate PR?

Good idea!

jrasm91:
Closing, as we've decided against using prisma at this time.

ferferga:
@jrasm91 May I ask why? Would love to get interop between different database providers.

jrasm91:
You wouldn't get that anyways as we use a lot of postgres specific features.

mertalev:
Prisma is too limited, both in the queries it can express, the database features it supports and the customization it offers.

PR #8035, opened 2024-03-18
fix(server): "view all" for cities only showing 12 cities

### LINKED ISSUE #7070: [Bug] Web Explore > Places > View All shows only first 12 places
### The bug

In the web client, under the `Explore` tab, the `Places` section initially shows 7 places. Clicking `View All` shows just 12 places, but there are 100s in the library. On the iOS mobile app, you can see all of the places.

Perhaps also worth noting that the places shown in both cases are the first N alphabetically, so all of my visible places in the web client start with `A`.

This bug has been mentioned in closed issues in a couple of places, but as far as I can tell there is no open issue for it. Related:
* #6092 which added the `View All` button, but didn't change the underlying API to surface > 12 places.
* #6676 which as of now has some changes requested and has no open issue tracking it.

Debian 12

### Version of Immich Server

v1.94.1

v.194.1

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
Standard: https://github.com/immich-app/immich/releases/download/v1.94.1/docker-compose.yml
```

### Your .env content

```Shell
Standard: https://github.com/immich-app/immich/releases/download/v1.94.1/example.env
```

### Reproduction steps

```bash
1. Ensure the test library has more than 12 places in the DB.
2. In the web client, click Explore.
3. Click View All. Only 12 places are shown.
```

### Additional information

_No response_

mertalev:
## Description

This PR adds a new endpoint for getting assets by distinct city to fix this issue. The query for the explore page is also updated to sort by the cities with the most assets, consistent with the query for people.

Fixes #7070

## How Has This Been Tested?

Tested that clicking "View All" shows a high number of thumbnails with the same styling as in the explore page. Also checked that the assets shown in the explore page are no longer sorted in alphabetical order.


mertalev:
Ironically, the query for getting all places is faster than the one for the explore page.

michelheusschen:
The order is all over the *place* for me. See screenshots below of the first six assets, with the count added to each place (only images and no invisible assets). Also notice the different thumbnail for Guadalupe Quistaj and Ristijärvi

**Explore page**
explore-page

**Places page**
places-page

jrasm91 (commented):
I think it would be more consistent if this was in the search controller.

mertalev:
> The order is all over the *place* for me. See screenshots below of the first six assets, with the count added to each place (only images and no invisible assets). Also notice the different thumbnail for Guadalupe Quistaj and Ristijärvi
> 
> 
> 
> **Explore page**
> 
> explore-page
> 
> 
> 
> **Places page**
> 
> places-page
> 
> 
> 
> 

Hmmm, the places page not being sorted might be because there's no order by in the outer query. But it's weird that the explore page isn't actually sorted by count.

mertalev:
The places query had a bug where it would do the filters *after* getting assets, so that should be fixed now. I reverted the change to the explore page query since getting the count right with filters is harder and can be done in a separate PR.

server/src/infra/repositories/search.repository.ts
```diff
@@ -220,6 +221,96 @@ export class SearchRepository implements ISearchRepository {
       .getMany();
   }
 
+  @GenerateSql({ params: [[DummyValue.UUID]] })
+  async getAssetsByCity(userIds: string[]): Promise<AssetEntity[]> {
```
jrasm91, line 225: Man this is ugly.

jrasm91 (approved):

server/src/infra/repositories/search.repository.ts
mertalev: It was too much for the query builder haha

alextran1502:
I think when the place count is under a certain number, clicking on the "View All" button i.e. `/places` cause an error

image

image

server/src/domain/search/search.service.ts
```diff
@@ -211,4 +211,10 @@ export class SearchService {
       }
     }
   }
+
+  async getAssetsByCity(auth: AuthDto): Promise<AssetResponseDto[]> {
+    const userIds = await this.getUserIdsToSearch(auth);
+    const assets = await this.searchRepository.getAssetsByCity(userIds);
+    return assets.map((asset) => mapAsset(asset));
+  }
```
danieldietzler, line 219: Can we move that method up? I think having all private methods at the bottom generally is preferred...?

danieldietzler (approved):
LGTM!

server/src/domain/search/search.service.ts
mertalev: Agreed. I moved it and the method above it above the private methods.

alextran1502:
@mertalev Just FYI, This issue of `View All` places where you have partner sharing with onTimeline enabled, and the partners user doesn't have any places is still there

mertalev:
Oh right, I'll take a look at that.

PR #8050, opened 2024-03-18
refactor(server): library service

jrasm91:
* require owner id when creating a new library (technically a breaking change for the library create endpoint)
* remove unused access checks
* remove unused method `getCount`
* fix issue with deleting secondary upload libraries


danieldietzler (approved):
👏🏼

PR #8055, opened 2024-03-19
chore(server): change `save` -> `update` in asset repository

mertalev:
## Description

The `save` method updates or creates a new asset, but its uses don't actually provide the fields necessary to create a new asset. As a result, this PR reduces the scope of this method to more accurately reflect its usage. This change is spurred by the transition to Prisma, which has much stricter typing.

### DIFF
--- a/server/src/domain/repositories/asset.repository.ts
+++ b/server/src/domain/repositories/asset.repository.ts
@@ -109,6 +109,8 @@ export type AssetWithoutRelations = {
 
 export type AssetUpdateOptions = Pick<AssetWithoutRelations, 'id'> & Partial<AssetWithoutRelations>;
 
+export type AssetUpdateAllOptions = Omit<Partial<AssetWithoutRelations>, 'id'>;
+
 export interface MonthDay {
   day: number;
   month: number;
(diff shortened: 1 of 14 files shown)


server/src/domain/repositories/asset.repository.ts
```diff
@@ -90,6 +90,27 @@ export type AssetCreate = Pick<
 > &
   Partial<AssetEntity>;
 
+export type AssetWithoutRelations = {
```
jrasm91, line 93: Why did you do it like this? Why not just use the Omit<AssetEntity, 'field1' | 'field2' > directly?

jrasm91 (approved):

server/src/domain/repositories/asset.repository.ts
mertalev: This is more reusable

jrasm91: You can have the result of omit as a type. I meant more of that I didn't understand why you need to deconstruct and reconstruct it into an object instead of just referencing export NewType = Omit<...> directly.

mertalev: Oh, that was because I was messing with ways to get this type without specifying all the fields, but gave up on that. I can change it to use the Omit part directly.

danieldietzler (approved):

PR #8077, opened 2024-03-19
chore(web): add asset store unit tests

michelheusschen:
Adds unit tests for the asset store, because it contains quite some logic and it's easy to introduce bugs. Not sure if there's a better way to mock the SDK. Covers about ~70% of the asset store.


jrasm91 (approved):
Awesome

PR #8078, opened 2024-03-19
feat(docs): Changes to the Administration section.

aviv926:
This PR adds information about the system settings, although most of the information already exists within the Immich browser interface, I saw fit to document it and add more information as I saw fit.

In addition, I deleted the password-login.md page because:

1. It is mentioned on the new page together with all the system settings.
2. The FAQ [mentions ](https://immich.app/docs/FAQ#how-can-i-reset-the-admin-password) how an administrator can reset his password.

I made appropriate changes to the _redirects page that will be sent in a separate PR because it contains references to pages that will be deleted in additional PRs that I will send soon.

In addition, a short explanation was added to the Jobs page.


docs/static/_redirects
```diff
@@ -24,3 +24,4 @@
 /docs/features/user-management /docs/administration/user-management 301
 /docs/developer/contributing /docs/developer/pr-checklist 301
 /docs/guides/machine-learning /docs/guides/remote-machine-learning 301
+/docs/administration/password-login /docs/administration/system-settings 301
```
jrasm91, line 27: Reorganizing content is great, but try to remember to add a record here so that we don't end up with broken links due to those changes :pray:

jrasm91 (approved):
Looks good. Thanks!

docs/static/_redirects
jrasm91: Oh, I just read your comment in the description about this. If we have separate PRs for different sections, just include the relevant line immediately. Otherwise, as soon as the PR is merged, the links will be broken.

PR #8079, opened 2024-03-19
feat(web): keyboard access for search dropdown, combobox fixes

ben-basten:
## Description

* Entering and leaving the search bar with a keyboard will show/hide the "Recent searches" dropdown.
* Moving focusout handling into a shared action for easier reuse
* Fixing minor usability issues in the combobox as a follow-up to PR #8007.

## How Has This Been Tested?

**Search dropdown**

1. Login to immich / refresh the page
2. Use the "Tab" key to navigate over to the search box
3. Observe "Recent searches" dropdown appears
4. Continue using "Tab" key to navigate out of the search bar and automatically close the dropdown

**Combobox clear button**

1. In the advanced search options, open a combobox and use arrow keys/enter to select an option
2. Press "Tab" once to move focus to the "Clear value" button
3. Press enter to clear text. Pressing the clear button should now deactivate the combobox and hide the magnifying glass icon.

## Checklist:

- [x] npm test (unit tests)


web/src/lib/components/shared-components/search-bar/search-bar.svelte
```diff
@@ -122,11 +122,11 @@
           ? 'rounded-t-3xl border  border-gray-200 bg-white dark:border-gray-800'
           : 'rounded-3xl border border-transparent bg-gray-200'}"
         placeholder="Search your photos"
-        required
```
michelheusschen, line 125: Removing `required` is an issue, because:
1. Empty searches are included in recent searches
2. The server request fails when searching for an empty string

web/src/lib/utils/click-outside.ts
```diff
@@ -42,13 +43,21 @@ export function clickOutside(node: HTMLElement, options: Options = {}): ActionRe
     }
   };
 
+  const handleFocusOut = (event: FocusEvent) => {
+    if (onFocusOut && event.relatedTarget instanceof Node && !node.contains(event.relatedTarget as Node)) {
+      onFocusOut();
```
michelheusschen, line 48: I think it's better to create a new action

web/src/lib/components/shared-components/search-bar/search-bar.svelte
ben-basten: Good to know, thanks for catching this!

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -101,6 +101,7 @@
   };
 
   const onClear = () => {
+    deactivate();
```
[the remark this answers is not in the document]
ben-basten: Makes sense! Added the following functionality to make the clear button behave more consistently:

* handle escape keypress while the clear button is focused
* move focus to input when clear button is clicked
* leave the dropdown closed if the user has already closed the dropdown and tabs over to the clear button
* activate the combobox if a user tabs backwards onto the clear button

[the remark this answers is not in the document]
michelheusschen: > activate the combobox if a user tabs backwards onto the clear button

Thank you, works great. I'm only unsure about this point. When tabbing backwards, there can only be one option which is already shown in the input field. That makes activating the combobox a bit redundant.

web/src/lib/utils/focus-change.ts
```diff
@@ -0,0 +1,21 @@
+interface Options {
+  onFocusOut?: () => void;
+}
+
+export function focusChange(node: HTMLElement, options: Options = {}) {
```
michelheusschen, line 5: Why not make the callback required and use similar naming to `clickOutside`?

`export function focusOutside(node: HTMLElement, onFocusOut: () => void) {`

ben-basten: I was trying to keep the naming generic enough such that if an opposite `onFocusIn` handler was ever needed, the name would still fit. Changing the name to `focusOutside` for consistency works for me :+1:

web/src/lib/components/shared-components/combobox.svelte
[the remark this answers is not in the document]
ben-basten: Reverted that change!

michelheusschen (approved):

PR #8080, opened 2024-03-19
chore(server): use absolute import paths

danieldietzler:


jrasm91 (approved):

server/package.json
```diff
@@ -155,16 +155,14 @@
       "./src/domain/": {
         "branches": 75,
         "functions": 80,
-        "lines": 90,
-        "statements": 90
+        "lines": 85,
+        "statements": 85
```
danieldietzler, line 159: Let's not talk about this... I've added too many new lines because of imports.

PR #8089, opened 2024-03-20
test: temporarily disable flaky audit e2e test until #7436 is fixed

zackpollard:
Audit service has a bug #7436 that shows files as untracked when it shouldn't. This behaviour seems flaky and is causing the audit e2e test to sporadically fail. Disabling the test until we fix that bug


jrasm91 (approved):
I didn't think the todo is necessarily the cause but we can skip it either way for now.

PR #8090, opened 2024-03-20
feat(web): rework context menus: add icons and reorder items

Ethan13310:
This PR adds icons to context menus and reorders some items in both context menus and nav bars.

- Nav bars now feature the "Add to favorites" icon instead of the "Delete" one (which has been moved into the context menu).

NavBar

- Album viewer:

AlbumViewer

- Album viewer when photos are selected:

AlbumViewerPhotosSelected

- Photos page:

PhotosPage

- People context menu:

PeopleContextMenu

- Asset viewer:

Nav bar remains unchanged.

AssetViewer

AssetViewerContextMenu

### DIFF
--- a/web/src/lib/components/shared-components/context-menu/context-menu.svelte
+++ b/web/src/lib/components/shared-components/context-menu/context-menu.svelte
@@ -23,12 +23,14 @@
 <div
   transition:slide={{ duration: 200, easing: quintOut }}
   bind:this={menuElement}
-  class="absolute z-10 w-[200px] overflow-hidden rounded-lg shadow-lg"
+  class="absolute z-10 min-w-[200px] w-max max-w-[300px] overflow-hidden rounded-lg shadow-lg"
   style="left: {left}px; top: {top}px;"
   role="menu"
   use:clickOutside
   on:outclick
   on:escape
 >
-  <slot />
+  <div class="flex flex-col rounded-lg">
+    <slot />
+  </div>
 </div>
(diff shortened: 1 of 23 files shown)


web/src/lib/components/shared-components/context-menu/context-menu.svelte
```diff
@@ -23,12 +23,14 @@
 <div
   transition:slide={{ duration: 200, easing: quintOut }}
   bind:this={menuElement}
-  class="absolute z-10 w-[200px] overflow-hidden rounded-lg shadow-lg"
+  class="absolute z-10 min-w-[200px] w-max max-w-[300px] overflow-hidden rounded-lg shadow-lg"
```
martabal, line 26: I think `w-max` does not exist

web/src/lib/components/shared-components/context-menu/menu-option.svelte
```diff
@@ -9,7 +12,14 @@
...
   {#if text}
-    {text}
+    {#if icon}
+      <p class="flex gap-2">
+        <Icon path={icon} size="18" />
+        {text}
+      </p>
+    {:else}
+      {text}
+    {/if}
```
martabal, line 22: Seems weird, maybe `:else if text` would be better ?

martabal: Mmmh, it's probably better the way it is now

martabal (approved):

jrasm91 (approved):
Looks great! Id be fine in the future requiring icons if this makes sense and everything is simpler then.

danieldietzler (approved):
Nice, thanks! Those icons look good

Issue #8073, opened 2024-03-19
OAuth immich_quota claim does not updates user quota on login


RedoverPlayer:
### The bug

The immich_quota claim works on user creation, but after changing the user group in authentik (and thus the quota) it is not updated in the app on the next login. When deleting and logging in, the newly created user has the new quota so it seems the claim is only used on user creation.

### The OS that Immich Server is running on

Debian

### Version of Immich Server

v1.98.2

### Version of Immich Mobile App

v1.98.2

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
version: "3"

services:
  immich-server:
    container_name: immich_server
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    command: [ "start.sh", "immich" ]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
    env_file:
      - .env
    ports:
      - 2283:3001
    depends_on:
      - redis
      - database
    restart: always

  immich-microservices:
    container_name: immich_microservices
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    # extends:
    #   file: hwaccel.yml
    #   service: hwaccel
    command: [ "start.sh", "microservices" ]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
    env_file:
      - .env
    depends_on:
      - redis
      - database
    restart: always

  immich-machine-learning:
    container_name: immich_machine_learning
    image: ghcr.io/immich-app/immich-machine-learning:${IMMICH_VERSION:-release}
    volumes:
      - model-cache:/cache
    env_file:
      - .env
    restart: always

  redis:
    container_name: immich_redis
    image: redis:6.2-alpine@sha256:70a7a5b641117670beae0d80658430853896b5ef269ccf00d1827427e3263fa3
    restart: always

  database:
    container_name: immich_postgres
    image: tensorchord/pgvecto-rs:pg14-v0.2.0@sha256:90724186f0a3517cf6914295b5ab410db9ce23190a2d9d0b9dd6463e3fa298f0
    env_file:
      - .env
    environment:
      POSTGRES_PASSWORD: ${DB_PASSWORD}
      POSTGRES_USER: ${DB_USERNAME}
      POSTGRES_DB: ${DB_DATABASE_NAME}
    volumes:
      - pgdata:/var/lib/postgresql/data
    restart: always

volumes:
[...]

PR #8151, opened 2024-03-21
fix(docs): Immich quota claim note

aviv926:
Added following https://github.com/immich-app/immich/issues/8073


docs/docs/administration/oauth.md
```diff
@@ -59,21 +59,21 @@ Before enabling OAuth in Immich, a new client application needs to be configured
...
+| Client Secret                                        | string  | (required)           | Required. Client Secret (previous step)                                                                                   |
+| Scope                                                | string  | openid email profile | Full list of scopes to send with the request (space delimited)                                                            |
+| Signing Algorithm                                    | string  | RS256                | The algorithm used to sign the id token (examples: RS256, HS256)                                                          |
+| Storage Label Claim                                  | string  | preferred_username   | Claim mapping for the user's storage label                                                                                |
+| Storage Quota Claim                                  | string  | immich_quota         | Claim mapping for the user's storage<br />**Note:** Claim is only used on user creation and cannot be changed afterwards. |
```
jrasm91, line 71:
```suggestion
| Storage Quota Claim                                  | string  | immich_quota         | Claim mapping for the user's storage<br />**Note:** Claim is only used on user creation and not synchronized after that. |
```

aviv926:
I still don't use this option personally, but is this note also valid for other things?
Like for example Storage Label Claim?

jrasm91 (approved):

jrasm91:
> I still don't use this option personally, but is this note also valid for other things? Like for example Storage Label Claim?

Yes, all these claims are set when the user is created and not synchronized after that. Maybe we could add a [1] and then move the note below the table instead.

aviv926:
> > I still don't use this option personally, but is this note also valid for other things? Like for example Storage Label Claim?
> 
> Yes, all these claims are set when the user is created and not synchronized after that. Maybe we could add a [1] and then move the note below the table instead.

 think it would be better, I changed accordingly.

PR #8152, opened 2024-03-21
refactor: asset v1, app.utils

jrasm91:


server/package.json
```diff
@@ -156,10 +156,10 @@
...
     "coverageThreshold": {
       "./src/": {
-        "branches": 75,
-        "functions": 80,
-        "lines": 85,
-        "statements": 85
+        "branches": 70,
+        "functions": 75,
+        "lines": 80,
+        "statements": 80
```
danieldietzler, line 162: lmao

danieldietzler (approved):

server/package.json
jrasm91: haha we added in asset v1 to the coverage report and it's not tested at all.

PR #8172, opened 2024-03-21
fix(server): validation events actually throwing an error

danieldietzler:


jrasm91 (approved):
Can you add an e2e test for this?

PR #8186, opened 2024-03-22
feat(cli) CLI when uploading photo.EXT, it detects both photo.EXT.xmp and photo.xmp

kokulshan:
immich CLI now can detect XMP sidecars of the formats photo.EXT.xmp and photo.xmp for a photo named photo.EXT. If both are present, it prefers photo.EXT.xmp

### DIFF
--- a/cli/src/commands/asset.ts
+++ b/cli/src/commands/asset.ts
@@ -86,7 +86,6 @@ class Asset {
     let sidecarData: Blob | undefined = undefined;
     if (sidecarPath) {
       try {
-        await access(sidecarPath, constants.R_OK);
         sidecarData = new File([await fs.openAsBlob(sidecarPath)], basename(sidecarPath));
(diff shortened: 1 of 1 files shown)


cli/src/commands/asset.ts
```diff
@@ -61,13 +61,35 @@ class Asset {
...
+    }
+
     let sidecarData: Blob | undefined = undefined;
-    try {
-      await access(sideCarPath, constants.R_OK);
-      sidecarData = new File([await fs.openAsBlob(sideCarPath)], basename(sideCarPath));
-    } catch {}
+    if (sidecarPath) {
+      try {
+        await access(sidecarPath, constants.R_OK);
```
jrasm91, line 89: Why do we need the check again?

jrasm91 (approved):

PR #8189, opened 2024-03-22
fix(mobile): album date range while adding or removing assets

### LINKED ISSUE #7643: [BUG] (add to/remove from) album - sorting
### The bug

When a new asset is uploaded and added/removed to/from an already existing album, that albums is not sorted according sorting settings on mobile. Web interface accepts that settings (Most recent photo).
Workaround is to reload all assets from server by double swipe down on Photos page. Then albums are sorted properly.

Fedora 39

### Version of Immich Server

v.1.97.0

v.1.97.0

### Platform with the issue

- [ ] Server
- [ ] Web
- [X] Mobile

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

```bash
1. sort albums by "Most recent photo", album with the newest photo at top.
2. take a photo
3. add it to some older albums (not at the first place)
4. album stays at the original position - at top expected
```

### Additional information

_No response_

waclaw66:
Fixes https://github.com/immich-app/immich/issues/7643
Replacement of https://github.com/immich-app/immich/pull/7670


mobile/lib/shared/services/sync.service.dart
```diff
@@ -62,6 +62,15 @@ class SyncService {
...
       _lock.run(() => _syncRemoteAlbumsToDb(remote, isShared, loadDetails));
 
+  /// Syncs single remote album to the database
+  /// returns `true` if there were any changes
+  Future<bool> syncRemoteAlbumToDb(
+    AlbumResponseDto remote,
+    Album album, {
+    required FutureOr<AlbumResponseDto> Function(AlbumResponseDto) loadDetails,
+  }) =>
+      _lock.run(() => _syncRemoteAlbum(remote, album, [], [], loadDetails));
```
fyfrey, line 72: This won't handle deletes at all and result in a garbage DB state.

fyfrey (changes requested):

mobile/lib/shared/services/sync.service.dart
waclaw66: Have you tested it? It works fine for me. Where supposed to be a problem?

waclaw66:
While testing it I've found another bug https://github.com/immich-app/immich/issues/8215

mobile/lib/shared/services/sync.service.dart
fyfrey: Well, this function can also be called for shared albums. So it needs to handle the delete candidates.

waclaw66: > Well, this function can also be called for shared albums. So it needs to handle the delete candidates.

Correct, therefore I don't know what problem do you see in my implementation. I will fix it if you can help me find what's wrong.

fyfrey: Let's have a shared album A owned by user O.
User O has shared A with user U.
User U opens the app, it syncs the asset and albums from the server. O removes assets from A, U adds a new asset to A. `syncRemoteAlbum` is called for this individual album A. The deletes are not correctly handled and remain in the database of user U's app although they maybe should not.

waclaw66: If you're plannig bigger changes that comprise solving this issue, feel free to close it in favour of your rework. Anyway thanks for giving me the bigger picture.

mobile/lib/modules/album/services/album.service.dart
```diff
@@ -170,6 +170,32 @@ class AlbumService {
...
 
+  /// Checks remote album for changes,
+  /// updates the local database and returns `true` if there were any changes
+  Future<bool> refreshRemoteAlbum({
+    required AlbumResponseDto remote,
+    required Album album,
+  }) async {
+    if (!_remoteCompleter.isCompleted) {
+      // guard against concurrent calls
+      return _remoteCompleter.future;
```
fyfrey, line 181: this is the same completer used for "all remote album refresh": if this new single album refresh is running and a full request is waiting, it will at some point resolve without actually refreshing all albums.

mobile/lib/shared/services/sync.service.dart
fyfrey: we could run a cleanup job once per day (or delayed on app start) to find assets not accessible by the logged-in user.

however, I just found another issue with the single album refresh :/ unless we find a solution, I'd propose to change this PR back into a draft.
In any case: Thank you for your work! Also for identifying that new bug...

mobile/lib/modules/album/services/album.service.dart
fyfrey: Thinking some more about it: we can remove the completer check in this method.

The sync service still has the lock. So running both all album refresh and an individual album refresh should be ok.

mobile/lib/shared/services/sync.service.dart
waclaw66: > Let's have a shared album A owned by user O. User O has shared A with user U. User U opens the app, it syncs the asset and albums from the server. O removes assets from A, U adds a new asset to A. `syncRemoteAlbum` is called for this individual album A. The deletes are not correctly handled and remain in the database of user U's app although they maybe should not.

It's not only update among multiple users, even updates from different clients of the same users do that. When an asset is added into album on web mobile doesn't sync album dates. Maybe album versioning would help. Each album change would raise album version number and clients can easily recognize changes and sync album no matter asset was added or deleted.

fyfrey:
superseded by #8324

PR #8190, opened 2024-03-22
refactor(cli): crawl service

jrasm91:
* use defaults from src/index.ts
* make crawl a normal function
* simplify


danieldietzler (approved):
LGTM

PR #8204, opened 2024-03-22
refactor: server events

jrasm91:
* rename communication to event
* use event emitter instead of callbacks for config update (cross-process event)
* user event emitter instead of callbacks for websocket connection event (in-process event)


server/src/repositories/event.repository.ts
```diff
@@ -25,14 +24,8 @@ import { ImmichLogger } from 'src/utils/logger';
   path: '/api/socket.io',
   transports: ['websocket'],
 })
-export class CommunicationRepository
-  implements OnGatewayConnection, OnGatewayDisconnect, OnGatewayInit, ICommunicationRepository
-{
-  private logger = new ImmichLogger(CommunicationRepository.name);
```
jrasm91, line 31: This is all basically gone - no more need to store, register, and invoke callbacks.

danieldietzler (approved):
LGTM!

PR #8210, opened 2024-03-23
chore(server): change upsert signature for search repo

mertalev:
## Description

The smart info table isn't actually used anymore, so this PR changes the upsert signature to be specific to the smart search table. This change is split off from the Prisma PR.


danieldietzler (approved):
LGTM

jrasm91 (approved):

PR #8228, opened 2024-03-23
feat(server): near-duplicate detection

mertalev:
## Description

This PR adds a new job to detect duplicate assets and aggregate them with a new `duplicateId` column. This PR only implements the backend for duplicate detection. It does not expose the results in the UI or take any actions relating to the assets: this is left for future work.

The data model is such that each `(duplicateId, assetId)` pair uniquely identifies a duplicate asset and each `duplicateId` can have many associated assets.

To do:
- [x] Handle edge case where multiple `duplicateIds` exist among the found duplicates
- [X] Better handling of concurrency
  - Disabled concurrency to avoid race conditions and improve accuracy
- [x] Confirm correctness of the results
- [x] Tune default threshold
- [x] Add migration
- [x] Add tests

Implements #1968

## How Has This Been Tested?

Tested by running the new job on all assets through the job panel and inspecting logs to confirm that some assets have duplicates.

Tested that the duplicates displayed in the web view are actually near-duplicates.

Tested that changing the duplicate threshold changes the strictness of the results.


alextran1502:
Can I help with anything regarding this PR, I am happy to work on UI

mertalev:
I cleaned it up so the backend part is essentially good to go (might need to adjust the response if you want them to be grouped by duplicates and not just sorted). The UI... has a lot of room for improvement haha. It'd be great if you could help with that 😄

klejejs:
In terms of UI, would it make sense if photo stacks were automatically created for near-duplicate photos? It's something that the App-Which-Must-Not-Be-Named introduced a while ago and I personally find it very useful.

mertalev:
The idea right now is to have the duplicates displayed in a dedicated page where there are options to convert them to stacks or deduplicate based on some criteria, but otherwise treat them as separate assets until the user elects to do this. Auto-stacking would be very useful and convenient, though, so a later PR to add this functionality would be nice.

AngelaDMerkel:
> Auto-stacking would be very useful and convenient, though, so a later PR to add this functionality would be nice.

Is it possible to have some kind of notification so that the user can interact with stacking? It's nice when images are stacked automatically, but I find that this sometimes occurs erroneously and I'd like to at least know when a stack has been made.

mertalev:
A notification in the web UI would be straightforward. But auto-stacking would be a later addition, so discussion on that is a bit out of scope for this PR.

PathToLife:
Thanks for working on this! It's awesome to read through the implementation here.

Just wanted to add my results, the 0.2 threshold (barely) didn't detect duplicates in my resized image test case.
I added in some console logs and set the distance to 1.0, to see `0.02084` as the distance 😭🤦‍♂️

Is it worth perhaps working on a multi alogrithm implementation? Experiementation shows it excels at resizing. I might have time soonish.

```
"duplicateId": "79b47ba0-28e5-479b-b745-9f2885299077",
immich_microservices     |       "assetId": "c8c4d6e9-8d72-4e55-be22-20b9019ffca6",
immich_microservices     |       "distance": 0.020842433
```

I was testing resized images here (download to reproduce):

[images-dupe-test.zip](https://github.com/immich-app/immich/files/15208343/images-dupe-test.zip)





```ts
async searchDuplicates({
    assetId,
    embedding,
    maxDistance,
    userIds,
  }: AssetDuplicateSearch): Promise<AssetDuplicateResult[]> {
    maxDistance = 1;
    this.logger.warn('searching duplicates', { assetId, maxDistance, userIds });
    const cte = this.assetRepository.createQueryBuilder('asset');
    cte
      .select('search.assetId', 'assetId')
      .addSelect('asset.duplicateId', 'duplicateId')
      .addSelect(`search.embedding <=> :embedding`, 'distance')
      .innerJoin('asset.smartSearch', 'search')
      .where('asset.ownerId IN (:...userIds )')
      .andWhere('asset.id != :assetId')
      .andWhere('asset.isVisible = :isVisible')
      .orderBy('search.embedding <=> :embedding')
      .limit(64)
      .setParameters({ assetId, embedding: asVector(embedding), isVisible: true, userIds });

    const builder = this.assetRepository.manager
      .createQueryBuilder()
      .addCommonTableExpression(cte, 'cte')
      .from('cte', 'res')
      .select('res.*')
      .where('res.distance <= :maxDistance', { maxDistance });

    const results = (await builder.getRawMany()) as any as Promise<AssetDuplicateResult[]>;

    this.logger.warn('found duplicates', { results });

    return results;
  }
```

image

mertalev:
The duplicate threshold will be exposed in the admin settings. I was debating between defaulting to 0.02 or 0.03, so maybe 0.03 is the better default after all.

NicholasFlamy:
> In terms of UI, would it make sense if photo stacks were automatically created for near-duplicate photos? It's something that the App-Which-Must-Not-Be-Named introduced a while ago and I personally find it very useful.

This

> > Auto-stacking would be very useful and convenient, though, so a later PR to add this functionality would be nice.
> 
> Is it possible to have some kind of notification so that the user can interact with stacking? It's nice when images are stacked automatically, but I find that this sometimes occurs erroneously and I'd like to at least know when a stack has been made.

And this, will be separate. Once this is implemented, these features can be worked on. So they will basically be built off of the code in this feature.

NicholasFlamy:
I would suggest taking some inspiration from Samsung Gallery for the UI. When you hit delete duplicates it selects all but one of each of the duplicates (so if there is 3 it will select 2). I'm pretty sure it takes date modified or something and if they are different resolutions or file sizes it selects the lower resolution or filesize. Then you can hit delete with all of the duplicates selected.

I think implementing something similar wouldn't be too difficult, and it doesn't even have to select for you but the side-by-side view is the most important thing. Having a button to select duplicates which could prefer selecting the lower resolution/filesize would be an added bonus.

Screenshot_20240507_110313_Gallery
Screenshot_20240507_110337_Gallery
Screenshot_20240507_110457_Gallery
Screenshot_20240507_110352_Gallery

FYI the Testing Immich Album has photos I copied over for testing immich and only select this album in the mobile app to protect my photos from bugs etc., so they are duplicates.

NicholasFlamy:
> Hey there, amazing work on this PR! Just a thought - how about we get rid of those JPEG duplicates when we've got the original HEIC files? What's your take on this?

If the files are basically identical then this should pick that up. If you're suggesting that it automatically prefer HEIC, that's for the UI which is coming eventually.

NicholasFlamy:
@mertalev what do you think?
Screenshot_20240507_185727_Chrome

I haven't done much but at least you can get out of there.

mertalev:
Nice! I'll reduce the scope of this PR to just be the backend changes so we can do the UI separately.

mertalev:
After removing the UI changes, this PR is ready for review. The current behavior is that it will run duplicate detection jobs internally, but won't expose this feature to the user. The only blocker is that a seemingly unrelated E2E test is failing.

NicholasFlamy:
Another complement about this functionality. Since it's AI based, it picks up 2 different pictures taken directly after one another at slightly different angles or distances. So when I take multiple pictures just in case one of them is blurry bur then later have a bunch of extra, this should be the solution. 

Long screenshot:
Screenshot_20240508_082317_Chrome

Ignore the sidebar, I tapped the button which scrolls down and adds to the screenshot and it did that.

AngelaDMerkel:
In addition to hashing, the exif spec contains a field for `OriginalFileName` which could be used to match duplicates created from an original. A lot of software writes this field and would resolve the need to determine whether heic or jpeg (for example) is the original

NicholasFlamy:
> In addition to hashing, the exif spec contains a field for `OriginalFileName` which could be used to match duplicates created from an original. A lot of software writes this field and would resolve the need to determine whether heic or jpeg (for example) is the original

I think that would be saved for later. For the UI. Alex and I discussed UI development and Alex will develop most of it but I'll try and start on it this week. We are thinking of a  Utilities page which has the deduplication page. I am taking not of what you said, I'm not sure how much logic will go into the deduplication page but that seems like a good idea.

server/src/services/search.service.ts
```diff
@@ -144,6 +156,97 @@ export class SearchService {
     }
   }
 
+  async handleQueueSearchDuplicates({ force }: IBaseJob): Promise<JobStatus> {
+    const { machineLearning } = await this.configCore.getConfig();
+    if (!machineLearning.enabled || !machineLearning.clip.enabled) {
```
danieldietzler, line 161:
```suggestion
    if (await this.configCore.hasFeature(FeatureFlag.SMART_SEARCH)) {
```

server/src/services/search.service.ts
```diff
@@ -144,6 +156,97 @@ export class SearchService {
...
+        assets.map((asset) => ({ name: JobName.DUPLICATE_DETECTION, data: { id: asset.id } })),
+      );
+    }
+
+    return JobStatus.SUCCESS;
+  }
+
+  async handleSearchDuplicates({ id }: IEntityJob): Promise<JobStatus> {
+    const { machineLearning } = await this.configCore.getConfig();
+    if (!machineLearning.enabled || !machineLearning.duplicateDetection.enabled) {
```
danieldietzler, line 182: Do you think it would be worth introducing a `DUPLICATE_DETECTION` feature flag?

danieldietzler (commented):
I don't understand why that test is failing either

server/src/dtos/model-config.dto.ts
```diff
@@ -25,6 +27,15 @@ export class CLIPConfig extends ModelConfig {
   mode?: CLIPMode;
 }
 
+export class DuplicateDetectionConfig extends TaskConfig {
+  @IsNumber()
+  @Min(0.01)
+  @Max(0.1)
```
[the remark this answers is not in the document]
NicholasFlamy: I'm not mertalev, but from testing I can say 0.1 is too high. So high enough maximum from my testing.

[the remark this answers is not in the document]
mertalev: Yeah, after a certain point it stops being duplicate detection and starts being a general "cluster my images" page.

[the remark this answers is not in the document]
NicholasFlamy: Here's an example where it's basically just detecting similar photos rather than duplicates/near-duplicates with the setting at 0.1:
Screenshot_20240509_123846_Chrome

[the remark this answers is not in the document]
danieldietzler: Ok nice, thanks for the explanation :)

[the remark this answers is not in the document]
nebriv: The clustering could be really neat as a separate feature down the road!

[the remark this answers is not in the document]
NicholasFlamy: > The clustering could be really neat as a separate feature down the road!

That's the plan, it would work the same way this does. I saw something about it and was thinking of helping make that feature from this one.

server/src/services/search.service.ts
NicholasFlamy: I added a flag in my test branch. I think we should and I think mertalev agrees with it.

server/src/services/search.service.ts
NicholasFlamy:
```suggestion
    if (await this.configCore.hasFeature(FeatureFlag.DUPLICATE_DETECTION)) {
```
I'm going to also make some reviews adding the feature flag if I can do that here.

server/src/services/search.service.ts
NicholasFlamy:
```suggestion
    if (await this.configCore.hasFeature(FeatureFlag.DUPLICATE_DETECTION)) {
```
Still need to add the flags (I already did and tested it on a fork.

server/src/services/search.service.ts
danieldietzler: @NicholasFlamy your change is wrong here.

open-api/immich-openapi-specs.json
```diff
@@ -7519,6 +7558,10 @@
       },
       "CLIPConfig": {
         "properties": {
+          "duplicateThreshold": {
+            "format": "float",
+            "type": "number"
+          },
```
NicholasFlamy, line 7564: Probably shouldn't be under CLIPConfig??? or is this still the case? Since you separated it from CLIP a little.

mertalev: Oh, I forgor to update the API

server/src/services/search.service.ts
mertalev: The condition needs to be inverted, right?

server/src/services/job.service.ts
```diff
@@ -116,6 +116,11 @@ export class JobService {
         return this.jobRepository.queue({ name: JobName.QUEUE_SMART_SEARCH, data: { force } });
       }
 
+      case QueueName.DUPLICATE_DETECTION: {
+        await this.configCore.requireFeature(FeatureFlag.SMART_SEARCH);
```
NicholasFlamy, line 120:
```suggestion
        await this.configCore.requireFeature(FeatureFlag.DUPLICATE_DETECTION);
```

NicholasFlamy: Needs my PR which I will make soon.

server/src/services/search.service.ts
NicholasFlamy: Crap, I see what I did. Yes I think so.

server/src/services/search.service.ts
NicholasFlamy: I think this is also wrong, I think it needs to be inverted.

server/src/services/search.service.spec.ts
```diff
@@ -95,4 +106,188 @@ describe(SearchService.name, () => {
       expect(result).toEqual(expectedResponse);
     });
   });
+
+  describe('handleQueueSearchDuplicates', () => {
+    it('should skip if machine learning is disabled', async () => {
+      configMock.load.mockResolvedValue([{ key: SystemConfigKey.MACHINE_LEARNING_ENABLED, value: false }]);
```
NicholasFlamy, line 112: Will likely need to be changed after the other changes.

server/src/services/search.service.ts
danieldietzler: Oh yeah sorry that is right. Fixed :)

NicholasFlamy (commented):
I'm going to put my suggestions in a PR to this PR's branch.

server/src/services/search.service.ts
NicholasFlamy: Huh?

NicholasFlamy: I'll test it on my test branch.

NicholasFlamy: I don't know if it's inverted but regardless it's wrong. It removes the constant `machineLearning`.

server/src/services/search.service.ts
NicholasFlamy: Is the purpose of the changes to this file so that you can search for duplicate files?

NicholasFlamy: I think it's for searching for the Duplicate?

NicholasFlamy: Figured it out.

server/src/services/search.service.ts
danieldietzler: Yes that was the idea, as it isn't used afaict.

NicholasFlamy: > Yes that was the idea, as it isn't used afaict.

It's used below for the `maxDistance`.

danieldietzler: I think I am just blind lmao. Could you link the line for me please? :)

server/src/services/search.service.ts
```diff
@@ -144,6 +156,97 @@ export class SearchService {
...
+
+    if (!asset.smartSearch?.embedding) {
+      this.logger.debug(`Asset ${id} is missing embedding`);
+      return JobStatus.FAILED;
+    }
+
+    const duplicateAssets = await this.searchRepository.searchDuplicates({
+      assetId: asset.id,
+      embedding: asset.smartSearch.embedding,
+      maxDistance: machineLearning.duplicateDetection.maxDistance,
```
NicholasFlamy, line 210: Okay, so it's only used here, so we could in fact get rid of the first one.

zackpollard (approved):
LGTM

server/src/entities/asset.entity.ts
```diff
@@ -165,6 +165,10 @@ export class AssetEntity {
 
   @OneToOne(() => AssetJobStatusEntity, (jobStatus) => jobStatus.asset, { nullable: true })
   jobStatus?: AssetJobStatusEntity;
+
+  @Index('IDX_assets_duplicateId')
+  @Column({ type: 'uuid', nullable: true })
+  duplicateId!: string | null;
```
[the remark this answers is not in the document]
zackpollard: We already have stacking separately to duplicate ids, this id is simply to allow for a future UI that allows the user to go through their images and find duplicates to remove them.

danieldietzler (approved):
Code looks good to me! Awesome :)

server/src/repositories/asset.repository.ts
```diff
@@ -207,6 +208,7 @@ export class AssetRepository implements IAssetRepository {
     let builder = this.repository.createQueryBuilder('asset');
     builder = searchAssetBuilder(builder, options);
     builder.orderBy('asset.createdAt', options.orderDirection ?? 'ASC');
+    builder.leftJoin('asset.smartSearch', 'smartSearch');
```
mertalev, line 211: I think the intent behind this inner join was to only queue assets with smart search embeddings. It doesn't do anything as a left join so this can be removed.

NicholasFlamy:
Man @mertalev I'm sorry for bringing this upset so late but I just checked #1968 and someone commented a really good point but was asking about the community project:
https://github.com/immich-app/immich/discussions/1968#discussioncomment-9432502

> for duplicates that are deleted, is there some kind of block to stop these from being re-uploaded (ie through mobile auto backup)? Or the immich db can somehow retain the record of that particular upload to flag future uploads as duplicates?

This is actually a really good idea, but how would something like this be implemented? There would have to be a blacklist of some kind and preferably a way to remove items from the blacklist. If deleting files on the web could delete them on mobile this wouldn't be a problem. But I'm worried that in the current state of immich, deleting a duplicate on the server that was backed up from the phone, will be reuploaded from the phone.

jrasm91 (approved):
I would prefer to see this moved to a `DuplicateController` and `DuplicateService` and served on the `/duplicates` route, but since it isn't used in the UI yet we can do this in a follow-up pull request.

zackpollard:
> Man @mertalev I'm sorry for bringing this up so late but I just checked #1968 and someone commented a really good point but was asking about the community deduplication project (but this prompted me to realize this could be an issue for our (by our I mean this PR) implementation) :
> https://github.com/immich-app/immich/discussions/1968#discussioncomment-9432502
> 
> > for duplicates that are deleted, is there some kind of block to stop these from being re-uploaded (ie through mobile auto backup)? Or the immich db can somehow retain the record of that particular upload to flag future uploads as duplicates?
> 
> This is actually a really good idea, but how would something like this be implemented? There would have to be a blacklist of some kind and preferably a way to remove items from the blacklist. If deleting files on the web could delete them on mobile this wouldn't be a problem. But I'm worried that in the current state of immich, deleting a duplicate on the server that was backed up from the phone, will be reuploaded from the phone. 

Imo this is a separate ongoing issue we need to address separately as it's impact is wider than just duplicates.

NicholasFlamy:
> Imo this is a separate ongoing issue we need to address separately as it's impact is wider than just duplicates.

Yeah, I guess so.

bo0tzz:
> deleting a duplicate on the server that was backed up from the phone, will be reuploaded from the phone.

I believe currently the mobile app will not reupload files that it has already uploaded in the past (but I could be wrong). If you reinstalled the app or something like that it still would though.

NicholasFlamy:
> > deleting a duplicate on the server that was backed up from the phone, will be reuploaded from the phone.
> 
> I believe currently the mobile app will not reupload files that it has already uploaded in the past (but I could be wrong). If you reinstalled the app or something like that it still would though.

Even then, you'd have the duplicates sitting on your phone and if you reinstalled the app or switched phones and copied everything over then all of a sudden the photos reappear. But that's a different problem because it doesn't just affect duplicates.

zackpollard:
> > > deleting a duplicate on the server that was backed up from the phone, will be reuploaded from the phone.
> > 
> > I believe currently the mobile app will not reupload files that it has already uploaded in the past (but I could be wrong). If you reinstalled the app or something like that it still would though.
> 
> Even then, you'd have the duplicates sitting on your phone and if you reinstalled the app or switched phones and copied everything over then all of a sudden the photos reappear. But that's a different problem because it doesn't just affect duplicates. 

We have a plan to sync deletions to synced devices, we will likely be delivering that before stable

ThreeAurora:
I'm looking forward to this feature, it's going to be a highlight of immich.

cuu508:
immich v1.106.1 was just released, the release includes this feature.

PR #8239, opened 2024-03-24
feat(server): add immich cli to container

danieldietzler:
In contrary to what was removed in #8224, this will always install the latest version from npm. By this, the cli will actually always be up to date (with the latest published version)


jrasm91 (approved):

martabal:
Couldn't this be an issue if a new version of the CLI is released at the same time as the server?

danieldietzler:
> Couldn't this be an issue if a new version of the CLI is released at the same time as the server?

The idea is to always publish the cli first, and then build the server afaik.

martabal:
> The idea is to always publish the cli first, and then build the server afaik.

Sounds good

PR #8255, opened 2024-03-25
feat(server): job metrics

mertalev:
## Description

This PR adds observability for jobs and queues. Specifically, it tracks job successes, skips and failures as well as the current number of active jobs in each queue. This is implemented in the job service through a new metric repository.

## How Has This Been Tested?

Tested that the active jobs in the queue show up in Prometheus and respond to changes in concurrency.

thumbnail-generation-active

### DIFF
--- a/server/src/interfaces/metric.interface.ts
+++ b/server/src/interfaces/metric.interface.ts
@@ -1,0 +1,13 @@
+import { MetricOptions } from '@opentelemetry/api';
+
+export interface CustomMetricOptions extends MetricOptions {
+  enabled?: boolean;
+}
+
+export const IMetricRepository = 'IMetricRepository';
+
+export interface IMetricRepository {
+  addToCounter(name: string, value: number, options?: CustomMetricOptions): void;
+  updateGauge(name: string, value: number, options?: CustomMetricOptions): void;
+  updateHistogram(name: string, value: number, options?: CustomMetricOptions): void;
+}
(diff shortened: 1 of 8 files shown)


alextran1502 (approved):

server/src/interfaces/metric.interface.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface CustomMetricOptions extends MetricOptions {
+  enabled?: boolean;
+}
+
+export const IMetricRepository = 'IMetricRepository';
+
+export interface IMetricRepository {
+  addToCounter(name: string, value: number, options?: CustomMetricOptions): void;
+  updateGauge(name: string, value: number, options?: CustomMetricOptions): void;
+  updateHistogram(name: string, value: number, options?: CustomMetricOptions): void;
```
jrasm91, line 12: Requiring passing in whether metrics is enabled or not send like an odd choice. Also making it optional send really weird because it doesn't do anything otherwise. Can you just move that logic globally to the repo instead? If anything add a method to enable or disable metrics, instead of requiring it in every call.

mertalev: The reason is because there are different kinds of metrics and it can be that only some are enabled. The `enabled` flag is just for ergonomics to avoid repeating `if (jobMetrics)`, etc.

I could change it to accept the metric group and have the repo own the decision to track the metric based on the group that's passed. How does that sound?

jrasm91: Isn't it being enabled a static property set once via an environment variable? Why would you need to pass it into the repo over and over? Id be fine with using it directly in the repository instead of requiring it to be passed in every method.

PR #8270, opened 2024-03-25
feat(web): better UX when creating a new album

Ethan13310:
This PR fixes something if find annoying:

Imagine you are sorting your thousands of photos into albums. You select `Add to... > Album`, then using the modal, you create a new album. You will be redirected to this album without being prompted. Now you have to go back and find yourself at the top of the "Photos" page, and have scroll again thousands of photos to where you previously were.

To fix this, a notification is now sent when adding assets to a new album, with a "View Album" button:

ViewAlbum

I also updated the search function to ignore accents:

SearchAccents

### DIFF
--- a/web/src/lib/components/shared-components/notification/notification.ts
+++ b/web/src/lib/components/shared-components/notification/notification.ts
@@ -30,13 +30,6 @@ export type Notification = {
 type DiscardAction = { type: 'discard' };
 type NoopAction = { type: 'noop' };
 
-export type LinkAction = {
-  type: 'link';
-  target: string;
-  newTab?: boolean;
-  button?: string;
-};
-
 export type NotificationAction = DiscardAction | NoopAction;
 
 export type NotificationOptions = Partial<Exclude<Notification, 'id'>> & { message: string };
--- a/web/src/lib/components/shared-components/notification/notification.ts
+++ b/web/src/lib/components/shared-components/notification/notification.ts
@@ -18,7 +18,14 @@ export type Notification = {
 
 type DiscardAction = { type: 'discard' };
 type NoopAction = { type: 'noop' };
-type LinkAction = { type: 'link'; target: string };
+
+export type LinkAction = {
+  type: 'link';
+  target: string;
+  newTab?: boolean;
+  button?: string;
+};
+
 export type NotificationAction = DiscardAction | NoopAction | LinkAction;
 
 export type NotificationOptions = Partial<Exclude<Notification, 'id'>> & { message: string };
--- a/web/src/lib/components/shared-components/notification/notification.ts
+++ b/web/src/lib/components/shared-components/notification/notification.ts
@@ -10,6 +10,11 @@ export type Notification = {
   id: number;
   type: NotificationType;
   message: string;
+  /**
+   * Allow HTML to be inserted within the message. Make sure to verify/encode
+   * variables that may be interpoalted into 'message'
+   */
+  html?: boolean;
   /** The action to take when the notification is clicked */
   action: NotificationAction;
   /** Timeout in miliseconds */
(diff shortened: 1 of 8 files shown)


jrasm91:
The way this pattern works in google photos is a notification is shown, which has a button/link you can click to "View X". Our actions already have the ability to perform and action when they are clicked, so maybe that would be a better implementation.

Ethan13310:
> The way this pattern works in google photos is a notification is shown, which has a button/link you can click to "View X". Our actions already have the ability to perform and action when they are clicked, so maybe that would be a better implementation.

That would indeed be way better, less intrusive. However it seems notification cards cannot display a button (unlike Google Photos' ones), they can only be clickable, without any obvious hint to the user. I'll add a way do display a button then.

waclaw66:
It would be nice to se new album name instead of "to new album" in the toast as well.

Added 11 assets into **My new album**
[Show album]

Ethan13310:
I edited the PR description. The notification card now looks like this when a button is added:

NotifCardButtons

To do this:

```ts
notificationController.show({
    // ...
    action: {
      type: 'link',
      button: 'Button Text',
      target: '/route/stuff'
    }
  });
```

If the `button` property is omitted, then the notification is clickable like it was before.

Ethan13310:
> It would be nice to se new album name instead of "to new album" in the toast as well.
> 
> Added 11 assets into **My new album** [Show album]

Like this?

NotifWithAlbumName

jrasm91:
Looks great but can it be View Album instead?

Ethan13310:
> Looks great but can it be View Album instead?

Yep, done!

web/src/lib/components/shared-components/notification/notification.ts
```diff
@@ -18,7 +23,14 @@ export type Notification = {
...
 type NoopAction = { type: 'noop' };
-type LinkAction = { type: 'link'; target: string };
+
+export type LinkAction = {
+  type: 'link';
+  target: string;
+  newTab?: boolean;
+  button?: string;
+};
+
```
jrasm91, line 33: It turns out there isn't actually any existing usage of the previous link functionality I think it was added for one scenario that has since been removed. What do you think about doing this instead?
- Remove the legacy link action entirely
- Add another option to show interface to add a button to the notification
```ts
  button?: NotificationButton // (type { text: string, onClick: () => void })
```

Ethan13310: Good idea!

web/src/lib/utils/asset-utils.ts
```diff
@@ -36,10 +37,11 @@ export const addAssetsToAlbum = async (albumId: string, assetIds: string[]) => {
...
         : `Asset${assetIds.length === 1 ? ' was' : 's were'} already part of the album`,
-    action: {
-      type: 'link',
-      button: 'View Album',
-      target: `${AppRoute.ALBUMS}/${albumId}`,
+    button: {
+      text: 'View Album',
+      onClick() {
+        return goto(`${AppRoute.ALBUMS}/${albumId}`);
+      },
```
jrasm91, line 44: Nice! This looks pretty clean to me. Although you could also do:
```ts
onClick: () => goto(...)
```

web/src/lib/components/shared-components/notification/notification.ts
```diff
@@ -6,20 +6,38 @@ export enum NotificationType {
...
-type LinkAction = { type: 'link'; target: string };
-export type NotificationAction = DiscardAction | NoopAction | LinkAction;
+
+export type LinkAction = {
+  type: 'link';
+  target: string;
+  newTab?: boolean;
+  button?: string;
+};
+
```
jrasm91, line 39: Can this be deleted no w?

jrasm91 (approved):
LGTM. I can try to test it out later unless someone picks it up first.

alextran1502 (approved):

PR #8278, opened 2024-03-25
refactor(server): metric repo

mertalev:
## Description

This PR changes the metric repo to work more like the access repo, faceted across several metric groups. This is cleaner and moves ownership of enabling/disabling metrics to the repo rather than the caller.


jrasm91 (approved):
LGTM! Thanks

PR #8329, opened 2024-03-28
refactor(server): extract add/remove assets logic to utility function

danieldietzler:


jrasm91 (approved):
Looks really nice! Thanks!

server/src/cores/access.core.ts
```diff
@@ -87,7 +87,7 @@ export class AccessCore {
   async checkAccess(auth: AuthDto, permission: Permission, ids: Set<string> | string[]) {
     const idSet = Array.isArray(ids) ? new Set(ids) : ids;
     if (idSet.size === 0) {
-      return new Set();
+      return new Set<string>();
```
jrasm91, line 90: Can you revert this change?

server/src/repositories/album.repository.ts
```diff
@@ -261,7 +255,7 @@ export class AlbumRepository implements IAlbumRepository {
   }
 
   @GenerateSql({ params: [{ albumId: DummyValue.UUID, assetIds: [DummyValue.UUID] }] })
-  async addAssets({ albumId, assetIds }: AlbumAssets): Promise<void> {
+  async addAssetIds(albumId: string, assetIds: string[]): Promise<void> {
```
jrasm91, line 258: You need to change the decorator to match

server/src/services/album.service.ts
```diff
@@ -164,33 +164,16 @@ export class AlbumService {
...
-    }
+    const results = await addAssets(
+      auth,
+      { accessRepository: this.accessRepository, repository: this.albumRepository },
+      { id, assetIds: dto.ids },
+    );
 
     const newAssetIds = results.filter(({ success }) => success).map(({ id }) => id);
     if (newAssetIds.length > 0) {
-      await this.albumRepository.addAssets({ albumId: id, assetIds: newAssetIds });
```
jrasm91, line 193: It might be better to use find instead the way it would abort as soon as possible instead of going through the whole list.

server/src/services/album.service.ts
```diff
@@ -206,31 +189,14 @@ export class AlbumService {
...
-    }
+    const results = await removeAssets(
+      auth,
+      { accessRepository: this.accessRepository, repository: this.albumRepository },
+      { id, assetIds: dto.ids, permissions: [Permission.ASSET_SHARE, Permission.ALBUM_REMOVE_ASSET] },
+    );
 
     const removedIds = results.filter(({ success }) => success).map(({ id }) => id);
     if (removedIds.length > 0) {
-      await this.albumRepository.removeAssets(id, removedIds);
```
jrasm91, line 233: Same here

server/src/repositories/album.repository.ts
```diff
@@ -203,7 +197,7 @@ export class AlbumRepository implements IAlbumRepository {
 
   @GenerateSql({ params: [DummyValue.UUID, [DummyValue.UUID]] })
   @Chunked({ paramIndex: 1 })
-  async removeAssets(albumId: string, assetIds: string[]): Promise<void> {
```
mertalev, line 206: Nit: I think I like `addAssets`/`removeAssets` better than `addAssetIds`/`removeAssetIds`.

server/src/utils/asset.util.ts
```diff
@@ -0,0 +1,91 @@
...
+      results.push({ id: assetId, success: false, error: BulkIdErrorReason.NO_PERMISSION });
+      continue;
+    }
+
+    results.push({ id: assetId, success: true });
+  }
+
+  const newAssetIds = results.filter(({ success }) => success).map(({ id }) => id);
+  if (newAssetIds.length > 0) {
+    await repository.addAssetIds(dto.id, dto.assetIds);
```
jrasm91, line 44: Is this a bug? Surely it should be the new asset IDs.

server/src/utils/asset.util.ts
```diff
@@ -0,0 +1,91 @@
...
+  addAssetIds: (id: string, assetIds: string[]) => Promise<void>;
+  removeAssetIds: (id: string, assetIds: string[]) => Promise<void>;
+}
+
+export const addAssets = async (
+  auth: AuthDto,
+  repositories: { accessRepository: IAccessRepository; repository: IBulkAsset },
+  dto: { id: string; assetIds: string[] },
+) => {
+  const { accessRepository, repository } = repositories;
```
mertalev, line 18: Nit: you can destructure the parameter directly.

server/src/utils/asset.util.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const newAssetIds = results.filter(({ success }) => success).map(({ id }) => id);
+  if (newAssetIds.length > 0) {
+    await repository.addAssetIds(dto.id, dto.assetIds);
+  }
+
+  return results;
+};
+
+export const removeAssets = async (
```
mertalev, line 50: I imagine these set operations have the potential to get expensive with huge libraries. Albums will generally have manageable sizes, but if we intend to reuse this elsewhere, we should be mindful of how many assets this might be handling and how often it might get called.

mertalev (approved):
LGTM!

server/src/utils/asset.util.ts
mertalev: Ooh, great catch

server/src/repositories/album.repository.ts
jrasm91: There is another one - get assets - which returns IDs. All three methods operate on IDs so it is a little weird to have them names differently imo.

mertalev: Ah, fair. `getAssetIds` is definitely better to distinguish between fetching the ID and the entity, and it's better for them to have consistent naming in that case.

jrasm91: Yeah that was my thought process as well.

server/src/utils/asset.util.ts
danieldietzler: I know, but I didn't want to cluster the parameters even more lol

server/src/utils/asset.util.ts
danieldietzler: I think - at least in the near future - the only use cases are stacks and memories. Both shouldn't be too large, right?

server/src/utils/asset.util.ts
danieldietzler: Dang you're good

server/src/utils/asset.util.ts
danieldietzler: You would prefer destructuring them directly?

mertalev: Personal preference! I only destructure at the top of a function body if doing it in the parameter will blow it up to multiple lines.

PR #8330, opened 2024-03-28
feat: persistent memories

jrasm91:
Dedicated endpoint, service, repository, and database entity for memories, with the goal to make it possible to share, save, and view in the future.

* create a new memory
* update an existing memory (mark is as saved, seenAt)
* delete a memory
* add assets to a memory
* remove assets from a memory

Other thoughts:
- it seems like we need a way to sort them, so there is a `memoryAt` date field to track the "date" the memory represents. Memories are sorted descending by default so newest memories appear first
- need to add a way to filter memories so "memory lane" is a cross section for the current time period
- need to add a job to auto generate memories based on the current query (and potentially expand this in the future)
- need to update the UI to all the user to "save" or "discard" a memory ( soft? delete)
- need to add a job to automatically delete memories after X days.


server/src/services/memory.service.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  describe('removeAssets', () => {
+    it('should require memory access', async () => {
+      await expect(sut.removeAssets(authStub.admin, 'not-found', { ids: ['asset1'] })).rejects.toBeInstanceOf(
+        BadRequestException,
+      );
+      expect(memoryMock.removeAssetIds).not.toHaveBeenCalled();
+    });
+
+    it('should skips assets not in the memory', async () => {
```
danieldietzler, line 187:
```suggestion
    it('should skip assets not in the memory', async () => {
```

danieldietzler (approved):
Nice job! :)

server/src/services/memory.service.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+        expect.arrayContaining([
+          expect.objectContaining({ id: 'memory1', assets: expect.any(Array) }),
+          expect.objectContaining({ id: 'memoryEmpty', assets: [] }),
+        ]),
+      );
+    });
+
+    it('should map ', async () => {
+      await expect(sut.search(authStub.admin)).resolves.toEqual([]);
+    });
```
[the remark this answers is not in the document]
jrasm91: It should map... Oh wait I can't be bothered to test the mapper...

server/src/entities/memory.entity.ts
```diff
@@ -0,0 +1,67 @@
...
+  @Column({ default: false })
+  isSaved!: boolean;
+
+  /** memories are sorted in ascending order by this value */
+  @Column({ type: 'timestamptz' })
+  memoryAt!: Date;
+
+  /** when the user last viewed the memory */
+  @Column({ type: 'timestamptz', nullable: true })
+  seenAt?: Date;
```
mertalev, line 62: If we want these memories to be shared, do we want this to be handled at a user level?

server/src/entities/memory.entity.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  @Column()
+  type!: T;
+
+  @Column({ type: 'jsonb' })
+  data!: MemoryData[T];
+
+  /** unless set to true, will be automatically deleted in the future */
+  @Column({ default: false })
+  isSaved!: boolean;
```
mertalev, line 54: Similarly, do we want to store who it's saved by?

server/src/dtos/memory.dto.ts
```diff
@@ -0,0 +1,65 @@
...
+import { IsEnum, IsObject } from 'class-validator';
+import { AssetResponseDto, mapAsset } from 'src/dtos/asset-response.dto';
+import { MemoryEntity, MemoryType } from 'src/entities/memory.entity';
+import { ValidateBoolean, ValidateDate, ValidateUUID } from 'src/validation';
+
+type MemoryData = any;
+
+class MemoryBaseDto {
+  @ValidateBoolean({ optional: true })
+  isSaved?: boolean;
```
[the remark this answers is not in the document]
jrasm91: Not really. Not passing it in an update call would change it from undefined to false and then it would always set it to false, even if it was previously set to true.

server/src/entities/memory.entity.ts
jrasm91: These are only going to be tracked for the owner. I'm not sure what functions shared users will be able to do besides view them in the future. Eventually we could do something like we have planned to do with assets and move owner id off of this table, but it seems premature to try to figure that out now, here.

server/src/entities/memory.entity.ts
jrasm91: I'm planning on having this work like Google. So, we'll generate memories and unless the user clicks something to "save" it (set this to true) it will get automatically deleted after 14 days.

alextran1502:
Do we address #8352 in this PR?

jrasm91:
> Do we address #8352 in this PR?

No. This PR adds the foundation to support persistent memories (adds a new API for them) but it doesn't include automatic memory generation yet. I was planning to do that part in a follow-up pull request.

server/src/dtos/memory.dto.ts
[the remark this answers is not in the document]
mertalev: Oh yeah, I was thinking about creation but updates would require it to be optional.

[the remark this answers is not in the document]
jrasm91: I usually just set the default value in the database and then make everything optional. Especially when it defaults to false, undefined is false when it is being checked like that.

alextran1502:
Let's resolve the conflicts and merge it!

PR #8332, opened 2024-03-28
fix(mobile): sync all album properties

fyfrey:
Currently some album properties are not synced to the Isar database when an album is updated. This PR adds the missing properties to be synced.


fyfrey:
@waclaw66 this should fix the issue with changes made to an album from another client

waclaw66:
> @waclaw66 this should fix the issue with changes made to an album from another client

Thank you, I will test in on Tuesday, currently out for extended weekend.

alextran1502 (approved):

waclaw66:
@fyfrey Tested, it works as expected, thanks!

PR #8341, opened 2024-03-28
fix(server): map style not being available for shared assets

### LINKED ISSUE #8031: Map not showing in image details in album shared via link
### The bug

Opening the image details ("i" icon), triggers a request to e.g. https://demo.immich.app/api/system-config/map/style.json?theme=light. However, this endpoint requires authentication as [explicitly tested](https://github.com/immich-app/immich/blob/148428a564ad5876ceb6b3991a7571e65dc15b38/e2e/src/api/specs/system-config.e2e-spec.ts#L18-L23). That is a problem when viewing a shared album as anonymous user (via link), because it seems that no kind of authentication (including the sharing key) is passed for this request, resulting in a 401.

Debian 12 / Docker

### Version of Immich Server

v1.98.2

n/a

### Platform with the issue

- [x] Server
- [X] Web
- [ ] Mobile

```YAML
n/a
```

### Your .env content

```Shell
n/a
```

### Reproduction steps

```bash
These steps show the problem on the demo site:
1. Create an album with at least one picture in it.
2. Share the album via link, make sure to include metadata.
3. Open the link in an incognito window.
4. Open Chrome Developer Tools.
5. Click on the picture.
6. Click the "i" icon to show the Info panel.
7. Note that the map isn't showing and Developer Tools complain about a 401 for the mentioned `/api/system-config/map/style.json`.
```

### Additional information

I'm wondering if there's really something so secret about the (not even customizable?) map styles that authentication must be required. Making the endpoint public should solve the issue.

danieldietzler:
Fixes #8031

### DIFF
--- a/e2e/src/api/specs/system-config.e2e-spec.ts
+++ b/e2e/src/api/specs/system-config.e2e-spec.ts
@@ -1,4 +1,4 @@
-import { LoginResponseDto, getConfig } from '@immich/sdk';
+import { AssetFileUploadResponseDto, LoginResponseDto, SharedLinkType, getConfig } from '@immich/sdk';
 import { createUserDto } from 'src/fixtures';
 import { errorDto } from 'src/responses';
 import { app, asBearerAuth, utils } from 'src/utils';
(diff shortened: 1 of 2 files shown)


jrasm91 (approved):
Can you add an e2e?

PR #8342, opened 2024-03-28
fix(server): add missing file extensions to library files

etnoy:
* Detect when the basename is invalid and force refresh
* Actually re-import new filename upon refresh

This adds to #8083 and #8052 and makes a retroactive fix for all affected assets


server/src/services/library.service.ts
```diff
@@ -521,6 +529,8 @@ export class LibraryService extends EventEmitter {
       await this.assetRepository.updateAll([existingAssetEntity.id], {
         fileCreatedAt: stats.mtime,
         fileModifiedAt: stats.mtime,
+        localDateTime: stats.mtime,
```
jrasm91, line 532: Does this need to be offset so it's in utc?

jrasm91 (approved):
Nice!

server/src/services/library.service.ts
etnoy: hm...good point. Maybe let's not touch that rabbit hole today

server/src/services/library.service.ts
```diff
@@ -453,6 +453,20 @@ export class LibraryService extends EventEmitter {
...
+    } else if (existingAssetEntity.originalFileName !== parse(assetPath).base) {
+      // Asset base filename is invalid.
+      // This cleans up an old bug where the file extension wasn't handled properly in previous versions
+      // We can likely remove this check in the second half of 2024
+      console.warn(
+        assetPath
+      );
+      console.warn(
+        existingAssetEntity.originalFileName
+      );
```
mertalev: Nit: remove console statements.

server/src/services/library.service.ts
```diff
@@ -453,6 +453,14 @@ export class LibraryService extends EventEmitter {
         `File modification time has changed, re-importing asset: ${assetPath}. Old mtime: ${existingAssetEntity.fileModifiedAt}. New mtime: ${stats.mtime}`,
       );
       doRefresh = true;
+    } else if (existingAssetEntity.originalFileName !== parse(assetPath).base) {
```
mertalev, line 456: Nit: the asset path is parsed twice. You can move it to a variable to reuse it.

server/src/services/library.service.ts
```diff
@@ -453,6 +453,14 @@ export class LibraryService extends EventEmitter {
...
+      // Asset base filename is invalid.
+      // This cleans up an old bug where the file extension wasn't handled properly in previous versions
+      // We can likely remove this check in the second half of 2024
```
mertalev, line 459: This should be a TODO

server/src/services/library.service.ts
```diff
@@ -453,6 +453,14 @@ export class LibraryService extends EventEmitter {
         `File modification time has changed, re-importing asset: ${assetPath}. Old mtime: ${existingAssetEntity.fileModifiedAt}. New mtime: ${stats.mtime}`,
       );
       doRefresh = true;
+    } else if (existingAssetEntity.originalFileName !== parse(assetPath).base) {
+      // Asset base filename is invalid.
```
mertalev, line 457: Nit: redundant comment.

server/src/services/library.service.ts
```diff
@@ -453,6 +453,14 @@ export class LibraryService extends EventEmitter {
...
+      // This cleans up an old bug where the file extension wasn't handled properly in previous versions
```
mertalev, line 458: Nit: this is a bit verbose

mertalev (approved):

server/src/services/library.service.ts
etnoy: Good point!

server/src/services/library.service.ts

server/src/services/library.service.ts
etnoy: Done, thanks

server/src/services/library.service.ts
etnoy: Done

server/src/services/library.service.ts
etnoy: Done

mertalev (approved):

PR #8350, opened 2024-03-29
Update remove-offline-files.md

Thoroslives:
I have amended the script as it only worked for assets that were owned by the admin user. it now prompts for admin and owner api. using the admin api to scan for orphan assets and using the owner api to send the delete request


jrasm91:
Hey, I think we're going to move to a model where we have a page with project descriptions, etc. and then link out to external pages/repositories. Do you have a page or repository where this code is available that we can link to instead?

Thoroslives:
> Hey, I think we're going to move to a model where we have a page with project descriptions, etc. and then link out to external pages/repositories. Do you have a page or repository where this code is available that we can link to instead?

Ive got a Gist if thats what youre after
https://gist.github.com/Thoroslives/ca5d8e1efd15111febc1e7b34ac72668

jrasm91:
Moving to #8641

PR #8354, opened 2024-03-29
feat(web, a11y): add labels!

ben-basten:
## Description

Labeling icons + buttons in various locations!

Notable changes:

* `title` is now a required prop for `CircleIconButton`, and it automatically adds an `aria-label` for screen reader users
* Add color labels for the "Select avatar color" modal
* Mark icons in the menu options as `aria-hidden`, because they already have perceivable text to go along with them

## Screenshots

N/A - UI looks visually the same.

## How Has This Been Tested?

Hover over any of the change buttons with a mouse to check that a popup with the title appears. Or try using a screen reader to focus the buttons, and check if it announces a title.

## Checklist:

- [x] npm test (unit tests)


web/src/lib/components/elements/buttons/circle-icon-button.svelte
```diff
@@ -6,7 +6,7 @@
   export let hoverColor = '#e2e7e9';
   export let padding = '3';
   export let size = '24';
-  export let title = '';
+  export let title: string;
```
bo0tzz, line 9: nit: Should required props be at the top of the list?

bo0tzz (approved):
Very nice!

web/src/lib/components/elements/buttons/circle-icon-button.svelte
ben-basten: Nice catch, fixed!

jrasm91 (approved):

PR #8415, opened 2024-03-31
renovate: mobile ignore *_plus plugins

shenlong-tanwen:
#### Changes made

- *_plus plugins are ignored from renovate updates until the jdk is migrated to 17

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -64,6 +64,6 @@
     }
   ],
   "ignorePaths": ["mobile/openapi/pubspec.yaml"],
-  "ignoreDeps": ["http", "intl"],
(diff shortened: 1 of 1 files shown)


bo0tzz (approved):

jrasm91:
You can instead just close the pull requests.

shenlong-tanwen:
> You can instead just close the pull requests.

Yes but it'll take a while for us to update our JDK to Java 11. None of the other plugins have such weird build requirement. We can remove this ignore rule when we update them.

jrasm91:
That's what I was thinking. It will still track the dependency and that way we don't forget about it, but we can ignore it until the sdk is updated.

PR #8422, opened 2024-03-31
feat: Improve error handling for Install Script

mmomjian:
- Improve error detection for `cd`, `mkdir`
- Add check that `curl` exists and downloads are successful
- Notify when overwriting an existing immich-app folder - may want to prompt for a warning in the future?
- Remove deprecated references to `docker-compose`

### DIFF
--- a/docs/docs/install/script.md
+++ b/docs/docs/install/script.md
@@ -17,12 +17,11 @@ curl -o- https://raw.githubusercontent.com/immich-app/immich/main/install.sh | b
 The script will perform the following actions:
 
 1. Download [docker-compose.yml](https://github.com/immich-app/immich/releases/latest/download/docker-compose.yml), and the [.env](https://github.com/immich-app/immich/releases/latest/download/example.env) file from the main branch of the [repository](https://github.com/immich-app/immich).
-2. Populate the `.env` file with necessary information based on the current directory path.
-3. Start the containers.
+2. Start the containers.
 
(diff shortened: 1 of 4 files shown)


docs/docs/install/script.md
```diff
@@ -17,12 +17,11 @@ curl -o- https://raw.githubusercontent.com/immich-app/immich/main/install.sh | b
 The script will perform the following actions:
 
 1. Download [docker-compose.yml](https://github.com/immich-app/immich/releases/latest/download/docker-compose.yml), and the [.env](https://github.com/immich-app/immich/releases/latest/download/example.env) file from the main branch of the [repository](https://github.com/immich-app/immich).
-2. Populate the `.env` file with necessary information based on the current directory path.
-3. Start the containers.
```
danieldietzler, line 21: I would say that people should still customize their env file. You don't think that's necessary?

danieldietzler (approved):
LGTM

docs/docs/install/script.md
mmomjian: I wouldn't say that's not necessary, but this script doesn't actually do that. It starts the containers right away so any customizations (such as `DB_PASSWORD` will be too late). We would have to change the behavior of the script to prompt for common common variables and write to the .env if we want to try this.

danieldietzler: Oh you are correct, fair.

PR #8443, opened 2024-04-01
mount postgres folder to local directory by default

danieldietzler:
> [!Caution]
> For people always pulling the latest compose file this is a *breaking change*!
> Disregarding the notes will result in (temporary) data loss!

## Background
In the past we've seen many cases where people accidentally deleted their postgres data by (unintentionally) deleting the docker volume (e.g. `docker compose down -v`).
This is unfortunate as there is *no* way to recover that data (if you don't have a backup; **MAKE BACKUPS!**).
We have been thinking about mounting the postgres data to a local folder for a while but always hesitated as this will break existing instances due to people not reading the change logs carefully. However, there have simply been too many issues with it that we ultimately decided we should make that change.

## What do I have to do?
Nothing. You shouldn't copy the compose file with every new release and generally, we don't recommend making changes to existing instances. If you never had issues before, attempting to migrate the data will put it at (an unnecessary) risk.

## I want to migrate my docker volume to a local folder
Unfortunately there isn't a "proper" way to export a docker volume.
The recommended method is to mount the volume and the directory (you want to copy your data to) to an arbitrary container, get a shell inside that container and copy the folder manually.

> [!Caution]
> Take backups before attempting this. Especially make sure you have a current database dump (`pg_dump`)

> [!Warning]
> Do *not* use a directory under `/mnt` for the postgres location if you are using WSL.
> Generally (on all operating systems) we recommend against using a network share for your database location. This is bound to break and cause all sorts of weird issues.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -69,9 +69,8 @@ services:
       POSTGRES_USER: ${DB_USERNAME}
       POSTGRES_DB: ${DB_DATABASE_NAME}
     volumes:
-      - pgdata:/var/lib/postgresql/data
+      - ${DB_DATA_LOCATION}:/var/lib/postgresql/data
     restart: always
 
 volumes:
-  pgdata:
   model-cache:
--- a/docker/example.env
+++ b/docker/example.env
@@ -14,5 +14,6 @@ DB_PASSWORD=postgres
 DB_HOSTNAME=immich_postgres
 DB_USERNAME=postgres
 DB_DATABASE_NAME=immich
+DB_DATA_LOCATION=./postgres
 
 REDIS_HOSTNAME=immich_redis


mmomjian:
Love this. Decreased risk of destruction of data with Docket commands. 

docs will need to updated to explain how to “reset” Immich. Currently we advise docker compose down -v

mmomjian:
Another thought - for people who currently backup the entire immich folder (including `./library` and now `./postgres`), they could have a false sense of security that their Postgres data is backed up (it's not, as a filesystem-level backup of a running DB is not OK - need to use pg_dump) - we should add some details / explanation of this.

danieldietzler:
> docs will need to updated to explain how to “reset” Immich. Currently we advise docker compose down -v

I would wait with that change until we make a release because we currently still deploy docs independently of releases.

jrasm91 (approved):

alextran1502:
> The recommended method is to mount the volume and the directory (you want to copy your data to) to an arbitrary container, get a shell inside that container and copy the folder manually.

Can you help provide detail instruction for this like I am 5?

mmomjian:
> > The recommended method is to mount the volume and the directory (you want to copy your data to) to an arbitrary container, get a shell inside that container and copy the folder manually.
> 
> Can you help provide detailed instructions for this, like I am 5 years old?

Something like: 

```
docker run -it --rm -v pgdata:/postgres-from -v /path/on/host:/postgres-to alpine cp -pr /postgres-from /postgres-to
```

danieldietzler:
> > The recommended method is to mount the volume and the directory (you want to copy your data to) to an arbitrary container, get a shell inside that container and copy the folder manually.
> 
> Can you help provide detailed instructions for this, like I am 5 years old?

I intentionally did not do that as I don't want it to be easy for people to change this. If it's a detailed instruction people will try it even though they shouldn't

bo0tzz (approved):

NICHOLAS85:
> ## I want to migrate my docker volume to a local folder
> Unfortunately there isn't a "proper" way to export a docker volume. The recommended method is to mount the volume and the directory (you want to copy your data to) to an arbitrary container, get a shell inside that container and copy the folder manually.



> > > The recommended method is to mount the volume and the directory (you want to copy your data to) to an arbitrary container, get a shell inside that container and copy the folder manually.
> > 
> > 
> > Can you help provide detailed instructions for this, like I am 5 years old?
> 
> Something like:
> 
> ```
> PGTARGET='/postgres-immich-data'
> mkdir "$PGTARGET" && \
>   docker run -it --rm \
>   -v pgdata:/postgres-from \
>   -v "$PGTARGET":/postgres-to \
>   alpine \
>   cp -prn /postgres-from/. /postgres-to
> ```

Any input on simply using [`docker cp`](https://docs.docker.com/reference/cli/docker/container/cp/)/[`podman cp`](https://docs.podman.io/en/latest/markdown/podman-cp.1.html)?

For now I've taken these steps:
Stopped immich (to ensure no issues with the database updating while migrating)
run `podman cp immich_postgres:/var/lib/postgresql/data [local path]`
Updated podman container file to point to the local path instead of pgdata volume
Restarted immich and made sure all my data is still there.

Any concerns with doing this over what @mmomjian suggested?

mmomjian:
`docker cp` is probably fine - I rarely work with Docker volumes so I wouldn't consider myself an expert though. As long as it works and Immich is functional it should be good.

mmomjian:
@NICHOLAS85 it's come to my attention that you can't use `docker cp` while a container is running. Was the immich_postgres container running when you ran that command? If so, I would not trust the integrity of that backup.

malavolti:
Dear All and thank you for your wonderful work! 🙏🏻 

I used already use bind-volume and these are the changes between the previous version (1.101.0):

```bash
# diff docker-compose-1.101.0.yml docker-compose-1.102.0.yml 
49,52c49,52
<     image: ghcr.io/immich-app/immich-machine-learning:${IMMICH_VERSION:-release}-openvino
<     extends: # uncomment this section for hardware acceleration - see https://immich.app/docs/features/ml-hardware-acceleration
<      file: hwaccel.ml.yml
<      service: openvino # set to one of [armnn, cuda, openvino, openvino-wsl] for accelerated inference - use the `-wsl` version for WSL2 where applicable
---
>     image: ghcr.io/immich-app/immich-machine-learning:${IMMICH_VERSION:-release}
>     # extends: # uncomment this section for hardware acceleration - see https://immich.app/docs/features/ml-hardware-acceleration
>     #   file: hwaccel.ml.yml
>     #   service: cpu # set to one of [armnn, cuda, openvino, openvino-wsl] for accelerated inference - use the `-wsl` version for WSL2 where applicable
54c54
<       - ./data/model-cache:/cache
---
>       - model-cache:/cache
61c61
<     image: registry.hub.docker.com/library/redis:6.2-alpine@sha256:51d6c56749a4243096327e3fb964a48ed92254357108449cb6e23999c37773c5
---
>     image: registry.hub.docker.com/library/redis:6.2-alpine@sha256:84882e87b54734154586e5f8abd4dce69fe7311315e2fc6d67c29614c8de2672
72c72
<       - ./data/pgdata:/var/lib/postgresql/data
---
>       - ${DB_DATA_LOCATION}:/var/lib/postgresql/data
73a74,76
> 
> volumes:
>   model-cache:
```

Can you help me to understand if I think correctly that I need only to do:

- `docker-compose pull`
- `docker-compose up -d`

to update my server instance?

Thank you so much! 🤗

NICHOLAS85:
> @NICHOLAS85 it's come to my attention that you can't use `docker cp` while a container is running. Was the immich_postgres container running when you ran that command? If so, I would not trust the integrity of that backup.

I'm my process I stopped all immich-related containers to ensure no problems, so immich_postgres was not running.

I'll have to do more research on permissions and the cp subcommand as in my case using podman I believe both the container and destination user were the same. Might be problematic in dockers case without looking at the manual.

mio-19:
`DB_DATA_LOCATION1` is in the `The values below this line do not need to be changed` section. I think it would be better to put it near the `UPLOAD_LOCATION`

poige:
@danieldietzler wrote:

> Disregarding the notes will result in (temporary) data loss!

Hold on, which "notes"? But ok, so the loss is _temporary_, let's read on:

> In the past we've seen many cases where people accidentally deleted their postgres data by (unintentionally) deleting the docker volume (e.g. docker compose down -v).

Ergh, are you talking about same kind of loss, which you called "(temporary)", or it's different loss?

> This is unfortunate as there is no way to recover that data (if you don't have a backup; MAKE BACKUPS!).

So now it's not _temporary_ but rather _irreversible_. Do "make backups" — what backups, do we need backups of Postgresql now?

All-in-all — what is going on, don't you notice there're too many contradictions and ambiguity even in this part of your texts?

Please provide clear and concise UPDATE/UPGRADE procedures in the docs, right now it's lacking it badly.
The release notes, as of now, is a mess and mixture of important and absolutely irrelevant (to end users) stuff — who needs the list of commits, for e. g., if it can be looked up at any time with Github or Git itself but if there're important things which needs to be noted it can be easily overlooked because of this redundant and irrelevant information.

poige:
image

did you ever have a definition of what is "breaking change" and how it's supposed to be handled? I don't see it in the docs except here:

image

where it's basically substituted by another vague term "breaking updates":

image


which is "important to follow".

Wow!

PR #8475, opened 2024-04-02
feat(web): paste photo from clipboard

TomixUG:
This adds the ability to upload a photo by pasting it from system clipboard. 
Just press CTRL + V and the photo from your clipboard starts uploading like it's in Google photos. It's awesome if you want to quickly upload a screenshot.

This is my first contribution, feel free to fix the bad implementation.

### Demonstration
Here I use Flameshot to take a screenshot into clipboard and then paste it using CTRL + V:
https://github.com/immich-app/immich/assets/46199769/8a1f1987-823c-4ebe-a431-6fd85af0465c

### DIFF
--- a/web/src/routes/(user)/+layout.svelte
+++ b/web/src/routes/(user)/+layout.svelte
@@ -41,8 +41,8 @@
   };
 </script>
 
-<div on:paste={pasteHandler}>
-  <slot {albumId} />
-</div>
+<slot {albumId} />
 
 <UploadCover {dropHandler} />
+
+<svelte:window on:paste={pasteHandler} />
--- a/web/src/routes/(user)/+layout.svelte
+++ b/web/src/routes/(user)/+layout.svelte
@@ -1,48 +1,6 @@
 <script lang="ts">
-  import { page } from '$app/stores';
   import UploadCover from '$lib/components/shared-components/drag-and-drop-upload-overlay.svelte';
-  import { dragAndDropFilesStore } from '$lib/stores/drag-and-drop-files.store';
-  import { fileUploadHandler } from '$lib/utils/file-uploader';
-
-  let albumId: string | undefined;
-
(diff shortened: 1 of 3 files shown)


web/src/routes/(user)/+layout.svelte
```diff
@@ -22,8 +22,27 @@
...
+
+    const filesArray: File[] = Array.from<File>(files);
+    albumId = ($page.route.id === '/(user)/albums/[albumId]' || undefined) && $page.params.albumId;
+
+    const isShare = $page.route.id === '/(user)/share/[key]' || undefined;
+    if (isShare) {
+      dragAndDropFilesStore.set({ isDragging: true, files: filesArray });
+    } else {
+      await fileUploadHandler(filesArray, albumId);
+    }
```
danieldietzler, line 40: Maybe it would be worthwhile to extract that logic to a private function as it's the exact same the `dropHandler` uses. Avoid some duplication etc

web/src/routes/(user)/+layout.svelte
```diff
@@ -22,8 +22,27 @@
...
+  };
 </script>
 
-<slot {albumId} />
+<div on:paste={pasteHandler}>
```
michelheusschen, line 44: Pretty nice feature. Instead of creating an empty div you can also listen for the event on window: `<svelte:window on:paste={pasteHandler} />`

TomixUG: Thanks, that's much better, I changed it in a new commit.

jrasm91 (commented):
This looks great. Maybe it would be better to move this logic into `drag-and-drop-upload-overlay.svelte` (and possibly rename the component). That seems to be where the other upload events are handled.

jrasm91 (approved):

danieldietzler (approved):
LGTM

Pebor:
Epic

PR #8479, opened 2024-04-03
feat(web,a11y): slider accessibility improvements

ben-basten:
## Description

Made the following accessibility changes to the `Slider` toggle component:

1. Added a focus outline around the currently focused slider, to make it easier for users to track where they are on the page. The color of the outline is automatically determined by the browser and can vary - open to ideas if there's a specific color in mind.
2. Linking labels to the slider using IDs, so screen readers will now announce the context that goes with it. 

## Screenshots

**Focused slider in dark mode (Firefox)**

focus-dark

**Focused slider in light mode (Firefox)**

focus-light

</details> 

## How Has This Been Tested?

- [x] MacOS + VoiceOver
    - [x] Chrome
    - [x] Firefox
    - [x] Safari
- [ ] Windows + NVDA - don't have this setup yet
    - [ ] Chrome
    - [ ] Firefox

Try tabbing through a form with sliders, such as the "Create link to share" modal. There should be visible focus outlines around the slider. 

## Checklist:

- [x] npm test (unit tests)


jrasm91 (approved):
LGTM

mertalev:
The outlines have some aliasing and color artifacts. Can we fix that?

Dark mode (left inner section):

outline-dark


Light mode (inner circle):

outline

alextran1502:
Can we please keep the slider's original styling?

mertalev:
I think it only changes the styling when the slider is focused. It makes it easier when you're tabbing through elements in the page, etc.

web/src/lib/components/elements/slider.svelte
```diff
@@ -1,26 +1,35 @@
...
+    on:blur={() => (isFocused = false)}
     {disabled}
   />
 
   {#if disabled}
-    <span class="slider slider-disabled cursor-not-allowed" />
+    <span class="slider slider-disabled cursor-not-allowed" class:focused={isFocused} />
   {:else}
-    <span class="slider slider-enabled cursor-pointer" />
+    <span class="slider slider-enabled cursor-pointer" class:focused={isFocused} />
```
michelheusschen, line 32: It's probably better to only show focus styling with `.slider:focus-visible` instead of regular focus. That also prevents the outline from flickering while clicking the slider. Also just using Tailwind or CSS would be easier instead of detecting focus with JS:

```svelte
<input class="peer" />
<span class="peer-focus-visible:outline" />
```

web/src/lib/components/shared-components/settings/setting-switch.svelte
```diff
@@ -30,9 +33,8 @@
       {/if}
     </div>
 
-    <p class="text-sm dark:text-immich-dark-fg">{subtitle}</p>
-    <slot />
+    <label class="text-sm dark:text-immich-dark-fg" for={sliderId}>{subtitle}</label>
```
michelheusschen, line 36: This shouldn't be a label. Using aria-describedby or something alike would be more appropriate

web/src/lib/components/shared-components/settings/setting-switch.svelte
```diff
@@ -4,20 +4,23 @@
...
   import Slider from '$lib/components/elements/slider.svelte';
 
+  export let id: string;
   export let title: string;
   export let subtitle = '';
   export let checked = false;
   export let disabled = false;
   export let isEdited = false;
 
+  const sliderId = `${id}-slider`;
```
michelheusschen, line 14: Reactivity breaks when the `id` prop is updated

```suggestion
  $: sliderId = `${id}-slider`;
```

web/src/lib/components/shared-components/settings/setting-switch.svelte
```diff
@@ -30,9 +33,8 @@
       {/if}
     </div>
 
-    <p class="text-sm dark:text-immich-dark-fg">{subtitle}</p>
-    <slot />
```
michelheusschen, line 34: Not safe to remove because the slot is used here:

https://github.com/immich-app/immich/blob/4ed68cf67385efac66d5c5035cd79f849a36959c/web/src/lib/components/user-settings-page/app-settings.svelte#L77-L84

web/src/lib/components/shared-components/settings/setting-switch.svelte
ben-basten: Yeah... so I was debating this too. Using `aria-describedby` makes sense to me, but many of the settings sliders are really using the subtitle as the label of the slider. How would you feel about a follow-up PR to improve the copy on some of the sliders to make the titles better labels, and only include a subtitle if it's needed to provide additional context?

slider-logging
slider-scanning
slider-sharing
slider-storage-template

ben-basten:
> The outlines have some aliasing and color artifacts. Can we fix that?

Yes, I'm seeing that too. I will do some investigation to see if there's a good solution for that.

web/src/lib/components/shared-components/settings/setting-switch.svelte
michelheusschen: Ah, I understand now why it was changed to a label. The contents of `aria-describedby` should get announced along with the label, but I'm also fine doing this in a separate PR

ben-basten: I've updated the PR to use `aria-describedby` instead. PR soon to start a discussion on label copy!

ben-basten:
> The outlines have some aliasing and color artifacts. Can we fix that?

So I've been looking into this @mert. Some things I've tried:

* **adding a box shadow to add color fill in the color artifacts:** no change
* **using border to indicate focus:** removes the color artifacts, but feels weird because the border makes the slider look smaller
* **enlarging the slider, to see if the color artifacts are caused by how small the slider is:** still seeing color artifacts when the slider is larger, see this screenshot for an example:

slider-outline-aliasing

So... the compromise solution I could think of is adding a subtle border in the same color as the background, to mask the color artifacting. It also has the added benefit of making the slider a little easier to see when it's focused, in my opinion.

What do you think about this solution?

**Dark Mode (Firefox)**

slider-outline-dark

**Light Mode (Firefox)**

slider-outline-light

mertalev:
That looks a lot better! It's also good in that it increases contrast when the outline and slider are both gray.

PR #8486, opened 2024-04-03
fix(web): prevent fetching asset info twice

michelheusschen:
Asset data is preloaded in the `assetViewingStore`, while it's already available in the `AssetStore`. This PR uses the data from the asset store instead of fetching it again.

### DIFF
--- a/web/src/lib/components/asset-viewer/asset-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/asset-viewer.svelte
@@ -64,7 +64,7 @@
 
   let reactions: ActivityResponseDto[] = [];
 
-  const { setAssetId } = assetViewingStore;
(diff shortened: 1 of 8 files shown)


jrasm91 (approved):

danieldietzler (approved):
Nice

chodthewacko:
I'm wondering if there is a glitch with this fix, because i'm semi frequently seeing incorrect stuff that seem to be cache errors.
Examples:
picture -> info:   shows a face but the name is wrong on the face.  I click on the face, and the "face's page" opens up and the name is correct there.

Every now and then, a page for a named face will open, and the number of shown under the name is wrong.  I'll then refresh and go back to that page, and it'll be correct the second time.

PR #8491, opened 2024-04-03
Add AV1 transcoding support

n00mkrad:
**Adds support for the free AV1 codec as an option for video transcoding.**

**Why?**
- AV1 is the most efficient codec that is widely adopted
- I was unhappy with the existing options: H264 is old, VP9 is horribly slow, HEVC browser support is bad
- The most popular encoder, SVT-AV1, is quite mature by now and keeps getting faster (faster than x265 or vpx-vp9)
- Feature has been requested before (e.g. https://github.com/immich-app/immich/discussions/5062)

**Playback compatibility?**
- Hardware encoding support on Nvidia (Ada+), Intel (Arc Alchemist+, 12th Gen+), AMD (RDNA3+)
- Hardware decoding support on Nvidia (Ampere+), Intel (11th Gen+), AMD (RDNA2+), Apple (M3+ A17 Pro+) 
- Software decoding support on basically anything that comes with libdav1d

**New dependencies?**
- None, everything is already included in immich's ffmpeg build

**Not implemented at the time of writing:**
- Hardware encoding (nvenc/qsv/amf)
- ~~Support for threads option~~ DONE
- ~~Optimized encoding paramters (e.g. tiles)~~ Defaults are fine

**Untested stuff:**
- ~~"Accepted video codecs" has the AV1 option, but I have not yet tested if it works as it should~~ WORKS
- No idea what happens on iPhone, I presume it won't be able to play back the transcoded videos


server/src/utils/media.ts
```diff
@@ -394,6 +394,33 @@ export class VP9Config extends BaseConfig {
...
+        `-crf ${this.config.crf}`,
+        `-svtav1-params mbr=${bitrates.max}${bitrates.unit}`, // Wrapping the params in quotes breaks them for some reason, so don't do that
+      ];
+    } else {
+      return [`-crf ${this.config.crf}`];
+    }
+  }
+
+  getThreadOptions() {
+    return [];
```
mertalev, line 419: You should only add this if `threads` is a positive number.

server/src/utils/media.ts
```diff
@@ -259,7 +259,7 @@ export class BaseHWConfig extends BaseConfig implements VideoCodecHWConfig {
   }
 
   getSupportedCodecs() {
-    return [VideoCodec.H264, VideoCodec.HEVC, VideoCodec.VP9];
+    return [VideoCodec.H264, VideoCodec.HEVC, VideoCodec.VP9, VideoCodec.AV1];
```
mertalev, line 262: If you're changing this, you should override it in VAAPI since it doesn't support AV1 until FFmpeg 6.1.

n00mkrad:
Just revered the commit for threading, as it's not possible to merge `-svtav1-params` by using it multiple times, only the last occurence will be used.

**THIS IS A FUNDAMENTAL ISSUE IN THE CODEBASE THAT AFFECTS X264 AND X265 AS WELL!** Will make an issue asap.

mertalev (commented):
Thanks for working on this! Besides the comments, we'll also need to add a few unit tests for this.

server/src/utils/media.ts
n00mkrad: Is this for readability?

Because SVT will default to using all threads anyway if 0 or a neg number is passed.
I can add a check for it anyway.

server/src/utils/media.ts
n00mkrad: Got it, done

server/src/utils/media.ts
mertalev: Yeah, it's just easier to know for a fact that it works a certain way than to leave it ambiguous. e.g. 0 is optimal, but is -1 the same as 0? Does it accept -2? Can there be a difference between passing `-threads 0` and omitting it? Do all the encoders have this same behavior?

server/src/utils/media.ts
mertalev: Oh, and NVENC overrides this since it doesn't support VP9. I think it'd be cleaner to change the base method to only include H.264 and HEVC and let the others add onto that depending on what else they support.

n00mkrad: Is this in the scope of this PR or should I leave it like this for now?

mertalev: Since we're touching this code anyway, we may as well change it here.

n00mkrad: Done - I also removed AV1 since it's only implemented for SW right now. Can be added back as soon as the HW configs are extended to enable AV1.

server/src/utils/media.ts
```diff
@@ -394,6 +394,33 @@ export class VP9Config extends BaseConfig {
...
+        `-svtav1-params mbr=${bitrates.max}${bitrates.unit}`, // Wrapping the params in quotes breaks them for some reason, so don't do that
+      ];
+    } else {
+      return [`-crf ${this.config.crf}`];
+    }
+  }
+
+  getThreadOptions() {
+    return [];
+    // return [`-svtav1-params lp=${this.config.threads}`]; // THIS WOULD OVERWRITE PREVIOUS SVTAV1-PARAMS! Can't be used multiple times.
```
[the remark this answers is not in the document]
n00mkrad: Using an array for the svt params now.
https://github.com/immich-app/immich/commit/1ded53bf71e360cdc241b1e9a6f1cc703ca63c36

server/src/utils/media.ts
```diff
@@ -394,6 +394,35 @@ export class VP9Config extends BaseConfig {
...
+    const svtparams = [];
+    if (this.config.threads > 0) {
+      svtparams.push(`lp=${this.config.threads}`)
+    }
+    if (bitrates.max > 0) {
+      svtparams.push(`mbr=${bitrates.max}${bitrates.unit}`)
+    }
+    return [
+      `-crf ${this.config.crf}`,
+      `-svtav1-params ${svtparams.join(':')}`
```
mertalev, line 417: You should only add this if `svtparams` isn't empty, else it will be an invalid command since `-svtav1-params` will expect an argument.

n00mkrad: Ah yeah that was an oversight, fixed.

n00mkrad: Oh damn, totally missed that and forgot to test that case, fixed

mertalev (commented):
This should be good to merge once we add unit tests.

mertalev:
Is it normal that AV1 produced a bigger video than H.264 at 32.5MiB vs 27.9MiB? I set the preset to veryslow for both and set the CRF for H.264 and AV1 to 23 and 28, respectively. I suppose it's an unfair comparison since veryslow maps to H.264's slowest preset (excluding placebo) but not for AV1. But it does feel underwhelming considering AV1 also took 62% longer. Is there maybe a setting we're missing?

n00mkrad:
> Is it normal that AV1 produced a bigger video than H.264 at 32.5MiB vs 27.9MiB? I set the preset to veryslow for both and set the CRF for H.264 and AV1 to 23 and 28, respectively. I suppose it's an unfair comparison since veryslow maps to H.264's slowest preset (excluding placebo) but not for AV1. But it does feel underwhelming considering AV1 also took 62% longer. Is there maybe a setting we're missing?

x265 and svt-av1 are entirely different encoders, so yes, this is normal. We can increase the default CRF maybe.

mertalev:
Increasing the CRF was the answer, at least for this video.

4K H264 with CRF 23 and veryslow preset: 210mb and 97.4 SSIM
4K AV1 with CRF 28 and preset 4: 240.5mb and 96.4 SSIM
4K AV1 with CRF 35 and preset 4: 117mb and 96.2 SSIM

mertalev (approved):
LGTM

jarylc:
I've done some testing for AV1 encoding with my Intel Arc A380 after this PR, works flawlessly.

PR to allow it for QSV and additional notes here: https://github.com/immich-app/immich/pull/8942

ppnplus:
Sorry for reviving an old PR, but after I played around, the default AV1 settings sometimes produce more noticeable artifacts than x264 especially in fast-moving, detail-rich scenes. Do you think more tuning or switching to svt-av1-hdr is needed?

Also, iOS app and browser works (iPhone 15 Pro Max), but on both, when fast-forwarding the video sometimes freezes with audio playing for up to several seconds, and would occasionally flash into colors (AV1 videos from YouTube and Netflix works fine). On an unsupported device (iPad Pro M2), the video plays in audio-only and shows thumbnail as video.

PR #8520, opened 2024-04-04
chore: remove unused port

jrasm91:


danieldietzler (approved):

jrasm91:
I think we're going to add health checks in, so we'll actually need this in a bit.

PR #8535, opened 2024-04-05
Make language gender neutral

iamwillbar:


mertalev (approved):
Thanks! This was an oversight.

PR #8546, opened 2024-04-05
docs: pg_dumpall refinements

mmomjian:
"fixes" #8459

- Switch to long-args, add --if-exists to prevent errors when dropping non-existent DB
- Switch Docker image to `pg_dumpall` to mirror behavior of given command, making sure that all information (schemas, perms, etc) is backed up.
- Do not need to specify a target database when restoring pg_dumpall
- Note that `pg_dumpall` (set by `POSTGRES_CLUSTER: true`) does not support compression, so these dumps will be larger than with the previous method. However, these are auto-rotate, and I believe that as a default we should provide a complete cluster backup as a database backup only can be harder to restore with extensions, schemas, etc.

### DIFF
--- a/docs/docs/FAQ.mdx
+++ b/docs/docs/FAQ.mdx
@@ -110,7 +110,7 @@ No. Our golden rule is that the original assets should always be untouched, so w
 
 ### How can I move all data (photos, persons, albums) from one user to another?
 
-This is not officially supported, but can be accomplished with some database updates. You can do this on the command line (in the PostgreSQL container using the psql command), or you can add for example an [Adminer](https://www.adminer.org/) container to the `docker-compose.yml` file, so that you can use a web-interface.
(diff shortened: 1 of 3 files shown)


alextran1502 (approved):

RenanGreca:
Hi, I just restored an Immich installation to a new computer and the procedure did not work until I added `-d immich` back into the command. The backup was generated by immich_db_dumper.

mmomjian:
> Hi, I just restored an Immich installation to a new computer and the procedure did not work until I added `-d immich` back into the command. The backup was generated by immich_db_dumper.

You are likely using an old setup of the DB dumper. Please check the Immich docs and make sure you have enabled POSTGRES_CLUSTER

RenanGreca:
Thanks for the quick reply @mmomjian, I'll update my setup. Perhaps this setup difference should be added as a note to the doc?

PR #8577, opened 2024-04-06
fix(web): reset to default button always being shown

mertalev:
## Description

The reset to default button is supposed to be hidden when the settings for a section are all default. However, some of the checks for this compare the entire config instead of a specific section, so the button is only hidden if the entire config is default. In the case of the transcoding settings, it was also comparing the transcoding settings to the full config and would never be hidden.

## How Has This Been Tested?

Tested that clicking `Reset` and saving makes the `Reset to default button` disappear, and changing a setting and saving makes it reappear.


danieldietzler (approved):
Sorry I missed that. Thanks for fixing!

jrasm91 (approved):

PR #8580, opened 2024-04-07
chore(server): better typing for system config key

mertalev:
## Description

The `SystemConfigKey` enum is a frequent source of bugs when changes are made to the config. One reason for this is because its values can be any string and don't necessarily need to be valid paths to the config object.

This PR tightens the typing for this enum such that any path not in `SystemConfig` is an error. To be able to do this, I changed it from an explicit enum to a constant object. 

This PR includes the fix in #8579.

## How Has This Been Tested?

Tested that the server starts up, that admin settings are changed successfully, and of course that invalid paths are caught by TypeScript.

### DIFF
--- a/server/src/entities/system-config.entity.ts
+++ b/server/src/entities/system-config.entity.ts
@@ -1,11 +1,11 @@
 import { ConcurrentQueueName } from 'src/interfaces/job.interface';
 import { Column, Entity, PrimaryColumn } from 'typeorm';
 
-export type SystemConfigValue = string | string[] | number | boolean;
+export type SystemConfigValue = string | number | boolean;
 
 // https://stackoverflow.com/a/47058976
 // https://stackoverflow.com/a/70692231
-type PathsToStringProps<T> = T extends SystemConfigValue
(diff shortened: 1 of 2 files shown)


server/src/entities/system-config.entity.ts
```diff
@@ -1,115 +1,136 @@
...
+
+type Join<T extends string[], D extends string> = T extends []
+  ? never
+  : T extends [infer F]
+    ? F
+    : T extends [infer F, ...infer R]
+      ? F extends string
+        ? `${F}${D}${Join<Extract<R, string[]>, D>}`
+        : never
+      : string;
```
jrasm91, line 22: What in the world. This is quite terrible lol. We should probably migrate to key value _per section_. Do like for the whole section json for a single key.

jrasm91 (approved):

server/src/entities/system-config.entity.ts
mertalev: I honestly don't know how it works except that it does lol

PR #8614, opened 2024-04-08
Cloudflare OAuth Guide

shanelord01:
Guide to setup immich with OAuth via Cloudflare Access


shanelord01:
> This is coming from #8299, right? How would you feel about instead adding a link to that discussion to the community projects page (https://immich.app/docs/community-projects)?

I'm just going off the request to add it to the guides section as per https://github.com/immich-app/immich/discussions/8299#discussioncomment-8947845

IMHO - it's not an add-on - it's a guide on how to configure OAuth.

alextran1502 (approved):

shanelord01:
Do I need to do something to fix this merge error? If so - can you let me know what?

mmomjian:
> This is coming from #8299, right? How would you feel about instead adding a link to that discussion to the community projects page (https://immich.app/docs/community-projects)?

I think this does fit better as a Guide, we have enough people on the team I believe that do use CloudFlare and it’s very popular with users as well. It’s not as esoteric as Podman, for example.

bo0tzz:
> we have enough people on the team I believe that do use CloudFlare

I'm not aware of anyone on the team that is specifically using cloudflare oauth. There are also plenty of other things that team members do use for which we wouldn't want to add guides. In a discussion with the other maintainers this morning we decided not to include tool-specific guides like this inside our docs.

#8812 adds a new community guides section that also includes a link to #8299, so I'll close this PR in favour of that one.

PR #8617, opened 2024-04-08
feat(mobile/web): add longer expirity for share link

akoscomp:
Sometimes peoples, like me, want to share an album to a longer period, but don't want to create an unlimited link.
For this option I add 1 year expirity for a share link.


waclaw66:
When making such changes, you should also provide web version update. Additionally a "3 months" option would be appretiated. "365 days" could be renamed to "1 year".

jrasm91:
I can add the web changes if you aren't familiar with web development.

akoscomp:
> I can add the web changes if you aren't familiar with web development.

Yes please. I'm not a developer, only a devops....

jrasm91 (approved):
LGTM

PR #8634, opened 2024-04-08
chore(web): specify that HDR videos will always be transcoded

patrontheo:
I spent quite a bit of time understanding why my videos were getting transcoded despite 'respecting' the transcoding policy.
I didn't see that written anywhere else so I thought it was a good place to inform users of that.


mertalev:
Thanks for adding this! I can see why the behavior here can be confusing.

mertalev (approved):

PR #8512, opened 2024-04-04
fix(server): motion photo asset linking & management

Ynng:
## Description

- close #8508 
- and close #8503 
  - link existing motion asset to motion photo if `checksum` and `library` matching
  - hide motion asset after linking (`isVisible` false)
  - prevent cross library motion asset linking
  	- refactor `getByChecksum` to take `libraryId` instead of `ownerId`. 
- close #8511
  - update storage usage on motion asset creation
  - exclude external library motion asset from storage usage
- close #8562 
  - skip non `isExternal` assets in external libraries during asset refresh
  - correctly delete external library motion assets 

## How Has This Been Tested?

Manually tested locally with MP, MVIMG and apple HEIF motion photo formats.
The video portion of each test photo was uploaded first before the image is uploaded.
The video asset is successfully hidden, and motion photo is linked properly.

Also tested with external library. The same motion photo is added to an external library and uploaded through the web gui. Immich correctly created 2 separate video assets.

## Discussion

This PR introduces a nuanced approach regarding the handling of video assets from motion photos in external libraries, potentially raising discussion.

Specifically, I mark video assets derived from motion photos in external libraries `isExternal = false`, despite their `library` being `EXTERNAL`. (what a mouthful)

It only makes sense to keep these video assets in the same external `library` as their corresponding motion photo, even though they are physically stored in the `encoded-video` folder internally. 

I'm interested to hear what your thoughts are

PR #8645, opened 2024-04-09
support livp photos

oldwang12:
Uploading photos in the format of livp through the iOS immich app will result in an error on the server.
```txt
[Nest] 1  - 04/08/2024, 1:22:53 PM   ERROR [AssetService] Unsupported file type IMG_4372.livp
```

I found that the code does not have the livp format here. Can you add the livp format here?

### DIFF
--- a/server/src/utils/mime-types.ts
+++ b/server/src/utils/mime-types.ts
@@ -19,6 +19,7 @@ const image: Record<string, string[]> = {
   '.gif': ['image/gif'],
   '.heic': ['image/heic'],
   '.heif': ['image/heif'],
+  '.livp': ['image/livp'],
   '.hif': ['image/hif'],
   '.iiq': ['image/iiq', 'image/x-phaseone-iiq'],
   '.insp': ['image/jpeg'],


mmomjian:
This PR is not complete - at a minimum it needs to be added to the spec.ts file as well. 

what testing have you done to confirm that the image libraries used in Immich support LIVP?

mertalev:
The livp format is apparently an archive containing a JPEG and a MOV file. This would require more than just adding it to the list.

mertalev:
Thanks for offering to help! The most relevant file is the [metadata service](https://github.com/immich-app/immich/blob/main/server/src/services/metadata.service.ts).

We handle something similar for motion photos: the video is embedded into the image so we extract it as a new asset and link the two. 

But there's an additional wrinkle in this case since the image portion also needs to be extracted. Every image and every video is its own asset, so thumbnail generation assumes the `originalPath` can be directly read, storage template can move that asset, etc. We could create a new asset for the image portion, link that to the video portion and hide the livp asset.

mertalev:
[This](https://github.com/immich-app/immich/pull/3175) PR is probably the most closely related.

We generally try to minimize the number of "special cases" to make job processing cleaner and more robust, so this is why I suggested making separate assets for the image and video. Using the livp asset as the image portion would have implications in other jobs as well as for things like asset serving.

@jrasm91 has the most context on live/motion photos, so they might have more thoughts on this.

mertalev:
The extracted videos for motion photos are put in the `encoded-videos` folder, separate from the external library. #8512 is a recent PR related to this that was later split up. I imagine it'd be similar in this case except that the image would go in the `thumbs` folder.

mertalev:
Closing this PR since it doesn't address the actual implementation needed to support this format.

PR #8656, opened 2024-04-09
adding documentation for quadlets

tbelway:
Adding documentation including sample files for quadlets with podman for ease of setup.

### DIFF
--- a/docs/docs/install/podman-quadlet.md
+++ b/docs/docs/install/podman-quadlet.md
@@ -1,0 +1,134 @@
+---
+sidebar_position: 90
+---
+
+# Podman deploy with quadlets
+
+You can deploy Immich on Podman using quadlets.
+
+Here are some sample rootless quadlet container files that can be placed in /etc/containers/systemd/users/${ID} where ID is the uid of whatever your rootless user is.
+
+Please note you'll need :z or :Z for selinux enabled hosts.
+
+immich-database.container
+```bash
+[Unit]
+Description=Immich Database
+Requires=immich-redis.service
+
+[Container]
+AutoUpdate=registry
+EnvironmentFile=${location_of_env_file}
+Image=registry.hub.docker.com/tensorchord/pgvecto-rs:pg16-v0.2.1
+Label=registry
+Network=slirp4netns:port_handler=slirp4netns
+PublishPort=5432:5432
+Volume=${host_database_directory}:/var/lib/postgresql/data:z
+Volume=/etc/localtime:/etc/localtime:ro
+
+[Service]
+Restart=always
+
+[Install]
+WantedBy=multi-user.target default.target
+```
+
+immich-microservices.container
+```bash
+[Unit]
+Description=Immich Microservices
+Requires=immich-redis.service immich-database.service
+
+[Container]
+#AddDevice=/dev/dri:/dev/dri #Needed for HWA
+#AddDevice=nvidia.com/gpu=0 #Needed for nvidia HWA, after setting up container tools
+AutoUpdate=registry
+EnvironmentFile=${location_of_env_file}
+Image=ghcr.io/immich-app/immich-server:release
+Label=registry
+Network=slirp4netns:port_handler=slirp4netns
+PublishPort=3002:3002
+Volume=${host_upload_directory}:/usr/src/app/upload:z
+Volume=/etc/localtime:/etc/localtime:ro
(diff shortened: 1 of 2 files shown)


bo0tzz:
None of the maintainers use podman/quadlet, so we don't have the knowledge needed to maintain this. It might be more appropriate to put this somewhere else (for example in the github discussions) and then maybe link to it from https://immich.app/docs/community-projects ?

andrewgdunn:
Maybe too much to ask but would it make sense to make a project under the immich group and have a couple podman users maintain a quadlet definition? Now that podman 5.x supports pod definitions in quadlet I was planning to move to that from my current systemd generated files.

Immich works great in podman, having a clear _adaptation_ to quadlet for folks may be interesting and it would create a set of users to __pin__ podman/quadlet support to (in discord).

docs/docs/install/podman-quadlet.md
```diff
@@ -0,0 +1,134 @@
...
+
+immich-microservices.container
+```bash
+[Unit]
+Description=Immich Microservices
+Requires=immich-redis.service immich-database.service
+
+[Container]
+#AddDevice=/dev/dri:/dev/dri #Needed for HWA
+#AddDevice=nvidia.com/gpu=0 #Needed for nvidia HWA, after setting up container tools
```
andrewgdunn, line 44: May want to consider [upstream documentation](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/latest/cdi-support.html) which I believe suggests that for podman use of CDI is suggested?

bo0tzz:
We've discussed this and decided we don't want to create a repository for external contributors under the immich-app organisation at the moment. However we will welcome a link on the community projects page 🙂

docs/docs/install/podman-quadlet.md
tbelway: If you are using CUDA you use AddDevice=nvidia.com/gpu=0 after setting up CDI, and if you're using VAAPI you would do AddDevice=/dev/dri:/dev/dri. You can find the information in the link you posted under: Running a Workload with CDI.

tbelway:
> Maybe too much to ask but would it make sense to make a project under the immich group and have a couple podman users maintain a quadlet definition? Now that podman 5.x supports pod definitions in quadlet I was planning to move to that from my current systemd generated files.
> 
> Immich works great in podman, having a clear _adaptation_ to quadlet for folks may be interesting and it would create a set of users to **pin** podman/quadlet support to (in discord).
> 
> Also @tbelway why not use of pod?

Not using a pod as I am in the process of doing a postgres replicated cluster, and I will be using that as well as a redis cluster, so this sample documentation makes it easier for anyone using quads to be up and running but also the flexibility to do as I am doing. :)

tbelway:
> None of the maintainers use podman/quadlet, so we don't have the knowledge needed to maintain this. It might be more appropriate to put this somewhere else (for example in the github discussions) and then maybe link to it from https://immich.app/docs/community-projects ?

It seems to be a bit hypocritcal to have a section that indicates podman can be used but may need additional setup, but then not to accept the PR for the actual instructions for said additional setup instructions (as found here: https://immich.app/docs/install/requirements/).

I'd be more than happy to update the quadlets section, and you can put a banner indicating the quadlets section is community maintained if you want, but using code as documentation I don't even see that as being necessary. Even if you switch away from redis to postgres queuing that just means the elimination of a quadlet file, or if you switch to KeyDB in place of redis, same thing... small change to fix.

bo0tzz:
If the quadlets config is hosted somewhere else like I suggested, it would make sense to add a link to that in the requirements note as well. I will stick with not including the quadlets configuration in the immich repository though, for the reasons discussed above.

tbelway:
> If the quadlets config is hosted somewhere else like I suggested, it would make sense to add a link to that in the requirements note as well. I will stick with not including the quadlets configuration in the immich repository though, for the reasons discussed above.

OK. I can work on that this evening and go from there. If that's the case though, I'll likely go a touch bit more comprehensive.

tbelway:
> If the quadlets config is hosted somewhere else like I suggested, it would make sense to add a link to that in the requirements note as well. I will stick with not including the quadlets configuration in the immich repository though, for the reasons discussed above.

https://github.com/tbelway/immich-podman-quadlets/blob/main/docs/install/podman-quadlet.md

PR #8659, opened 2024-04-09
fix(server): use correct file extension for motion photo videos

### LINKED ISSUE #8658: Motion photo video extension is .jpg in archive downloads
### The bug

When downloading an archive that includes motion photos, two assets per motion photo are added to the archive.

The file extension of the video is `jpg`.

To be consistent with the single file download, it should be `mp4` (or something similar)

```
PXL_20240407_130345095.MP+1.jpg       <---- is a video file
PXL_20240407_130345095.MP.jpg
```

Alpine

### Version of Immich Server

v1.101.0

v1.101.0

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
Not relevant
```

### Your .env content

```Shell
Not relevant
```

### Reproduction steps

```bash
1. Upload a motion photo from a Pixel
2. Download that photo in an archive
3. The video part has `jpg` extension
```

### Additional information

_No response_

lukashass:
Closes #8658 

Set `originalFileName` extension of motion assets to `mp4`

## TODO

- [x] migration for existing assets


lukashass:
@alextran1502 I saw that since #7679 the file extension for the archive download is only taken from `originalFileName`. Maybe a better fix than this one would be to set the correct extension in that property in the first place :thinking:

danieldietzler:
Yeah, IMO `originalFileName` should always have the correct file name.

lukashass:
On the other hand I'm wondering if the video should be downloaded at all, since it seems to still be embedded in the downloaded jpg and thus downloaded twice.

jrasm91 (commented):
I think we should probably fix the original filename which is wrong, no?

lukashass:
The video is extracted from the jpg on upload, so the filename is kind of correct (since it says "original").

It is a little ambiguous, that's why I'm questioning the download of the separate video altogether 🤔

lukashass:
So from what I'm reading there are motion photo formats which
- embed the video in the jpg
- are two separate files from the beginning

This means the correct fix for the issue is to only download a separate video when the jpg does not contain the video, right?

How would we detect that? I could think of two ways:
- save this information (separate/embedded video) in a new property
- compare the file extension of `originalFileName` of video and photo, if the same, do not download the video

server/src/services/download.service.ts
```diff
@@ -96,7 +96,8 @@ export class DownloadService {
 
       const { originalPath, originalFileName } = asset;
 
-      let filename = originalFileName;
+      const extension = parse(originalPath).ext;
+      let filename = `${parse(originalFileName).name}${extension}`;
```
jrasm91, line 100: Why does it have the wrong extension to begin with? I think I would rather fix the root cause instead of patching it everyone it is used.

lukashass: @jrasm91 I agree (sort of). Hence my comments https://github.com/immich-app/immich/pull/8659#issuecomment-2046140774 and https://github.com/immich-app/immich/pull/8659#issuecomment-2072999502.

What is your opinion on that?

jrasm91: iOS doesn't have it embedded but Android does. Either way getting the extracted video might be nice when you download it. The name for the motion asset should come from the original when it is uploaded, but that should not include the extension...

justinvdk: A filename includes the extension (by definition), which makes it confusing as that does not match what users perceive as such.

Having said that I'd say its fair to transform the (original)filename of files when converted (thus, if I am following, never having the incorrect extension for these records at time of download).

jrasm91: Pretty sure the old extension is coming from here:

https://github.com/immich-app/immich/blob/main/server/src/services/metadata.service.ts#L441

If you are going to fix anything, it should be here instead. It is probably possible to write a migration to fix affected files as well. 

IMO we should store the correct filename and extension when the asset is created and then never touch or need to touch/modify if moving forward. The current solution also introduces the possibility of the extension case changing between upload and download.

server/src/migrations/1715435221124-MotionAssetExtensionMP4.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export class MotionAssetExtensionMP41715435221124 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `UPDATE "assets" SET "originalFileName" = regexp_replace("originalFileName", '\\.[a-zA-Z0-9]+$', '.mp4') WHERE "originalPath" LIKE '%.mp4' AND "isVisible" = false`,
+    );
+  }
+
+  public async down(): Promise<void> {
+    // TODO
```
jrasm91, line 11: probably can leave blank

server/src/services/metadata.service.ts
```diff
@@ -438,7 +438,7 @@ export class MetadataService {
           checksum,
           ownerId: asset.ownerId,
           originalPath: motionPath,
-          originalFileName: asset.originalFileName,
+          originalFileName: `${path.parse(asset.originalFileName).name}.mp4`,
```
jrasm91, line 441: perfect

jrasm91 (commented):
Looks great imo. Once the tests are passing I think we'll be good to go.

lukashass:
@jrasm91 I added the `originalFileName` to the asset stub that is used in the metadata test.

This tripped up another a "storage template move" test because moving uses asset ids (`const filename = asset.originalFileName || asset.id;`) as a fallback for non-existent `originalFileName`.

Outside of tests, I don't see how an asset can __not__ have an `originalFileName`. This feels a little like a bug, but maybe I'm missing something. 

Should I just create a new stub to not interfere with other tests?

jrasm91:
> Should I just create a new stub to not interfere with other tests?
This is fine.

jrasm91 (approved):

danieldietzler (approved):
LGTM

PR #8665, opened 2024-04-09
docs: Pin to Postgres v14 in postgres-backup-local

mmomjian:
We currently don't pin to a Docker image with the database dumper, which means they are created using PG v16. This is not good as Immich uses v14. We now pin to PG v14.

### DIFF
--- a/docs/docs/administration/backup-and-restore.md
+++ b/docs/docs/administration/backup-and-restore.md
@@ -63,7 +63,7 @@ services:
   ...
   backup:
     container_name: immich_db_dumper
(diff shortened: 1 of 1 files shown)


bo0tzz (approved):
Since this runs as a normal postgres client it's probably fine, but it's still good to be precise about it. Thanks!

mmomjian:
> Since this runs as a normal postgres client it's probably fine, but it's still good to be precise about it. Thanks!

I just helped someone on Discord who had a backup that couldn't be imported because `LOCALE_PROVIDER` was added in PG 15. It errored out trying to create the databases.

PR #8720, opened 2024-04-11
feat: readonly album sharing

mgabor3141:
First significant contribution.

This PR adds the ability for album owners to share an album with a specific user without letting them add images to or remove images from the album. It includes database, backend, API as well as frontend changes to achieve this.

Screenshot 2024-04-12 at 19 00 10

Relates to #1607


server/src/entities/album.entity.ts
```diff
@@ -53,7 +53,7 @@ export class AlbumEntity {
   albumThumbnailAssetId!: string | null;
 
   @ManyToMany(() => UserEntity)
-  @JoinTable()
+  @JoinTable({name: "album_permissions"})
```
alextran1502, line 56: I don't think this is correct and it will break existing installation

mgabor3141: Yeah, I'm still checking

server/src/repositories/album-permission.repository.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  async update(userId: string, albumId: string, dto: Partial<AlbumPermissionEntity>): Promise<AlbumPermissionEntity> {
+    // @ts-expect-error I'm pretty sure I messed something up with the entity because
+    // if I follow what typescript says I get postgres errors
+    await this.repository.update({ users: userId, albums: albumId }, dto);
+    return this.repository.findOneOrFail({
+      where: { users: Equal(userId), albums: Equal(albumId) },
+      relations: { users: true },
+    });
```
mgabor3141, line 26: This is when I don't include an `Equals` call. (See other comment)

server/src/repositories/album.repository.ts
```diff
@@ -23,7 +23,7 @@ export class AlbumRepository implements IAlbumRepository {
   getById(id: string, options: AlbumInfoOptions): Promise<AlbumEntity | null> {
     const relations: FindOptionsRelations<AlbumEntity> = {
       owner: true,
-      sharedUsers: true,
+      albumPermissions: { users: true },
```
mgabor3141, line 26: How "deep" these selects go tends to be very crucial for some calls as far as I can tell. I'm not very confident about changing them, but adding `users: true` has worked well in my tests for a bit now.

server/src/services/album.service.ts
```diff
@@ -212,14 +215,14 @@ export class AlbumService {
   async addUsers(auth: AuthDto, id: string, dto: AddUsersDto): Promise<AlbumResponseDto> {
     await this.access.requirePermission(auth, Permission.ALBUM_SHARE, id);
 
-    const album = await this.findOrFail(id, { withAssets: false });
+    const album = await this.findOrFail(id, { withAssets: true });
```
mgabor3141, line 218: The old `return` query of this function returned assetCount, so I adapted that to happen here now instead. The mapping still happens without assets in both cases.

server/src/services/album.service.ts
```diff
@@ -262,11 +261,25 @@ export class AlbumService {
...
 
-    await this.albumRepository.update({
-      id: album.id,
-      updatedAt: new Date(),
-      sharedUsers: album.sharedUsers.filter((user) => user.id !== userId),
-    });
+    await this.albumPermissionRepository.delete(userId, id);
+  }
+
+  async setAlbumPermission(
```
mgabor3141, line 267: Similar to the controller, this could in theory have its own service. Question is does it need to?

web/src/routes/(user)/albums/[albumId]/+page.svelte
```diff
@@ -448,14 +461,14 @@
             {/if}
 
             {#if album.assetCount > 0}
+              <CircleIconButton title="Slideshow" on:click={handleStartSlideshow} icon={mdiPresentationPlay} />
```
Comment (mgabor3141) [inline web/src/routes/(user)/albums/[albumId]/+page.svelte:464]: This one line being moved has nothing to do with the rest of the PR. Let me know if I should separate it into its own PR.

Issue: the album slideshow is only available to album owners because the button is under the [...] button which is owner only

I don't think this makes sense, I see two solutions:
- make the [...] button available to everyone and make the specific options underneath (such as delete) owner only
- move the slideshow out from under the [...] button.

This is the latter.

Comment (mgabor3141) [reply web/src/routes/(user)/albums/[albumId]/+page.svelte:464]: I moved this to #8757

server/src/migrations/1712905931092-SetReadonlyDefault.ts
```diff
@@ -0,0 +1,14 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class SetReadonlyDefault1712905931092 implements MigrationInterface {
+    name = 'SetReadonlyDefault1712905931092'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "albums_shared_users_users" ALTER COLUMN "readonly" SET DEFAULT true`);
```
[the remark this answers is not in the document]
jrasm91: I think it would be cleaner to dry the default to false and run a SQL update afterwards.

server/src/repositories/access.repository.ts
```diff
@@ -118,18 +119,22 @@ class AlbumAccess implements IAlbumAccess {
...
+        relations: { albumPermissions: true },
+        // -@ts-expect-error asd
         where: {
           id: In([...albumIds]),
-          sharedUsers: {
-            id: userId,
+          albumPermissions: {
+            users: Equal(userId),
+            // If write is needed we check for it, otherwise both are accepted
+            readonly: readWrite === 'write' ? false : undefined,
```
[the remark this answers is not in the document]
jrasm91: You should match on user.id

server/src/repositories/album-permission.repository.ts
jrasm91: You should add user id and album id to the entity

server/src/services/album.service.ts
jrasm91: Can you remember this to update permission?

server/src/repositories/album.repository.ts
jrasm91: You could replace everywhere you see shared users with permissions/users: true

server/src/repositories/album-permission.repository.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  async delete(userId: string, albumId: string): Promise<void> {
```
jrasm91, line 29: Can you make a new type? AlbumPermissionId { album id: string, userId: string } and use the instead? I don't like two string IDs that might get sent in the wrong order.

server/src/repositories/album-permission.repository.ts
```diff
@@ -0,0 +1,32 @@
...
+import { InjectRepository } from '@nestjs/typeorm';
+import { Equals } from 'class-validator';
+import { AlbumPermissionEntity } from 'src/entities/album-permission.entity';
+import { IAlbumPermissionRepository } from 'src/interfaces/album-permission.interface';
+import { Instrumentation } from 'src/utils/instrumentation';
+import { Equal, Repository } from 'typeorm';
+
+@Instrumentation()
+@Injectable()
+export class AlbumPermissionRepository implements IAlbumPermissionRepository {
```
jrasm91, line 11: I might just put this in the album repository?
- addSharedUser
- updateSharedUser 
- removeSharedUser

server/src/entities/album.entity.ts
```diff
@@ -52,9 +53,8 @@ export class AlbumEntity {
   @Column({ comment: 'Asset ID to be used as thumbnail', nullable: true })
   albumThumbnailAssetId!: string | null;
 
-  @ManyToMany(() => UserEntity)
-  @JoinTable()
-  sharedUsers!: UserEntity[];
+  @OneToMany(() => AlbumPermissionEntity, (permission) => permission.albums)
+  albumPermissions!: AlbumPermissionEntity[];
```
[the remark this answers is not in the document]
jrasm91: I think it would make more sense to call the entity AlbumUser instead.

server/src/entities/album-permission.entity.ts
```diff
@@ -0,0 +1,20 @@
...
+@Index('IDX_427c350ad49bd3935a50baab73', ['albums'])
+@Index('IDX_f48513bf9bccefd6ff3ad30bd0', ['users'])
+export class AlbumPermissionEntity {
+  @PrimaryColumn({ type: 'uuid', name: 'albumsId' })
+  @ManyToOne(() => AlbumEntity, { onDelete: 'CASCADE', onUpdate: 'CASCADE', nullable: false })
+  albums!: AlbumEntity;
+
+  @PrimaryColumn({ type: 'uuid', name: 'usersId' })
+  @ManyToOne(() => UserEntity, { onDelete: 'CASCADE', onUpdate: 'CASCADE', nullable: false })
+  users!: UserEntity;
```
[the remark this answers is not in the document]
jrasm91: You can set the name explicitly if you didn't want to use the default one. I think it is with `@JoinTable()`

server/src/dtos/album.dto.ts
```diff
@@ -92,7 +102,9 @@ export class AlbumResponseDto {
   updatedAt!: Date;
   albumThumbnailAssetId!: string | null;
   shared!: boolean;
+  @ApiProperty({ deprecated: true, description: 'Deprecated in favor of albumPermissions' })
   sharedUsers!: UserResponseDto[];
+  albumPermissions!: AlbumPermissionResponseDto[];
```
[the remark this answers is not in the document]
jrasm91: I think that is a good idea

server/src/entities/album-permission.entity.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  @Column({ default: true })
+  readonly!: boolean;
```
jrasm91, line 19: I think this should be a varchar with values like read, write, and maybe admin later

jrasm91 (commented):
Pretty good start and seems mostly correct. Just needs some clean up.

server/src/controllers/album.controller.ts
```diff
@@ -96,4 +97,14 @@ export class AlbumController {
   ) {
     return this.service.removeUser(auth, id, userId);
   }
+
+  @Put(':id/permission/:userId')
+  setAlbumPermission(
```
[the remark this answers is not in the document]
danieldietzler: Since it's only that one endpoint it's fine IMO

server/src/repositories/album.repository.ts
mgabor3141: That's what I did, yes.

server/src/migrations/1712905931092-SetReadonlyDefault.ts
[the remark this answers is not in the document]
mgabor3141: I'm not sure what you mean

server/src/repositories/album-permission.repository.ts
mgabor3141: I would prefer to keep it, I think it's more readable this way.

server/src/migrations/1712905931092-SetReadonlyDefault.ts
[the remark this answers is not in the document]
mgabor3141: I think I got it, please check the latest migration and confirm

server/src/repositories/access.repository.ts
```diff
@@ -118,18 +119,21 @@ class AlbumAccess implements IAlbumAccess {
...
 
     return this.albumRepository
       .find({
         select: { id: true },
+        relations: { sharedUsers: true },
         where: {
           id: In([...albumIds]),
           sharedUsers: {
-            id: userId,
+            user: { id: userId },
```
mgabor3141, line 134: `sql:generate` still gives the following message here:

```
AccessRepository.album.checkSharedAlbumAccess error: QueryFailedError: invalid input syntax for type uuid: "{"id":"00000000-0000-4000-a000-000000000000"}"
```

Same thing happens when calling the endpoint. Strange, because this is what TS is happy with.

`relations: { sharedUsers: { user: true } },` doesn't help either.

@jrasm91 can you please take a look, I don't know how to proceed.

server/src/migrations/1713298646379-AddAlbumUserReadonly.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class AddAlbumUserReadonly1713298646379 implements MigrationInterface {
+    name = 'AddAlbumUserReadonly1713298646379'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "albums_shared_users_users" ADD "readonly" boolean NOT NULL DEFAULT false`);
+        await queryRunner.query(`ALTER TABLE "albums_shared_users_users" ALTER COLUMN "readonly" SET DEFAULT true`);
```
mgabor3141, line 8: First row was generated, I added the second row manually after.

server/src/repositories/access.repository.ts
mgabor3141: Figured it out! 🙂

Found the solution here:
https://github.com/typeorm/typeorm/pull/8616#:~:text=You%20still%20have%20an%20ability%20to%20use%20foreign%20keys%20as%20your%20primary%20keys

mgabor3141:
Only tests remain AFAIK

server/src/cores/access.core.ts
```diff
@@ -142,6 +144,12 @@ export class AccessCore {
           : new Set();
       }
 
+      case Permission.ALBUM_WRITE: {
+        return sharedLink.allowUpload
+          ? await this.repository.album.checkSharedLinkAccess(sharedLinkId, ids)
+          : new Set();
+      }
+
```
jrasm91, line 152: I don't think the album details should should be updated via share link authentication.

server/src/cores/access.core.ts
```diff
@@ -215,7 +223,21 @@ export class AccessCore {
 
       case Permission.ALBUM_READ: {
         const isOwner = await this.repository.album.checkOwnerAccess(auth.user.id, ids);
-        const isShared = await this.repository.album.checkSharedAlbumAccess(auth.user.id, setDifference(ids, isOwner));
+        const isShared = await this.repository.album.checkSharedAlbumAccess(
+          auth.user.id,
+          setDifference(ids, isOwner),
+          AlbumUserRole.Viewer,
```
jrasm91, line 229: this looks slick.

server/src/dtos/album.dto.ts
```diff
@@ -92,7 +105,9 @@ export class AlbumResponseDto {
   updatedAt!: Date;
   albumThumbnailAssetId!: string | null;
   shared!: boolean;
+  @ApiProperty({ deprecated: true, description: 'Deprecated in favor of sharedUsersV2' })
   sharedUsers!: UserResponseDto[];
+  sharedUsersV2!: AlbumUserResponseDto[];
```
jrasm91, line 110: I like the term `AlbumUsers` that you have used everywhere. What do you think permanently changing this to `albumUsers`. It would basically mean `album.albumUser.user` instead of `album.users.user` when it is accessed in the code.

server/src/entities/album-user.entity.ts
```diff
@@ -0,0 +1,31 @@
+import { AlbumEntity } from 'src/entities/album.entity';
+import { UserEntity } from 'src/entities/user.entity';
+import { Column, Entity, Index, JoinColumn, ManyToOne, PrimaryColumn } from 'typeorm';
+
+export enum AlbumUserRole {
+  Editor = 'editor',
+  Viewer = 'viewer',
```
jrasm91, line 7: Enums in the server are all caps right now, even though I think I might prefer to move to this style. Can you make them all caps to be consistent though?

server/src/entities/album-user.entity.ts
```diff
@@ -0,0 +1,31 @@
...
+@Index('IDX_427c350ad49bd3935a50baab73', ['album'])
+@Index('IDX_f48513bf9bccefd6ff3ad30bd0', ['user'])
+export class AlbumUserEntity {
+  @PrimaryColumn({ type: 'uuid', name: 'albumsId' })
+  albumId!: string;
+
+  @PrimaryColumn({ type: 'uuid', name: 'usersId' })
+  userId!: string;
+
+  @JoinColumn({ name: 'albumsId' })
```
jrasm91, line 21: Nice it looks like you figured this out, which is great. We can also do a migration later to rename the columns if we want to (and then remove this configuration)

server/src/interfaces/access.interface.ts
```diff
@@ -1,5 +1,9 @@
+import { AlbumUserRole } from 'src/entities/album-user.entity';
+
 export const IAccessRepository = 'IAccessRepository';
 
+export type ReadWrite = 'read' | 'write';
```
jrasm91, line 5: I think you can delete this now.

server/src/migrations/1713337511945-AddAlbumUserRole.ts
```diff
@@ -0,0 +1,15 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class AddAlbumUserRole1713337511945 implements MigrationInterface {
+    name = 'AddAlbumUserRole1713337511945'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "albums_shared_users_users" ADD "role" character varying NOT NULL DEFAULT 'editor'`);
```
jrasm91, line 7: Looks perfect

server/src/repositories/access.repository.ts
```diff
@@ -81,7 +82,8 @@ class ActivityAccess implements IActivityAccess {
     return this.albumRepository
       .createQueryBuilder('album')
       .select('album.id')
-      .leftJoin('album.sharedUsers', 'sharedUsers')
+      .leftJoin('album.sharedUsers', 'albumSharedUsers')
+      .leftJoin('albumSharedUsers.user', 'sharedUsers')
```
jrasm91, line 86: this looks good, but maybe in the future we can look at renaming sharedUsers to albumUsers in the entity itself.

server/src/repositories/access.repository.ts
```diff
@@ -120,18 +122,22 @@ class AlbumAccess implements IAlbumAccess {
...
-  async checkSharedAlbumAccess(userId: string, albumIds: Set<string>): Promise<Set<string>> {
+  async checkSharedAlbumAccess(userId: string, albumIds: Set<string>, access: AlbumUserRole): Promise<Set<string>> {
     if (albumIds.size === 0) {
       return new Set();
     }
 
     return this.albumRepository
       .find({
         select: { id: true },
+        relations: { sharedUsers: true },
```
jrasm91, line 133: I don't think this line is necessary.

server/src/repositories/album.repository.ts
```diff
@@ -62,9 +62,9 @@ export class AlbumRepository implements IAlbumRepository {
     return this.repository.find({
       where: [
         { ownerId, assets: { id: assetId } },
-        { sharedUsers: { id: ownerId }, assets: { id: assetId } },
+        { sharedUsers: { user: Equal(ownerId) }, assets: { id: assetId } },
```
jrasm91, line 65: fix this one please. it should be { user: { id: ownerId } } I think. Or potentially { userId: ownerId } since that column is on the album user entity now.

server/src/repositories/album.repository.ts
```diff
@@ -141,11 +141,11 @@ export class AlbumRepository implements IAlbumRepository {
   @GenerateSql({ params: [DummyValue.UUID] })
   getShared(ownerId: string): Promise<AlbumEntity[]> {
     return this.repository.find({
-      relations: { sharedUsers: true, sharedLinks: true, owner: true },
+      relations: { sharedUsers: { user: true }, sharedLinks: true, owner: true },
       where: [
-        { sharedUsers: { id: ownerId } },
+        { sharedUsers: { user: Equal(ownerId) } },
```
jrasm91, line 146: fix this one as well

server/src/services/album.service.ts
```diff
@@ -190,7 +192,7 @@ export class AlbumService {
   async removeAssets(auth: AuthDto, id: string, dto: BulkIdsDto): Promise<BulkIdResponseDto[]> {
     const album = await this.findOrFail(id, { withAssets: false });
 
-    await this.access.requirePermission(auth, Permission.ALBUM_READ, id);
+    await this.access.requirePermission(auth, Permission.ALBUM_WRITE, id);
```
jrasm91, line 195: This is an interesting change, but I think it is probably fine.

server/src/services/album.service.ts
jrasm91: Yeah we can't be using withAssets: true here. For large albums this loads way too much data from the database.

server/src/services/album.service.ts
```diff
@@ -262,11 +258,20 @@ export class AlbumService {
...
 
-    await this.albumRepository.update({
-      id: album.id,
-      updatedAt: new Date(),
-      sharedUsers: album.sharedUsers.filter((user) => user.id !== userId),
-    });
+    await this.albumUserRepository.delete({ albumId: id, userId });
+  }
+
+  async updateAlbumUser(auth: AuthDto, id: string, userId: string, dto: Partial<AlbumUserEntity>): Promise<void> {
```
jrasm91, line 264: IMO when we are in the album service, using album in the name is redundant. We have addUser, removeUser, so updateUser would be fine I think.

server/src/services/album.service.ts
```diff
@@ -262,11 +258,20 @@ export class AlbumService {
...
+    await this.access.requirePermission(auth, Permission.ALBUM_SHARE, id);
+
+    const album = await this.findOrFail(id, { withAssets: false });
```
jrasm91, line 267: This is probably fine for now, but since we have an entity for this intersection table now, we'll eventually want to transition to a `getUser` repository method that gets a user by albumId and userId. Rather than loading every single user and filtering through them.

jrasm91 (commented):
Server code looks really good. A few more things to clean up and I think it'll be ready to merge. Thanks for you work on this it looks great!

server/src/entities/album-user.entity.ts
jrasm91: Actually I don't care if you change this or not. I'm going to start using this format myself I think.

server/src/cores/access.core.ts
mgabor3141: I believe this doesn't allow that. This is `ALBUM_WRITE`, which gives the shared link, if `allowUpload` has been enabled, the ability to add photos to the album. Modifying album details are behind `ALBUM_UPDATE` as seen here:

https://github.com/mgabor3141/immich/blob/cecb99b4b6edadcb3e4e841df4a83318f580db04/server/src/services/album.service.ts#L139

server/src/repositories/access.repository.ts
mgabor3141: Did that too for consistency

server/src/repositories/album.repository.ts
mgabor3141: Thanks, I missed these

server/src/services/album.service.ts
mgabor3141: Should I open a ticket? How are these tracked?

jrasm91: You can add a comment in the code here if you want and/or create a new issue on github.

web/src/lib/components/album-page/share-info-modal.svelte
```diff
@@ -78,7 +99,15 @@
           <p class="text-sm">Owner</p>
         </div>
       </div>
-      {#each album.sharedUsers as user}
+      {#each album.albumUsers.toSorted((a, b) => {
```
jrasm91, line 102: Maybe it would be better to just have this order applied server-side instead?

server/src/interfaces/access.interface.ts
```diff
@@ -20,7 +22,7 @@ export interface IAccessRepository {
 
   album: {
     checkOwnerAccess(userId: string, albumIds: Set<string>): Promise<Set<string>>;
-    checkSharedAlbumAccess(userId: string, albumIds: Set<string>): Promise<Set<string>>;
+    checkSharedAlbumAccess(userId: string, albumIds: Set<string>, access: AlbumUserRole): Promise<Set<string>>;
```
danieldietzler, line 25: Nit: What do you think of renaming `access` to `role`? To me this fits better into the general naming you chose

danieldietzler (approved):
This looks great to me!

server/src/interfaces/access.interface.ts
mgabor3141: I gave this a different name intentionally, because this parameter will be used a little differently than other places. This is to communicate that we are not looking for that exact role, but rather Viewer access (Viewer and Editor roles) or Editor access (only Editor role)

server/src/dtos/album.dto.ts
```diff
@@ -109,13 +124,29 @@ export class AlbumResponseDto {
...
+  albumUsers.sort((a, b) => {
+    if (a.role === AlbumUserRole.VIEWER && b.role === AlbumUserRole.EDITOR) {
+      return 1;
+    }
+    if (a.role === AlbumUserRole.EDITOR && b.role === AlbumUserRole.VIEWER) {
+      return -1;
+    }
+    return a.user.name.localeCompare(b.user.name);
+  });
+
```
jrasm91, line 149: can you just use lodash orderBy? `_.orderBy(albumUsers, ['role', 'user.name'], ['DESC', 'ASC'])`

jrasm91 (approved):
Noice!

alextran1502:
Hello, I just ran a functional test on this PR and had a few thoughts.

- It would be more intuitive to let the album owner set permission when adding a user to a shared album instead of adding it later, thus changing the known default behavior.
- What plan do you have for the mobile UI?

mgabor3141:
> It would be more intuitive to let the album owner set permission when adding a user to a shared album instead of adding it later, thus changing the known default behavior.

That's a good point, but I don't think this is a dealbreaker for the PR. I think that it is a good followup immediately after. Do you agree?

> What plan do you have for the mobile UI?

This on the other hand could be a dealbreaker, but I'm not too familiar with the mobile stack. I can take a look, but I wouldn't mind some help 🙂

alextran1502:
> That's a good point, but I don't think this is a dealbreaker for the PR. I think that it is a good followup immediately after. Do you agree?

I disagree, as changing a known behavior without clear indication will make a lot of users confused

mgabor3141:
Screenshot 2024-04-24 at 12 58 26

This might look a little weird, but is a lot more usable. Both because of accessibility as well as just with less clicking needed.

mgabor3141:
Screenshot 2024-04-24 at 13 26 23Screenshot 2024-04-24 at 13 26 30

Played around with the design a bit. I think this is better for accessibility, and also for usability as less clicking needs to happen. Thoughts?

alextran1502:
> Screenshot 2024-04-24 at 13 26 23Screenshot 2024-04-24 at 13 26 30
> 
> Played around with the design a bit. I think this is better for accessibility and also for usability as less clicking needs to happen. Thoughts?
> 
> I'm not too sure about the wording. Would Editor**s** (plural) look a little better?

I think with this design, you cannot add UserA as Editor and UserB as Viewer in a single operation

mgabor3141:
> I think with this design, you cannot add UserA as Editor and UserB as Viewer in a single operation

Yes, and I like that. Say you want to add users as 50/50 editors and viewers. Adding one group first and then the rest in another operation is much less clicks than having to click three times for one half before clicking add.

The possibility is also there to just add everyone then change the role of a few users on the other screen.

But the most common usecase I think is to add all users as either one, which works a lot better here.

alextran1502:
I think there might be a misunderstanding with my previous comment; I would like to be able to set the permission for users individually, similar to your first UI work

image

mgabor3141:
What do you think about adding a `Custom` option to the bottom dropdown that would enable the individual dropdowns on each row? That would be the best of both solutions.

Also I won't have too much time to work on this for the rest of this week. I'll see what I can do.

alextran1502:
> What do you think about adding a Custom option to the bottom dropdown that would enable the individual dropdowns on each row? That would be the best of both solutions.

Please don't. I'd prefer a streamlined operation. Introducing many variations is not a good design; too many choices will confuse the common user. I understand that this serves your specific use case; however, when we add a feature, we have to think about the majority of the user group.

jrasm91:
image
image
image

mgabor3141:
To add some more references to this topic, this is how Google Docs does its sharing modal:

Screenshot 2024-04-24 at 22 54 45

Here it also takes multiple operations to add users as multiple different roles, but it is very clear in what's going on.

Compared to Google Photos:
Screenshot 2024-04-24 at 22 58 04

Google Photos doesn't have readonly sharing as a concept, at least not from the add screen. It lists all users and one click simply adds them to the shared album. Even if you do link sharing, whoever opens the link has a `Join` button that they can click to add themselves just as if they were directly invited.

"Readonly" happens in the options panel, where the album owner can toggle adding and removing photos GLOBALLY for all joined users (they call it "collaborate"):

Screenshot 2024-04-24 at 23 03 57

I think the pattern I have in mind is a really effective one. People who are present on a trip or at an event can all join an album as Editors and upload all their photos. They can then add whoever they want as Viewers. The really cool thing is that everyone's Albums page will be populated with all the various albums they have access to (as any role). This is something that link sharing doesn't really achieve.

This pattern depends on being able to add many viewers easily. This version with three clicks per viewer user per album is I think a non-starter for many users who would like to adopt a similar use pattern with regards to sharing.

mgabor3141:
@jrasm91 I really like what you did with the modal. I think this one should be the "options/edit users" modal, and there could be a separate add users modal that's more closer to the Google Docs style interface.

That's all I wanted to say, whatever we go with is going to be better than it was before. Thanks for reading me ramble about it.

alextran1502 (approved):
Thank you very much! I know there were a lot of back and forth about the UI. I am open for additional user feedback, and we will make changes if need

PR #8724, opened 2024-04-11
fix(server): link motion photo with existing video asset

### LINKED ISSUE #8503: Some Live Photos are Split (Again)
### The bug

I have found another instance of bug #3228.

I setup Immich and used my phone to upload all the photos locally stored there, then i downloaded all photos from google photos using takeout and then imported them to Immich.

Live photos uploaded from the phone worked flawlessly but when the import from google photos happened an additional video file was added. Photos that where not on the phone but where on the google photos zip file did not become live photos and kept the photo and video file separated.

After reading issue #3228, I verified that all jobs where done and then tried to run metadata job but this did not fix the issue.

Here is an example (I don't mind showing the location since is an old picture and its not very specific):

image
image

As you can see both images have metadata and they share the same name(minus the extension), same location and same capture time. So I don't know why they wouldn't merge (perhaps MP4 is not supported?).

unraid 6.12.9

### Version of Immich Server

1.100.0

N/A

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1. Set up Immich
2. Back up photos from phone app.
3. Download same photos from google takeout
4. Import downloaded photos from google take out.
```

### Additional information

_No response_
```[tasklist]
### Tasks
```

Ynng:
Closes #8508, closes #8503.
Originally part of #8512.

Links existing video asset with motion photo if their checksum matches.

Should NOT be merged until #8719 is merged to prevent cross library linking.

### DIFF
--- a/server/src/services/metadata.service.spec.ts
+++ b/server/src/services/metadata.service.spec.ts
@@ -439,7 +439,7 @@ describe(MetadataService.name, () => {
       });
       cryptoRepository.hashSha1.mockReturnValue(randomBytes(512));
       assetMock.getByChecksum.mockResolvedValue(null);
-      assetMock.create.mockResolvedValue(assetStub.livePhotoMotionAsset);
+      assetMock.create.mockImplementation((asset) => Promise.resolve({ ...assetStub.livePhotoMotionAsset, ...asset }));
 
       await sut.handleMetadataExtraction({ id: assetStub.livePhotoStillAsset.id });
       expect(jobMock.queue).toHaveBeenNthCalledWith(2, {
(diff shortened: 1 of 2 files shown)


server/src/services/metadata.service.spec.ts
```diff
@@ -439,7 +439,7 @@ describe(MetadataService.name, () => {
       });
       cryptoRepository.hashSha1.mockReturnValue(randomBytes(512));
       assetMock.getByChecksum.mockResolvedValue(null);
-      assetMock.create.mockResolvedValue(assetStub.livePhotoMotionAsset);
+      assetMock.create.mockImplementation((asset) => Promise.resolve({ ...assetStub.livePhotoMotionAsset, ...asset }));
```
danieldietzler, line 442: AFAICT the create function is only called once, correct? Why can't we use `mockResolvedValue` then, given that we already know which asset we pass in.

danieldietzler (approved):
Your PRs are really good, thank you! :)

server/src/services/metadata.service.ts
```diff
@@ -438,11 +444,14 @@ export class MetadataService {
         this.storageCore.ensureFolders(motionPath);
         await this.storageRepository.writeFile(motionAsset.originalPath, video);
         await this.jobRepository.queue({ name: JobName.METADATA_EXTRACTION, data: { id: motionAsset.id } });
-        await this.assetRepository.update({ id: asset.id, livePhotoVideoId: motionAsset.id });
+      }
 
+      if (asset.livePhotoVideoId !== motionAsset.id) {
```
jrasm91, line 449: does this whole method run even when an asset is already linked?

jrasm91 (approved):
LGTM

PR #8729, opened 2024-04-11
fix(mobile): show error details in the log when available

shenlong-tanwen:
#### Changes made:

- Updated the export format of log from the mobile app and gave it a bit more structure to easily skim through it
- Added the additional detail to the exported log if available


jrasm91 (approved):
Tyyyyyyyy

PR #8730, opened 2024-04-11
fix(web): focus escaping from modals

ben-basten:
## Description

Fixes included in this change:

1. On initial load, the external library "Add exclusion pattern" modal was incorrectly showing the "Add" button as actionable, even without any path input.
2. Update the focus trap to exclude any elements that are disabled, since they do not accept focus. This was causing focus to escape out of the External Library "Add exclusion pattern" and "Add import path" modals.

## Screenshots

N/A

## How Has This Been Tested?

**Scenario 1: "Add import path" modal**

1. Open "External Library" admin settings: https://demo.immich.app/admin/library-management
2. On a library, click the ellipses "..." > Edit Import Paths > Add path
3. Press the Tab key to move focus through the modal. Observe that it moves outside of the modal. Shift+Tab also gets stuck going backwards in focus.

**Scenario 2: "Add exclusion pattern" modal**

1. Open "External Library" admin settings: https://demo.immich.app/admin/library-management
2. On a library, click the ellipses "..." > Scan Settings > Add Exclusion Pattern
3. Press the Tab key to move focus through the modal. Observe that it moves outside of the modal. Shift+Tab also gets stuck going backwards in focus.

## Checklist:

- [x] npm test (unit tests)


web/src/lib/components/forms/library-exclusion-pattern-form.svelte
```diff
@@ -17,7 +17,8 @@
   });
 
   $: isDuplicate = exclusionPattern !== null && exclusionPatterns.includes(exclusionPattern);
-  $: canSubmit = exclusionPattern !== '' && exclusionPattern !== null && !exclusionPatterns.includes(exclusionPattern);
+  $: canSubmit =
+    exclusionPattern !== '' && exclusionPattern !== undefined && !exclusionPatterns.includes(exclusionPattern);
```
danieldietzler, line 21: Since both `''` and `undefined` are falsy you could probably just write
```suggestion
  $: canSubmit = exclusionPattern && !exclusionPatterns.includes(exclusionPattern);
```

ben-basten: Good call, that's much simpler!

alextran1502 (approved):

PR #8759, opened 2024-04-13
docs: Add community project

mmomjian:
I know we are working on an internal version of this in #8228, but I think it's still reasonable to link to the community version of it.


jrasm91 (approved):
Nice!

PR #8771, opened 2024-04-13
chore: added 'logs' field to bug template

SandiyosDev:
Lots of issues are missing logs, people are not submitting them proactively, making things hard to debug, so a new field is added


bo0tzz (approved):
nit: I would put this above the "additional information" field

jrasm91 (approved):

PR #8773, opened 2024-04-13
feat(server): use embedded preview from raw images

mertalev:
## Description

Some cameras embed a preview image into raw files. There can be differences in interpretation between these previews and the ones we generate. For example, ProRAW images from iPhones can appear very underexposed in our previews.

This PR adds a setting to check raw formats for embedded images. If one is found, it extracts it to a temporary file and uses it as the input to the image processing pipeline instead of the raw file. It still needs to be processed because the "preview" can be huge and contains metadata.

~~This is enabled by default to more closely align with user expectations of what the images should look like.~~ But I'm unsure if there might be cases where the preview is low quality or too small.

Edit: Testing with an image from an iPhone XS, the preview is aliased and quite small. I think this setting can lead to worse images for some cases, so leaving it off by default.

Olympus Disabled:

3a3d43bf-5b48-4e70-bd74-b5f2ece0a888

Olympus Enabled:

02e7f859-36c3-4e0f-bed1-13eb32c58451

iPhone XS Disabled:

1c06db4a-92cc-4c6c-950b-ea1dee7477bc

iPhone XS Enabled:

a400daf7-fc74-4948-b7e3-e601003a961a


danieldietzler (approved):
LGTM!

jrasm91 (commented):
We used to have this feature actually but removed it when we got better support for raw files. Mostly because like you found the embedded pictures are usually quite low quality. I would be maybe ok keeping this if we only used it for specific scenarios or something like that. Or smartly discarded previews that are too small. Otherwise, I don't know if it makes sense to add more code and another setting for something that generally is for a niche scenario

mertalev:
The main issue for me is that libraw underexposes ProRAW images to preserve highlights, which then has the effect of crushing blacks. I haven't been able to get the image to look like the preview. I've tried building their latest snapshot from a few weeks ago, patching it with certain flags and even building it with the Adobe DNG SDK, but nothing short of outright boosting exposure seems to help.

mertalev:
I've played around with their CLI (dcraw_emu) and these settings actually look close to the preview (without the sharpening filters etc.): `-c 0 -w -W -aexpo 6 1 -T -o 7 -4`. But this is for one image, so need to do more testing.

mertalev:
I looked into it more and it seems ProRAW uses fairly new fields in the DNG spec: `ProfileGainTableMap` and `ProfileToneCurve`. It basically intentionally underexposes the raw image so it can bring out more dynamic range with this extra metadata. libraw doesn't use this metadata, which is why the previews don't come out right. Let me see if I can upstream this.

mertalev:
So it looks like I wasn't building with the DNG SDK correctly before so it wasn't getting used. Unfortunately it's a royal pain to get it to work for a number of reasons.

I'm inclined to just continue with the preview approach and call it a day. There's also a request to use the embedded image because it's developed a certain way, so this would cover that use-case as well.

I can also make it smarter so it checks if the preview is big enough before using it.

PR #8778, opened 2024-04-13
feat(server):  trusted header authentication

truongsinh:
This PR implements trusted header authentication

To quote from https://www.authelia.com/integration/trusted-header-sso/introduction/

> This authentication method is referred to by many names; notably trusted header authentication, header authentication, header sso, and probably many more.

API test included.

To use it:
- `IMMICH_TRUSTED_REMOTE_NETWORKS` must be set, for example "10.3.3.0/24, fd00:3::/32, 172.3.4.0/24, 172.3.2.0/24"
- `remote-email` (note, email, not user) must exist in the HTTP request.

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -43,6 +43,7 @@
         "handlebars": "^4.7.8",
         "i18n-iso-countries": "^7.6.0",
         "ioredis": "^5.3.2",
+        "ipaddr.js": "^2.1.0",
         "joi": "^17.10.0",
         "js-yaml": "^4.1.0",
         "lodash": "^4.17.21",
@@ -8968,11 +8969,11 @@
       }
     },
     "node_modules/ipaddr.js": {
-      "version": "1.9.1",
-      "resolved": "https://registry.npmjs.org/ipaddr.js/-/ipaddr.js-1.9.1.tgz",
-      "integrity": "sha512-0KI/607xoxSToH7GjN1FfSbLoU0+btTicjsQSWQlh/hZykN8KpmMf7uYwPW3R+akZ6R/w18ZlXSHBYXiYUPO3g==",
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/ipaddr.js/-/ipaddr.js-2.1.0.tgz",
+      "integrity": "sha512-LlbxQ7xKzfBusov6UMi4MFpEg0m+mAm9xyNGEduwXMEDuf4WfzB/RZwMVYEd7IKGvh4IUkEXYxtAVu9T3OelJQ==",
       "engines": {
-        "node": ">= 0.10"
+        "node": ">= 10"
       }
     },
     "node_modules/is-arrayish": {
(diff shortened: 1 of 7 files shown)


bo0tzz:
What is the use-case or problem solved by this? What are the security implications, for example if a potentially malicious client has the ability to set this header themselves?

truongsinh:
> What is the use-case or problem solved by this?

One use case is to use with reverse proxy and dedicated IAM. For example, Caddy as reverse proxy, and authelia/authentik/keyclock as IAM. See:
- https://www.authelia.com/integration/trusted-header-sso/introduction/
- https://www.authelia.com/integration/proxies/caddy/

To quote from https://caddy.community/t/securing-web-apps-with-caddy-and-authelia-in-docker-compose-an-opinionated-practical-and-minimal-production-ready-login-portal-guide/20465:

> Tight integration to the server itself may reduce footguns (of which there were indeed some encountered in the course of formulating this guide!), and may reduce complexity (of which there is indeed an added amount by virtue of configuring, running, and maintaining multiple separate services).

There are some other use cases of having auth header as well https://github.com/immich-app/immich/discussions/1604, though I may misunderstand what the OP means.

>  What are the security implications, for example if a potentially malicious client has the ability to set this header themselves?

The implication would be that the operator of the stack (proxy, IAM server, and immich), if choosing to use this feature, must configure correctly, for example, https://www.authelia.com/integration/trusted-header-sso/introduction/#trusted-remote-networks . That is, if  the operator blindly set IMMICH_TRUSTED_REMOTE_NETWORKS, and the proxy does not use IAM or sanitize `remote-email` header, the malicious client can successfully set this header themselves. When configured correctly, any `remote-*` header would be sanitize by proxy before being forwarded to immich.

On the other hand, be default, this feature is not enabled, and if the malicious client set the `header`, it it not considered at all.

jrasm91 (commented):
I think we can add this feature, but can you clean up the pull request before I look at it? 
- All the automated checks are failing
- There is a describe.only block
- The web changes likely don't work in the production build (env variables don't mean anything for a static build). 
- Like all other auth settings, this should be system config instead of an environment variable. 
- You should add an e2e test for this feature

bo0tzz:
I don't really see a case for this feature. Authentication via external IAM is already better covered with the existing oauth implementation, and the request in #1604 is not solved here - that request is for the ability to configure a header inside of the mobile app.

jrasm91:
@truongsinh - can you explain why you would need to use this authentication mechanism over the already existing OAuth implementation and integration? Does that also not allow external IAM via authelia/authentik/keyclock?

truongsinh:
@bo0tzz @jrasm91 
> Authentication via external IAM is already better covered with the existing oauth implementation
> can you explain why you would need to use this authentication mechanism over the already existing OAuth implementation and integration

While Immich already supports OAuth, trusted header authentication presents unique advantages for our self-hosted community. This method allows seamless integration into personal network setups with reverse proxies that handle authentication, enhancing user convenience significantly.

1. Ease of Use in Secured LANs: In a secure and trusted LAN, where ease of access is crucial, especially for users like elderly family members who prefer simplicity over complex logins, trusted headers allow a password-free (neither immich nor "sign in with google" password), "click-and-see" experience. This setup lets them easily access content like family videos without remembering complex passwords.
2. Decentralized Security Management: By leveraging trusted header authentication, the responsibility for security and cryptography can shift from Immich to dedicated, specialized apps designed for security. This reduces the security burden on Immich while utilizing more robust, dedicated tools that are already managing network security.
3. Efficient Single Logout (SLO): Unlike OAuth 2.0, trusted header authentication supports true single logout capabilities, making it simpler and more straightforward than implementing SLO with SAML or OIDC. This ease extends not only to Immich contributors but also to operators, simplifying system management and enhancing security.

Given we're discussing the need of trusted auth header, I'll hold off for now. If there is a consensus, I'll continue to address all the comment above (btw I'm a little embarrassed about having `describe.only` in the PR 😅)

jrasm91:
We talked about this internally and we are fine adding this feature to immich. However can you make the following changes? 
- add the network IP settings to the system config so it can be configured in the UI or via a config file
- write some e2e tests for this feature
- auto detect and login the web without using .server.page.ts, since we currently use adapter-static

jrasm91:
It's been 6 weeks with no response from OP. If you are still interested in adding this feature, please re-open with the requested changes. Thanks!

PR #8799, opened 2024-04-15
POC of short asset ids

midzelis:
This is a proof of concept of what it would take to implement short asset ids using nanoid, as mentioned in https://github.com/immich-app/immich/pull/8532#issuecomment-2054087850. 

Its currently set to use nanoid(12), which gives approximately 9B (9,000,000,000) records. Its not in the POC, but when/if this runs out, or if there is a collision with the random ID generator (rare), the operation could be retried, or the length of the id could be increased, which will exponentially (probably) grow the number of records. 

As you can see, most of the work is in updating the validation of input to loosen the UUID format checks to be just length based, and the migration script. 

Also not implemented, is the down() part of the migration, and a way to "upgrade" the old 36 char ids to be 12 chars long. I did a quick skim, and it looks like most foreign keys have a `ON UPDATE CASCADE` clause, so this could be as simple as just updating all the records to a nanoid(12) id. 

All of the not-implemented things ought to be finished if the thought of "saving" 67% of the characters of an asset. 

This was only done for assetIds, but could also be done for albumIds, and maybe all uuids?

### DIFF
--- a/open-api/immich-openapi-specs.json
+++ b/open-api/immich-openapi-specs.json
@@ -8159,6 +8159,7 @@
         "properties": {
           "assetIds": {
             "items": {
+              "format": "uuid",
               "type": "string"
             },
             "type": "array"
@@ -8416,7 +8417,6 @@
             "type": "string"
           },
           "entityId": {
-            "format": "uuid",
             "type": "string"
           },
           "entityType": {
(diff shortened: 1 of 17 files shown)


zackpollard:
Hey, to add a little more context, there are some performance optimisations that are made around using UUID type for primary IDs with regards to the way postgres stores the rows on disk and within indexes. These same optimisations can't be made when using varchar and so you can end up with big gaps in data pages and sparse indexes. These downsides aren't worth the reduced response size in requests, so we decided against making this change a while ago. There are other ways we can go about reducing response sizes that we may look into in future which will bring much greater improvements too.

PR #8808, opened 2024-04-15
docs: Database backup in Truenas

aviv926:
I recently started using Truenas to deploy Immich, I ran into a lot of missing information about how it's deployed.
As part of this their layout includes a different name for the database (immich&immich) rather than immich and postgres which has led to a lot of confusion.
In addition the database [password is generated randomly](https://github.com/truenas/charts/blob/master/community/immich/3.0.19/templates/_configuration.tpl#L8) without the user knowing what it is (It does not appear in the Truenas documentation).

This PR adds relevant information to the documentation that explains how to back up the Immich database using pgAdmin (and explains how the user can know what their password is).

**How is it tested?**
The backup was created on Immich with an environment of 260 images (including machine learning running)
and successful restoration of the backup.

**Problems:**
During the restoration process I encountered a problem that the names of the places were not restored at all, this is a known problem https://github.com/immich-app/immich/discussions/7004.

### DIFF
--- a/docs/docs/administration/backup-and-restore.md
+++ b/docs/docs/administration/backup-and-restore.md
@@ -62,21 +62,21 @@ Note that for the database restore to proceed properly, it requires a completely
 
 By default when using the Immich distribution of TrueNAS the database password is randomly generated, to back up Immich the database password is needed.
 
-1. [HeavyScript](https://github.com/Heavybullets8/heavy_script) and pgAdmin must be installed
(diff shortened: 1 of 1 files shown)


bo0tzz:
Rather than mixing platform-specific instructions into our docs, would this be better suited as a guide in the github discussions with a link from the community page? (Maybe under a new 'community guides' header?)

bo0tzz:
Wonderful, thank you!

PR #8818, opened 2024-04-15
fix(web): album description height

martabal:
fix for #8783


danieldietzler (approved):
Bug fixing day today :D

martabal:
> Bug fixing day today :D

It feels good :sunglasses:

Issue #8590, opened 2024-04-07
Return 404 on non-existing paths


Funk66:
### The bug

Immich shows a server error (500) with no information (defaults to 'Hmm, not sure about that. Check the logs or open a ticket?') whenever the user tries to access an undefined path. It should return a 404 'not found' error instead.

### The OS that Immich Server is running on

N/A

### Version of Immich Server

v1.101.0

### Version of Immich Mobile App

v1.101.0

### Platform with the issue

- [X] Server
- [X] Web
- [X] Mobile

### Your docker-compose.yml content

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1. Go to any random path such as https://<immich domain>/whatever
2. The user is shown an unhelpful error message suggesting to open a bug ticket
```

shubhrad1:
Can I work on this?

shubhrad1:
I think the issue has already been solved. Accessing an undefined path already shows error 404. 

<img width="1438" alt="Screenshot 2024-07-22 at 11 07 04 PM" src="https://github.com/user-attachments/assets/5afb70e2-0844-45ba-9d70-4ff951f32278">

PR #8820, opened 2024-04-15
fix(web): 404 error

martabal:
fix for #8590

## Screenshots

image

### DIFF
--- a/web/src/hooks.client.ts
+++ b/web/src/hooks.client.ts
@@ -3,10 +3,13 @@ import type { HandleClientError } from '@sveltejs/kit';
 
 const DEFAULT_MESSAGE = 'Hmm, not sure about that. Check the logs or open a ticket?';
 
-const parseError = (error: unknown) => {
+const parseError = (error: unknown, statusCode: number) => {
   const httpError = isHttpError(error) ? error : undefined;
-  const statusCode = httpError?.status || httpError?.data?.statusCode || 500;
-  const message = httpError?.data?.message || (httpError?.data && String(httpError.data)) || httpError?.message;
+
+  const message =
+    statusCode === 404
+      ? 'This page does not exist'
+      : undefined || httpError?.data?.message || (httpError?.data && String(httpError.data)) || httpError?.message;
 
   console.log({
     status: statusCode,
@@ -20,8 +23,8 @@ const parseError = (error: unknown) => {
   };
 };
 
-export const handleError: HandleClientError = ({ error }) => {
-  const result = parseError(error);
+export const handleError: HandleClientError = ({ error, status }) => {
+  const result = parseError(error, status);
   console.error(`[hooks.client.ts]:handleError ${result.message}`);
   return result;
 };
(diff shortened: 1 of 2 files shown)


martabal:
> While you're here, can you update the error page to also have a way back to the main timeline?

Maybe it's better to have a different error page for a 404 error with a "go back" button ?

martabal:
Something like that ?

| Light | Dark |
| :---: | :---: |
| image | image |

danieldietzler:
I personally like the the design but I'm not sure if it fits into the immich "design language"

martabal:
> I personally like the the design but I'm not sure if it fits into the immich "design language"

Yeah, this is clearly not the immich "design language", I'm just wondering what's the best option between a full page error and the traditional error in a modal.

bo0tzz:
For now I'd say stick with the existing error layout for simplicity, and it can get a more focused cleanup in the future. For now the error page is great to steer all the technical users in the right direction but when we head towards stable we might want to make it a bit more mom-friendly ;)

jrasm91:
Personally, it we know it's a 404 error I think a full page thing would make sense.

web/src/hooks.client.ts
```diff
@@ -3,10 +3,13 @@ import type { HandleClientError } from '@sveltejs/kit';
...
-const parseError = (error: unknown) => {
+const parseError = (error: unknown, statusCode: number) => {
   const httpError = isHttpError(error) ? error : undefined;
-  const statusCode = httpError?.status || httpError?.data?.statusCode || 500;
-  const message = httpError?.data?.message || (httpError?.data && String(httpError.data)) || httpError?.message;
+
+  const message =
+    statusCode === 404
+      ? 'This page does not exist'
+      : undefined || httpError?.data?.message || (httpError?.data && String(httpError.data)) || httpError?.message;
```
danieldietzler, line 12: I would prefer splitting this up into two variables like it was before. This is quite long right now

danieldietzler (approved):

web/src/hooks.client.ts
martabal: Is it even possible to have different error messages with the current implementation ?

It's maybe easier to set ourselves different error messages with a switch case

jrasm91: If the api server is down or unreachable it shows a different error.

web/src/routes/+error.svelte
```diff
@@ -54,6 +55,11 @@
                 <label for="stacktrace">Stacktrace</label>
                 <pre id="stacktrace" class="text-xs">{$page.error?.stack || 'No stack'}</pre>
               {/if}
+              {#if $page.error?.code}
+                <a href="/">
+                  <Button size="sm">Go back</Button>
+                </a>
+              {/if}
```
jrasm91, line 62: should we not show this button all the time? I don't think it needs to be limited to when there is an error code.

martabal: I think we should keep it only for 404 errors, the button is at best useless for other errors at worst misleading

martabal:
Closing in favor of #8901

PR #8821, opened 2024-04-15
fix(web): delete trashed item

martabal:
fix for #8507, do not try to trash items which are already trashed

### DIFF
--- a/web/src/lib/components/photos-page/asset-grid.svelte
+++ b/web/src/lib/components/photos-page/asset-grid.svelte
@@ -71,7 +71,17 @@
   };
 
   const onDelete = () => {
-    if (!isTrashEnabled && $showDeleteModal) {
+    const hasTrashedAsset = Array.from($selectedAssets).some((asset) => asset.isTrashed);
+
+    if (hasTrashedAsset) {
+      if ($showDeleteModal) {
+        isShowDeleteConfirmation = true;
+        return;
+      } else {
+        handlePromiseError(trashOrDelete(true));
+        return;
+      }
+    } else if (!isTrashEnabled && $showDeleteModal) {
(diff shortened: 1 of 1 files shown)


web/src/lib/components/photos-page/asset-grid.svelte
```diff
@@ -71,7 +71,17 @@
...
+
+    if (hasTrashedAsset) {
+      if ($showDeleteModal) {
+        isShowDeleteConfirmation = true;
+        return;
+      } else {
+        handlePromiseError(trashOrDelete(true));
+        return;
+      }
+    } else if (!isTrashEnabled && $showDeleteModal) {
```
danieldietzler, line 84: When you're in the else case `isTrashEnabled` is already false, so the first part of the and is tautological.

danieldietzler (approved):

web/src/lib/components/photos-page/asset-grid.svelte
martabal: Why is it already false ?

danieldietzler: Oh nvm, I misread `hasTrashedAsset` in l. 76 for `isTrashEnabled`. Sorry!

danieldietzler: But your new commit still looks better haha

martabal: > Oh nvm, I misread `hasTrashedAsset` in l. 76 for `isTrashEnabled`. Sorry!

np

> But your new commit still looks better haha

yeah, my first commit was a quick and ugly fix :sweat_smile:

jrasm91 (approved):

PR #8823, opened 2024-04-15
feat(server): correlation id via injected logger

jrasm91:
* make logger repository
* inject correlation id to the log message when at least verbose level
* accept external correlation id via `x-immich-cid`

image


danieldietzler (approved):

bo0tzz (approved):
Awesome! Is there an easy way to also return the `x-immich-cid` header in responses?

jrasm91:
Yes... Forgot about that. We can add it to a middleware.

PR #8835, opened 2024-04-16
chore: cleanup library watching

jrasm91:
* delete library watcher e2e (too many bugs and problems with the tests to be bothered to migrate them)
* delete code to watch libraries on create/update, since it doesn't do anything
* avoid re-initializing the reverse geocode
* remove unused dependencies
* remove test-only event emitter functionality


zackpollard (approved):

e2e/src/utils.ts
```diff
@@ -407,8 +413,14 @@ export const utils = {
       },
     ]),
 
-  deleteTempFolder: () => {
+  resetTempFolder: () => {
     rmSync(`${testAssetDir}/temp`, { recursive: true, force: true });
+    mkdirSync(`${testAssetDir}/temp`, { recursive: true });
+  },
```
danieldietzler, line 419: This is not what we want as this will keep dummy files. That results in changes in git which I personally find very annoying

jrasm91: This deletes all the files in the directory.

danieldietzler: Yeah nvm I'm stupid.

PR #8842, opened 2024-04-16
feat(mobile): use efficient sync

fyfrey:
follow-up for #8755

enable the use of the more efficient endpoints to perform sync operations.

only merge after #9189 has been in a release for 1-2 weeks.


mobile/lib/shared/services/asset.service.dart
```diff
@@ -83,38 +91,30 @@ class AssetService {
...
-  Future<List<Asset>?> _getRemoteAssets(User user) async {
+  Future<List<Asset>?> _getRemoteAssets(User user, DateTime until) async {
     const int chunkSize = 10000;
     try {
-      final DateTime now = DateTime.now().toUtc();
       final List<Asset> allAssets = [];
-      for (int i = 0;; i += chunkSize) {
+      DateTime? lastCreationDate;
+      String? lastId;
+      for (;;) {
```
alextran1502, line 100: `while` loop?

alextran1502: I see this a lot in embedded system code, not used to seeing it in higher level language 😝

fyfrey: I personally use this for loops when I break out using other means.

While(true).. I use for endless tasks. Like a server listening indefinitely until it's killed etc.

martyfuhry: Can this be rewritten to a `do while` loop on the condition `assets.length < chunkSize`? I think that's the only break condition I see here. I concur that `while (true)` loops are odd and potentially dangerous. I also think `for (;;)` is less readable. I did not know what it meant until I read some of the other comments.

martyfuhry (commented):
I wish I could be more help, as I'm not tremendously familiar with the backup sync. But I'm happy to do any testing if you'd like me to, as well!

mobile/test/modules/shared/sync_service_test.dart
```diff
@@ -75,8 +75,12 @@ void main() {
         makeAsset(checksum: "c", remoteId: "1-1"),
       ];
       expect(db.assets.countSync(), 5);
-      final bool c1 =
-          await s.syncRemoteAssetsToDb(owner, _failDiff, (u) => remoteAssets);
+      final bool c1 = await s.syncRemoteAssetsToDb(
```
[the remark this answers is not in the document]
alextran1502: I agree with Marty here, I found a lot of these method with passed in function rather hard to read and follow

alextran1502:
I ran into this error on the server when running the PR

image

mobile/lib/shared/services/asset.service.dart
fyfrey: sadly, I can't put the break condition in the loop condition check: it would not even run the first pass of the loop - i can only to while(true) instead of for(;;). The loop has two break conditions that cannot be handled in the loop header

fyfrey:
> I ran into this error on the server when running the PR

It seems some more server changes are required for stacked assets (my test instance probably had no stacked assets at the time...). I converted this PR back to a draft and will work on a new PR for the required server changes.

jrasm91 (approved):
LGTM

alextran1502:
Hi @fyfrey I tested this again and found a problem with rendering memory lane. Here is the error message

image

From my debugging, the API returned the correct data. Still, the call to get the asset from the database when building the model probably happened before the synchronization process finished, so it could not parse all data correctly, leading to the error.

image

fyfrey:
> Hi @fyfrey I tested this again and found a problem with rendering memory lane. Here is the error message

Thanks @alextran1502 ! I've pushed a new commit to fix the bug in the memory lane. We were just lucky it did not occur previously..
Solution: Do not add a memory lane if assets are not yet in DB. Refresh memory lane provider after remote asset sync.

alextran1502:
Hi @fyfrey, the change doesn't entirely fix the issue. I have multiple memories on a day but only the first memory gets rendered, so I checked the database. On `main`, I see 69000 assets in Isar database, while on this branch, I only see 64000 assets, so the fetching mechanism somehow doesn't fetch all assets

fyfrey:
> Hi @fyfrey, the change doesn't entirely fix the issue. I have multiple memories on a day but only the first memory gets rendered, so I checked the database. On `main`, I see 69000 assets in Isar database, while on this branch, I only see 64000 assets, so the fetching mechanism somehow doesn't fetch all assets

Thank you for testing again @alextran1502 . Very interesting. I did not notice any sync discrepancy (asset count) with my initial server changes. In the newer fix PR (#9100) we changed it to use the common asset query builder. It turns out it is incompatible with the app with stacked assets as the app expects all assets individually while the server returns them in an array with of the  parent asset. This is probably the same issue you encountered a while ago and reverted server refactoring changes in #7752.

made another server PR to fix it: #9189

alextran1502:
I tried again with #9189 on my prod (FOR SCIENCE!!) and I still see missing memories. Checking the database again, now I see 68000 assets are fetched on #9189 while still 69000 assets on `main`

fyfrey:
Something must still be different between the old and the new server query.
We need some internal test instance that has all the edge cases like your prod instance does :-P

The old method also fetched archived assets for partners, maybe this accounts for the difference. But memories should not show partner assets.... I'll take another look at the code, or better the generated SQL

Issue #8805, opened 2024-04-15
BUG: Search bar doesn't autofocus when adding photos to album.


halphalp:
### The bug

i'm not sure how recently this started happening, but when selecting photos to add to an album, the search bar doesn't auto-focus. i keep accidentally stacking multiple photos when trying to add them to an album because i start typing after clicking 'add'.

### The OS that Immich Server is running on

Docker

### Version of Immich Server

v1.101.0

### Version of Immich Mobile App

v1.101.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
same as provided
```

### Your .env content

```Shell
same as provided
```

### Reproduction steps

```bash
1. Select multiple photos
2. Click + sign
3. Click add to album
4. Note that the search bar doesn't autofocus
```

ben-basten:
> i'm not even sure what the keyboard shortcut is for stacking photos, but i apparently keep hitting it when attempting to type my album names.

Yeah, the shortcut is "s" so it would be easy to unintentionally stack some photos! Looks to me like this is a similar issue to #8804.

PR #8846, opened 2024-04-16
fix(web): Prevent DELETE key from clearing DateInput in modal

### LINKED ISSUE #8804: [BUG] Web : Change date modale / DEL key trash the selected pictures
### The bug

Pressing the DEL key when the "Change date" modale is visible delete the pictures instead of clearing a field.

*I did not tried on the "Change location" modale*

image

TrueNas

### Version of Immich Server

v1.101.0

v1.101.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
TrueNas app
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1. Open an album
2. Select some pictures
3. Click on the "..." menu (top right)
4. Click on "Change date"
5. Select a field and click "delete" key on your keyboard
```

### Relevant log output

_No response_

### Additional information

_No response_

uniqueeest:
This commit modifies the keydown event handler to stop the propagation of non-navigation key events in the change date modal. 
So the DELETE key prevented the error of erasing the selected pictures.


web/src/lib/components/shared-components/change-date.svelte
```diff
@@ -57,6 +57,8 @@
   const handleKeydown = (event: KeyboardEvent) => {
     if (['ArrowUp', 'ArrowDown', 'ArrowLeft', 'ArrowRight'].includes(event.key)) {
       event.stopPropagation();
+    } else {
+      event.stopPropagation();
     }
```
danieldietzler, line 62: This right now could be reduced to `=> event.stopPropagation()`, right? :P
I don't think that's what is wanted here...?

uniqueeest: @danieldietzler I'll reduce that code to one line. I think the delete key issue is caused by event bubbling, is there anything I need to add?

jrasm91: I don't think we want to do event propagation for every single event.

ben-basten:
This might be out of scope for the current PR, but I'm wondering if this event propagation fix should be pulled into the `FullScreenModal` level. Keyboard shortcuts affecting the content behind a modal is an issue in multiple places, so that would also fix issues like #8805.

danieldietzler:
> This might be out of scope for the current PR, but I'm wondering if this event propagation fix should be pulled into the `FullScreenModal` level. Keyboard shortcuts affecting the content behind a modal is an issue in multiple places, so that would also fix issues like #8805.

Yeah I think you're right. Before we solve the same issue in a dozen places we should fix the actual issue at the root.

uniqueeest:
@danieldietzler I fixed the keyboard issue that occurs in modals by adding stop propagation to the full-screen-modal component. What do you think?

danieldietzler:
You don't want to essentially disable all events. Something like escape should still work there. https://github.com/immich-app/immich/pull/8846#discussion_r1571694954 applies here as well.

web/src/lib/components/shared-components/full-screen-modal.svelte
```diff
@@ -45,9 +45,13 @@
 
 <FocusTrap>
```
ben-basten, line 46: Looks like the `FocusTrap` will have to be moved, to make sure that it's still receiving the keyboard events needed to manage focus within the modal. Making it a direct child of the `section` should fix this!

uniqueeest: @ben-basten Thank you, and I'll give it a try!

jrasm91: I just changed this to just ignore events that originate from an input type that includes date and that seemed to fix the issue without needing to make any other changes.

jrasm91 (approved):

jrasm91:
This fixes the issue for keyboard shortcuts while focus is in the input. It doesn't address the issue for shortcuts triggering events while the modal is open and the focus is _not_ in an input field. I think that is fine for now and can be left for a follow-up PR.

PR #8855, opened 2024-04-16
Migrate ImmichLogger over to injected ILoggerRepository

### LINKED ISSUE #8824: Migrate ImmichLogger over to injected ILoggerRepository
#8823 introduces a new ILoggerRepository. All remaining usages of ImmichLogger should be migrated to use this new injected logger.

iamamansharma:
fixes  #8824

### DIFF
--- a/server/src/cores/storage.core.ts
+++ b/server/src/cores/storage.core.ts
@@ -1,3 +1,4 @@
+import { Inject } from '@nestjs/common';
 import { dirname, join, resolve } from 'node:path';
 import { APP_MEDIA_LOCATION } from 'src/constants';
 import { SystemConfigCore } from 'src/cores/system-config.core';
@@ -7,11 +8,11 @@ import { PersonEntity } from 'src/entities/person.entity';
 import { ImageFormat } from 'src/entities/system-config.entity';
 import { IAssetRepository } from 'src/interfaces/asset.interface';
 import { ICryptoRepository } from 'src/interfaces/crypto.interface';
+import { ILoggerRepository } from 'src/interfaces/logger.interface';
 import { IMoveRepository } from 'src/interfaces/move.interface';
 import { IPersonRepository } from 'src/interfaces/person.interface';
 import { IStorageRepository } from 'src/interfaces/storage.interface';
 import { ISystemConfigRepository } from 'src/interfaces/system-config.interface';
-import { ImmichLogger } from 'src/utils/logger';
 
 export enum StorageFolder {
   ENCODED_VIDEO = 'encoded-video',
@@ -41,7 +42,6 @@ export type GeneratedAssetType = GeneratedImageType | AssetPathType.ENCODED_VIDE
 let instance: StorageCore | null;
 
 export class StorageCore {
-  private logger = new ImmichLogger(StorageCore.name);
   private configCore;
   private constructor(
     private assetRepository: IAssetRepository,
@@ -50,8 +50,10 @@ export class StorageCore {
     private cryptoRepository: ICryptoRepository,
     private repository: IStorageRepository,
     systemConfigRepository: ISystemConfigRepository,
+    @Inject(ILoggerRepository) private logger: ILoggerRepository,
   ) {
-    this.configCore = SystemConfigCore.create(systemConfigRepository);
+    this.logger.setContext(StorageCore.name);
+    this.configCore = SystemConfigCore.create(systemConfigRepository, this.logger);
   }
 
   static create(
@@ -61,6 +63,7 @@ export class StorageCore {
     cryptoRepository: ICryptoRepository,
     configRepository: ISystemConfigRepository,
     repository: IStorageRepository,
+    logger: ILoggerRepository,
   ) {
     if (!instance) {
       instance = new StorageCore(
@@ -70,6 +73,7 @@ export class StorageCore {
         cryptoRepository,
         repository,
         configRepository,
+        logger,
       );
     }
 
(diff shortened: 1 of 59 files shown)


server/src/cores/storage.core.ts
```diff
@@ -50,8 +50,10 @@ export class StorageCore {
     private cryptoRepository: ICryptoRepository,
     private repository: IStorageRepository,
     systemConfigRepository: ISystemConfigRepository,
+    @Inject(ILoggerRepository) private logger: ILoggerRepository,
```
danieldietzler, line 53: You can't and don't want to have the inject there

server/src/cores/system-config.core.ts
```diff
@@ -169,16 +169,20 @@ let instance: SystemConfigCore | null;
...
 export class SystemConfigCore {
-  private logger = new ImmichLogger(SystemConfigCore.name);
   private configCache: SystemConfigEntity<SystemConfigValue>[] | null = null;
 
   public config$ = new Subject<SystemConfig>();
 
-  private constructor(private repository: ISystemConfigRepository) {}
+  private constructor(
+    private repository: ISystemConfigRepository,
+    @Inject(ILoggerRepository) private logger: ILoggerRepository,
```
danieldietzler, line 178: Here the same thing. You don't want to get something injected when you call the constructor yourself (using the create method)

server/src/repositories/media.repository.ts
```diff
@@ -21,8 +22,9 @@ sharp.cache({ files: 0 });
 
 @Instrumentation()
```
danieldietzler, line 23: You'll also need `@Injectable()` for it to work

server/src/repositories/metadata.repository.ts
```diff
@@ -26,9 +26,10 @@ export class MetadataRepository implements IMetadataRepository {
     @Inject(ISystemMetadataRepository)
     private readonly systemMetadataRepository: ISystemMetadataRepository,
     @InjectDataSource() private dataSource: DataSource,
-  ) {}
-
-  private logger = new ImmichLogger(MetadataRepository.name);
+    @Inject(ILoggerRepository) private readonly logger: ILoggerRepository,
```
danieldietzler, line 29: Why do you give it the readonly attribute here?

server/src/repositories/search.repository.ts
```diff
@@ -37,7 +36,9 @@ export class SearchRepository implements ISearchRepository {
     @InjectRepository(AssetFaceEntity) private assetFaceRepository: Repository<AssetFaceEntity>,
     @InjectRepository(SmartSearchEntity) private smartSearchRepository: Repository<SmartSearchEntity>,
     @InjectRepository(GeodataPlacesEntity) private readonly geodataPlacesRepository: Repository<GeodataPlacesEntity>,
+    @Inject(ILoggerRepository) private readonly logger: ILoggerRepository,
```
danieldietzler, line 39: Here as well?

server/src/repositories/storage.repository.ts
```diff
@@ -13,13 +15,13 @@ import {
   WatchEvents,
 } from 'src/interfaces/storage.interface';
 import { Instrumentation } from 'src/utils/instrumentation';
-import { ImmichLogger } from 'src/utils/logger';
 import { mimeTypes } from 'src/utils/mime-types';
 
 @Instrumentation()
```
danieldietzler, line 20: `@Injectable()`

danieldietzler (commented):
I wonder if we should call the logger variables `loggerRepository` for consistency or no...? @jrasm91

server/src/cores/storage.core.ts
iamamansharma: it seems to me that we wanna inject the new logger .
ref: https://github.com/immich-app/immich/pull/8823

jrasm91:
> I wonder if we should call the logger variables `loggerRepository` for consistency or no...? @jrasm91

I don't have a strong preference, but that might make it more consistent, so I'm not against doing it in a follow-up PR.

server/src/repositories/logger.repository.ts
```diff
@@ -1,10 +1,10 @@
-import { Injectable } from '@nestjs/common';
+import { Injectable, Scope } from '@nestjs/common';
 import { ClsService } from 'nestjs-cls';
 import { LogLevel } from 'src/entities/system-config.entity';
 import { ILoggerRepository } from 'src/interfaces/logger.interface';
 import { ImmichLogger } from 'src/utils/logger';
 
-@Injectable()
+@Injectable({ scope: Scope.TRANSIENT })
```
jrasm91, line 7: This is necessary so each injected instance is independent and `.setContext` in one service doesn't override that of another

server/src/services/database.service.spec.ts
```diff
@@ -23,30 +25,18 @@ describe(DatabaseService.name, () => {
...
     [{ vectorExt: DatabaseExtension.VECTOR, extName: 'pgvector', minVersion: new Version(0, 5, 0) }],
   ] as const)('init', ({ vectorExt, extName, minVersion }) => {
-    let fatalLog: MockInstance;
-    let errorLog: MockInstance;
-    let warnLog: MockInstance;
-
     beforeEach(() => {
-      fatalLog = vitest.spyOn(ImmichLogger.prototype, 'fatal');
-      errorLog = vitest.spyOn(ImmichLogger.prototype, 'error');
-      warnLog = vitest.spyOn(ImmichLogger.prototype, 'warn');
```
jrasm91, line 33: these tests are so much nicer now that the logger in dependency injected instead of using a spy

jrasm91 (approved):
Ty!

danieldietzler (approved):
LGTM!

PR #8860, opened 2024-04-16
docs: fix vectors grant... again

mmomjian:
@mertalev

I did some testing in my DB with two instances of Immich running. The following worked, without affecting the other Immich installation.

```
ALTER SCHEMA vectors OWNER TO immich;
ALTER TABLE pg_vector_index_stat OWNER TO immich;
```

If we alter the schema owner first, the table owner will inherit upon creation, yes? Rendering the second command unnecessary.

@VoVAllen we keep running into the error `error: permission denied for view pg_vector_index_stat`. Will changing the owner of the vectors schema to the Immich user fix this without causing other issues?


mertalev:
> If we alter the schema owner first, the table owner will inherit upon creation, yes?

Yes, that's right.

mertalev (approved):

VoVAllen:
alter owner should fix this. We've included this patch in the pgvecto.rs but haven't made new releases yet

PR #8862, opened 2024-04-17
refactor(server): immich-admin list-users

jrasm91:


e2e/src/utils.ts
```diff
@@ -59,14 +59,19 @@ export const testAssetDirInternal = '/data/assets';
...
-    stdio: 'pipe',
+export const immichCli = (args: string[]) =>
+  executeCommand({ command: 'node', args: ['node_modules/.bin/immich', '-d', `/${tempDir}/immich/`, ...args] });
+export const immichAdmin = (args: string[]) =>
+  executeCommand({
+    command: 'docker',
+    args: ['exec', '-i', 'immich-e2e-server', '/bin/bash', '-c', `immich-admin ${args.join(' ')}`],
   });
 
+const executeCommand = ({ command, args }: { command: string; args: string[] }) => {
```
danieldietzler, line 70: I'm wondering if we need an object here or if `(command: string, ...args: string[])` is equally good. I think I prefer the latter

danieldietzler (approved):

e2e/src/utils.ts
jrasm91: Yeah that's probably fine

jrasm91: I kept args as `args: string[]`. I think `...args: string[]` is a bit of an anti-pattern. There is a limitation on the number of params you can pass to a function and maybe some other considerations. The `Promise.all( [list] )` used to be like that, but has migrated to a list instead of individual args.

PR #8915, opened 2024-04-19
refactor(server): sessions

jrasm91:
* Rename auth-device/user-token to session
* Move session CRUD out of the auth service and into a dedicated controller and service

This is technically a breaking API change, although I doubt anyone has been using `/auth/devices`.


mertalev (approved):

danieldietzler (approved):
LGTM!

PR #8924, opened 2024-04-19
feat(web): new look option for slideshow

martabal:
This PR adds a new look option for the slideshow: blurred instead of black background and fix an issue where videos always have a margin-bottom

## Screenshots

Screenshot from 2024-04-19 20-31-41
Screenshot from 2024-04-19 20-31-47


danieldietzler (approved):
This looks nice!

PR #8935, opened 2024-04-19
fix(web): restore button in asset viewer 

### LINKED ISSUE #8861: No restore button when viewing images in the Trash
### The bug

You can't restore trashed images from the view screen, you can only restore them by selecting the images in the Trash.

The thumbnail sometimes doesn't have enough detail to discriminate between images so restoring a specific image from a bunch of similar looking images is cumbersome.

Overall, the view screen doesn't seem to discriminate much between trashed images and those that aren't in the trash: except for the "share" button, the user interface when viewing an image is very similar in both cases. Actions like "favorite", "archive" and "add to album" don't make sense in the context of an image that is about to be deleted, while a "restore" action would be useful.

The mobile app seems to have the same issue.

Raspberry PI OS

### Version of Immich Server

v1.101.0

v1.101.0

### Platform with the issue

- [ ] Server
- [X] Web
- [X] Mobile

```YAML
...
```

### Your .env content

```Shell
...
```

### Reproduction steps

```bash
1. Go to https://demo.immich.app
2. Delete any image
3. Go to the Trash
4. Open the deleted image by clicking on it
5. Observe the lack of any restore function in that screen
```

### Relevant log output

_No response_

### Additional information

_No response_

ConnerWithAnE:
## Description
Fixes #8861 

This fixes the web version of this issue. This pull request #8919 fixed the mobile version.

## How Has This Been Tested?

- ran  `npm run check:all` and all tests passed successfully.
- Inserted an image from the web client and deleted it, restored from the trash.

## Checklist:

### DIFF
--- a/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
+++ b/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
@@ -20,6 +20,7 @@
     mdiFolderDownloadOutline,
     mdiHeart,
     mdiHeartOutline,
+    mdiHistory,
     mdiImageAlbum,
     mdiImageMinusOutline,
     mdiImageOutline,
@@ -52,6 +53,7 @@
 
   type MenuItemEvent =
     | 'addToAlbum'
+    | 'restoreAsset'
     | 'addToSharedAlbum'
     | 'asProfileImage'
     | 'setAsAlbumCover'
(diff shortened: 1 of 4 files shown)


jrasm91 (approved):
Noice. I moved it to advance the view, just like it does when you trash the item initially. Also, it removes it from the store, so no refresh is required.

PR #8939, opened 2024-04-20
Fix backup and restore docker compose issue

bercribe:
For the backup and restore docker compose config to work, the `POSTGRES_DB` environment variable needs to be specified

### DIFF
--- a/docs/docs/administration/backup-and-restore.md
+++ b/docs/docs/administration/backup-and-restore.md
@@ -71,6 +71,7 @@ services:
       POSTGRES_CLUSTER: 'TRUE'
       POSTGRES_USER: ${DB_USERNAME}
       POSTGRES_PASSWORD: ${DB_PASSWORD}
+      POSTGRES_DB: ${DB_DATABASE_NAME}
       SCHEDULE: "@daily"
       POSTGRES_EXTRA_OPTS: '--clean --if-exists'
       BACKUP_DIR: /db_dumps


mmomjian:
I don't believe this is necessary/wanted. What is your YML for the backup container and what issue are you running into? With `POSTGRES_CLUSTER: 'TRUE'` set, the behavior of the backup container should be similar to `pg_dumpall`, which does not take a specific database as an argument.

bercribe:
Here's my backup container YML:
```
  backup:
    image: prodrigestivill/postgres-backup-local:14
    env_file:
      - .env
    environment:
      POSTGRES_HOST: database
      POSTGRES_CLUSTER: 'TRUE'
      POSTGRES_USER: ${DB_USERNAME}
      SCHEDULE: "@daily"
      POSTGRES_EXTRA_OPTS: '--clean --if-exists'
      BACKUP_DIR: /db_dumps
    volumes:
      - type: volume
        source: db-dumps
        target: /db_dumps
        volume:
          nocopy: true
    depends_on:
      - database
```
Prior to adding this line I tried it as suggested by the Immich docs and got an error log stating that `POSTGRES_DB` was required. (Lost the exact logs after restarting the container). The main repo does state that it's a required field so it made sense to me: https://github.com/prodrigestivill/docker-postgres-backup-local

mmomjian:
Thanks for testing. I will open an issue on the upstream repo but for now I guess we need to re add this.

mmomjian:
prodrigestivill/postgres-backup-local will update the documentation to specify that `POSTGRES_DB` will be used as the connecting database for pg_dumpall, not the target backup database. We need to update our example compose YML to re-introduce this variable.

mmomjian (approved):

mertalev (approved):

mmomjian:
Upstream project has added documentation to confirm that this is a required flag even with PG_CLUSTER set.

PR #8977, opened 2024-04-20
refactor(server): session interface types

jrasm91:


bo0tzz (approved):
Can this pattern be extended to other repos?

jrasm91:
Definitely, but not today :-)

server/src/interfaces/session.interface.ts
```diff
@@ -2,10 +2,12 @@ import { SessionEntity } from 'src/entities/session.entity';
 
 export const ISessionRepository = 'ISessionRepository';
 
+type E = SessionEntity;
```
danieldietzler, line 5: That was necessary lmao

danieldietzler (approved):

server/src/interfaces/session.interface.ts
jrasm91: I was considering making all interfaces implement a `BaseInterface` that has the correct typings for `create`, `update`, `delete`, and `get`. It would be something like `BaseInterface<E>` and then the methods would indeed reference `E`.

PR #8992, opened 2024-04-21
Feature: Read-only property for External Libraries

rokiden:
Closes #5449. Original [thread ](https://github.com/immich-app/immich/discussions/5449#discussioncomment-8900247).

This PR adds  "Read Only" property for External library.

In database it stored as column "libraries.isReadOnly". This property set at library creation time and can't be modified. Asset.isReadOnly inherits library's value at creation time. At frontend it was added as toggle to "Select library owner" dialog.  Mobile app doesn't require changes. 

Asset modification and deletion require to check only isReadOnly property, isExternal becomes only informational.

Tested:
* delete to trash
* empty trash
* modification
* rescan of library

TODO:
* automated tests
* indication of ReadOnly property in web/mobile
* ? cmdline tool to migrate library to RW mode and back


web/src/lib/components/forms/library-user-picker-form.svelte
```diff
@@ -37,6 +39,7 @@
     <p class="p-5 text-sm">NOTE: This cannot be changed later!</p>
 
     <SettingSelect bind:value={ownerId} options={userOptions} name="user" />
+    <SettingSwitch title="Read Only" subtitle="Disable deletion and modification" bind:checked={isReadOnly} />
```
benmccann, line 42: does this disable modification? I saw you updated `handleAssetDeletion` in `asset.service.ts`, but I did not see changes to any other methods such as `update`

rokiden: There are no server-side checks of isExternal or isReadOnly before modification. I have modified checks on web-side. It was tested, modification works, creates sidecar file in external lib.

benmccann: Should the text here be updated to just say "Disable deletion" instead of "Disable deletion and modification" then?

rokiden: It disables modifications because frontend checks isReadOnly.

ThreeAurora:
Thanks, that's what I need most about this feature, otherwise I wouldn't even know how to do my job!

web/src/lib/components/forms/library-user-picker-form.svelte
benmccann: Ah, I misunderstood. It seems like it should also disable it on the server. Otherwise you can modify via the mobile app or third-party integrations using the API. It'd be weird to only disallow it in the web frontend, but allow it from these other locations.

rokiden: Disabled deletion and modification for External Library's assets is default behavior of immich.

This PR allows this actions by disabling ReadOnly. Now I can disable highlighted switch and cleanup my External Library. When I started this fork, I looked at asset.service and didn't find any restrictions, then tested and it worked. 

I agree that frontend-only restrictions is not okay, but think that it's not this PR's problem. I can enforce server-side checks but in another PR.

You can't _see changes to any other methods such as update_ because it is not required for this task. I made minimal changes allowing deletion and modification of ExtLib assets. 

Regards

benmccann:
Thanks! I discussed this PR with the other maintainers and we're happy with it overall, but adding that additional server-side check would be a blocker. Then I think the only other change needed would be some updates to the tests

> TODO:
> indication of ReadOnly property in web/mobile

I believe we already have some indications such as delete being disabled for readonly assets

> ? cmdline tool to migrate library to RW mode and back

Probably better to make the readonly property modifiable in the UI rather than a cmdline tool, but that's probably better as a separate PR. I think this is fine for now

ThreeAurora:
Thanks for keeping an eye on this feature, we're excited to see it coming soon!

jrasm91:
#8943 seems more complete

benmccann:
@jrasm91 in what way does it seem more complete? It does have tests, but the logic on this one I think is closer to the desired logic because it uses the readonly flag on the asset

jrasm91:
Both look to be doing essentially the same thing. The other PR was created first, has some tests, and allows the property to be updated after the library is created.

rokiden:
@benmccann 
I've added server-side checks to update and updateAll.
Also added library's isReadOnly indication using alternative icon for read-write libs:
Screenshot from 2024-04-28 16-57-36

server/src/services/asset.service.ts
```diff
@@ -252,11 +252,19 @@ export class AssetService {
   async update(auth: AuthDto, id: string, dto: UpdateAssetDto): Promise<AssetResponseDto> {
     await this.access.requirePermission(auth, Permission.ASSET_UPDATE, id);
 
+    let asset = await this.assetRepository.getById(id);
```
benmccann, line 255: this is going to cause a second query right after having done a query on the line above. Perhaps it should just be part of that initial permission check. see `AssetAcess.checkOwnerAccess` in `access.repository.ts`

jrasm91: That would be fine with me. Asset update access implies the asset is not read only.

web/src/lib/components/forms/library-user-picker-form.svelte
```diff
@@ -37,6 +39,7 @@
     <p class="p-5 text-sm">NOTE: This cannot be changed later!</p>
```
benmccann, line 39: Assuming https://github.com/immich-app/immich/pull/9188 is merged, then you will be able to change the readonly status, so we might need to clarify what "This" is referring to in this message

jrasm91:

rokiden:
This is good news, for me it’s much better that the team took it upon themselves. Thanks! I was glad to help.

PR #8994, opened 2024-04-21
feat(server): partial config updates

danieldietzler:
`PUT /system-config` now accepts partial config dtos.
Has been tested by making various (partial) API requests against the server


danieldietzler:
Apparently this breaks the open api definitions and the nestjs/swagger equivalent (`PartialType`) does not allow for deep partial objects

danieldietzler:
Apparently it isn't as easy, so we'll just postpone this until we have a better way of storing config stuff.

PR #9001, opened 2024-04-22
chore: test more formats in e2e

mertalev:
## Description

Adds tests for ARW, AVIF, JXL and RW2 images. This does add a few seconds to the e2e tests since each asset upload is done sequentially.


jrasm91 (approved):
We can change them to upload in parallel in another pr.

PR #9021, opened 2024-04-22
docs: default to sed pg_catalog for Linux, document restore/reset for PG bind mount

mmomjian:
- Add `sed` as default for restore - multiple issues of people trying to restore recently and having geocoding issues or errors.
  - I do not know the equivalent command for Windows, would welcome someone adding it
- Collapse some code blocks in the FAQ
- Wrap the library-change instructions in a transaction
- Update docs to explain how to reset the DB now that Docker volumes are no longer used

### DIFF
--- a/docs/docs/FAQ.mdx
+++ b/docs/docs/FAQ.mdx
@@ -267,6 +267,9 @@ The initial backup is the most intensive due to the number of jobs running. The 
 
 By default, a container has no resource constraints and can use as much of a given resource as the host's kernel scheduler allows. To limit this, you can add the following to the `docker-compose.yml` block of any containers that you want to have limited resources.
 
+<details>
+<summary>docker-compose.yml</summary>
+
 ```yaml
 deploy:
   resources:
@@ -276,7 +279,7 @@ deploy:
       # Gigabytes of memory
       memory: '1G'
 ```
-
+</summary>
 For more details, you can look at the [original docker docs](https://docs.docker.com/config/containers/resource_constraints/) or use this [guide](https://www.baeldung.com/ops/docker-memory-limit).
 
 ### How can I boost machine learning speed?
--- a/docs/docs/FAQ.mdx
+++ b/docs/docs/FAQ.mdx
@@ -232,7 +232,10 @@ to increase the bar for what the algorithm considers a "core face" for that pers
 
 ### The immich_model-cache volume takes up a lot of space, what could be the problem?
 
-If you installed several models and chose not to use some of them, it might be worth deleting the old models that are in immich_model-cache. To do this you can mount the model cache and remove the undesired models:
+If you installed several models and chose not to use some of them, it might be worth deleting the old models that are in immich_model-cache. To do this you can mount the model cache and remove the undesired models.
+
+<details>
+<summary>Steps</summary>
 
 ```bash
 docker run -it --rm -v immich_model-cache:/mnt-models alpine sh
@@ -240,6 +243,7 @@ cd /mnt-models
 ls
 # delete unused models with 'rm -r <model_name>'
 ```
+</details>
 
 ---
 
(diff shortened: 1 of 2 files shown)


danieldietzler:
Regarding Windows I'd personally be fine with just leaving that out lmao. Alternatively we could also run that in the database container...? Not sure if that comes with sed preinstalled but I'd assume so.

bo0tzz:
Why is this pg_catalog line wrong in the first place? Is that something we can fix further up?

mertalev:
It could theoretically be fixed for future users if we changed `CREATE EXTENSION cube` to `CREATE EXTENSION cube SCHEMA pg_catalog` in the migration (also for earthdistance). But I'm not sure if this is allowed for non-superusers.

mmomjian:
> It could theoretically be fixed for future users if we changed `CREATE EXTENSION cube` to `CREATE EXTENSION cube SCHEMA pg_catalog` in the migration (also for earthdistance). But I'm not sure if this is allowed for non-superusers.

Immich has superuser in the default deployment, yes? I have no qualms about adding an extra step for people (like myself) who run an external PG instance - we know what we signed up for. My suggestion would be to have some kind of permission check +log message that brings the container down if this extension change is unsuccessful, to alert the admin that a change need to be made.

mertalev:
Oh right, I forgot that these aren't trusted extensions so they need admin intervention anyway.

I think we can update the instructions to include `SCHEMA pg_catalog` and change the migration as well. This will solve this issue for new instances without breaking anything for existing ones.

mmomjian:
Will installing the extension into pg_catalog cause any issues for users or companies that run multiple instances of Immich in the same DB or alongside other services?

I think an alternate option would be to run the migration and if it fails, simply log an error and expect that the DB admin can manage their own pg_restore with a search path change if needed. The data is all there, this just makes the restore process a little cleaner. Of course if it's no issue to have it in pg_catalog that doesn't hurt either.

mertalev:
Each database has its own `pg_catalog` schema, so it shouldn't affect anything else. That being said, there is a risk that it could negatively impact managed deployments with potentially exotic setups.

For your suggestion, we can't just catch an exception and move on if a migration fails. All migrations are in the same transaction that will be aborted if anything throws an error. But it could be handled as a startup check outside of the migrations instead.

mmomjian:
Do you want me to remove the `sed` changes from this PR and open a new issue instead to track this fix? Let me know what you think is best.

mertalev:
The `sed` command is fine. I think we can add that now and remove it at a later time when the code change is in and most people have updated.

docs/docs/administration/backup-and-restore.md
```diff
@@ -15,7 +15,7 @@ Immich saves [file paths in the database](https://github.com/immich-app/immich/d
 Refer to the official [postgres documentation](https://www.postgresql.org/docs/current/backup.html) for details about backing up and restoring a postgres database.
 :::
 
-The recommended way to backup and restore the Immich database is to use the `pg_dumpall` command.
+The recommended way to backup and restore the Immich database is to use the `pg_dumpall` command. When restoring, you need to delete the `DB_DATA_LOCATION` folder (if exists) to reset the database.
```
mertalev, line 18:
```suggestion
The recommended way to backup and restore the Immich database is to use the `pg_dumpall` command. When restoring, you need to delete the `DB_DATA_LOCATION` folder (if it exists) to reset the database.
```

docs/docs/FAQ.mdx
```diff
@@ -230,14 +232,19 @@ to increase the bar for what the algorithm considers a "core face" for that pers
...
 
-To do this you can run:
+<details>
+<summary>Steps</summary>
+
+```bash
+docker run -it --rm -v immich_model-cache:/mnt-models alpine sh
+cd /mnt-models
+ls
+# delete unused models with 'rm -r <model_name>'
```
mertalev, line 244: The models are under `clip` and `facial-recognition`, so they should either `cd` to those folders first or adjust the `rm -r` command to include that part of the path.

mertalev (approved):

mertalev (approved):

PR #9022, opened 2024-04-22
chore(server): enable swagger to take class validator rules into account

danieldietzler:


open-api/immich-openapi-specs.json
```diff
@@ -8286,6 +8299,7 @@
           },
           "quotaSizeInBytes": {
             "format": "int64",
+            "minimum": 1,
```
danieldietzler, line 8302: Do we want to allow 0 for this?

bo0tzz: In the docs/UI we specify setting 0 for unlimited quota, and the final quota check also implicitly relies on that.
https://github.com/immich-app/immich/blob/c9a079201a38a4b5d272e8120c6daf7acb30bba5/server/src/services/asset-v1.service.ts#L363

danieldietzler: Yeah I was thinking the same, but then I wonder if that even was possible in the past since `@IsPositive()` should only allows numbers > 0 lol
Screenshot 2024-04-22 at 21 08 11

bo0tzz: I think we should make this consistent - either we allow 0 and use that to signify unlimited quota, or we make it explicit that the field being null/undefined means unlimited quota.

danieldietzler: Yeah I allow 0 now, but I was primarily just confused that it was working before.

jrasm91: You set it to null to reset it. This PR just documents the current rules/enforcement in the spec, it doesn't actually change the implementation.

jrasm91: `quotaSizeInBytes: quotaSize ? convertToBytes(Number(quotaSize), 'GiB') : null,` the web sends null when the user puts in 0.

jrasm91 (approved):
Personally, I think this PR should just add the shim, not change any implementation. We can always do that in a separate PR since it is unrelated.

open-api/immich-openapi-specs.json
danieldietzler: Ok that works for me!

PR #9043, opened 2024-04-23
chore(docs): Update backup-and-restore.md

r7o3gko8:
Removed additional note about sed command as it is included in the standard procedure


mmomjian:
Thank you for your PR, and for noticing this so quickly! My recommendation is to hold off on changing this for now, 1) because the Windows restore does not have the command yet and 2) in the near future we may remove the sed command from the default restore but will need to keep it documented for people with older installs.

PR #9051, opened 2024-04-24
fix(server): Improve reverse geocoded location metadata

### LINKED ISSUE #8941: Reverse geocoding is frequently incorrect
### The bug

I have a bunch of photos from one place that get incorrect location tags. This makes it difficult to correctly search for photos from one particular city.  Most of the times it's neighborhood names that get classified as cities, cities as states.

Here are some examples:
image
- Actual:
  - City: Paseos del Valle
  - State: Tlajomulco de Zuñiga, Jalisco
  - Country: Mexico
Expected:
  - City: Tlajomulco de Zuñiga
  - State: Jalisco
  - Country: México
  
  ---
  
image
- Actual:
  - City: Brooklyn Heights
  - State: Kings County, New York
  - Country: United States of America
- Expected:
  - City: New York City (Brooklyn might be alright too)
  - State: New York
  - Country: United States of America

Linux raspberrypi 6.1.0-rpi7-rpi-2712

### Version of Immich Server

v1.101.0

v1.101.0 build.147

### Platform with the issue

- [X] Server
- [X] Web
- [X] Mobile

```YAML
Not relevant?
```

### Your .env content

```Shell
Not relevant?
```

### Reproduction steps

```bash
1. Open up place search.
2. Search for a place that is not a city inside "City" or open a photo and see the location tag.
```

### Relevant log output

_No response_

### Additional information

_No response_

hermesespinola:
## Description

- Limit state field to admin1Name (1st level country divisions).
- Add filter to remove 'PPLX' Feature codes from cities, which is classified as "section of populated place" (see https://www.geonames.org/export/codes.html).
- Updated tests to reflect these changes.

- Improves the reverse geocoding features. There are still edge cases.

Fixes #8941 

## How Has This Been Tested?

- [ ] Download photos from my Immich server.
- [ ] Upload theses photos to the dev version.
- [ ] Compare the difference between the two.

**After**:

image
image
image

**Before**:

image
image
image

## Checklist:


hermesespinola:
I see there are still some incorrect label but it boils down to the GeoNames database:

image

Ideally this should be tagged as "Tlajomulco de Zuñiga, Jalisco, México", but there are a lot of places that are tagged as 'PPL' rather than 'PPLX'.  I wonder if it'd be good idea to add some custom regex filter to the reverse geocoding settings or some other customization on how the feature works.

danieldietzler (approved):

alextran1502 (approved):

alextran1502:
@zackpollard Feel free to merge this one after you have taken a look

zackpollard:
I like this change for the immediate effect, however in future would it be more beneficial to keep these in but process them differently, so have more fields for storing this geo data at a lower level than city?

hermesespinola:
What would be the use case for this data? I could create a migration to add this field to the geodata entities and filter them during reverse geocoding.

One of the main problems is that sometimes a pplx entity is close to the media location (i.e.: the San Francisco photo) even if that photo is not taken in that place (I guess because the reverse geocoding is based on proximity rather than geofencing).

zackpollard:
> What would be the use case for this data? I could create a migration to add this field to the geodata entities and filter them during reverse geocoding.
> 
> One of the main problems is that sometimes a pplx entity is close to the media location (i.e.: the San Francisco photo) even if that photo is not taken in that place (I guess because the reverse geocoding is based on proximity rather than geofencing).

Yea, perhaps the real problem is that we don't have enough of that type of data. I say we merge this and revisit it once we look at importing the entire allCountries dataset which includes way more of this finer detail data 😄

Nixon506E:
I am seeing the API exif data responding with the county name in the state exif while the web and mobile do not show it. Is there something that could be added from this PR that the API isn't following?

marcules:
Hi @zackpollard @hermesespinola

I'm living in a "PPLX" https://www.geonames.org/2814835/waldperlach.html and the way this is currently coded (by distance, not geofencing?) I'm closer to the nearest municipality (Neubiberg) instead of the proper city center (Munich).

I would suspect this to be the case for all locations that are not near the "center" of a city but rather at the periphery.

I have not yet looked into this, but would you be open for a PR that changes the determination of where the location belongs to to be contained in a geofence first for PPLAs (and up) and then fall back to nearest location as it is currently implemented? Or alternatively removing the filter again, but then selecting the PPLA from the hierarchy of the PPLX?

Because the PPLA my PPLX is contained in would be "more correct" to display as the location:
https://www.geonames.org/2867714/munich.html

If you're open to this, I would look into implementing this.

Here's an example of GEO Coordinates that do not doxx me: lat 48,100386 lon 11,648861
I would expect "Neuperlach (Munich)" or "Munich" to be the location but not "Neubiberg"

Would be nice if you could give me a short message if you're interested in such a contribution or rather not.

zackpollard:
> Hi @zackpollard @hermesespinola
> 
> I'm living in a "PPLX" https://www.geonames.org/2814835/waldperlach.html and the way this is currently coded (by distance, not geofencing?) I'm closer to the nearest municipality (Neubiberg) instead of the proper city center (Munich).
> 
> I would suspect this to be the case for all locations that are not near the "center" of a city but rather at the periphery.
> 
> I have not yet looked into this, but would you be open for a PR that changes the determination of where the location belongs to to be contained in a geofence first for PPLAs (and up) and then fall back to nearest location as it is currently implemented? Or alternatively removing the filter again, but then selecting the PPLA from the hierarchy of the PPLX?
> 
> Because the PPLA my PPLX is contained in would be "more correct" to display as the location:
> https://www.geonames.org/2867714/munich.html
> 
> If you're open to this, I would look into implementing this.
> 
> Here's an example of GEO Coordinates that do not doxx me: lat 48,100386 lon 11,648861
> I would expect "Neuperlach (Munich)" or "Munich" to be the location but not "Neubiberg"
> 
> Would be nice if you could give me a short message if you're interested in such a contribution or rather not.

We generally want to move to geofences for everything, however the main problem is sourcing that data and getting it into a format that we can use in postgres as well as the infrastructure to deliver that to end users.

marcules:
@zackpollard thank you for the fast reply - if I'm reading this right, you're saying that this is not the right place to introduce something that you want to have everywhere and for everything, like geofencing, and I think I would agree.

---

Nevertheless, would you be open for the second idea, which was, not filtering PPLX anymore but then using its hierarchical PPLA for the reverse geocoding (i.e. "PPLX Waldperlach" has a parent "PPLA Munich" so we return "Munich Germany"). This would also be a very minimal and non-invasive change, if I'm not mistaken.

Also, that kind of change would satisfy the initial complaint about Chinatown and Brooklyn Heights, as those have parents to PPL(A)s San Francicso and New York City

For example:

Chinatown (PPLX) -> San Francisco (PPLA2) -> No Further PPL Parent
image


Brooklyn Heights (PPLX) -> Brooklyn (PPLA) -> New York City (PPL) -> No Further PPL Parent




Waldperlach (PPLX) -> Munich (PPLA) -> No Further PPL Parent

hermesespinola:
It'd be also interesting if photos could be tagged with other feature metadata for points of interest: HIST, RUIN, CH, TPL, MNMT, etc. For example: we'd be able to search for "Roman Colosseum" and get relevant photos.

Issue #5644, opened 2023-12-11
[BUG] Documentation for a exclusion of a folder in external library is wrong


Miluttin:
### The bug

In description for adding exclusion path to external library it is written '... To ignore an absolute path, use "/path/to/ignore".'
It's not working without ** at the end of the path.
So it should be: To ignore an absolute path, use "/path/to/ignore/**"

### The OS that Immich Server is running on

Docker

### Version of Immich Server

v1.88.2

### Version of Immich Mobile App

v.1.90

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
services:
  immich-server:
    container_name: immich_server
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    command: [ "start.sh", "immich" ]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - /media/nas/private:/media/nas/private
      - /etc/localtime:/etc/localtime:ro
    env_file:
      - .env
    ports:
      - 2283:3001
    depends_on:
      - redis
      - database
      - typesense
    restart: always

  immich-microservices:
    container_name: immich_microservices
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    # extends:
    #   file: hwaccel.yml
    #   service: hwaccel
    command: [ "start.sh", "microservices" ]
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - /media/nas/private:/media/nas/private
      - /etc/localtime:/etc/localtime:ro
    env_file:
      - .env
    depends_on:
      - redis
      - database
      - typesense
    restart: always

  immich-machine-learning:
    container_name: immich_machine_learning
    image: ghcr.io/immich-app/immich-machine-learning:${IMMICH_VERSION:-release}
    volumes:
      - model-cache:/cache
    env_file:
      - .env
    restart: always

  typesense:
    container_name: immich_typesense
    image: typesense/typesense:0.24.1@sha256:9bcff2b829f12074426ca044b56160ca9d777a0c488303469143dd9f8259d4dd
    environment:
      - TYPESENSE_API_KEY=${TYPESENSE_API_KEY}
      - TYPESENSE_DATA_DIR=/data
      # remove this to get debug messages
      - GLOG_minloglevel=1
    volumes:
      - tsdata:/data
    restart: always

  redis:
    container_name: immich_redis
[...]

PR #9057, opened 2024-04-24
Update community-projects.tsx

Thoroslives:
I have made a repo for the remove offline assets python script that was linked as one of my gists. This repo has much more explanation as the the usage and troubleshooting and allows for the contribution from other community members.

PR #9059, opened 2024-04-24
docs: small changes&updates

aviv926:
This PR adds the following:

- [x] - Starting with the new release it seems [necessary to specify](https://discord.com/channels/979116623879368755/994044917355663450/1231894723078127626) the database name at restore time.
- [x] - Adding and updating new guides and projects to the relevant pages.
- [x] - Added a new "documentation" section to the PR Checklist.
~- [ ] - Added 360 support to the README file by #8762.~
- [x] - fix for #5644.


README.md
```diff
@@ -94,7 +95,7 @@ Spec: Free-tier Oracle VM - Amsterdam - 2.4Ghz quad-core ARM64 CPU, 24GB RAM
 | OAuth support                                | Yes    | Yes |
 | API Keys                                     | N/A    | Yes |
 | LivePhoto/MotionPhoto backup and playback    | Yes    | Yes |
-| Support 360 degree image display             | No     | Yes |
+| Support 360 degree image display             | Yes    | Yes |
```
bo0tzz, line 98: This is marking mobile as supporting panoramas, but the PR you linked just affected support for 360 videos on the web.

docs/src/components/community-guides.tsx
```diff
@@ -33,6 +33,11 @@ const guides: CommunityGuidesProps[] = [
     description: 'Documentation for simple podman setup using quadlets.',
     url: 'https://github.com/tbelway/immich-podman-quadlets/blob/main/docs/install/podman-quadlet.md',
   },
+  {
+    title: 'gphotos import + albums',
+    description: 'Import your G photos files into Immich and add your albums',
+    url: 'https://github.com/tbelway/immich-podman-quadlets/blob/main/docs/install/podman-quadlet.md',
+  },
```
bo0tzz, line 40: Bad copy paste?

docs/src/components/community-projects.tsx
```diff
@@ -26,7 +26,12 @@ const projects: CommunityProjectProps[] = [
   {
     title: 'Remove offline files',
     description: 'A Python script to remove offline files.',
-    url: 'https://gist.github.com/Thoroslives/ca5d8e1efd15111febc1e7b34ac72668',
+    url: 'https://github.com/Thoroslives/immich_remove_offline_files/tree/main',
```
bo0tzz, line 29: dupe: #9057

docs/src/components/community-projects.tsx
```diff
@@ -26,7 +26,12 @@ const projects: CommunityProjectProps[] = [
...
+  },
+  {
+    title: 'Move untracked files',
+    description: 'Script to move untracked files from within Immich directories.',
+    url: 'https://github.com/clumsyCoder00/Immich-Tools/tree/main?tab=readme-ov-file#move_untracked_filessh',
   },
```
bo0tzz, line 35: This should just be a general link to the immich-tools repo

bo0tzz (changes requested):

README.md
aviv926: my bad

bo0tzz (approved):
ty :)

docs/docs/administration/backup-and-restore.md
```diff
@@ -33,7 +33,7 @@ docker start immich_postgres    # Start Postgres server
 sleep 10    # Wait for Postgres server to start up
 gunzip < "/path/to/backup/dump.sql.gz" \
 | sed "s/SELECT pg_catalog.set_config('search_path', '', false);/SELECT pg_catalog.set_config('search_path', 'public, pg_catalog', true);/g" \
-| docker exec -i immich_postgres psql --username=postgres    # Restore Backup
+| docker exec -i immich_postgres psql --username=postgres --dbname=immich    # Restore Backup
```
mmomjian, line 36: @aviv926 it is not necessary to specify a DB name at restore time with pg_dumpall, which has been the default for a while. Likely that user has an old backup created with tehe backup docker container that doesn't have the POSTGRES_CLUSTER set to true.

docs/docs/administration/backup-and-restore.md
```diff
@@ -51,7 +51,7 @@ docker compose pull     # Update to latest version of Immich (if desired)
 docker compose create   # Create Docker containers for Immich apps without running them.
 docker start immich_postgres    # Start Postgres server
 sleep 10    # Wait for Postgres server to start up
-gc "C:\path\to\backup\dump.sql" | docker exec -i immich_postgres psql --username=postgres    # Restore Backup
+gc "C:\path\to\backup\dump.sql" | docker exec -i immich_postgres psql --username=postgres --dbname=immich   # Restore Backup
```
mmomjian, line 54: Same here

mmomjian (changes requested):

docs/docs/administration/backup-and-restore.md
aviv926: Is there any downside to this? If not, I think it doesn't hurt to have it because it can solve future problems without affecting future restores (if there is no downside).

mmomjian: Hm. I'm doing some testing to see. The biggest issue I see is that when doing a database restore, you can't dump the connecting database. This means that instead of DROP DATABASE Postgres being the one command that fails, now the DROP DATABASE immich command will fail. It's not clear to me if that will cause issues - if it's a fresh DB it shouldn't? 

As an aside, if we do end up re adding it, there is another instance on line 92 that would need it.

aviv926: Thanks for checking, as far as I know the database **must** be completely clean when performing a restore.
That's why we use

```
docker compose create # Create Docker containers for Immich apps without running them.
docker start immich_postgres # Start Postgres server
...
```

Restore and then bring up Immich to start writing data to the database.
For these reasons I don't see it as a problem.

mmomjian: We do currently advise that. However, there are installations where it is not possible to start a container (Portainer for example). In these instances, adding dbname=immich will make it harder to restore into these systems. I think whatever we specify it will not work for all users. 

Maybe we can add a note to the page that "if you receive an error when restoring, add / remove dbname=immich". I don't have a strong opinion as to which way, but I think we need to provide both options.

docs/docs/developer/pr-checklist.md
```diff
@@ -9,6 +9,11 @@ When contributing code through a pull request, please check the following:
 - [ ] `npm run check:svelte` (Type checking via SvelteKit)
 - [ ] `npm test` (unit tests)
 
+## Documentation
+
+- [ ] `npm run format` (formatting via Prettier)
```
danieldietzler, line 14: This does not actually fix the formatting errors, only `format:fix` does this. I'm not sure if that becomes clear here

docs/src/components/community-guides.tsx
```diff
@@ -33,6 +33,11 @@ const guides: CommunityGuidesProps[] = [
     description: 'Documentation for simple podman setup using quadlets.',
     url: 'https://github.com/tbelway/immich-podman-quadlets/blob/main/docs/install/podman-quadlet.md',
   },
+  {
+    title: 'gphotos import + albums',
+    description: 'Import your G photos files into Immich and add your albums',
```
danieldietzler, line 38: I would prefer writing either "GPhotos" or "Google Photos". I think "Google Photos" would be the best, but I'm open for other opinions :)

danieldietzler (approved):

docs/src/components/community-guides.tsx
aviv926: Agreed, I wasn't sure how problematic it would be to write the word Google in our documentation, but I guess that's fine.

docs/docs/developer/pr-checklist.md
danieldietzler: I see :D Maybe add the something about the existence of format:fix here as well then?

docs/docs/administration/backup-and-restore.md
zackpollard: I think if it's not required, lets leave this change out for now.

zackpollard (changes requested):
Revert dbname changes to restore command

zackpollard (approved):

PR #9068, opened 2024-04-24
fix(server): activity with deleted assets / users

martabal:
Currently, you get an error if a deleted user have commented / liked an asset in the album and there's still the comments of a trashed asset.

This PR adds the ability to navigate to an asset by clicking on the thumbnail of a reaction

### Errors

Trashed asset:
Screenshot from 2024-04-24 23-51-17

Deleted user commented one asset :
Screenshot from 2024-04-24 23-53-42

### Feature

https://github.com/immich-app/immich/assets/74269598/44cb907c-34ff-4f85-a888-9205b6e8cddd

### DIFF
--- a/server/src/queries/activity.repository.sql
+++ b/server/src/queries/activity.repository.sql
@@ -32,8 +32,28 @@ FROM
   AND (
     "ActivityEntity__ActivityEntity_user"."deletedAt" IS NULL
   )
+  LEFT JOIN "assets" "ActivityEntity__ActivityEntity_asset" ON "ActivityEntity__ActivityEntity_asset"."id" = "ActivityEntity"."assetId"
+  AND (
+    "ActivityEntity__ActivityEntity_asset"."deletedAt" IS NULL
+  )
 WHERE
-  (("ActivityEntity"."albumId" = $1))
+  (
+    ("ActivityEntity"."albumId" = $1)
+    AND (
+      (
+        (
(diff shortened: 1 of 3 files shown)


server/src/entities/activity.entity.ts
```diff
@@ -40,10 +40,10 @@ export class ActivityEntity {
   @Column({ type: 'boolean', default: false })
   isLiked!: boolean;
 
-  @ManyToOne(() => AssetEntity, { onDelete: 'CASCADE', onUpdate: 'CASCADE', nullable: true })
+  @ManyToOne(() => AssetEntity, { onDelete: 'CASCADE', nullable: true })
   asset!: AssetEntity | null;
 
-  @ManyToOne(() => UserEntity, { onDelete: 'CASCADE', onUpdate: 'CASCADE' })
+  @ManyToOne(() => UserEntity, { onDelete: 'CASCADE' })
```
jrasm91, line 46: This change seems unnecessary. Why did you make it?

martabal: Because when an asset is trashed, that row becomes `NULL`, so you can't differentiate a global comment from an asset comment.

jrasm91: But that has nothing to do with this cascade.

jrasm91 (approved):

PR #9069, opened 2024-04-25
feat(server): env variable to skip migrations on startup

mertalev:
## Description

Some deployment methods, like Portainer and TrueNAS SCALE, make it difficult to start the database without starting everything else as well. This makes it awkward to restore from a backup as the server will run migrations that then cause duplicate key violations, etc.

This PR adds an env variable to ease this process, skipping migrations to allow the admin to restore to a clean database.

### DIFF
--- a/docs/docs/install/environment-variables.md
+++ b/docs/docs/install/environment-variables.md
@@ -58,15 +58,16 @@ These environment variables are used by the `docker-compose.yml` file and do **N
 
 ## Database
 
-| Variable                            | Description                                                   |   Default    | Services              |
-| :---------------------------------- | :------------------------------------------------------------ | :----------: | :-------------------- |
-| `DB_URL`                            | Database URL                                                  |              | server, microservices |
-| `DB_HOSTNAME`                       | Database Host                                                 | `localhost`  | server, microservices |
-| `DB_PORT`                           | Database Port                                                 |    `5432`    | server, microservices |
-| `DB_USERNAME`                       | Database User                                                 |  `postgres`  | server, microservices |
-| `DB_PASSWORD`                       | Database Password                                             |  `postgres`  | server, microservices |
-| `DB_DATABASE_NAME`                  | Database Name                                                 |   `immich`   | server, microservices |
-| `DB_VECTOR_EXTENSION`<sup>\*1</sup> | Database Vector Extension (one of [`pgvector`, `pgvecto.rs`]) | `pgvecto.rs` | server, microservices |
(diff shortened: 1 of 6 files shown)


mmomjian:
Ty. Can we update line 76 which states "When `DB_URL` is defined, the other database (`DB_*`) variables are ignored, with the exception of `DB_VECTOR_EXTENSION`."  The same note is in the pre existing Postgres docs.

mertalev:
> Ty. Can we update line 76 in backup and restore which states "When `DB_URL` is defined, the other database (`DB_*`) variables are ignored, with the exception of `DB_VECTOR_EXTENSION`." The same note is in the pre existing Postgres docs.

Good catch! I changed it to just list the relevant variables so it won't get outdated as more variables are added.

mmomjian:
> > Ty. Can we update line 76 in backup and restore which states "When `DB_URL` is defined, the other database (`DB_*`) variables are ignored, with the exception of `DB_VECTOR_EXTENSION`." The same note is in the pre existing Postgres docs.
> 
> 
> 
> Good catch! I changed it to just list the relevant variables so it won't get outdated as more variables are added.

Ty! I misspoke its also on line 92 of env vars docs that also contains the same language. I like switching to explicitly saying which ones are ignored.

docs/docs/install/environment-variables.md
```diff
@@ -58,21 +58,22 @@ These environment variables are used by the `docker-compose.yml` file and do **N
...
+| `DB_DATABASE_NAME`                  | Database Name                                                            |   `immich`   | server, microservices |
+| `DB_VECTOR_EXTENSION`<sup>\*1</sup> | Database Vector Extension (one of [`pgvector`, `pgvecto.rs`])            | `pgvecto.rs` | server, microservices |
+| `DB_SKIP_MIGRATIONS`                | Whether to skip running migrations on startup (one of [`true`, `false`]) |   `false`    | server, microservices |
+
+\*1: This setting cannot be changed after the server has successfully started up.
 
 :::info
 
-When `DB_URL` is defined, the other database (`DB_*`) variables are ignored, with the exception of `DB_VECTOR_EXTENSION`.
+When `DB_URL` is defined, the `DB_HOSTNAME`, `DB_PORT`, `DB_USERNAME`, `DB_PASSWORD` and `DB_DATABASE_NAME` database variables are ignored.
```
mmomjian, line 76: Ty!

mmomjian (approved):

jrasm91 (approved):

PR #9071, opened 2024-04-25
fix(cli): `--ignore` getting ignored

mertalev:
## Description

The internal typing for this flag expects `exclusionPatterns`, so it will always consider it to be `undefined`.

## How Has This Been Tested?

Tested that it actually ignores the pattern specified in the command.

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -9,6 +9,7 @@
       "version": "2.2.0",
       "license": "GNU Affero General Public License version 3",
       "dependencies": {
+        "fast-glob": "^3.3.2",
         "lodash-es": "^4.17.21"
       },
       "bin": {
@@ -31,7 +32,6 @@
         "eslint-config-prettier": "^9.1.0",
         "eslint-plugin-prettier": "^5.1.3",
         "eslint-plugin-unicorn": "^52.0.0",
-        "glob": "^10.3.1",
         "mock-fs": "^5.2.0",
         "prettier": "^3.2.5",
         "prettier-plugin-organize-imports": "^3.2.4",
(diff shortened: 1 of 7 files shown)


jrasm91 (approved):
Can you add an e2e test for this?

mertalev:
It turns out there was more than one issue here:
- `--ignore` has confusing behavior: it's sensitive to the order of the flags (particularly whether it comes before or after the upload path(s)) because it can take multiple arguments. The behavior is predictable if it can only accept one string.
- The behavior is different when a search pattern has `..`: it ignores the `--ignore` flag in this case. Resolving the path fixes this.
- Lastly, the `glob` library still failed to ignore files even with the above changes. `fast-glob` does behave as expected. As a nice bonus, it seems something else depends on `fast-glob` anyway so replacing `glob` with it shrunk the build by about 65kb.

danieldietzler (approved):

jrasm91 (approved):

TransRapid:
immich upload -n -i **/Recently/** -r /media/Vault/Apple\ iCloud\ Photos/7/iCloud\ Photos\ Part\ 68\ of\ 68/
Crawling for assets...
Checking files | ████████████████████████████████████████ | 100% | ETA: 0s | 2/2 assets
Found 2 new files and 0 duplicates
Would have uploaded 2 assets (9.5 MB)

Expected behavior would be 0 assets as the only 2 things are in a folder titled. _Recently Deleted_

danieldietzler:
The CLI does not randomly ignore folders because of their name. And IMO that's a good thing, in case that's what you're referring to @TransRapid

TransRapid:
No, the issue is the `--ignore` just does not work as described . The ignore pattern does not ignore using the CLI, it does however work using the GUI.

mertalev:
Can you confirm that you're actually using a CLI build that has the fixes in this PR? It isn't in the latest CLI release.

macross5:
Do we know when the fix will be released?

JustABoringUser:
Tried with the most recent release of CLI tool but --ignore patterns do not work as soon as I add more than just a single pattern.

Works:
`docker exec -it immich_server immich upload --ignore **/*.arw --recursive "import_ds218/pics"`

Nothing of this works:
`docker exec -it immich_server immich upload --ignore **/*.{arw,dng} --recursive "import_ds218/pics"`

`docker exec -it immich_server immich upload --dry-run --ignore **/*.arw **/*.dng --recursive "import_ds218/pics"`

`docker exec -it immich_server immich upload --dry-run --recursive "import_ds218/pics" --ignore **/*.{arw,dng}`

`docker exec -it immich_server immich upload --dry-run --recursive "import_ds218/pics" --ignore **/*.arw **/*.dng`

mertalev:
That's because multiple `--ignore` patterns isn't supported. But you can use any pattern that [micromatch](https://github.com/micromatch/micromatch) supports, including [extglobs](https://github.com/micromatch/micromatch?tab=readme-ov-file#extglobs).

mmvr65:
@mertalev Can you give a command line example how to exclude *.nef and *.raw please ?
I can't figure out how to put it together.
Thanks in advance.
Mike

PR #9074, opened 2024-04-25
fix(web): add debounce to location search

nghduc97:
The search input on change location dialog ignore changes when search is in place leading to issue https://github.com/immich-app/immich/issues/9061. This PR fixes it by adding debounce on the input.

### DIFF
--- a/web/src/lib/components/shared-components/change-location.svelte
+++ b/web/src/lib/components/shared-components/change-location.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
   import { createEventDispatcher } from 'svelte';
   import ConfirmDialogue from './confirm-dialogue.svelte';
-  import { timeBeforeShowLoadingSpinner } from '$lib/constants';
+  import { timeDebounceOnSearch } from '$lib/constants';
   import { handleError } from '$lib/utils/handle-error';
 
   import { clickOutside } from '$lib/utils/click-outside';
@@ -22,7 +22,7 @@
   let places: PlacesResponseDto[] = [];
   let suggestedPlaces: PlacesResponseDto[] = [];
   let searchWord: string;
-  let isSearching = false;
+  let latestSearchTimeout: number;
   let showSpinner = false;
   let suggestionContainer: HTMLDivElement;
   let hideSuggestion = false;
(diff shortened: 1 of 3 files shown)


web/src/lib/components/shared-components/change-location.svelte
```diff
@@ -66,24 +66,35 @@
...
+        }
+      } catch (error) {
+        // skip error when a newer search is happening
+        if (latestSearchTimeout === searchTimeout) {
+          places = [];
+          handleError(error, "Can't search places");
+          showSpinner = false;
+        }
+      }
+    }, timeDebounceOnSearch);
```
benmccann, line 96: you should just ignore errors with eslint-disable. you probably need to write the code differently. maybe something like:
```suggestion
    const searchTimeout = window.setTimeout(() => {
      searchPlaces({ name: searchWord }).then(
        (searchResult) => {
          // skip result when a newer search is happening
          if (latestSearchTimeout === searchTimeout) {
            places = searchResult;
            showSpinner = false;
          }
        },
        (error) => {
          // skip error when a newer search is happening
          if (latestSearchTimeout === searchTimeout) {
            places = [];
            handleError(error, "Can't search places");
            showSpinner = false;
          }
        },
      );
    }, timeDebounceOnSearch);
```

martabal:
Can't we just abort the previous request ?

web/src/lib/components/shared-components/change-location.svelte
```diff
@@ -66,24 +66,35 @@
...
-      places = await searchPlaces({ name: searchWord });
-    } catch (error) {
-      places = [];
-      handleError(error, "Can't search places");
-    } finally {
-      clearTimeout(timeout);
-      isSearching = false;
-      showSpinner = false;
+    if (latestSearchTimeout) {
+      clearTimeout(latestSearchTimeout);
```
nghduc97, line 75: > Can't we just abort the previous request ?

Yes, abort if no requests have been sent, if already sent then ignore the result.

web/src/lib/components/shared-components/change-location.svelte
nghduc97: Other parts of code use await/async style so I just followed. I modified since you prefer this way.

benmccann: There's generally no problem with async/await. The issue here was that `window.setTimeout` expects you to pass it a synchronous function and you were passing it an async function

alextran1502 (approved):

web/src/lib/components/shared-components/change-location.svelte
martabal: No, I mean using an AbortController. The timeout is used to avoid the loading spinner from blinking.

PR #9100, opened 2024-04-26
fix(server): stacked assets for full sync, userIds as array for delta sync

fyfrey:
server changes required for #8842


server/src/dtos/sync.dto.ts
```diff
@@ -27,7 +27,9 @@ export class AssetFullSyncDto {
 export class AssetDeltaSyncDto {
   @ValidateDate()
   updatedAfter!: Date;
+
   @ValidateUUID({ each: true })
+  @ApiProperty({ name: 'userIds[]' })
```
danieldietzler, line 32: You'd want `@IsArray()` instead. That also gives you proper validation

fyfrey: Are you sure? I think, I tried that first. IsArray is already set in the open API json description.

This change makes the client send a single value so that express sees it as an array in any case.

danieldietzler: Apparently you are right, sorry. I think you want both annotations then though, right?

jrasm91: You should be able to delete the API property without impacting the spec file

jrasm91 (approved):

server/src/dtos/sync.dto.ts
jrasm91: After you delete this line, feel free to merge it.

jrasm91: This was specifically for a query param, which indeed was needed. I just changed it to post to avoid these types of serialization issues entirely.

server/src/interfaces/asset.interface.ts
```diff
@@ -138,6 +138,8 @@ export interface AssetFullSyncOptions {
   lastCreationDate?: Date;
   lastId?: string;
   updatedUntil: Date;
+  isArchived?: false;
```
fyfrey, line 141: Do we need these two options? I think we can drop them and let the server automatically set them: true for owner, false for partner

jrasm91: I think the server does set them. This is just the interface for the repository method. It means we can assert the right values are passed in unit tests, not rely on the implementation itself.

server/src/repositories/asset.repository.ts
```diff
@@ -800,21 +800,23 @@ export class AssetRepository implements IAssetRepository {
...
+    const builder = this.getBuilder({
+      userIds: [ownerId],
+      exifInfo: true,
+      withStacked,
+      isArchived,
+    });
+
     if (lastCreationDate !== undefined && lastId !== undefined) {
-      builder = builder.andWhere('(asset.fileCreatedAt, asset.id) < (:lastCreationDate, :lastId)', {
+      builder.andWhere('(asset.fileCreatedAt, asset.id) < (:lastCreationDate, :lastId)', {
```
danieldietzler, line 812: Are you sure this works? We assign the (presumably new) builder in other places too

danieldietzler (approved):

server/src/repositories/asset.repository.ts
jrasm91: Yes, it doesn't need to be re-assigned.

danieldietzler: Ah ok nice, maybe we could refactor that then in all places?

jrasm91: Go for it!

danieldietzler: #9119 :)

PR #9104, opened 2024-04-27
chore(server): remove old asset search

jrasm91:


mertalev (approved):
I think you can also remove the `search` method in the search service and the `searchMetadata` method in the asset repo.

danieldietzler (approved):
🚀

PR #9109, opened 2024-04-27
chore: for redis, use service name as the default

jrasm91:
* default to service name instead of container name


danieldietzler (approved):

bo0tzz:
This works without changing `redis.container_name`?

jrasm91:
> This works without changing `redis.container_name`?

I don't need to change the container name and it still works because `redis` is now referring to the service name of the container. Under the `services` section of the compose file we have an entry for `redis` and this is referring to that now.

PR #9118, opened 2024-04-27
chore(server): default postgres hostname to compose service name

jrasm91:
* change default back to `database` (reverse change made [here](https://github.com/immich-app/immich/pull/1381/files#diff-69986fee491fd46475173d3925e5d1a56cbdce049f16a5c30cf39b93d6c8c380)).
* npm run typeorm:migrations:* commands still work
* updated warning for export data source and fixed one stray usage
* number of required env variables goes down by one

### DIFF
--- a/docker/example.env
+++ b/docker/example.env
@@ -13,6 +13,5 @@ DB_PASSWORD=postgres
 
 # The values below this line do not need to be changed
 ###################################################################################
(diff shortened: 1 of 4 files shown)


danieldietzler (approved):
LGTM!

bo0tzz (approved):

PR #9121, opened 2024-04-27
feat(server): remove inactive sessions

jrasm91:
* Remove session tokens that haven't been used in 90 days.
* Fix an issue with `make sql`


server/src/services/session.service.spec.ts
```diff
@@ -26,6 +27,14 @@ describe('SessionService', () => {
     expect(sut).toBeDefined();
   });
 
+  describe('search', () => {
```
bo0tzz, line 30: Should we test deleting outdated tokens and preserving recent tokens?

server/src/services/session.service.ts
```diff
@@ -19,6 +21,25 @@ export class SessionService {
     this.access = AccessCore.create(accessRepository);
   }
 
+  async handleCleanup() {
+    const sessions = await this.sessionRepository.search({
+      updatedBefore: DateTime.now().minus({ days: 90 }).toJSDate(),
```
bo0tzz, line 26: Do we want to make the timeout configurable?

bo0tzz (approved):

server/src/services/session.service.spec.ts
jrasm91: Yes, had to run though and that's all I finished on the first pass lol

server/src/services/session.service.ts
jrasm91: Maybe?

danieldietzler: @bo0tzz where would you put that in the admin settings? User Settings?

jrasm91: We maybe should move password and auth and oauth into a parent auth category in the UI.

server/src/services/session.service.spec.ts
bo0tzz: All good. Should this PR still be draft then? 😛

server/src/services/session.service.ts
bo0tzz: An overarching auth section is probably the cleanest, but under server or user settings would work for me as well.

server/src/services/session.service.spec.ts
danieldietzler: Nope I'm on it :)

danieldietzler: Ideally I'd also like to have e2e tests for testing the repository method but since this is a nightly job that's a bit tricky I think.

server/src/services/session.service.ts
danieldietzler: Moving the auth/oauth settings around and adding the delay for deleting sessions will be its own PR

PR #9135, opened 2024-04-28
feat(server, web): album order preference

martabal:
The goal is to have a new user settings to choose the default album order (the option is saved in db). The options are `Backward` and `Forward`. 

## Screenshots

User settings :
Screenshot from 2024-04-28 15-17-37


server/src/services/search.service.ts
```diff
@@ -83,14 +84,27 @@ export class SearchService {
...
     const size = dto.size || 250;
-    const enumToOrder = { [AssetOrder.ASC]: 'ASC', [AssetOrder.DESC]: 'DESC' } as const;
+
+    const enumToOrder = { [AssetOrderPreference.ASC]: 'ASC', [AssetOrderPreference.DESC]: 'DESC' } as const;
+
+    const mapOrder = (order?: AssetOrder) => {
+      if (order) {
+        if (order === AssetOrder.PREFERENCE) {
+          return enumToOrder[auth.user.preferedAlbumOrder];
+        }
```
danieldietzler, line 94: This does not need to be in the outer if. You can simply move it in front of it :)

server/src/services/timeline.service.ts
```diff
@@ -40,7 +41,9 @@ export class TimelineService {
   private async buildTimeBucketOptions(auth: AuthDto, dto: TimeBucketDto): Promise<TimeBucketOptions> {
     const { userId, ...options } = dto;
     let userIds: string[] | undefined = undefined;
-
+    if (dto?.order === AssetOrder.PREFERENCE) {
+      options.order = auth.user.preferedAlbumOrder as unknown as AssetOrder;
```
danieldietzler, line 45: What is wrong with those types so that you need to cast it to `unknown` first?

danieldietzler (approved):
I like this much more now! Thank you :)

server/src/services/timeline.service.ts
martabal: It's because we have 
```ts
export enum AssetOrderPreference {
  ASC = 'asc',
  DESC = 'desc',
}
``` 
and 
```ts
export enum AssetOrder {
  ASC = 'asc',
  DESC = 'desc',
  PREFERENCE = 'preference',
}
```

danieldietzler: `AssetOrderPreference` is a subset of `AssetOrderPreference` though, right? I don't understand why TS doesn't get this :D

martabal: Extending an enum would be cool ...

jrasm91:
Why don't you convert this to ready for review after you have removed the per album/user sorting setting. I think we can keep the default sort order per person, and default to that setting for any new albums, but we don't need per user sort settings for each album.

danieldietzler:
> Why don't you convert this to ready for review after you have removed the per album/user sorting setting. I think we can keep the default sort order per person, and default to that setting for any new albums, but we don't need per user sort settings for each album.

Martabal reverted that, no? Afaict this PR now only has (1) default setting per user and (2) sorting order per album (not per album per user).

martabal:
Yep, I think that's how we decided it to be :thinking:

PR #9151, opened 2024-04-29
fix(web): update assets after delete

### LINKED ISSUE #8338: [Bug] Removing photo from search form doesn't show next photo
### The bug

When you delete an open photo from the search form, the photo does not change to the next one from the list, although it is actually deleted.

TrueNAS-SCALE-23.10.2

### Version of Immich Server

v1.99.0

v1.99.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
https://github.com/truenas/charts/tree/master/community/immich
```

### Your .env content

```Shell
https://github.com/truenas/charts/tree/master/community/immich
```

### Reproduction steps

```bash
1. Go to the photo search, apply any filter
2. Open the photo in the filtered list of photos
3. Click "Delete", we will see the message "Moved to Trash" but the photo is still there
4. Return to the filtered photo list, the deleted photo is still in the list (but the photo is actually deleted)
5. When the window is refreshed, deleted photos disappear from the list. We can find the deleted photos in the Trash
```

### Additional information

_No response_

Xiayucheng1212:
fixes #8338 

## Current Behavior
After click the delete button on the viewer page, it will show the next photo by "navigateAsset('next')". Also, the assets shown in the gallery viewer will also filter out the deleted photo, and update itself.

### DIFF
--- a/web/src/lib/components/shared-components/gallery-viewer/gallery-viewer.svelte
+++ b/web/src/lib/components/shared-components/gallery-viewer/gallery-viewer.svelte
@@ -11,7 +11,8 @@
   import { getAssetRatio } from '$lib/utils/asset-utils';
   import { calculateWidth } from '$lib/utils/timeline-util';
   import { navigate } from '$lib/utils/navigation';
-  import { AssetAction } from '$lib/constants';
+  import { AppRoute, AssetAction } from '$lib/constants';
+  import { goto } from '$app/navigation';
 
   const dispatch = createEventDispatcher<{ intersected: { container: HTMLDivElement; position: BucketPosition } }>();
 
@@ -21,7 +22,7 @@
   export let showArchiveIcon = false;
   export let viewport: Viewport;
 
-  let { isViewing: showAssetViewer, asset: viewingAsset, setAsset } = assetViewingStore;
+  let { isViewing: isViewerOpen, asset: viewingAsset, setAsset } = assetViewingStore;
 
   let currentViewAssetIndex = 0;
   $: isMultiSelectionMode = selectedAssets.size > 0;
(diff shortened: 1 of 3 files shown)


jrasm91 (approved):

web/src/lib/components/shared-components/gallery-viewer/gallery-viewer.svelte
```diff
@@ -65,6 +66,20 @@
     }
   };
 
+  const handleAction = (action: AssetAction, asset: AssetResponseDto) => {
+    switch (action) {
+      case AssetAction.DELETE:
```
jrasm91, line 71: Should we handle the archive action as well like this?

Xiayucheng1212: Yes, archieve is added. Do you think we should also navigate to next photo for action arvhieve?

Xiayucheng1212: Yes, archieve is added. Do you think we should also navigate to next photo for action arvhieve?

jrasm91: The idea is that if you are on the viewer and click "archive", "delete" or "trash" the asset no longer belongs to the current collection, so auto navigating forward seems to make a lot of sense. So yeah, I would have all of them to that same action.

jrasm91: It turns out that it is better if the gallery component handles the navigation to next/previous, in order to not double up on events (the asset grid also does navigation on trash/delete - the parent component of the asset viewer in some situations)

danieldietzler (approved):

PR #9153, opened 2024-04-29
feat(web,a11y): replace IconButton with CircleIconButton

ben-basten:
## Description

Migrate all usages of `IconButton` to `CircleIconButton`, or other similar alternatives.

In my opinion, `CircleIconButton` is preferred because its colors are consistent with other buttons throughout the application, implementation is simpler, and accessible labeling is built-in by default.

New in this PR: 

- The magnifying glass "Search" button in the main search bar now shows an outline when hovered, to make it more obvious that it's clickable.
- The ellipses menu when hovering over a person or album now has a white background, for visibility.
- Added a screen reader only label to the main search box.
- Added a button `type` prop to `CircleIconButton`, to make it more flexible for usage in any scenario. This was a required change to get the search bar buttons working as expected.

<p>

**Light theme toggle**

light

**Dark theme toggle**

dark

**Album options ellipses**

album-options

**Hover over search button**

search-button-hover

</p>
</details> 

## How Has This Been Tested?

Checked the buttons in light mode, light mode hovered, dark mode, and dark mode hovered.

There are also different icons in the top navigation in narrow mobile views, so the same testing needs to be done there too.

## Checklist:

- [x] npm run lint
- [x] npm run format
- [x] npm run check:svelte
- [x] npm test


ben-basten (commented):
FYI @Ethan13310, this will affect your new implementation of the responsive top navigation in #8776. For example, the `mdiMenu` hamburger button in the top navigation will need to be moved over to a `CircleIconButton`. Open to your thoughts on this too!

jrasm91 (approved):

danieldietzler (approved):
Nice!

ben-basten:
Hey @waclaw66, thanks for the feedback!

The more prominent background on the icon was intentional, to give the button guaranteed color contrast in photos where the upper portion is brighter, such as an overexposed sky. To demonstrate, here's what the icon looks like on a white background:

**Before the PR (light mode):**

before-light-v2

**After the PR (light mode):**

after-light

I do see what you mean about it being jarring having a light icon in dark mode. What would you think about leaving the more prominent background, but making it responsive to light mode vs dark mode?

waclaw66:
The main (subjective) problem I have with this PR is that the three dot button background is shown immediately after album/face is hovered. When you hover mouse across multiple albums that moving big three dot button is like crosshair and distorts the impression of elegance. I would preffer to not show the three dot button background immediately when album/face is hovered, but when the button itself is hovered, like it was before and also Google Photos does it. The three dot button background color is just a small thing, it can stay as it is now always in light theme. Thanks for consideration my proposal.

waclaw66:
@ben-basten The only change you need to do is to remove `color="light"` attribute from `CircleIconButton` element from album and face card and add `transparent-primary` to its class to work as below. Thanks.

https://github.com/immich-app/immich/assets/15554561/a7a4e919-9d7b-4932-b561-0e4ac559e1c8

ben-basten:
I have another icon-related PR in the works, and I can add it there once I get some time to test it out.

PR #9188, opened 2024-04-30
Option to disable readonly status of external libraries

benmccann:
This is an updated version of @valpackett's https://github.com/immich-app/immich/pull/8943 with the code review comments addressed

closes https://github.com/immich-app/immich/pull/8943
closes https://github.com/immich-app/immich/discussions/5449
ref https://github.com/immich-app/immich/discussions/8438


server/src/migrations/1714513984514-AddLibraryReadOnly.ts
```diff
@@ -0,0 +1,14 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class AddLibraryReadOnly1714513984514 implements MigrationInterface {
+    name = 'AddLibraryReadOnly1714513984514'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "libraries" ADD "isReadOnly" boolean NOT NULL`);
```
mertalev, line 7: I think you need to set a default here (probably `true`) for the not null constraint to work.

web/src/routes/admin/library-management/+page.svelte
```diff
@@ -243,6 +243,19 @@
...
+
+    try {
+      await updateLibrary({ id: library.id, updateLibraryDto: { isReadOnly: !prevValue } });
+      closeAll();
+      await readLibraryList();
+    } catch (error) {
+      handleError(error, 'Unable to update library');
+    }
+  };
+
```
mertalev, line 258: Just wondering, but how would this code behave if a user clicked multiple times in quick succession?

web/src/routes/admin/library-management/+page.svelte
```diff
@@ -393,12 +406,19 @@
...
+                  {#if showContextMenu && selectedLibrary}
                     <Portal target="body">
                       <ContextMenu {...contextMenuPosition} on:outclick={() => onMenuExit()}>
                         <MenuOption on:click={() => onRenameClicked()} text={`Rename`} />
-
-                        {#if selectedLibrary && selectedLibrary.type === LibraryType.External}
+                        <MenuOption
+                          on:click={() => onToggleReadOnlyClicked()}
+                          text={selectedLibrary.isReadOnly ||
+                          (selectedLibrary.isReadOnly == null && selectedLibrary.type === LibraryType.External)
```
mertalev, line 416: When would `isReadOnly` be null-ish?

alextran1502 (approved):

mertalev:
I see the library's `isReadOnly` getting updated, but don't see where that changes the asset `isReadOnly`. That would have to either be a separate query or a DB trigger. The library's value would also need to be applied to assets that get added afterwards.

server/src/migrations/1714513984514-AddLibraryReadOnly.ts
benmccann: this file is generated, so I don't think I should manually change it? any idea how I'd get it to generate that?

benmccann: and really we should probably default it to true for external libraries and false for upload libraries

web/src/routes/admin/library-management/+page.svelte
benmccann: yeah, it shouldn't be (assuming we do the migration correctly). I've updated it to remove that

server/src/migrations/1714513984514-AddLibraryReadOnly.ts
mertalev: You can just add `default: true` to the `@Column` decorator options. Doing a conditional default is more complicated; It should be fine to add a query here like `UPDATE libraries SET "isReadOnly" = false WHERE type = 'UPLOAD'`

mertalev: I think the query + adding `isReadOnly: false` [here](https://github.com/immich-app/immich/blob/2cf63eeeab5ced19f1764545d7b1c0e0684dbd82/server/src/cores/user.core.ts#L104) would get the job done.

benmccann: thanks! updated this

web/src/routes/admin/library-management/+page.svelte
benmccann: it might be a bit wasteful to do, but nothing jumps out at me as being particularly problematic if that were to occur

server/src/services/asset.service.ts
```diff
@@ -358,7 +358,7 @@ export class AssetService {
 
   async handleAssetDeletion(job: IAssetDeletionJob): Promise<JobStatus> {
     const { id, fromExternal } = job;
-
+    console.error('hanndling asset deletion\n\n\n');
```
jrasm91, line 361: Remove this.

web/src/lib/components/photos-page/actions/delete-assets.svelte
```diff
@@ -34,7 +34,7 @@
 
   const handleDelete = async () => {
     loading = true;
-    const ids = [...getOwnedAssets()].filter((a) => !a.isExternal).map((a) => a.id);
+    const ids = [...getOwnedAssets()].map((a) => a.id);
```
jrasm91, line 37: this might lead to weird behavior as it will actually delete assets that are from read only external libraries and then they will get added back on the next scan, but it's fine with me.

web/src/routes/admin/library-management/+page.svelte
```diff
@@ -243,6 +243,19 @@
     updateLibraryIndex = selectedLibraryIndex;
   };
 
+  const onToggleReadOnlyClicked = async () => {
+    const library = libraries[selectedLibraryIndex];
+    const prevValue = library.isReadOnly == null ? library.type === LibraryType.External : library.isReadOnly;
```
jrasm91, line 248: isReadOnly can't be null. I think you just want if(!external) { return } otherwise flip the value.

jrasm91 (commented):
Looks like you are still missing the server implementation:
1. to set the value when assets are created (via a library scan)
2. to flip the flag on affected assets when the library is updated.

server/src/services/asset.service.ts
benmccann: thanks. leftover from debugging

web/src/routes/admin/library-management/+page.svelte
benmccann: ah, yep

web/src/lib/components/photos-page/actions/delete-assets.svelte
benmccann: updated to check that they're not readonly

benmccann:
Thanks @mertalev and @jrasm91 for pointing out that the original PR didn't actually flip the relevant flag. I've updated that now

server/src/services/library.service.ts
```diff
@@ -249,12 +249,17 @@ export class LibraryService {
...
         break;
       }
       case LibraryType.UPLOAD: {
         if (!dto.name) {
           dto.name = 'New Upload Library';
         }
+        dto.isReadOnly ??= false;
+        if (dto.isReadOnly) {
+          throw new BadRequestException('Upload libraries cannot be readonly');
+        }
```
mertalev, line 262: This dto field doesn't seem to get used - I think you mean to use it at the end.

Also just a preference thing but I think `isReadOnly: dto.isReadOnly ?? dto.type !== LibraryType.UPLOAD` at the end reads better than `dto.isReadOnly ??= true/false`.

server/src/services/library.service.ts
```diff
@@ -503,7 +512,7 @@ export class LibraryService {
         type: assetType,
         originalFileName,
         sidecarPath,
-        isReadOnly: true,
+        isReadOnly: library?.isReadOnly ?? false,
```
mertalev, line 515: I think we can just throw an exception above this if the library is missing: that really shouldn't happen.

server/src/services/library.service.ts
```diff
@@ -359,6 +364,10 @@ export class LibraryService {
       }
     }
 
+    if (dto.isReadOnly !== undefined) {
+      await this.assetRepository.setReadOnlyForLibrary(id, dto.isReadOnly);
```
mertalev, line 368: I think you can just use `updateAll` for this.

benmccann: I believe that `updateAll` requires ids of the items to update, so first I'd need to do a query to get a list of all the ids for the library and then I'd need to do a second query to update them all. This does the update in a single query, which is much easier in my view

server/src/services/library.service.ts
benmccann: thanks!! updated

server/src/services/library.service.ts
benmccann: yeah, I'd actually tried to make that change, but it seems there's quite a few existing tests that are relying on this to work with no library. I'm afraid it's not a small change. It's not a bad idea, but probably better to do in a separate PR if we want to do it to keep this one from growing too large

server/src/services/library.service.ts
jrasm91: A dedicated repo method probably makes the most sense.

server/src/repositories/asset.repository.ts
```diff
@@ -201,6 +201,10 @@ export class AssetRepository implements IAssetRepository {
     );
   }
 
+  async setReadOnlyForLibrary(libraryId: string, isReadOnly: boolean): Promise<void> {
+    await this.repository.update({ library: { id: libraryId } }, { isReadOnly });
```
jrasm91, line 205: External libraries can have isVisible:false motion parts of extracted live photos and they need to be excluded from this update.

server/src/services/library.service.ts
mertalev: Oh, I might have misremembered how it works. I'm fine with a separate repo method.

server/src/services/library.service.ts
mertalev: Hmm gotcha. We can do that in a later PR if it involves a lot of other changes.

server/src/repositories/asset.repository.ts
benmccann: okay. I've updated it, but you should check that I understood correctly: https://github.com/immich-app/immich/pull/9188/commits/3e19aef1f8f05adff6c8fbc248dc85ab1384ad77

jrasm91: Yeah, I think that looks good. Will have to test later with motion photos to make sure are working as expected.

jrasm91:

PR #9200, opened 2024-05-01
chore: Move mobile backup FAQ to correct section

bo0tzz:


danieldietzler (approved):

benmccann (approved):
you could arguably move "### Why are only photos and not videos being uploaded to Immich?" to the mobile section as well, but I think it's fine in either place

jrasm91 (approved):

PR #9229, opened 2024-05-02
feat(server): immich checksum header

jrasm91:
check for `x-immich-checksum` and dedupe before handling the full file upload

### DIFF
--- a/server/src/utils/request.ts
+++ b/server/src/utils/request.ts
@@ -1,0 +1,5 @@
+export const fromChecksum = (checksum: string): Buffer => {
+  return Buffer.from(checksum, checksum.length === 28 ? 'base64' : 'hex');
+};
+
+export const fromMaybeArray = (param: string | string[] | undefined) => (Array.isArray(param) ? param[0] : param);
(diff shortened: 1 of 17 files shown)


server/src/utils/request.ts
```diff
@@ -0,0 +1,5 @@
+export const fromChecksum = (checksum: string): Buffer => {
+  return Buffer.from(checksum, checksum.length === 28 ? 'base64' : 'hex');
```
danieldietzler, line 2: Why can it be guaranteed that base64 is always 28 character long? lol

danieldietzler (approved):

server/src/utils/request.ts
jrasm91: https://stackoverflow.com/a/13378842

benmccann:
Perhaps "hash" would be more accurate than "checksum"? I usually think of a checksum as being used for data integrity

jrasm91:
> Perhaps "hash" would be more accurate than "checksum"? I usually think of a checksum as being used for data integrity

We could also use this for that in the future.

PR #9233, opened 2024-05-02
fix(web): Configure web's immich sdk to use event.fetch

midzelis:
Currently, the console is filled with warnings saying that fetches should be made using event.fetch, not window.fetch. This PR will configure the immich server client to use event.fetch.


danieldietzler (approved):
Oh nice!

PR #9235, opened 2024-05-03
feat(server): read-write external assets

jrasm91:
Breaking change: external libraries and associated assets are no longer read only and can now be deleted.

After a bunch  of back and forth within the developer team, this is the final implementation we landed on:
* Make external assets read/writeable, including able to be deleted.
* Trashing an external asset works like normal and all associated files will be deleted, unless they are mounted as RO
* drops the `isReadOnly` column
* deprecates the `isExternal` and `isReadOnly` asset response properties
* unrelated, but adds a generated sql for adjusting a user's quota (saw that it was missing)

Supersedes: #9188, #8992, #8943
Closes: https://github.com/immich-app/immich/discussions/5449
Reference: https://github.com/immich-app/immich/discussions/8438


server/src/entities/asset.entity.ts
```diff
@@ -106,9 +106,6 @@ export class AssetEntity {
   @Column({ type: 'boolean', default: false })
   isExternal!: boolean;
```
benmccann, line 107: I noticed you removed this from the open API definition. Do we still need it here? If you end up making your change to have null libraryId indicate an upload library then this becomes redundant because you can just check if there's a library id present or not

ThreeAurora:
Thank you so much for all the work you've done on the external libraries, it's the feature I need most at the moment and it's finally here!

server/src/services/storage-template.service.ts
```diff
@@ -170,7 +170,7 @@ export class StorageTemplateService {
   }
 
   async moveAsset(asset: AssetEntity, metadata: MoveAssetMetadata) {
-    if (asset.isReadOnly || asset.isExternal || StorageCore.isAndroidMotionPath(asset.originalPath)) {
+    if (asset.isExternal || StorageCore.isAndroidMotionPath(asset.originalPath)) {
       // External assets are not affected by storage template
       // TODO: shouldn't this only apply to external assets?
```
danieldietzler, line 175:
```suggestion
```
?

danieldietzler (approved):

PR #9253, opened 2024-05-04
Update asset.ts - fix file path translation from Windows > Linux

Donovoi:
Some of my folders in Windows had special characters like "@" which were not being handled correctly and stopped everything else from being processed. This uses the library "upath" to fix the issue. I'm not a typescript developer so please let me know if I'm doing something incorrect. 

Thank you


JW-CH:
Can you update your PR? The build fails and you have some strange changes in the imports etc (changed " to ') etc. Please fix the build errors :)

Donovoi:
HI I believe I sorted the issues, but how do I test if the build is ok? I've tried to follow 
https://immich.app/docs/developer/pr-checklist
I also ran make e2e and it seems to run

Donovoi:
Sweet thank you! Let me know when you do and I'll fix up any issues 😁

zackpollard:
You seem to have an extra dependency "install" that is not required, also the builds and tests are failing. Can you fix these two problems please?

Donovoi:
hopefully that's all done!

cli/package.json
```diff
@@ -59,7 +59,9 @@
   },
   "dependencies": {
     "fast-glob": "^3.3.2",
-    "lodash-es": "^4.17.21"
+    "install": "^0.13.0",
```
zackpollard, line 62: You still have an extra dependency here that has been added erroneously it seems

Donovoi: Awesome thank you for pointing it out. I removed it and re-ran npm install. hopefully its all fixed now 😁

bo0tzz:
I think this has been superseded by #10430?

Issue #8887, opened 2024-04-17
Individual Photo Sharing with Album Link


bitbytenybble110:
### The bug

When an album is shared using the "Sharing" link option viewers of individual photos have the option to click the "Share" button.

Attempting to create a "Sharing" link to a photo on a shared album generates an Authentication Error on the front end web site.

We believe this is mostly a cosmetic error.

Either the "Sharing" link should be removed when not authenticated to avoid the error or "Sharing" links to photos within a shared album can be created by unauthenticated third parties since the album is already public.

Erring on the side of caution, removing the "Sharing" link from unauthenticated sessions feels like the better option.

### The OS that Immich Server is running on

Ubuntu 22.04

### Version of Immich Server

v1.101.0

### Version of Immich Mobile App

v1.101.0 build.147

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
#
# WARNING: Make sure to use the docker-compose.yml of the current release:
#
# https://github.com/immich-app/immich/releases/latest/download/docker-compose.yml
#
# The compose file on main may not be compatible with the latest release.
#

version: "3.8"

services:
  immich-server:
    container_name: immich_server
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    command: ['start.sh', 'immich']
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - /etc/localtime:/etc/localtime:ro
    ports:
      - 2283:3001
    depends_on:
      - redis
      - database
    restart: always
    environment:
      DB_USERNAME: ${DB_USERNAME}
      DB_PASSWORD: ${DB_PASSWORD}
      DB_DATABASE_NAME: ${DB_DATABASE_NAME}
      DB_HOSTNAME: ${DB_HOSTNAME}
    deploy:
      placement:
        constraints: [node.labels.zwave.hardware != true]

  immich-microservices:
    container_name: immich_microservices
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    # extends: # uncomment this section for hardware acceleration - see https://immich.app/docs/features/hardware-transcoding
    #   file: hwaccel.transcoding.yml
    #   service: cpu # set to one of [nvenc, quicksync, rkmpp, vaapi, vaapi-wsl] for accelerated transcoding
    command: ['start.sh', 'microservices']
    volumes:
[...]

PR #9264, opened 2024-05-05
fix(web):  Individual Photo Sharing with Album Link #8887

iamamansharma:
fixes issue #8887 -  Individual Photo Sharing with Album Link


danieldietzler (approved):
LGTM

PR #9279, opened 2024-05-06
fix(web): stream 360 video instead of fetching entire video

TapuCosmo:
The Immich web client fetches entire 360 videos before beginning to play them. This causes long/large 360 videos to be unplayable as the request to fetch a very large (tested with a > 30 GB video) 360 video eventually fails.

This PR fixes that issue by passing the URL to the video asset to the Photo Sphere Viewer module instead of fetching the entire video and converting it to a Blob that is then passed to the aforementioned module.


bo0tzz (approved):
I'm not too familiar with the web code, but I think this approach might be applicable to all the asset serving (also non-panorama images and such), right?

TapuCosmo:
> I'm not too familiar with the web code, but I think this approach might be applicable to all the asset serving (also non-panorama images and such), right?

The existing non-panoramic video player already uses this approach. (Hence why large regular videos do not have the same issue playing.)

jrasm91 (approved):

PR #9282, opened 2024-05-06
fix install.sh add random password for .env

xiagw:


install.sh
```diff
@@ -22,6 +23,12 @@ download_dot_env_file() {
   "${Curl[@]}" "$RepoUrl"/example.env -o ./.env
 }
 
+generate_random_password() {
+  echo "Generate random password for .env file..."
+  rand_pass=$(openssl rand -base64 20 | tr -dc A-Za-z0-9 | head -c10)
+  sed -i -e "s/DB_PASSWORD=postgres/DB_PASSWORD=${rand_pass}/" ./.env
```
mmomjian, line 29: Thanks. I like these changes. I would recommend adding some error handling here. We don’t know for sure if OpenSSL will be present on every system. Can we split the OpenSSL and head commands into 2 lines and then add a `|| return` for all 3 new commands?

mmomjian (changes requested):

danieldietzler (approved):

mmomjian (approved):

install.sh
```diff
@@ -22,6 +23,17 @@ download_dot_env_file() {
   "${Curl[@]}" "$RepoUrl"/example.env -o ./.env
 }
 
+generate_random_password() {
+  echo "Generate random password for .env file..."
+  # rand_pass=$(openssl rand -base64 20 | tr -dc A-Za-z0-9 | head -c10)
```
zackpollard, line 28: I don't think we need to keep the other implementation commented here, lets just remove it if we're going with the other, more compatible implementation
```suggestion
```

zackpollard (changes requested):

zackpollard (approved):

PR #9297, opened 2024-05-07
feat(server,web,mobile): Add external upload endpoint

ddshd:
Adding setting for external upload endpoint.

See more on my own FR: https://github.com/immich-app/immich/discussions/9125

Creating a draft PR if it were to get considered to be merged - otherwise just to get help with some components in server code and I'll keep the changes to my local version.

The `.well_known` and the `uploadDomain` URL would be the only two URL which would need to be publicly accessible.

- [x] Add `uploadDomain` support to server backend code
- [x] Add `uploadDomain` to `.well_known` to allow clients to find and use the correct endpoint.
- [x] Add `uploadDomain` setting to web UI
- [ ] Find a solution to forward main domain cookies to `uploadDomain` so it can authenticate and upload
- [x] Add `uploadDomain` usage to web upload
- [ ] Add `uploadDomain` usage to mobile app
- [ ] Add docs

Users should expose:
```
/asset/exist
/asset/device/{deviceId}
/asset/upload
/.well-known/immich or /server-info/config
```

### DIFF
--- a/mobile/lib/models/server_info/server_config.model.dart
+++ b/mobile/lib/models/server_info/server_config.model.dart
@@ -4,33 +4,38 @@ class ServerConfig {
   final int trashDays;
   final String oauthButtonText;
   final String externalDomain;
+  final String uploadEndpoint;
 
   const ServerConfig({
     required this.trashDays,
     required this.oauthButtonText,
     required this.externalDomain,
+    required this.uploadEndpoint,
   });
 
   ServerConfig copyWith({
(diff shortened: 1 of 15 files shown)


zackpollard:
Hey, we discussed this and this isn't something we want to support within the app officially. Thanks for the PR but this is a very specialised case. If we were to add support for this, it would probably be by simply having the app uploads follow redirects which would allow you to control this setup from your reverse proxy instead.

PR #9315, opened 2024-05-07
refactor: create album users

jrasm91:
Add the option to specify a user's shared permission when creating an album.

Basically replace `sharedUserIds: [user1]` with `albumUsers: [{ userId: user1, role: editor }]` in album create.


danieldietzler (approved):
LGTM

PR #9325, opened 2024-05-08
chore: arabic translations

Me-Mr:
This PR adds translations for the Readme file, and completes the translations that were added in #9287


jrasm91 (approved):

JW-CH:
The links to the other translation pages are broken. Link contains `/readme_i18n/` twice. Also, the links from other translation pages to Arabic are missing. For example → There is no way to get from the German translation page to the Arabic one, because it is missing. @Me-Mr @jrasm91

PR #9334, opened 2024-05-08
chore(server): remove unused deps

jrasm91:
* remove packages we don't use directly
* add a missing `@ApiProperty` decorator


server/src/dtos/album.dto.ts
```diff
@@ -38,6 +38,7 @@ class AlbumUserCreateDto {
   userId!: string;
 
   @IsEnum(AlbumUserRole)
+  @ApiProperty({ enum: AlbumUserRole, enumName: 'AlbumUserRole' })
```
danieldietzler, line 41: It isn't necessary but you still prefer having it there?

danieldietzler (approved):

server/src/dtos/album.dto.ts
jrasm91: It is necessary actually. Otherwise it defaults to strings instead of a named enum.

PR #9344, opened 2024-05-08
refactor(server): auth route metadata

jrasm91:
* Collapse `SharedLinkRoute`, `PublicRoute`, and `AdminRoute` into a single `Authenticated` decorator
* Transition `Authenticated` to be applied per route in preparation for fine-grained permissions

### DIFF
--- a/server/src/middleware/auth.guard.ts
+++ b/server/src/middleware/auth.guard.ts
@@ -10,7 +10,7 @@ import {
 import { Reflector } from '@nestjs/core';
 import { ApiBearerAuth, ApiCookieAuth, ApiOkResponse, ApiQuery, ApiSecurity } from '@nestjs/swagger';
 import { Request } from 'express';
-import { AuthDto } from 'src/dtos/auth.dto';
+import { AuthDto, ImmichQuery } from 'src/dtos/auth.dto';
 import { ILoggerRepository } from 'src/interfaces/logger.interface';
 import { AuthService, LoginDetails } from 'src/services/auth.service';
 import { UAParser } from 'ua-parser-js';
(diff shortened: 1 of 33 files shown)


server/src/middleware/auth.guard.ts
```diff
@@ -19,42 +19,30 @@ export enum Metadata {
...
+    SetMetadata(Metadata.AUTH_ROUTE, options || {}),
   ];
 
-  if (options.admin) {
-    decorators.push(AdminRoute());
-  }
-
-  if (options.isShared) {
-    decorators.push(SharedLinkRoute());
+  if ((options as SharedLinkRoute)?.sharedLink) {
```
danieldietzler, line 37: This isn't pretty lol. I feel like checking the type first would be cleaner

danieldietzler: Maybe we could add an `isSharedRoute` method or something that basically has the type `(options: AuthenticatedOptions) options is SharedLinkRoute => boolean`? (and then a similar thing for AdminRoute)

danieldietzler (approved):
Looks like you haven't missed a single one, nice!

server/src/middleware/auth.guard.ts
jrasm91: This is the only place it is used though.

jrasm91: There is no way to check the type besides doing exactly this.

danieldietzler: Oh because it's just a type, yeah you're right. It is used a couple of lines below as well though

PR #9364, opened 2024-05-09
Add and Use Duplicate Detection Feature Flag

NicholasFlamy:
So I created the `DUPLICATE_DETECTION` feature flag and applied it to the places that it could be used based on the discussion is #8228.

### DIFF
--- a/open-api/typescript-sdk/src/fetch-client.ts
+++ b/open-api/typescript-sdk/src/fetch-client.ts
@@ -170,7 +170,7 @@ export type AlbumResponseDto = {
     updatedAt: string;
 };
 export type AlbumUserCreateDto = {
-    role: AlbumUserRole;
+    role: Role;
     userId: string;
 };
 export type CreateAlbumDto = {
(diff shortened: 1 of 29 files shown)


NicholasFlamy:
Pretty sure the e2e is inherited from #8228 and I need help with the server test.

NicholasFlamy:
Undid some changes because they made it worse.

mertalev (approved):

NicholasFlamy:
Yay, all but the e2e (which is failing the same as with your PR) are successful!

open-api/typescript-sdk/src/fetch-client.ts
```diff
@@ -2852,6 +2857,10 @@ export enum AssetOrder {
     Asc = "asc",
     Desc = "desc"
 }
+export enum Role {
```
jrasm91, line 2860: Why is this here?

NicholasFlamy: It was added by running `make open-api`.

PR #9371, opened 2024-05-10
docs: warn against use on NTFS / WSL

mmomjian:


docs/docs/install/requirements.md
```diff
@@ -15,12 +15,14 @@ Hardware and software requirements for Immich
...
-:::info Podman
-You can also use Podman to run the application. However, additional configuration might be required.
-:::
-
 ## Hardware
 
-- **OS**: Preferred unix-based operating system (Ubuntu, Debian, MacOS, etc). Windows works too, with [Docker Desktop on Windows](https://docs.docker.com/desktop/install/windows-install/)
-- **RAM**: At least 4GB, preferred 6GB.
-- **CPU**: At least 2 cores, preferred 4 cores.
+- **OS**: Recommended Unix-compatible operating system (Ubuntu, Debian, MacOS, etc).
```
danieldietzler, line 20: I would remove MacOS from that list, as FreeBSD does not support LXCs (only some other container thing, Darwin did not assume that feature either though lol), so you end up using docker desktop and a Linux VM anyways.

danieldietzler (approved):

docs/docs/install/requirements.md
mmomjian: Good call. I also change Unix -> Linux because Docker on BSD isn't seamless either.

jrasm91 (approved):

PR #9377, opened 2024-05-10
perf: cache `getConfig`

mertalev:
### Description

When not using a config file, we currently fetch the config on each call. This adds up to quite a few unnecessary queries since this method is called so frequently. Even when using a config file, we apply the same overrides and re-validate the same resulting object each time. This PR caches the config object, only updating it when a config update event is triggered.

### How Has This Been Tested

Tested that jobs process as normal with and without a config file, that updating the config through the web UI updates the config object, and that multiple concurrent calls to `getConfig` result in only one query.

### DIFF
--- a/server/src/interfaces/database.interface.ts
+++ b/server/src/interfaces/database.interface.ts
@@ -20,6 +20,7 @@ export enum DatabaseLock {
   StorageTemplateMigration = 420,
   CLIPDimSize = 512,
   LibraryWatch = 1337,
+  GetSystemConfig = 304,
 }
 
 export const extName: Record<DatabaseExtension, string> = {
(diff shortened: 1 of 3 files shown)


jrasm91 (commented):
Did you test changing config at runtime and verifying the microservices uses the new config? Specifically for stuff like storage template migration and concurrency?

mertalev:
Yep. I ran thumbnail generation and changed the preview format midway.  The success logs immediately started using the new format.

server/src/interfaces/database.interface.ts
```diff
@@ -20,6 +20,7 @@ export enum DatabaseLock {
   StorageTemplateMigration = 420,
   CLIPDimSize = 512,
   LibraryWatch = 1337,
+  GetSystemConfig = 304,
```
danieldietzler, line 23:
```suggestion
  GetSystemConfig = 69,
```

danieldietzler (approved):
LGTM :)

jrasm91:
> Yep. I ran thumbnail generation and changed the preview format midway.  The success logs immediately started using the new format.

Those don't use config$ though

PR #9378, opened 2024-05-10
feat(web): hide all unnamed

martabal:
When you click the eye button in `Show and Hide People` mode, you toggle between "hiding only unnamed people", "hiding all people", "showing all people".

#5928

https://github.com/immich-app/immich/assets/74269598/70a4cee9-8ea3-4c14-bb89-a6e4aac66bab


web/src/lib/components/faces-page/show-hide.svelte
```diff
@@ -1,24 +1,46 @@
...
-
   export let showLoadingSpinner: boolean;
-  export let toggleVisibility: boolean;
+  export let toggleVisibility: ToggleVisibilty = ToggleVisibilty.VIEW_ALL;
   export let screenHeight: number;
   export let countTotalPeople: number;
+  export let onClose = () => {};
+  export let onReset = () => {};
+  export let onChange = () => {};
+  export let onDone = () => {};
```
danieldietzler, line 35:
```suggestion
  export let onClose: () => void;
  export let onReset: () => void;
  export let onChange: () => void;
  export let onDone: () => void;
```
This is what you want, right?

web/src/lib/components/faces-page/show-hide.svelte
```diff
@@ -1,24 +1,46 @@
+<script lang="ts" context="module">
+  export enum ToggleVisibilty {
+    HIDE_ALL = 'hide-all',
+    HIDE_UNNANEMD = 'hide-unnamed',
+    VIEW_ALL = 'view-all',
+  }
+
+  export const changeVisibility = (toggleVisibility: ToggleVisibilty) => {
```
danieldietzler, line 8: This method and the whole data flow because of this is kinda confusing to me. I think I've got an idea to make it more intuitive/simpler:

- `onChange` passes the next visibility option (and the old one if necessary), so `onChange: (next: ToggleVisibility) => void;`
- This `changeVisibility` will be moved "inside" the component and not get exported
- `onChange`will be called with something like `getNextVisibility(toggleVisibility)`

web/src/lib/components/faces-page/show-hide.svelte
martabal: We use both in the codebase, what should we use ?

web/src/lib/components/faces-page/show-hide.svelte
martabal: That's better, thanks !

web/src/lib/components/faces-page/show-hide.svelte
danieldietzler: I would personally prefer this, as it better aligns with other exports, e.g.`export let assetId: string`

danieldietzler (approved):
Thank you, LGTM! :)

zackpollard (approved):

waclaw66:
I've tested this PR, however it doesn't work like described for me. It only hides faces without name, that button doesn't work as toggle.

PR #9382, opened 2024-05-10
fix(web): Fixed video unmutes when scrubbing

### LINKED ISSUE #9367: Video unmutes when scrubbing
### The bug

After muting a video in the player, using the scrub bar to move around in the video will cause it to unmute.

### Version of Immich Server

v1.103.1

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Reproduction steps

```bash
1. Open video in player
2. Press mute
3. Use scrub bar
```

Tushar-Harsora:
Fixes #9367


danieldietzler (approved):
Haven't tested it but the code looks like it should work

NicholasFlamy:
Confirmed fix works.

jrasm91 (approved):

PR #9403, opened 2024-05-12
chore(web): icon consistency and improvements

NicholasFlamy:
Add change from outline to regular icon in sidebar when page selected to more of the icons (previously only applied to Photos, Sharing, and Favorites). Also change Favorites to single heart consistent with mobile app. (I can change anything if y'all would like.)

### DIFF
--- a/web/src/lib/components/shared-components/side-bar/side-bar.svelte
+++ b/web/src/lib/components/shared-components/side-bar/side-bar.svelte
@@ -11,8 +11,6 @@
     mdiArchiveArrowDownOutline,
     mdiHeart,
     mdiHeartOutline,
-    mdiHeartMultiple,
-    mdiHeartMultipleOutline,
     mdiImageAlbum,
     mdiImageMultiple,
     mdiImageMultipleOutline,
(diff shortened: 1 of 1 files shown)


alextran1502:
Can you post before and after screenshot?

NicholasFlamy:
Yes.
Before:


<br><br>

After:

alextran1502 (approved):

jrasm91:
I think favorites (plural) should be a double icon, simpler to photos. In fact, it was specifically changed to this originally. The mobile app probably just needs to be updated to match it.

NicholasFlamy:
> I think favorites (plural) should be a double icon, similar to photos. In fact, it was specifically changed to this originally. The mobile app probably just needs to be updated to match it.

I was really curious about this. I had a feeling it might have been on purpose. If y'all agree on it I'll make the change on the web and the mobile app. I do think that with the style of the button for Favorites on the mobile app a single heart looks better but for the web it doesn't matter to me as much, though on small screens with the sidebar closed I think it looks better.

PR #9409, opened 2024-05-12
fix(web): prevent navigation from overlapping video controls

### LINKED ISSUE #9174: Web: Comments and video player controls inaccessible due to carousel in the way
### The bug

The carousel controls seem to have higher z-index and are overlapping the video player, making it impossible to enter `FULLSCREEN` or use the kebab menu.
It also overlaps the image ❤️ and comments buttons, preventing users from clicking it.

image

----

image

Ubuntu 24

### Version of Immich Server

v1.103.1

1.102.2 build.151

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
n/a
```

### Your .env content

```Shell
n/a
```

### Reproduction steps

```bash
1. Open web interface
2. Open any video
3. Try switching to full screen
```

### Relevant log output

```shell
n/a
```

### Additional information

_No response_

### LINKED ISSUE #9405: Video playback: Play/pause button overlayed by "previous" button
### The bug

If a video is wide enough that aligns with the left of the viewport the play/pause button becomes unclickable and is overlaid by the "previous" button (chevron left)

unrelated

### Version of Immich Server

v1.103.1

unrelated

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
unrelated
```

### Your .env content

```Shell
unrelated
```

### Reproduction steps

```bash
1. Open video that is wide
2. Clicking pause does either not work (if first item in list) or navigates to the previous item
```

### Relevant log output

_No response_

### Additional information

_No response_

Snowknight26:
Fixes #9405, fixes #9174.

Can't really say I like how the fix is done but there are a couple challenges here:
* ActivityStatus component takes up real estate
  * it's not always shown
* To accommodate for that, video controls have to be shifted up
* Navigation areas have to account for that
* Navigation area icon should stay centered

In my opinion there are several things that could be done differently:
* ActivityStatus component should be implemented differently and not be floating in the bottom right. I'm no designer though so it stays where it is.
  * the video container wouldn't need any bottom margins
* The navigation bars shouldn't have bottom margins
  * but then how do we ensure video controls are accessible?

Then there's also mobile and video aspect ratio/viewport aspect ratio to consider. At the moment this change causes the navigation area to be even smaller (since the bottom margins have been doubled):

image

### DIFF
--- a/web/src/lib/components/asset-viewer/asset-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/asset-viewer.svelte
@@ -600,7 +600,7 @@
     {/if}
 
     {#if $slideshowState === SlideshowState.None && showNavigation}
-      <div class="z-[1001] column-span-1 col-start-1 row-span-1 row-start-2 justify-self-start">
+      <div class="z-[1001] column-span-1 col-start-1 row-span-1 row-start-2 mb-[128px] justify-self-start">
         <NavigationArea onClick={(e) => navigateAsset('previous', e)} label="View previous asset">
           <Icon path={mdiChevronLeft} size="36" ariaHidden />
         </NavigationArea>
@@ -691,7 +691,7 @@
     </div>
 
     {#if $slideshowState === SlideshowState.None && showNavigation}
-      <div class="z-[1001] col-span-1 col-start-4 row-span-1 row-start-2 justify-self-end">
+      <div class="z-[1001] col-span-1 col-start-4 row-span-1 row-start-2 mb-[128px] justify-self-end">
         <NavigationArea onClick={(e) => navigateAsset('next', e)} label="View next asset">
           <Icon path={mdiChevronRight} size="36" ariaHidden />
         </NavigationArea>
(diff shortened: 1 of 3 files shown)


danieldietzler (dismissed):
This is a working fix, but long term this should be refactored as part of a major refactoring. I'd like to hear @alextran1502 thoughts on this though.

zackpollard:
I'm not a big fan of using absolute margins or paddings anywhere, this should be solvable without that and will likely just lead to more problems in the future if we did merge this. I think as you say it probably makes sense to just move the components currently overlapping with the bar to a different area of the screen where they won't overlap.

Thinkscape:
@Snowknight26 Another workaround would be to not make the navigation panes be 100% height. 
If they were i.e. the size of the arrow images + buffer (i.e. 50 px), they would have big enough hit-zone to be useful, but would not overlap any controls.

Here's the node from Google Photos:

image

Thinkscape:
Example with video player in google photos:

image

Snowknight26:
Whoops, deleted the branch by accident. I think that's definitely a better approach. If we're happy with something like that I can attempt to implement something similar on this PR.

Snowknight26:
Closing in favor of https://github.com/immich-app/immich/pull/9455.

PR #9426, opened 2024-05-13
feat(server): run microservices in worker thread

zackpollard:
POC for bringing microservices into the server instance by using Node.JS worker threads.

### DIFF
--- a/server/src/main.ts
+++ b/server/src/main.ts
@@ -4,8 +4,9 @@ import { json } from 'body-parser';
 import cookieParser from 'cookie-parser';
 import { CommandFactory } from 'nest-commander';
 import { existsSync } from 'node:fs';
+import { Worker } from 'node:worker_threads';
 import sirv from 'sirv';
-import { ApiModule, ImmichAdminModule, MicroservicesModule } from 'src/app.module';
+import { ApiModule, ImmichAdminModule } from 'src/app.module';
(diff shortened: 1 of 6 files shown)


server/src/main.ts
```diff
@@ -86,15 +73,28 @@ async function bootstrapImmichAdmin() {
...
+      process.exit(exitCode);
+    }
+  });
+}
+
 function bootstrap() {
   switch (immichApp) {
     case 'immich': {
       process.title = 'immich_server';
+      if (process.env.INTERNAL_MICROSERVICES === 'true') {
```
bo0tzz, line 90: Do we eventually want to have an env var for each worker, or rather something like `IMMICH_ENABLED_WORKERS=api,web,microservices`?

bo0tzz (approved):
🚀

server/src/repositories/logger.repository.ts
```diff
@@ -10,6 +10,12 @@ export class LoggerRepository extends ImmichLogger implements ILoggerRepository
     super(LoggerRepository.name);
   }
 
+  static appName?: string = undefined;
```
danieldietzler, line 13: Shouldn't this be private? Or is that implicit?

danieldietzler (approved):

server/src/main.ts
zackpollard: We can figure this out later, but a pattern like that might make sense

PR #9431, opened 2024-05-13
chore(server): remove unused column

jrasm91:
As far as I can tell this isn't used anywhere, so it was probably from the old search implementation.


bo0tzz (approved):
This was indeed from the old search.

mertalev (approved):

PR #9470, opened 2024-05-14
fix(web): admin settings number input validation

### LINKED ISSUE #9361: Web admin settings page number input broken
### The bug

The number input field value is changed to the minimum when going under the minimum value and changed to the maximum when going over the maximum value. This is intended. The problem with this is that if you try and backspace the value to type in a new one, it'll become 0 and instantly update to the minimum.

What should happen is it waits until you tap outside of the input box (deselecting/defocusing the box) so that you're able to set the value.

Note: Ignore the fact I'm in an unreleased part of the settings, this applies to the other settings with number inputs, including job concurrency, Image Quality, ML Settings, etc.

This screenshot is where I hit backspace on the value and it instantly resets to the minimum:
Screenshot_20240509_125740_Chrome

This is the only workaround, putting your cursor in front or behind the value and typing some digits to make it your value with the old value now added to yours and you can backspace what was the old value:
Screenshot_20240509_125823_Chrome

N/A

### Version of Immich Server

All

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1. Go to administration page.
2. Open a menu like Machine Learning and go to one of the options with a number.
3. Backspace the number and see the minimum keeps appearing when you hit backspace. 
...
```

### Relevant log output

_No response_

### Additional information

I'm using a Samsung Galaxy S21 on Android 14 with Google Chrome. Also tested on Brave on Linux Mint.

michelheusschen:
Fixes #9361 by performing input validation `on:blur` instead of `on:input`

### DIFF
--- a/web/src/lib/components/shared-components/settings/setting-input-field.spec.ts
+++ b/web/src/lib/components/shared-components/settings/setting-input-field.spec.ts
@@ -1,0 +1,29 @@
+import { render } from '@testing-library/svelte';
+import userEvent from '@testing-library/user-event';
+import SettingInputField, { SettingInputFieldType } from './setting-input-field.svelte';
+
+describe('SettingInputField component', () => {
+  it('validates number input on blur', async () => {
+    const { getByRole } = render(SettingInputField, {
+      props: {
+        label: 'test-number-input',
+        inputType: SettingInputFieldType.NUMBER,
+        value: 0,
+        min: 0,
+        max: 100,
+        step: '0.1',
+      },
+    });
+    const user = userEvent.setup();
+
+    const numberInput = getByRole('spinbutton') as HTMLInputElement;
+    expect(numberInput.value).toEqual('0');
+
+    await user.click(numberInput);
+    await user.keyboard('100.1');
+    expect(numberInput.value).toEqual('100.1');
+
+    await user.click(document.body);
+    expect(numberInput.value).toEqual('100');
+  });
+});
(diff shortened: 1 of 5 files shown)


jrasm91 (approved):

web/src/lib/components/shared-components/settings/setting-input-field.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { render } from '@testing-library/svelte';
+import userEvent from '@testing-library/user-event';
+// @ts-expect-error import works
```
zackpollard, line 3: This should be unnecessary, I pulled the branch locally and after an npm install the error went away

zackpollard (changes requested):

web/src/lib/components/shared-components/settings/setting-input-field.spec.ts
zackpollard: Right, I see that now, why that doesn't error in the IDE that also runs typescript against the file, I have no idea...

zackpollard: The annoying part for me is that the ts-expect-error line errors because my IDE doesn't believe there is actually an error there :laughing:

zackpollard (approved):

NicholasFlamy:
Thank you, thank you, thank you! This is a huge relief. It's working great.

web/src/lib/components/shared-components/settings/setting-input-field.spec.ts
jrasm91: I have only seen and run into issues when using `<script lang="ts" context="module">`. Personally, I just prefer to externalize it into a dedicated ts file.

PR #9492, opened 2024-05-14
refactor(server): feature flags

jrasm91:
Use the config directly, instead of the "feature flags", which should just be a dto for use by the clients.


danieldietzler (approved):

NicholasFlamy:
@mertalev you might want to see this and might want to update #8228

mertalev:
Thanks for the heads up!

PR #9499, opened 2024-05-15
feat(web): add archive shortcut to grid

### LINKED ISSUE #9466: timeline view - shortcut Shift-a not working to move assets to archive
### The bug

I can archive selected assets in the main timeline view only by using the 3-dot-menu > archive
But when I use shift-a, then nothing happens.
When I press "?" I get the hint, which shortcuts can be used. Shift-A is listed.
But it looks like, the information, which is displayed when pressing "?" doesn't match the implementation.
I suggest, that the information window content and the implementation should match.

docker in synology containers

### Version of Immich Server

v1.104.0

not relevant

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

```bash
1. open the timeline view
2. select one or more assets
3. press shift-A => nothing happens
4. Use the 3-dot-menu > archive => assets will be archived
```

### Relevant log output

_No response_

### Additional information

_No response_

Snowknight26:
Closes #9466.

Per the keyboard shortcut modal, `Shift`+`A` now works on the grid.


Snowknight26:
Not sure if the keyboard shortcut modal is actually supposed to appear globally since some of the actions don't apply in all contexts (`i` for show/hide info is only applicable when viewing assets), but c'est la vie.

danieldietzler (approved):
Code looks good to me!

jrasm91 (approved):
Seems to work well.

Snowknight26:
Need to merge main in again but I'm wondering.. is this superseded by #9712?

jrasm91:
Probably, but this one works and doesn't have any conflicts. That one has a bunch and other unresolved comments, so I'm just going to merge it.

PR #9517, opened 2024-05-15
refactor(server): system config

jrasm91:
* Move server config implementation to just store a single diff in system-metadata instead of a dedicated table of single key/value overrides
* Make it much easier to add new configuration properties in the future


server/src/cores/system-config.core.ts
jrasm91, line 1: This file

server/src/migrations/1715787369686-RemoveSystemConfigTable.ts
jrasm91, line 1: This file

jrasm91 (commented):
This PR has a lot of noise, but the only meaningful changes are in the system-config.core.ts file and the migration.

danieldietzler (approved):

zackpollard (approved):
Haven't tested this but the changes look good, if you want to merge it go ahead, I will give it an actual test tomorrow regardless :smile:

PR #9525, opened 2024-05-15
refactor: library type

jrasm91:
* Remove option to create upload libraries from the api
* Convert upload libraries into `libraryId: null`
* No longer create an upload library when a new user is created
* Drop `library.type` column

Tested by adding a second user and creating an external library for each.

Before:
image

After:
image


server/src/services/asset-v1.service.ts
```diff
@@ -245,36 +249,16 @@ export class AssetServiceV1 {
...
-    dto: CreateAssetDto & { libraryId: string },
+    dto: CreateAssetDto,
     file: UploadFile,
     livePhotoAssetId?: string,
     sidecarPath?: string,
   ): Promise<AssetEntity> {
     const asset = await this.assetRepository.create({
       ownerId: auth.user.id,
-      libraryId: dto.libraryId,
+      libraryId: null,
```
zackpollard, line 261: This could be omitted here right? This should be possibly undefined at this point as it comes from AssetCreate which picks from AssetEntity

open-api/typescript-sdk/src/fetch-client.ts
```diff
@@ -129,7 +129,7 @@ export type AssetResponseDto = {
     /** This property was deprecated in v1.104.0 */
     isReadOnly?: boolean;
     isTrashed: boolean;
-    libraryId: string;
+    libraryId?: string | null;
```
zackpollard, line 132: Should this be libraryId?, do we want it to be possibly undefined in the response? Surely it should always be string or null, feels like it should remain as libraryId!: but with string | null type

zackpollard (approved):
LGTM overall, couple of minor comments

server/src/migrations/1715804005643-RemoveLibraryType.ts
```diff
@@ -0,0 +1,29 @@
...
+    await queryRunner.query(`DELETE FROM "libraries" WHERE "type" = 'UPLOAD'`);
+    await queryRunner.query(`ALTER TABLE "libraries" DROP COLUMN "type"`);
+    await queryRunner.query(`CREATE INDEX "IDX_originalPath_libraryId" ON "assets" ("originalPath", "libraryId")`);
+    await queryRunner.query(`CREATE UNIQUE INDEX "UQ_assets_owner_checksum" ON "assets" ("ownerId", "checksum") WHERE "libraryId" IS NULL`);
+    await queryRunner.query(`CREATE UNIQUE INDEX "UQ_assets_owner_library_checksum" ON "assets" ("ownerId", "libraryId", "checksum") WHERE "libraryId" IS NOT NULL`);
+    await queryRunner.query(`ALTER TABLE "assets" ADD CONSTRAINT "FK_9977c3c1de01c3d848039a6b90c" FOREIGN KEY ("libraryId") REFERENCES "libraries"("id") ON DELETE CASCADE ON UPDATE CASCADE`);
+  }
+
+  public async down(): Promise<void> {
+    // not implemented
```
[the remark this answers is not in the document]
jrasm91: Yes. Do I need to implement it? I would need to load the users and create a default library for each one and then and then SQL update nulls to refer to the newly created upload libraries and it seems a bit of a pain to do, especially if nobody is ever going to use it.

server/src/entities/asset.entity.ts
```diff
@@ -52,10 +57,10 @@ export class AssetEntity {
   ownerId!: string;
 
   @ManyToOne(() => LibraryEntity, { onDelete: 'CASCADE', onUpdate: 'CASCADE', nullable: false })
```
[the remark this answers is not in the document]
jrasm91: This is an alias for the column, but I think I changed the column directly. It can probably be removed.

server/src/services/asset-v1.service.ts
jrasm91: Yeah it can be removed.

open-api/typescript-sdk/src/fetch-client.ts
jrasm91: This is the auto generated client. It will always match the asset response dto type.

server/src/dtos/asset-response.dto.ts
```diff
@@ -26,7 +26,7 @@ export class AssetResponseDto extends SanitizedAssetResponseDto {
   deviceId!: string;
   ownerId!: string;
   owner?: UserResponseDto;
-  libraryId!: string;
+  libraryId?: string | null;
```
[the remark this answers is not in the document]
jrasm91: I think we don't need this in the response at all, do we? I made it optional so we could remove it in the future.

open-api/typescript-sdk/src/fetch-client.ts
zackpollard: Yea... missed that. Should we deprecate libraryId like we have with isReadOnly so we know when we can remove it in future?

jrasm91: We can do that yeah.

server/src/migrations/1715804005643-RemoveLibraryType.ts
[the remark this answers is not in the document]
jrasm91: Sounds good to me. If it is convenient we can add the down to use during testing, etc. But otherwise I'm fine skipping it.

danieldietzler (approved):

PR #9565, opened 2024-05-17
fix(server): Disable duplicate detection when smart search disabled

NicholasFlamy:
While testing #9540, I noticed that after some changes to #8228, before it was merged, Duplicate Detection stays enabled with Smart Search disabled and Machine Learning enabled and Duplicate Detection enabled.

It was intended that any cobination of Machine Learning is disabled and/or Smart Search is disabled and/or Duplicate Detection is disabled that Duplicate Detection would be disabled. This PR accomplishes this intent.

To test this PR, I borrowed the functioning Duplicate Detection Job UI which is part of the work-in-progress Duplicate Detection UI to confirm that the change in this PR works.

### DIFF
--- a/server/src/utils/misc.ts
+++ b/server/src/utils/misc.ts
@@ -63,7 +63,7 @@ export const isSmartSearchEnabled = (machineLearning: SystemConfig['machineLearn
 export const isFacialRecognitionEnabled = (machineLearning: SystemConfig['machineLearning']) =>
   isMachineLearningEnabled(machineLearning) && machineLearning.facialRecognition.enabled;
 export const isDuplicateDetectionEnabled = (machineLearning: SystemConfig['machineLearning']) =>
-  isMachineLearningEnabled(machineLearning) && machineLearning.duplicateDetection.enabled;
+  isMachineLearningEnabled(machineLearning) && isSmartSearchEnabled(machineLearning) && machineLearning.duplicateDetection.enabled;
 
 export const isConnectionAborted = (error: Error | any) => error.code === 'ECONNABORTED';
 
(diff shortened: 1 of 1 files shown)


server/src/utils/misc.ts
```diff
@@ -63,7 +63,7 @@ export const isSmartSearchEnabled = (machineLearning: SystemConfig['machineLearn
 export const isFacialRecognitionEnabled = (machineLearning: SystemConfig['machineLearning']) =>
   isMachineLearningEnabled(machineLearning) && machineLearning.facialRecognition.enabled;
 export const isDuplicateDetectionEnabled = (machineLearning: SystemConfig['machineLearning']) =>
-  isMachineLearningEnabled(machineLearning) && machineLearning.duplicateDetection.enabled;
+  isMachineLearningEnabled(machineLearning) && isSmartSearchEnabled(machineLearning) && machineLearning.duplicateDetection.enabled;
```
NicholasFlamy, line 66:
```suggestion
  isMachineLearningEnabled(machineLearning) && machineLearning.clip.enabled && machineLearning.duplicateDetection.enabled;
```
This should also accomplish the goal, someone who know about making good code should determine whether my change or this alternative change is better but they should both be identical on functionality.

NicholasFlamy: @mertalev, you might know which is best since you made the Duplicate Detection and know a lot about the AI features of immich.

mertalev: I think you can just change `isMachineLearningEnabled` in the original check to `isSmartSearchEnabled`.

NicholasFlamy: I'll test that and get back to you.

NicholasFlamy: > I think you can just change `isMachineLearningEnabled` in the original check to `isSmartSearchEnabled`.

This works.
```suggestion
  isSmartSearchEnabled(machineLearning) && machineLearning.duplicateDetection.enabled;
```

danieldietzler (approved):

mertalev (approved):

PR #9566, opened 2024-05-17
docs: use `npm` in README

roschaefer:
Motivation
----------
Looks like `npm` is being used, `package-lock.json` is checked in whereas `yarn.lock` is gitignored.

So, let's use `npm` in the README.

How to test
-----------
1. Follow the README.
2. It behaves just like before.


bo0tzz (approved):
I think this is some autogenerated docs that we never really scrutinized. Thank you for the catch 😄

PR #9583, opened 2024-05-18
feat: add docker healthchecks to server and ml

CodaBool:
## 🐋 What is a docker healthcheck
the docker spec has a way to determine if a service is [healthy](https://docs.docker.com/compose/compose-file/05-services/#healthcheck). This is useful for servers. The command `docker ps` can be used to get a quick summary of if a container is responding positively. 

Platforms like AWS have built in integration for this spec and show the health status in their UI when running containers in ECS Fargate.

There probably are other benefits but I only know of those. I personally like to use the `lazydocker` [cli](https://github.com/jesseduffield/lazydocker) to manage my docker services. Which also displays if they are passing healthchecks

lazy

## 🏗️ My solution
One thing to know about the docker healthcheck is it uses CLI on the image to perform it. This can be tricky to accomplish in an era of distroless images. The images that Immich puts out seem to be in that category (which is a good thing). The following common CLI that allow docker healthchecks to easily be done are absent on the image for both `immich_server` and `immich_machine_learning`

> common commands used for Docker healthchecks

- `curl`
- `wget`
- `netcat` or `nc`
- `pgrep`

This means to support docker healthchecks you have to go one of two routes. Add wget from the busybox image. Or use a programming language to script a healthcheck command. 

If the end image has `node` or `python`, like is the case here, then the later is the most slim and secure way to do this.

So, that's what I've added for the two images

###  Lint
I ran `npm run lint`  but got the following error

ignore_lint

Healthchecks do need to follow the Docker spec and [return exactly what Docker wants](https://docs.docker.com/reference/dockerfile/#healthcheck). 

```
0: success - the container is healthy and ready for use
1: unhealthy - the container isn't working correctly
2: reserved - don't use this exit code
```

So, since using `process.exit()` is a requirement. I added this to the ignore patterns of eslint instead of resolving it.

pass

### Something I haven't tested
Healthchecks by default will run every 30 seconds. This in some apps could cause an excess of logs. I did not check for this. If anyone wants to give that concern a look. That would be appreciated.

##  😶‍🌫️Keep Optional
It would be possible to add these to the installation `docker-compose.yml` file. But I don't think it would be good to force this out (by adding it to `/docker/docker-compose.yml`). It can always be something that people optionally add onto their compose file if they want it (it's more for people who know what they are doing anyways). For example postgres and redis have a built in way to do healthchecks. Which I'm adding to my docker compose file.

I guess something in the docs would be useful for this. But curious what the maintainers think on this.

> An example of what my compose file looks like with healthchecks

```yaml
services:
  immich-server:
    ...
    healthcheck:
      test: npm run healthcheck

  immich-microservices:
    ...
      # I don't even know where this code is. So, I didn't add a healthcheck

  machine-learning:
    ...
    healthcheck:
      test: python3 healthcheck.py

  redis:
    ...
    healthcheck:
      test: redis-cli ping

  database:
    ...
    healthcheck:
      test: ["CMD-SHELL", "pg_isready", "-d", "${DB_DATABASE_NAME}"]
```


mmomjian:
I think this is a great idea. Currently I run healthchecks by using a static curl binary mounted inside the container. A built in healthcheck would be great.

server/src/utils/healthcheck.ts
```diff
@@ -0,0 +1,21 @@
+const http = require("http");
+
+const options = {
+  host: "localhost",
```
mertalev, line 4: Would it be better to use the `/api/server-info/ping` endpoint?

machine-learning/app/healthcheck.py
```diff
@@ -0,0 +1,13 @@
+import sys
+import requests
+import os
+
+port = os.getenv("IMMICH_PORT", 3003)
+
+try:
+  response = requests.get(f"http://localhost:{port}")
```
mertalev, line 8: Similarly `/ping` here?

mertalev, line 8: The Node.js health check has a timeout set, but this one doesn't. They should probably be the same.

mertalev (commented):
Thanks for the detailed explanation! I like the idea of using scripts to avoid adding dependencies.

`immich-microservices` doesn't have a separate codebase; it shares the same codebase as the server. It's also just been removed, so no need to worry about it :)

mertalev:
Looks like you need to format the node script through Prettier and the Python script through ruff.

mmomjian:
I have drafted some documentation. According to the [PG docs](https://www.postgresql.org/docs/current/app-pg-isready.html) pg_isready can return values other than 0 or 1. Can we add a `|| exit 1` to that test?

server/src/utils/healthcheck.ts
bo0tzz: Yes, and we should probably also assert on the response body, not just the status code.

server/src/utils/healthcheck.ts
```diff
@@ -0,0 +1,21 @@
+const http = require("http");
+
```
jrasm91, line 2: This should use an import statement

server/src/utils/healthcheck.ts
```diff
@@ -0,0 +1,21 @@
...
+const options = {
+  host: "localhost",
+  port: 3001,
```
jrasm91, line 5: This should use IMMICH_PORT env variable

jrasm91 (changes requested):

server/.eslintrc.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
   env: {
     node: true,
   },
-  ignorePatterns: ['.eslintrc.js'],
+  ignorePatterns: ['.eslintrc.js', 'healthcheck.ts'],
```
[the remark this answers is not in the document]
jrasm91: You can ignore the process exit rule per line or per file instead.

server/src/utils/healthcheck.ts
CodaBool: ### Expected value
I first tested with a wget call for a successful call to `/api/server-info/ping` gave. It responded with this:

```
2024-05-19 20:47:35 URL: http://localhost:3001/api/server-info/ping 200 OK
```
Next I attempted to read that data inside the healthcheck.ts file


### What I tried
I'm having some trouble using `node.js` `http.request()` function for the purpose of getting the response body. Looking at their [docs](https://nodejs.org/api/http.html#httprequesturl-options-callback). It seems something like this should work. When placed inside the callback function of `http.request()`:

```javascript
res.on('data', (chunk) => {
  console.log(`BODY: ${chunk}`);
});
```

but this never gets triggered. Not sure what I'm doing wrong here. If anyone is familiar with http.request(), I need some help on reading the body. :bow: 

### Alternative
I didn't see any http requesting packages. So, I would like to make http.request() work here.

This only other useful piece of data I saw in the response was `statusMessage`. Which has the value of "OK". This seems to be like it could be used as a replacement for the body. Any thoughts on this idea?

CodaBool:
Thanks for the help everyone. I have implemented your suggestions. With the exception of @bo0tzz. I was not able to find a way to read the response body with node.js http.request() callback. I left a reply to your comment with more details

server/src/utils/healthcheck.ts
jrasm91: Just use fetch instead

CodaBool: oh nice fetch does work. Seems I have to do a weird `signal` thing to implement a timeout :face_exhaling: 

But I was able to read the body this way.

mertalev:
I believe you can fix the `process.exit` linting error by adding `#!/usr/bin/env node` to the top of the file.

mertalev:
May as well add it to the Python script too for consistency (except `python3` instead of `node`).

jrasm91 (approved):

mertalev (approved):

bo0tzz (dismissed):
Can you rebase on main to fix the status checks?

I'm also thinking we should probably document the existence of these healthchecks somewhere, though I'm not sure where would be a good place at the moment 🤔

zackpollard:
> Can you rebase on main to fix the status checks?
> 
> I'm also thinking we should probably document the existence of these healthchecks somewhere, though I'm not sure where would be a good place at the moment 🤔

Dockerfiles have a HEALTHCHECK instruction, we should add these endpoints into there so that by default everyone gets healthchecks. @CodaBool

jrasm91:
@zackpollard @bo0tzz #9590

mmomjian:
@zackpollard I agree with adding it to the `Dockerfile`. That won't cover Redis or Postgres. One option (my preference) is to add it to the docker-compose.yml for those two services (non-breaking). Thoughts?

mmomjian:
I have added an enhanced PG healthcheck in #9590 that uses database checksums

CodaBool:
> One option is to add it to the docker-compose.yml

I'm happy either way

zackpollard:
> I have added an enhanced PG healthcheck in #9590 that uses database checksums. My preference would still be to modify that PR into one that just adds the checks to the docker-compose.yml

Happy to add that into the docker compose and have our own ones directly in the dockerfiles

mmomjian:
> > I have added an enhanced PG healthcheck in #9590 that uses database checksums. My preference would still be to modify that PR into one that just adds the checks to the docker-compose.yml
> 
> Happy to add that into the docker compose and have our own ones directly in the dockerfiles

#9590 updated.

server/src/utils/healthcheck.ts
```diff
@@ -0,0 +1,31 @@
...
+  const timeout = setTimeout(() => controller.abort(), 2000);
+  try {
+    const response = await fetch(`http://localhost:${port}/api/server-info/ping`, {
+      signal: controller.signal,
+    });
+
+    if (response.ok) {
+      const body = await response.json();
+      if (body.res === 'pong') {
+        console.log('Server is running');
```
mmomjian, line 15: Will this write to the docker logs every time the HC is run (default 30 seconds)? Is that desired?

CodaBool: Doing a watch on the `immich_server` container for 2 minutes and nothing showed up (healthchecks should by default happen every 30 seconds). There is no other logs than the container's stdout right?

mmomjian: Hm, interesting. I guess that should be fine then

zackpollard: Imo just remove it anyway, if it doesn't get written anywhere then there's no point having it

CodaBool: removed 2 of the 3 console calls. I left the last one in because there is a very minor usefulness to them. If healthchecks are failing when they shouldn't be you could exec in and run the file with node. Then look at the output for debugging.

Let me know what you think. I'm good either way.

zackpollard (approved):

server/src/utils/healthcheck.ts
zackpollard: Looks good to me, thanks for making all these changes!

bo0tzz: It's not really important here, but for future reference, healthcheck logs are captured separately by docker and listed in the container status next to healthcheck status changes.

PR #9391, opened 2024-05-11
docs: add info about database checksums

mmomjian:
Stage for next release (1.104.0 ?)

PR #9590, opened 2024-05-19
feat(server): healthchecks for PG and redis

mmomjian:
implements #9583 in non-Immich services

this is a non breaking change. Users who add this will get the benefit of the healthcheck, if they have checksums enabled (new installs starting with 1.104.0)

Users with older setups will have a `unhealthy` status. This can be fixed by removing everything after the first `exit 1` in the database health check test: line


zackpollard:
In #9583 I am requesting that we make the healthchecks active by default using the HEALTHCHECK command in the dockerfiles, so we shouldn't need the documentation around the docker compose changes. If that change makes it in the docs will need to be updated to reflect that

docs/docs/FAQ.mdx
```diff
@@ -360,6 +360,33 @@ cap_drop:
...
+
+  redis:
+    ...
++   healthcheck:
++     test: redis-cli ping
+
+  database:
+    ...
++   healthcheck:
++     test: ["CMD-SHELL", "pg_isready", "--dbname=${DB_DATABASE_NAME}"]
```
aviv926, line 387: Out of curiosity, does it perform the database checksums as you added in #9391?

mmomjian: Good idea! I just added this.

docker/docker-compose.prod.yml
```diff
@@ -55,7 +57,13 @@ services:
       - ${UPLOAD_LOCATION}/postgres:/var/lib/postgresql/data
     ports:
       - 5432:5432
+    healthcheck:
+      test: pg_isready --dbname='${DB_DATABASE_NAME}' || exit 1; Chksum="$$(psql --dbname='${DB_DATABASE_NAME}' --username='${DB_USERNAME}' --tuples-only --no-align --command='SELECT SUM(checksum_failures) FROM pg_stat_database')"; echo "checksum failure count is $$Chksum"; [ "$$Chksum" = '0' ] || exit 1
+      interval: 5m
+      start_interval: 30s
+      start_period: 5m
     command: ["postgres", "-c" ,"shared_preload_libraries=vectors.so", "-c", 'search_path="$$user", public, vectors', "-c", "logging_collector=on", "-c", "max_wal_size=2GB", "-c", "shared_buffers=512MB", "-c", "wal_compression=on"]
+    restart: always
```
mmomjian, line 66: Is there any reason the database was the only core service not `restart: always` in prod?

Same question for grafana and prometheus, though I did not make a change there.

bo0tzz (approved):
LGTM. I do worry a bit about the default compose file growing large again, but I don't think there's any other way for us to apply these healthchecks right?

zackpollard:
> LGTM. I do worry a bit about the default compose file growing large again, but I don't think there's any other way for us to apply these healthchecks right? 

Yea, this is the only way without pushing our own containers

zackpollard (approved):

PR #9604, opened 2024-05-19
docs: community docs typos

mmomjian:


docs/src/components/community-guides.tsx
```diff
@@ -10,17 +10,17 @@ interface CommunityGuidesProps {
 const guides: CommunityGuidesProps[] = [
   {
     title: 'Cloudflare Tunnels with SSO/OAuth',
-    description: `Setting up Cloudflare Tunnels and a SaaS App for immich.`,
+    description: `Setting up Cloudflare Tunnels and a SaaS App for Immich.`,
```
danieldietzler, line 13: Actually, afaik it isn't even super clear if it should be Immich or immich lol. I think Alex mentioned once that immich was the intention but in most places we have Immich 😅

danieldietzler (approved):

docs/src/components/community-guides.tsx
mmomjian: Hm. Last time I asked I think he said to use Immich in the docs and/or he didn’t care but just be consistent haha.  @alextran1502 ?

jrasm91: Capital everywhere except logos is fine imo

jrasm91 (approved):

PR #9615, opened 2024-05-20
refactor(server): version logic

jrasm91:
Use semver library for version compatibility checking and parsing.


zackpollard (approved):

server/src/services/database.service.spec.ts
```diff
@@ -13,211 +12,170 @@ describe(DatabaseService.name, () => {
...
   });
 
   it('should work', () => {
     expect(sut).toBeDefined();
   });
 
-  describe.each([
-    [{ vectorExt: DatabaseExtension.VECTORS, extName: 'pgvecto.rs', minVersion: new Version(0, 1, 1) }],
-    [{ vectorExt: DatabaseExtension.VECTOR, extName: 'pgvector', minVersion: new Version(0, 5, 0) }],
-  ] as const)('init', ({ vectorExt, extName, minVersion }) => {
```
mertalev, line 28: The tests should still test everything with pgvector and pgvecto.rs. There's only one test for pgvector now from what I can see.

server/src/services/database.service.ts
```diff
@@ -1,163 +1,134 @@
...
+    if (!current || !semver.satisfies(current, POSTGRES_VERSION_RANGE)) {
+      throw new Error(
+        `Invalid PostgreSQL version. Found ${version}, but needed ${POSTGRES_VERSION_RANGE}. Please use a supported version.`,
+      );
+    }
+
     await this.databaseRepository.withLock(DatabaseLock.Migrations, async () => {
-      await this.createVectorExtension();
-      await this.updateVectorExtension();
-      await this.assertVectorExtension();
```
mertalev, line 35: Why did you inline all of this? It looks really messy this way.

mertalev (changes requested):
This is a good (and necessary) change, but I think the reduced test coverage and messy database service `init` should be resolved.

server/src/services/database.service.ts
jrasm91: I removed all the logging and consolidated the shared variables, so it looks a lot better now. Can you take another look?

server/src/services/database.service.spec.ts
jrasm91: This is a unit test not a e2e test. Can you tell me what code path in the database service is not tested anymore?

server/src/services/database.service.ts
```diff
@@ -1,38 +1,126 @@
...
+        Otherwise, please update the version of ${name} in the Postgres instance to a compatible version.`,
+  createFailed: ({ name, extension, otherName }: CreateFailedArgs) => `
+        Failed to activate ${name} extension.
+        Please ensure the Postgres instance has ${name} installed.
+
+        If the Postgres instance already has ${name} installed, Immich may not have the necessary permissions to activate it.
+        In this case, please run 'CREATE EXTENSION IF NOT EXISTS ${extension}' manually as a superuser.
+        See https://immich.app/docs/guides/database-queries for how to query the database.
+
+        Alternatively, if your Postgres instance has ${name}, you may use this instead by setting the environment variable 'DB_VECTOR_EXTENSION=${otherName}'.
```
mertalev, line 48: `name` should be `otherName` here.

jrasm91: Ah, good catch. Thank you.

danieldietzler (approved):
LGTM!

server/src/services/database.service.spec.ts
mertalev: The checks for updating, enforcing version constraints and displaying the right extension in logs aren’t being tested. The tests would still pass if we only handled them for pgvecto.rs.

server/src/services/database.service.ts
mertalev: Much better now, thanks.

server/src/services/database.service.spec.ts
jrasm91: I can add tests to verify the extension in the log message and some to verify the semver ranges work as expected

jrasm91: Alright, I've updated the tests to check semver and log messages for the extension name.

mertalev: Sweet, thanks!

mertalev (approved):

PR #9645, opened 2024-05-21
refactor: deprecate /server-info and replace with /server-info/storage

zackpollard:
/server-info root endpoint was used for storage info only, moving this to a more sensible place


open-api/immich-openapi-specs.json
```diff
@@ -4337,14 +4337,16 @@
     },
     "/server-info": {
       "get": {
+        "deprecated": true,
+        "description": "This property was deprecated in v1.106.0",
```
jrasm91, line 4341: oh this might be a little bug that it should say endpoint instead of property. looks like it uses the same description for both.

jrasm91 (approved):

danieldietzler (approved):
Except for the tests this looks good ;)

zackpollard:
> Except for the tests this looks good ;)

I'll fix those in the morning, brain was mush by the time this was done 🤣

open-api/immich-openapi-specs.json
danieldietzler: Yup, fixing this is actually not super-trivial. I think I'd prefer a neutral name that works for both properties and endpoints

PR #9649, opened 2024-05-21
feat: support tls protected postgres

bzp2010:
## Background
Currently, immich does not support connections to postgres databases protected by TLS or mTLS, which makes it impossible to connect to database instances that enforce TLS.

I looked at the documentation for the postgres driver in nodejs, and it mentions that using ?sslmode=require can require the driver to connect to the database using TLS, but after testing, it prevents node from establishing a trusted connection due to the database using self-signed CA certificates, although it is possible to do so with `NODE_TLS_REJECT_UNAUTHORIZED` environment variable to force TLS authentication to be turned off, but this actually invalidates TLS.

## News
The PR adds new configuration entries and reads from a new set of environment variables and passes them to the driver. This allows the user to manually specify a CA certificate to be used to validate the server certificate and establish a truly trusted connection. And, it also supports the mTLS use case in passing, which I don't use, but does have potential usability.

I added the matrix to the E2E test, which now connects to the database using unencrypted and TLS/mTLS respectively, and it passes all of the tests without modifying any of the test cases, so the modification didn't break any existing functionality.


mmomjian:
I assume you have already tried `?sslmode=require&sslmode=no-verify` [link](https://immich.app/docs/administration/postgres-standalone/) and you prefer to have an implementation with a trusted cert?

bzp2010:
> I assume you have already tried `?sslmode=require&sslmode=no-verify` [link](https://immich.app/docs/administration/postgres-standalone/) and you prefer to have an implementation with a trusted cert?

Hi, @mmomjian. Thanks for your quick reply.

I really didn't notice this usage mentioned in the documentation for this project, I mainly referred to this document [https://node-postgres.com/features/ssl](https://node-postgres.com/features/ssl) and there was nothing about it, which was an oversight on my part.

I recognize that there are several ways to do this if one is simply preventing node.js from validating server certificates, including, but not limited to, the environment variables mentioned above and the ones you point to in the documentation.

But the new configuration introduced by this PR helps to make sense of TLS, rather than just ignoring any problems the server certificate might have in order to establish a connection.

mmomjian:
Makes sense. I agree with you that ideally, we should provide authentication (which this PR adds) in addition to just encryption (which is what we currently have).

bzp2010:
Agree with you, for PostgreSQL authentication, the user&password-based one and the mTLS-based one are decoupled, and we can use both or only one of them. This PR completes the rest.

| Password | TLS/mTLS | Description |
|-----------|-------|---------------|
| Yes           | No    | Safe, without encryption |
| No           | Yes (with mTLS)    | Safe, encrypted |
| Yes          | Yes     | Safest |

bo0tzz:
For my understanding: What exactly is this adding that cannot already be done through the URL parameters?

mmomjian:
> For my understanding: What exactly is this adding that cannot already be done through the URL parameters?

The current implementation is equivalent to using HTTPS, but when the page loads and you get a certificate mismatch you say "connect anyway". The data is still encrypted, but you have no idea if you are talking to the server that you expected to reach.

bzp2010:
Hi, @bo0tzz.

> For my understanding: What exactly is this adding that cannot already be done through the URL parameters?

According to the [documentation](https://node-postgres.com/features/ssl), certificates themselves can't be implemented using a URL parameters in a DSN, and although the driver itself supports some [environment variables](https://www.postgresql.org/docs/9.1/libpq-envars.html), it doesn't quite work the same way (nest.js's) as our existing configuration management.

jrasm91:
You can programatically pass options to the driver through the connection string still.

https://github.com/brianc/node-postgres/tree/master/packages/pg-connection-string#tcp-connections

bzp2010:
Hi, @jrasm91.

Indeed, I ignored the documentation in that separate package, and the documentation dedicated to describing that section is not associated here.

This applies to the mode that uses DSNs, but not the one that uses separate environment variables, and only supports passing in a file path. It seems to work enough, but not very well.

So if you think it is possible to not support or pass in some kind of anti-pattern through these [https://www.postgresql.org/docs/9.1/libpq-envars.html-specified](https://www.postgresql.org/docs/9.1/libpq-envars.html-specified) environment variables, then that PR is not needed and I would close it. :)

jrasm91:
If anything, I am for removing all the DB_* environment variables and only using DB_URL. I think if there is a way to pass the same information via the url as individual environment variables then we don't need to copy/duplicate them. We'd essentially be copying what the internal connection string library parser is already doing.

We could definitely update the documentation to make it more clear how to use the connection string to connect to different ssl configurations though.

jrasm91:
I am going to go ahead and close this, as I believe all the same options can be passed in using a connection string instead. Any help with adding some documentation in this area would be appreciated though.

PR #9659, opened 2024-05-22
fix(mobile): Missing space in norwegian i18n

dagstuan:


zackpollard:
Hey, we have a translation platform called Localizely that we use to manage translations, I have made this change over there for you but if you plan to make more contributions to the translations we can get you setup there with an account :slightly_smiling_face:

PR #9663, opened 2024-05-22
feat(web): s

jrasm91:


danieldietzler (approved):
LMAO I just mention that I am slowly leaning towards `s()` and here we go :P

PR #9667, opened 2024-05-22
refactor(server): plural endpoints

jrasm91:
Change endpoints to use plural version instead. Will need to do asset in a follow up PR. It is a bit more involved as a lot of the file serving and also upload paths are manually generated in the clients.
 
| Before | After |
| - | - |
| `/activity` | `/activities` |
| `/album` | `/albums` |
| `/api-key` | `/api-keys` |
| `/face` | `/faces` |
| `/report` | `/reports` |
| `/library` | `/libraries` |
| `/partner` | `/partners` |
| `/person` | `/people` |
| `/shared-link` | `/shared-links` |
| `/tag` | `/tags` |
| `/user` | `/users` |


zackpollard (approved):
Seems fine, providing the app only varies for assets then it's probably fine, might be worth us testing it before we merge this

mmomjian:
Looks like the PR description is missing the s for partners. :)

jrasm91:
> Seems fine, providing the app only varies for assets then it's probably fine, might be worth us testing it before we merge this

Sounds good. We can do testing now, or a full mobile regression test run after all the breaking api changes have been merged into main.

PR #9670, opened 2024-05-22
chore(server): remove unused code

jrasm91:


danieldietzler (approved):
I *love* those "remove unused code" PRs. Almost feels better than adding new code :D

PR #9684, opened 2024-05-23
feat: API operation replaceAsset, POST /api/asset/:id/file 

midzelis:
For history, see PR #8480. 

This PR was split from above, it contains only new API (no breaking changes) for the implementation of replaceAsset, the rest endpoint POST /api/asset/:id/file, and tests.


server/src/dtos/asset-media-response.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export class DefaultAssetMediaResponseDto {
+  @ApiProperty({
+    type: String,
+    enum: AssetMediaStatusEnum,
+    required: true,
+    enumName: 'AssetMediaStatus',
+  })
+  readonly status?: string;
+  @IsOptional()
```
jrasm91, line 17: You need to use `Optional` instead of `IsOptional` because the latter has some unintended behavior.

server/src/dtos/asset-media-response.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export class DefaultAssetMediaResponseDto {
+  @ApiProperty({
+    type: String,
+    enum: AssetMediaStatusEnum,
+    required: true,
+    enumName: 'AssetMediaStatus',
+  })
+  readonly status?: string;
```
jrasm91, line 16: This should be AssetMediaStatusEnum (without `readonly`)

server/src/dtos/asset-media-response.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+    this.copyId = copyId;
+  }
+}
+export class DuplicateAssetResponseDto {
+  readonly status = 'duplicate';
+  duplicateId: string;
+  constructor(duplicateId: string) {
+    this.duplicateId = duplicateId;
+  }
+}
```
jrasm91, line 40: Generally, we don't actually instantiate instances of these dtos, they act purely as types. The only reason they have to be classes is because they need to exist at runtime for the swagger plugin to work correctly. I would prefer to just send { status: 'duplicate', duplicateId: assetId } instead of creating an instance and having code for the constructor.

server/src/dtos/asset-media-response.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+import { IsOptional } from 'class-validator';
+
+export enum AssetMediaStatusEnum {
+  UPDATED = 'updated',
+  DUPLICATE = 'duplicate',
+}
+
+export class DefaultAssetMediaResponseDto {
+  @ApiProperty({
+    type: String,
```
jrasm91, line 11: Remove this

server/src/dtos/asset-media-response.dto.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export class AssetMediaUpdateResponseDto {
+  readonly status = 'updated';
+  assetId: string;
+  copyId: string;
+  constructor(assetId: string, copyId: string) {
+    this.assetId = assetId;
+    this.copyId = copyId;
+  }
+}
```
jrasm91, line 33: Let's simplify this to just status, id.

server/src/services/asset-media.service.ts
```diff
@@ -0,0 +1,193 @@
...
+}
+
+export interface UploadFile {
+  uuid: string;
+  checksum: Buffer;
+  originalPath: string;
+  originalName: string;
+  size: number;
+}
+
```
jrasm91, line 45: These look to be copied twice.

server/src/services/asset-media.service.ts
```diff
@@ -0,0 +1,193 @@
...
+    await this.assetRepository.upsertExif({ assetId, fileSizeInByte: file.size });
+    await this.jobRepository.queue({
+      name: JobName.METADATA_EXTRACTION,
+      data: { id: assetId, source: 'upload' },
+    });
+  }
+
+  /**
+   * Create a 'shallow' copy of the specified asset record creating a new asset record in the database.
+   * Uses only vital properties likxcluding things like: stacks, faces, smart search info, etc,
```
jrasm91, line 162: typo `likxcluding`

web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
```diff
@@ -243,6 +247,13 @@
                 icon={asset.isArchived ? mdiArchiveArrowUpOutline : mdiArchiveArrowDownOutline}
                 text={asset.isArchived ? 'Unarchive' : 'Archive'}
               />
+              {#if showDevUI}
+                <MenuOption
+                  icon={mdiUpload}
+                  on:click={() => openFileUploadDialog({ multiple: false, assetId: asset.id })}
+                  text="Replace with upload"
+                />
+              {/if}
```
jrasm91, line 256: I think we can probably remove this, but maybe it is useful for testing it in the short term.

web/src/lib/utils.ts
```diff
@@ -158,23 +159,33 @@ const createUrl = (path: string, parameters?: Record<string, unknown>) => {
...
 
 export const getProfileImageUrl = (...[userId]: [string]) => {
-  const path = `/users/${userId}/profile-image`;
+  const path = `/user/profile-image/${userId}`;
   return createUrl(path);
 };
 
 export const getPeopleThumbnailUrl = (personId: string) => {
-  const path = `/people/${personId}/thumbnail`;
+  const path = `/person/${personId}/thumbnail`;
```
jrasm91, line 188: these changes need to get reverted, probably a bad merge.

jrasm91 (approved):

PR #9689, opened 2024-05-23
feat(server): fully accelerated qsv

mertalev:
### Descripion

This PR adds end-to-end acceleration for Quick Sync transcoding. Similar to NVENC and RKMPP, tone-mapping is done with OpenCL. Based on my testing, OpenCL was even faster than native VPP tone-mapping while having more features and higher quality.

### Testing

Tested with hardware decoding enabled and disabled, with tone-mapping enabled and disabled, and scaling set to either 720p or original.

Note: My processor is too new for the `intel-compute-runtime` in the server image, so I had to manually install the latest release [here](https://github.com/intel/compute-runtime/releases/tag/24.17.29377.6). I'll make a PR to apply this in the base image, but in the meantime it should only affect very new processors and only when hardware decoding is enabled.

Current transcoding time: 470s
With hardware decoding: 34s

### DIFF
--- a/server/src/utils/media.ts
+++ b/server/src/utils/media.ts
@@ -652,7 +652,7 @@ export class QsvHwDecodeConfig extends QsvSwDecodeConfig {
     const options = ['-hwaccel qsv', '-async_depth 4', '-threads 1'];
     const hwDevice = this.getPreferredHardwareDevice();
     if (hwDevice) {
-      options.push(`-qsv_device ${hwDevice}`);
+      options.push(`-qsv_device /dev/dri/${hwDevice}`);
     }
 
     return options;
@@ -661,7 +661,7 @@ export class QsvHwDecodeConfig extends QsvSwDecodeConfig {
   getFilterOptions(videoStream: VideoStreamInfo) {
     const options = [];
     if (this.shouldScale(videoStream) || !this.shouldToneMap(videoStream)) {
-      let scaling = `scale_qsv=${this.getScaling(videoStream)}:mode=hq:async_depth=4`;
+      let scaling = `scale_qsv=${this.getScaling(videoStream)}:async_depth=4:mode=hq`;
       if (!this.shouldToneMap(videoStream)) {
         scaling += ':format=nv12';
       }
--- a/server/src/utils/media.ts
+++ b/server/src/utils/media.ts
@@ -703,7 +703,7 @@ export class VAAPIConfig extends BaseHWConfig {
     }
 
     let hwDevice = this.getPreferredHardwareDevice();
-    if (hwDevice === null) {
+    if (!hwDevice) {
       hwDevice = `/dev/dri/${this.devices[0]}`;
     }
 
--- a/server/src/utils/media.ts
+++ b/server/src/utils/media.ts
@@ -313,7 +313,7 @@ export class BaseHWConfig extends BaseConfig implements VideoCodecHWConfig {
       throw new Error(`Device '${device}' does not exist`);
     }
 
-    return device;
+    return `/dev/dri/${deviceName}`;
   }
 }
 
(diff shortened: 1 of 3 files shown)


server/src/utils/media.ts
```diff
@@ -643,14 +643,67 @@ export class QSVConfig extends BaseHWConfig {
...
+
 export class VAAPIConfig extends BaseHWConfig {
   getBaseInputOptions() {
     if (this.devices.length === 0) {
       throw new Error('No VAAPI device found');
     }
 
     let hwDevice = this.getPreferredHardwareDevice();
-    if (hwDevice === null) {
+    if (!hwDevice) {
```
jrasm91, line 706: :pray:

jrasm91 (approved):

server/src/utils/media.ts
```diff
@@ -643,14 +643,67 @@ export class QSVConfig extends BaseHWConfig {
...
+      `primaries=${colors.primaries}`,
+      'range=pc',
+      `tonemap=${this.config.tonemap}`,
+      `transfer=${colors.transfer}`,
+    ];
+
+    return [
+      'hwmap=derive_device=opencl',
+      `tonemap_opencl=${tonemapOptions.join(':')}`,
+      'hwmap=derive_device=vaapi:reverse=1',
```
nyanmisaka, line 694: `hwmap=derive_device=vaapi:reverse=1`
This will be not allowed in the future. It should be `hwmap=derive_device=qsv:reverse=1`
The qsv encoder's ability to accept vaapi fmt is experimental and was decided by Intel not to be promoted to upstream.

mertalev: I used `vaapi` there because `qsv` throws an error:
```
[AVHWDeviceContext @ 0x24e400304c0] 0.0: Intel(R) OpenCL Graphics / Intel(R) Arc(TM) Graphics
[AVHWDeviceContext @ 0x24e400304c0] Intel QSV to OpenCL mapping function found (clCreateFromVA_APIMediaSurfaceINTEL).
[AVHWDeviceContext @ 0x24e400304c0] Intel QSV in OpenCL acquire function found (clEnqueueAcquireVA_APIMediaSurfacesINTEL).
[AVHWDeviceContext @ 0x24e400304c0] Intel QSV in OpenCL release function found (clEnqueueReleaseVA_APIMediaSurfacesINTEL).
[AVHWFramesContext @ 0x24e401d4740] The hardware pixel format 'vaapi' is not supported by the device type 'QSV'
[Parsed_hwmap_2 @ 0x24e40173000] Failed to initialise target frames context: -38.
[Parsed_hwmap_2 @ 0x24e40173000] Failed to configure output pad on Parsed_hwmap_2
Error reinitializing filters!
Failed to inject frame into filter network: Function not implemented
Error while processing the decoded data for stream #0:0
[AVIOContext @ 0x24e401907c0] Statistics: 0 bytes written, 0 seeks, 0 writeouts
Terminating demuxer thread 0
[AVIOContext @ 0x24e401902c0] Statistics: 1982895 bytes read, 2 seeks
Conversion failed!
```
Is there something else I would need to change to make this work? The full command is this:
```
ffmpeg -hwaccel qsv -async_depth 4 -threads 1 -i HDR.mp4 \
    -c:v h264_qsv -c:a copy -movflags faststart -fps_mode passthrough -map 0:0 -bf 7 -refs 5 -g 256 -v verbose \
    -vf hwmap=derive_device=opencl,tonemap_opencl=desat=0:format=nv12:matrix=bt709:primaries=bt709:range=pc:tonemap=hable:transfer=bt709,hwmap=derive_device=qsv:reverse=1 \
    -preset 7 -global_quality 23 SDR.mp4
```

nyanmisaka: I think you have to specify the `format=qsv` to enable the indirect hwmap.

`hwmap=derive_device=qsv:reverse=1,format=qsv`

mertalev: Thanks, that worked! I'll change it to use `qsv`.

PR #9699, opened 2024-05-23
fix(mobile): use correct Focus Node for latitude and longitude

foux:
Today on the mobile app, we can't change the location of pictures due to a bug in setting latitude and longitude : 
CleanShot 2024-05-23 at 16 35 52

This PR fixes that bug by assigning the correct focus nodes to latitude and longitude Picker Input


bo0tzz (approved):

zackpollard (approved):
Lol nice, thanks for the fix

PR #9700, opened 2024-05-23
fix(web): hide detail panel for shared links with hidden metadata

michelheusschen:
The detail panel for shared links can still be opened with the shortcut `i` when created with `showMetadata: false`. Fixed by always hiding the detail panel when metadata is hidden.


jrasm91 (approved):
Thanks for the tests!

PR #9704, opened 2024-05-23
fix(web): show api key copy button in Firefox

lukashass:
Since all other text copy buttons do not have a check either, I removed this.

`writeText()` compatibility: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText#browser_compatibility

An alternative would be to check `navigator.clipboard.writeText` for all buttons, but since there is a good error message that encourages people to use https, I don't think there is a need to hide the buttons.


bo0tzz (approved):

jrasm91 (approved):
Noice. Why in the world were we requiring "copy image to clipboard" for copy text lol.

PR #9705, opened 2024-05-23
fix(web): allow copying text in photo viewer

### LINKED ISSUE #8583: Copy keyboard shortcuts do not work on Web
### The bug

In both Firefox and Chrome, the "Ctrl+C" keyboard shortcut does not function to copy text. You can still copy by right clicking and selecting Copy. Having the Ctrl C shortcut is useful when working with filenames and GPS coordinates. Immich web UI should not preempt this keyboard combination, which is what I believe is currently happening, and can lead to user confusion.

Debian 12

### Version of Immich Server

v1.100.0

v1.100.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
N/A
```

### Additional information

_No response_

lukashass:
Closes #8583

### DIFF
--- a/web/src/lib/components/asset-viewer/photo-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/photo-viewer.svelte
@@ -3,7 +3,7 @@
   import { boundingBoxesArray } from '$lib/stores/people.store';
   import { alwaysLoadOriginalFile } from '$lib/stores/preferences.store';
   import { photoZoomState } from '$lib/stores/zoom-image.store';
-  import { copyToClipboard, downloadRequest, getAssetFileUrl, handlePromiseError } from '$lib/utils';
+  import { downloadRequest, getAssetFileUrl, handlePromiseError } from '$lib/utils';
   import { isWebCompatibleImage } from '$lib/utils/asset-utils';
   import { getBoundingBox } from '$lib/utils/people-utils';
   import { shortcuts } from '$lib/actions/shortcut';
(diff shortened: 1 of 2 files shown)


web/src/lib/components/asset-viewer/photo-viewer.svelte
```diff
@@ -130,8 +130,12 @@
     }
   });
 
-  const onCopyShortcut = () => {
-    if (window.getSelection()?.type === 'Range') {
+  const onCopyShortcut = async () => {
+    const selection = window.getSelection();
+    if (selection?.type === 'Range') {
+      if (selection.toString().length > 0) {
+        await copyToClipboard(selection.toString(), true);
```
bo0tzz, line 137: Alternatively we can set `preventDefault = false` on the shortcut ([see here](https://github.com/immich-app/immich/blob/57d94bce6865d222ed56078224076a682017869e/web/src/lib/actions/shortcut.ts#L63-L65)), just return from this `if type == range`, and then call preventDefault after the if block.

bo0tzz: (thus letting the browser just go to the default copy behaviour)

lukashass: That is indeed better, didn't realize that `preventDefault` was true by default.

bo0tzz (approved):
Lovely, thank you!

jrasm91 (commented):
Should this not be something that is implemented in the use:shortcuts implementation? Not anywhere/everywhere we might use a shortcut that has a default behavior? IMO all custom short cuts should be ignored if there is a current range selection, not just this one.

mmomjian:
Thank you!

zackpollard (approved):

PR #9715, opened 2024-05-24
refactor: Move checkExistingAssets(), checkBulkUpdate() remove getAllAssets()

midzelis:
This was split from PR https://github.com/immich-app/immich/pull/8480.

This is relocating a few controller.v1 methods into the asset-media controller: 
```
POST /asset/exist (checkExistingAssets)
POST /asset/bulk-upload-check (checkBulkUpload)
```

This one is deleted, as its unused.
```
GET /asset (getAllAssets)
```


server/src/controllers/asset-media.controller.ts
```diff
@@ -15,13 +17,19 @@ import { Response } from 'express';
...
+  AssetBulkUploadCheckDto,
+  CheckExistingAssetsDto
+} from 'src/dtos/asset-media.dto';
 import { AuthDto } from 'src/dtos/auth.dto';
 import { ILoggerRepository } from 'src/interfaces/logger.interface';
 import { Auth, Authenticated } from 'src/middleware/auth.guard';
 import { FileUploadInterceptor, Route, UploadFiles, getFiles } from 'src/middleware/file-upload.interceptor';
 import { AssetMediaService } from 'src/services/asset-media.service';
 import { FileNotEmptyValidator, UUIDParamDto } from 'src/validation';
-
```
danieldietzler, line 24: I don't think removing this newline was intentional

server/src/services/asset.service.ts
```diff
@@ -224,6 +228,13 @@ export class AssetService {
     return this.assetRepository.getAllByDeviceId(auth.user.id, deviceId);
   }
 
+  public async getAllAssets(auth: AuthDto, dto: AssetSearchDto): Promise<AssetResponseDto[]> {
+    const userId = dto.userId || auth.user.id;
+    await this.access.requirePermission(auth, Permission.TIMELINE_READ, userId);
+    const assets = await this.assetRepositoryV1.getAllByUserId(userId, dto);
```
danieldietzler, line 234: I am not a big fan of having asset repo v1 stuff in here tbh.

danieldietzler (approved):

server/src/services/asset.service.ts
jrasm91: The mobile app just stopped using it, yeah. I think it is still being used in production today though

danieldietzler: Then it is fine removing it with this release ;)

jrasm91 (approved):

PR #9723, opened 2024-05-24
refactor(web): Moved global video player mute status control from the video player mute button to a user preference

IMBeniamin:
Added an option in the user preferences to mute the video player by default as there was no way of setting this currently.

I believe that the video player should be muted by default, but I understand that someone else might not want this, so I think a setting is the best way to handle this

Maybe it would make sense to create another section for the video player, but I thought that for now that isn't necessary and would simply pollute the page

### DIFF
--- a/web/src/lib/components/asset-viewer/video-native-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/video-native-viewer.svelte
@@ -52,9 +52,9 @@
     class="h-full object-contain"
     on:canplay={handleCanPlay}
     on:ended={() => dispatch('onVideoEnded')}
-    bind:muted={$videoViewerMuted}
     bind:volume={$videoViewerVolume}
     poster={getAssetThumbnailUrl(assetId, ThumbnailFormat.Jpeg, checksum)}
+    muted={$videoViewerMuted}
   >
     <source src={assetFileUrl} type="video/mp4" />
     <track kind="captions" />
--- a/web/src/lib/components/user-settings-page/app-settings.svelte
+++ b/web/src/lib/components/user-settings-page/app-settings.svelte
@@ -11,6 +11,7 @@
     playVideoThumbnailOnHover,
     showDeleteModal,
     sidebarSettings,
+    videoViewerMuted,
   } from '$lib/stores/preferences.store';
   import { findLocale } from '$lib/utils';
   import { onMount } from 'svelte';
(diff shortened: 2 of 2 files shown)


michelheusschen:
The PR looks incomplete, did you forget to commit some files?

>Maybe it would make sense to create another section for the video player, but I thought that for now that isn't necessary and would simply pollute the page

An alternative approach is to remember the value of `video.muted` which doesn't require a user setting

IMBeniamin:
> The PR looks incomplete, did you forget to commit some files?
It should be complete, all I added was the preferences option

> An alternative approach is to remember the value of video.muted which doesn't require a user setting
This could be a viable alternative, I honestly don't mind as long as there is a way to have videos muted by default

One drawback of the video.muted is that when I want to turn the volume on I only want it for that video in that specific context (say I'm showing something to a friend)
In all other situations I want everything to stay muted so I don't get surprises in public places/sensitive situations

What is missing is the mobile implementation for this

michelheusschen:
I didn't realize `videoViewerMuted` already exists and it already remembers the value of `video.muted`. That means the video mute button and the user option from this PR do the same thing and the video won't be muted by default if you change either.

IMBeniamin:
> the video won't be muted by default if you change either.

Considering that this option is persistent
https://github.com/immich-app/immich/blob/056a3ce3b6aaf72f161a323820966888d1baadef/web/src/lib/stores/preferences.store.ts#L68
it should persist between sessions.
In my testing the set state has been kept after closing and opening the site.

> That means the video mute button and the user option from this PR do the same thing

I have now noticed that this is also implemented in the mute button and it does behave as expected

michelheusschen:
You can see the issue I'm trying to describe with the following steps:
1. Enable the `Muted video player` option
2. Play a video and unmute it
3. The option `Muted video player` is now disabled

The option description says the video player will be muted *by default*, but that's not the case.

IMBeniamin:
> You can see the issue I'm trying to describe with the following steps:
> 
> 1. Enable the `Muted video player` option
> 2. Play a video and unmute it
> 3. The option `Muted video player` is now disabled
> 
> The option description says the video player will be muted _by default_, but that's not the case.

They share the same variable so this is expected, I was not aware that the video player toggle had this effect.
It now becomes a question of preference; for the reasons mentioned above I prefer a global toggle (which would mean removing the current global behaviour from the mute button i.e. mute button doesn't change `$videoViewerMuted`), but this is again a personal preference.

I will update the pr with another commit that moves the functionality from the mute button to the settings toggle and I guess leave the choice to the maintainers

michelheusschen:
>It now becomes a question of preference; for the reasons mentioned above I prefer a global toggle (which would mean removing the current global behaviour from the mute button i.e. mute button doesn't change $videoViewerMuted), but this is again a personal preference.

You can keep the functionality of both by introducing something like `videoViewerMutedDefault`

```svelte
<video
    muted={$videoViewerMutedDefault || $videoViewerMuted}
    on:volumechange={(e) => ($videoViewerMuted = e.currentTarget.muted)}
 />
 ```

IMBeniamin:
I think that the current implementation is fine for now.
Closing this in order to remove clutter, if this is needed in the future feel free to re-open/add the secondary variable.

PR #9729, opened 2024-05-24
fix(web): shared link isOwner check

michelheusschen:
The buttons favorite, download and more options are visible for guests visiting a shared link with hidden metadata. The check currently fails because `asset.ownerId` is undefined, even though it's a required property of `AssetResponseDto`.


jrasm91 (approved):
Can you add a test for this?

jrasm91 (approved):
Awesome, thanks!

PR #9747, opened 2024-05-25
fix(server): search duplicates of the same asset type

mertalev:
## Description

Duplicate detection currently treats images and videos as duplicates so long as their previews look similar. This is nonsensical since they're known to be different asset types. This PR filters by asset type to correct this.


danieldietzler (approved):
LGTM

jrasm91 (approved):

Issue #9676, opened 2024-05-22
Microservices error contacting Machnie Learning when URL has trailing slash


AlmightyFrog:
### The bug

I was doing tests doing load balancing ML jobs to several systems in parallel. When i changed it back from nginx load balancer to direct container configured it like this `http://immich-machine-learning:3003/` including a trailing slash
![image](https://github.com/immich-app/immich/assets/2421469/df36fdac-82bc-4b8b-a984-68a5cbcc01c5)

This seems to break Microservices container issuing Machinelearning
```
immich_microservices           | [Nest] 7  - 05/22/2024, 7:35:18 PM   ERROR [ImmichMicroservices] [JobService] Object:
immich_microservices           | {
immich_microservices           |   "id": "d63c6a02-d2d7-4df9-be46-17c3432d4281"
immich_microservices           | }
immich_microservices           |
immich_microservices           | [Nest] 7  - 05/22/2024, 7:35:18 PM   ERROR [ImmichMicroservices] [JobService] Unable to run job handler (smartSearch/smart-search): Error: Machine learning request for clip failed with status 404: Not Found
immich_microservices           | [Nest] 7  - 05/22/2024, 7:35:18 PM   ERROR [ImmichMicroservices] [JobService] Error: Machine learning request for clip failed with status 404: Not Found
immich_microservices           |     at MachineLearningRepository.predict (/usr/src/app/dist/repositories/machine-learning.repository.js:23:19)
immich_microservices           |     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
immich_microservices           |     at async SmartInfoService.handleEncodeClip (/usr/src/app/dist/services/smart-info.service.js:86:31)
immich_microservices           |     at async /usr/src/app/dist/services/job.service.js:145:36
immich_microservices           |     at async Worker.processJob (/usr/src/app/node_modules/bullmq/dist/cjs/classes/worker.js:394:28)
immich_microservices           |     at async Worker.retryIfFailed (/usr/src/app/node_modules/bullmq/dist/cjs/classes/worker.js:581:24)
```

When I change it to `http://immich-machine-learning:3003` everything works fine.

Can someone else please confirm that this is a real issue, and if so, either sanitize the input or give a warning or at least document it somewhere that traling slash seems to make an issue?
Thanks!

---

I tried to get the exact url which is making issues by switching back to nginx load balancer and logging all access+error, but there then now only following popped up which i guess is something completely else.
```
==> access.log <==
[...]

PR #9751, opened 2024-05-25
fix(server): Properly build ML predict URL

safehome-jdev:
#9676 - A trailing slash at the end of the machine learning URL breaks the requests to the ML endpoint.

- Created a new enum to filter for URL specific types
- Added a `validationMessage` variable to provide information to the front-end for user feedback
- Added an `isInvalid` variable to provide a state for the tailwind classes
- These changes do not prevent a user from pressing save

validation error with message:
image

valid input:
image

### DIFF
--- a/server/src/repositories/machine-learning.repository.ts
+++ b/server/src/repositories/machine-learning.repository.ts
@@ -19,9 +19,11 @@ export class MachineLearningRepository implements IMachineLearningRepository {
   private async predict<T>(url: string, input: TextModelInput | VisionModelInput, config: ModelConfig): Promise<T> {
     const formData = await this.getFormData(input, config);
 
-    const res = await fetch(`${url}/predict`, { method: 'POST', body: formData }).catch((error: Error | any) => {
-      throw new Error(`${errorPrefix} to "${url}" failed with ${error?.cause || error}`);
-    });
+    const res = await fetch(new URL('/predict', url), { method: 'POST', body: formData }).catch(
+      (error: Error | any) => {
+        throw new Error(`${errorPrefix} to "${url}" failed with ${error?.cause || error}`);
+      },
+    );
 
     if (res.status >= 400) {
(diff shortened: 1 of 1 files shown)


bo0tzz (commented):
Shouldn't this be serverside validation instead?

Furthermore, validating URLs with complex regex is not great. It'd be much better to use something like [URL.canParse()](https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static) instead, or an actual validation library.

martabal:
yep, IMO that's should be a server-side validation. Technically, `http://immich-machine-learning:3000/` is a valid URL and it would be misleading to not accept it especially when the input title and the description are `URL` and `URL of the machine learning server`. 

But having a way to alert/block when a URL is not a valid URL like in your first example would be amazing.

safehome-jdev:
@bo0tzz @martabal making this server-side would mean validation at both locations to the same input, no? I say that because in order to provide real-time user feedback, we need to manipulate the DOM; otherwise we'd need to provide an API response that their input is incorrect after pressing save and highlight the objects that are incorrect with some form of object that identifies the form inputs.

Also I didn't know that `urlParse()` accepted non-Latin characters. I'll have to give that a test

bo0tzz:
I just tested and Immich already rejects invalid machine-learning URLs at the API level, with a clear error message in the UI, so none of these clientside changes are necessary. What is still necessary, and what is really the problem in #9676, is to properly build the URL when calling the machine learning service in a way that does not break if there is a trailing slash. That would be done by building a URL object, rather than the current string concatenation:
```javascript
let baseUrl = "http://immich-machine-learning:3003/";
let url = new URL("/predict", baseUrl);
// use url in fetch call
```

safehome-jdev:
Removed previous UI validation suggestions and input highlighting due to API validation performing this. Modified server-side code to use the `new URL` constructor instead of string con-🐈.

bo0tzz (approved):
Thanks!

jrasm91 (approved):

Issue #9657, opened 2024-05-22
Blank album appears in Add to Album modal


Snowknight26:
### The bug

If you go to the Sharing page, start creating an album, and then close out of it, a blank album appears in the Add to Album modal.

### The OS that Immich Server is running on

Windows 10 22H2 (19045.4291)

### Version of Immich Server

v1.105.1

### Version of Immich Mobile App

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1. Go to Sharing
2. Click Create Album in the top right
3. Click the back ("Close") icon or press escape
4. Go to Photos
5. Select a photo
6. Click the plus ("Add to...") icon in the top right
7. Select Add to Album
```

### Additional information

![image](https://github.com/immich-app/immich/assets/5608270/149ca1bf-f2de-4592-a3b3-e81517fb8dd3)

Refreshing the page makes and then doing steps 4-7 makes them go away, so they seem to be temporary albums in memory instead of in the database.

aviv926:
I can't reproduce the problem (Android, latest version.)
Are you using a reverse proxy?

Snowknight26:
No reverse proxy, but this only affects the web version.

aviv926:
Yes, my bad, I didn't notice that you mentioned it on the Web

remshams:
Navigating to the albums page (after adding it in sharing) also "fixes" the bug. You can quickly see the album popping up and then it's gone (also from the context menu when selecting a photo).

Snowknight26:
Fixed by #9771.

PR #9771, opened 2024-05-26
fix(web): empty album stored

remshams:
I was annoyed by #9657 so I fixed it on may local version.

The idea is to remove and empty album (without assets) when navigating back (clicking the close button) from the album create page instead of cleaning up when navigating to the albums page.
However I'm just starting to get familiar with the code, so I don't know if there are any side effects I'm not aware of...

The alternative would be to do the same cleanup when e.g. when adding a a photo to an album (in the "photos" section).
The advantage of doing it on closing is that it only needs to done once and it also avoids the flickering when opening the "albums" section (the empty album is visible just before it is deleted).

Please let me know if it is useful for you.

### DIFF
--- a/web/src/lib/components/album-page/albums-list.svelte
+++ b/web/src/lib/components/album-page/albums-list.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
   import { onMount } from 'svelte';
   import { groupBy, orderBy } from 'lodash-es';
-  import { addUsersToAlbum, deleteAlbum, type AlbumUserAddDto, type AlbumResponseDto } from '@immich/sdk';
+  import { addUsersToAlbum, deleteAlbum, type AlbumUserAddDto, type AlbumResponseDto, isHttpError } from '@immich/sdk';
   import { mdiDeleteOutline, mdiShareVariantOutline, mdiFolderDownloadOutline, mdiRenameOutline } from '@mdi/js';
   import Icon from '$lib/components/elements/icon.svelte';
   import EditAlbumForm from '$lib/components/forms/edit-album-form.svelte';
@@ -267,9 +267,19 @@
   };
 
   const handleDeleteAlbum = async (albumToDelete: AlbumResponseDto) => {
-    await deleteAlbum({
-      id: albumToDelete.id,
-    });
+    try {
+      await deleteAlbum({
+        id: albumToDelete.id,
+      });
+    } catch (error) {
+      // In rare cases deleting an album completes after the list of albums has been requested,
+      // leading to a bad request error.
+      // Since the album is already deleted, the error is ignored.
+      const isBadRequest = isHttpError(error) && error.status === 400;
+      if (!isBadRequest) {
+        throw error;
+      }
+    }
 
     ownedAlbums = ownedAlbums.filter(({ id }) => id !== albumToDelete.id);
     sharedAlbums = sharedAlbums.filter(({ id }) => id !== albumToDelete.id);
--- a/web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte
+++ b/web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { afterNavigate, goto } from '$app/navigation';
+  import { afterNavigate, goto, onNavigate } from '$app/navigation';
   import AlbumDescription from '$lib/components/album-page/album-description.svelte';
   import AlbumOptions from '$lib/components/album-page/album-options.svelte';
   import AlbumSummary from '$lib/components/album-page/album-summary.svelte';
@@ -405,6 +405,12 @@
       handleError(error, 'Unable to update album cover');
     }
   };
+
+  onNavigate(async () => {
+    if (album.assetCount === 0 && !album.albumName) {
+      await deleteAlbum(album);
+    }
+  });
 </script>
 
 <div class="flex overflow-hidden" bind:clientWidth={globalWidth}>
@@ -559,7 +565,7 @@
             {#if viewMode !== ViewMode.SELECT_THUMBNAIL}
               <!-- ALBUM TITLE -->
               <section class="pt-24">
-                <AlbumTitle id={album.id} albumName={album.albumName} {isOwned} />
+                <AlbumTitle id={album.id} bind:albumName={album.albumName} {isOwned} />
 
                 {#if album.assetCount > 0}
                   <AlbumSummary {album} />


web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.ts
```diff
@@ -6,7 +6,7 @@ import type { PageLoad } from './$types';
 export const load = (async ({ params }) => {
   await authenticate();
   const [album, asset] = await Promise.all([
-    getAlbumInfo({ id: params.albumId, withoutAssets: true }),
+    getAlbumInfo({ id: params.albumId, withoutAssets: false }),
```
Comment (remshams) [inline web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.ts:9]: This is necessary as otherwise the albums with assets get deleted when clicking the back button.

Not sure if this is a performance problem/how expensive the query is in the backend.

web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte
```diff
@@ -275,12 +276,15 @@
     return;
   };
 
-  const refreshAlbum = async () => {
+  const refreshAlbum = async (options: { assets?: AssetResponseDto[] } = {}) => {
```
Comment (remshams) [inline web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte:279]: The optional parameter could be removed if `withoutAssets` is set to `false` (depending on how expensive the query is in the backend).

web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.ts
Comment (jrasm91) [reply web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.ts:9]: It is and it is quite expensive. Album info should have the asset count in it though.

Comment (remshams) [reply web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.ts]: This works perfectly, no idea why I missed the property, sorry and thx.

web/src/lib/components/album-page/albums-list.svelte
```diff
@@ -301,11 +294,6 @@
     }
   };
 
-  const removeAlbumsIfEmpty = async () => {
-    const albumsToRemove = ownedAlbums.filter((album) => album.assetCount === 0 && !album.albumName);
-    await Promise.allSettled(albumsToRemove.map((album) => handleDeleteAlbum(album)));
```
michelheusschen, line 306: We should keep this in case the user creates an empty album and then closes the tab. In the long term, it might be better to only create albums that have a title or contain assets.

web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte
```diff
@@ -399,6 +400,13 @@
       handleError(error, 'Unable to update album cover');
     }
   };
+
+  const checkAlbumAndGoBack = async () => {
+    if (album.assetCount === 0) {
```
Comment (michelheusschen) [inline web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte:405]: This check changes the behavior slightly. Previously, an empty album name was required before the album gets deleted. Is that an intentional change?

Comment (remshams) [reply web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte]: No, that was by mistake. Thanks for pointing it out.

web/src/lib/components/album-page/albums-list.svelte
remshams: Good point, I have not considered the "tab close" case. The change has been reverted.

I had the same thought/idea about not creating the album, but if I'm not mistaken, that requires more/farther reaching changes. Maybe I can work on it in the future...

web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte
Comment (remshams) [reply web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte]: The album is still deleted even when setting a title., it looks like the property is not updated before the page is closed and the album reloaded.

I have to look into it.

Comment (michelheusschen) [reply web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte]: You can use bind to make sure the album name is updated:

```diff
- <AlbumTitle id={album.id} albumName={album.albumName} {isOwned} />
+ <AlbumTitle id={album.id} bind:albumName={album.albumName} {isOwned} />
```

Maybe using [beforeNavigate](https://kit.svelte.dev/docs/modules#$app-navigation-beforenavigate) would be a more robust approach, because there are several ways to navigate away from this page including the browser back button, pressing escape and clicking the back arrow.

Comment (remshams) [reply web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte]: I had another look into it. The reason is, that after a full page reload, the `backUrl` is set to `ALBUMS` instead of `SHARING` (the `from` route gets lost due to the reload). So when clicking the "close" button, the user is redirected to "Albums" instead of "Sharing" (this is also the case on `main`).

The `load` method in "albums" is triggered/finished before the `delete` call in `onNavigate` which leads to the aforementioned error (`getAllAlbums` loads a soon to be deleted album). It seems like SvelteKit executes the `load` function before `onNavigate` completes (this was surprising to me but, as mentioned, I'm not an expert in Svelte).

If I'm not mistaken, the only way to solve the issue is to create the album in the "createAlbum" page itself, but this is bigger change. Not sure if it's worth the effort...

Another option would be to go back to the previous "onClose" solution with the drawback that it only covers the close by button case.

Any other ideas?

Comment (michelheusschen) [reply web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte]: The back url defaulting to /albums is expected, because the state isn't persisted.

It's annoying that load functions run first. I did come up with a solution that seems to work by first cancelling navigation and then calling `goto`, but it's nothing pretty:

```ts
let albumDeleted = false;

beforeNavigate((navigation) => handlePromiseError(handleBeforeNavigate(navigation)));

const handleBeforeNavigate = async (navigation: BeforeNavigate) => {
  const url = navigation.to?.url;
  if (!url || albumDeleted) {
    return;
  }

  if (album.assetCount === 0 && !album.albumName) {
    navigation.cancel();
    albumDeleted = true;

    try {
      await deleteAlbum(album);
    } finally {
      await goto(url);
    }
  }
};
```

Another solution would be to treat `400 Bad Request` as a success when deleting empty albums on the page `/albums`. Typically delete calls for REST APIs are idempotent and this wouldn't be an issue, but ours isn't.

Comment (remshams) [reply web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte]: I also thought about loading the albums outside of the `load` function, but this feels hacky as well. In addition, if I'm not mistaken, the `load` function runs server side so it would probably have performance implications.

If I had to choose, I would go with ignoring the error on the album page, as it may cover other such cases (although I can't think of any). Since I don't like either solution (no offense), I'd go with the one that's easiest to implement. 

Of course, it's your final decision.

Comment (michelheusschen) [reply web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte]: Ignoring 400 errors seems fine to me

web/src/lib/components/album-page/albums-list.svelte
```diff
@@ -267,9 +267,15 @@
...
-    });
+    try {
+      await deleteAlbum({
+        id: albumToDelete.id,
+      });
+    } catch {
+      // In rare cases deleting an album completes after the list of albums has been requested,
+      // leading to a bad request error.
+      // Since the album is already deleted, the error is ignored.
+    }
```
michelheusschen, line 278: I woud rethrow the error in all other cases

```ts
const isBadRequest = isHttpError(error) && error.status === 400;
if (!isBadRequest) {
  throw error;
}
 ```

remshams: I wanted just wanted to ask this..., thanks.

michelheusschen (approved):

PR #9773, opened 2024-05-26
feat(web): Option to assign people to unassigned faces

Jan108:
Added the possibility to assign a person to an unassigned face. This includes:
- Show in the detail panel that there are unassigned faces
- Show all faces in the person-side-panel, and highlight faces without a person assigned

grafik
grafik


web/src/lib/components/asset-viewer/detail-panel.svelte
```diff
@@ -75,6 +78,10 @@
     if (newAsset.id && !isSharedLink()) {
       const data = await getAssetInfo({ id: asset.id });
       people = data?.people || [];
+      await getFaces({ id: asset.id }).then((data) => {
+        faces = data;
+        hasUnassignedFaces = faces.length > people.length;
```
mertalev, line 83: This won't always work. The same person can appear multiple times in an image (mirrors, collages, etc.).

web/src/lib/components/asset-viewer/detail-panel.svelte
```diff
@@ -75,6 +78,10 @@
     if (newAsset.id && !isSharedLink()) {
       const data = await getAssetInfo({ id: asset.id });
       people = data?.people || [];
+      await getFaces({ id: asset.id }).then((data) => {
```
mertalev, line 81: `data` already comes with the faces for the asset. No need to query for them again.

web/src/lib/components/asset-viewer/detail-panel.svelte
```diff
@@ -118,6 +128,11 @@
   const handleRefreshPeople = async () => {
     await getAssetInfo({ id: asset.id }).then((data) => {
       people = data?.people || [];
+      hasUnassignedFaces = faces.length > people.length;
+    });
+    await getFaces({ id: asset.id }).then((data) => {
+      faces = data;
+      hasUnassignedFaces = faces.length > people.length;
```
mertalev, line 135: Same here.

web/src/lib/components/faces-page/person-side-panel.svelte
```diff
@@ -207,91 +204,125 @@
...
-                      curve
-                      shadow
-                      url={selectedPersonToCreate[face.id]}
-                      altText={selectedPersonToCreate[face.id]}
-                      title={'New person'}
-                      widthStyle={thumbnailWidth}
-                      heightStyle={thumbnailWidth}
-                    />
-                  {:else if selectedPersonToReassign[face.id]}
+          {@const person_name = face.person ? face.person?.name : 'unassigned'}
```
mertalev, line 207: `person_name` should be `personName` (or just `name`). `unassigned` should be `Unassigned`.

web/src/lib/components/faces-page/person-side-panel.svelte
```diff
@@ -207,91 +204,125 @@
...
+                    title={'New person'}
+                    widthStyle={thumbnailWidth}
+                    heightStyle={thumbnailWidth}
+                  />
+                {:else if selectedPersonToReassign[face.id]}
+                  <ImageThumbnail
+                    curve
+                    shadow
+                    url={getPeopleThumbnailUrl(selectedPersonToReassign[face.id].id)}
+                    altText={selectedPersonToReassign[face.id]?.name || selectedPersonToReassign[face.id].id}
```
mertalev, line 233: I know this is existing code, but I don't see the point of having an ID for alt text. It's arguably worse than not having anything there because it will make a screen reader speak gibberish.

web/src/lib/components/faces-page/person-side-panel.svelte
```diff
@@ -207,91 +204,125 @@
...
-                      title={getPersonNameWithHiddenValue(
-                        selectedPersonToReassign[face.id].name,
-                        face.person?.isHidden,
-                      )}
-                      widthStyle={thumbnailWidth}
-                      heightStyle={thumbnailWidth}
-                      hidden={selectedPersonToReassign[face.id].isHidden}
+                      url="/src/lib/assets/no-thumbnail.png"
+                      altText="unassigned"
+                      title="unassigned"
```
mertalev, line 260: Also capitalized.

web/src/lib/components/faces-page/person-side-panel.svelte
```diff
@@ -207,91 +204,125 @@
...
+                    widthStyle={thumbnailWidth}
+                    heightStyle={thumbnailWidth}
+                    hidden={selectedPersonToReassign[face.id].isHidden}
+                  />
+                {:else if face.person}
+                  <ImageThumbnail
+                    curve
+                    shadow
+                    url={getPeopleThumbnailUrl(face.person.id)}
+                    altText={face.person.name || face.person.id}
```
mertalev, line 247: The `id` alt text should be removed here too.

web/src/lib/utils/people-utils.ts
```diff
@@ -69,3 +71,52 @@ export const getBoundingBox = (
   }
   return boxes;
 };
+
+export const zoomImageToBase64 = async (face: AssetFaceResponseDto, assetId: string): Promise<string | null> => {
```
mertalev, line 75: The current code has different behavior for images vs videos while this doesn't. Is there a particular reason for that?

mertalev (commented):
Thanks for working on this! This is something I've wanted to add for some time.

web/src/lib/components/asset-viewer/detail-panel.svelte
Jan108: Completely forgot about that. Changed it to compare to the amount of faces assigned to people.

web/src/lib/components/faces-page/person-side-panel.svelte
Jan108: I agree with that and removed the ID from every altText you mentioned.

web/src/lib/utils/people-utils.ts
Jan108: I moved the method to a utils files, because it's used in two different files. The method used the open image in the photoeditor for the thumbnail generation if the asset was an image. If the method is in the utils file, I can't be sure that it will be called only when the photoeditor is open. I think this dependency would be bad, so I simplified it to always query the thumbnail from the server.

web/src/lib/components/asset-viewer/detail-panel.svelte
Jan108: I don't think so. `data` contains the people, and their assigned faces, but not the unassigned faces.

mertalev: Is that so? We should change that in the server then.

Jan108: I added faces to the getAssetInfo endpoint, but didn't update the client sdks yet.

web/src/lib/utils/people-utils.ts
martabal: That dependency exists because it avoids to do a network request for a ressource that has already been fetched.

Jan108: Ok, I added the dependency again.

web/src/lib/components/asset-viewer/detail-panel.svelte
Jan108: Client sdks are not up-to-date.

server/src/dtos/asset-response.dto.ts
```diff
@@ -116,6 +122,7 @@ export function mapAsset(entity: AssetEntity, options: AssetMapOptions = {}): As
     livePhotoVideoId: entity.livePhotoVideoId,
     tags: entity.tags?.map(mapTag),
     people: peopleWithFaces(entity.faces),
+    faces: entity.faces?.map((a) => mapFacesWithoutPerson(a)),
```
mertalev, line 125: This duplicates the info in `people`. Could you filter to only include unassigned faces and maybe change `faces` to `facesWithoutPerson`?

web/src/lib/components/asset-viewer/detail-panel.svelte
```diff
@@ -92,6 +94,9 @@
   $: people = asset.people || [];
   $: showingHiddenPeople = false;
 
+  $: faces = asset.faces || [];
+  $: hasUnassignedFaces = faces.length > people.reduce((count, person) => count + person.faces.length, 0);
```
mertalev, line 98: With the above change, you can just check `facesWithoutPerson.length`.

server/src/dtos/asset-response.dto.ts
jrasm91: I think if `people` includes faces + people and `unassignedFaces` includes the rest of the faces, that would be ok. This is a bit annoying though, as this is adding more and more data to the asset response dto, which is used everywhere, including in the main timeline. We should probably change it to not return people/faces by default and only load those relations in the asset detail endpoint. Or, load those in a separate query like /people?assetId=<id> and /faces?assetId=<id>&state=unassigned

Jan108: That was what I was doing in the first place, using the `/face?id=`. For now, I filtered the faces to only include unassigned ones.

mertalev: > We should probably change it to not return people/faces by default and only load those relations in the asset detail endpoint.

The time bucket query doesn't have these relations, so I believe this is the current behavior.

mertalev (approved):

server/src/dtos/asset-response.dto.ts
jrasm91: Oh, it looks like you are right. That works for me then.

jrasm91 (approved):
LGTM

danieldietzler (approved):

PR #9778, opened 2024-05-26
docs: Add Google OAuth example

aviv926:
This PR adds screenshots for Google OAuth setup.


jrasm91:
I think we can include the screenshot for Google cloud, but for the immich side, let's just include the example values in plain text instead, since screenshots quickly become outdated.

aviv926:
> I think we can include the screenshot for Google cloud, but for the immich side, let's just include the example values in plain text instead, since screenshots quickly become outdated.

Yes, it's better, I updated accordingly.

docs/docs/administration/oauth.md
```diff
@@ -110,8 +110,45 @@ Immich has a route (`/api/oauth/mobile-redirect`) that is already configured to
...
+
+Configuration of Authorised redirect URIs (Google Console)
+
+<img src={require('./img/google-example.webp').default} width='50%' title="Authorised redirect URIs" />
+
+Configuration of OAuth in System Settings
+
+| Setting                      | Value                                                                                                  |
+| ---------------------------- | ------------------------------------------------------------------------------------------------------ |
+| Enable                       | Login with OAuth                                                                                       |
```
jrasm91, line 139: Is this supposed to have the value of true?

aviv926: A mistake, I fixed it now. 👌

alextran1502 (approved):

jrasm91 (approved):

PR #9784, opened 2024-05-27
feat(server): better video thumbnails

mertalev:
## Description

We currently use the first frame to generate video thumbnails. However, this can be a poor choice as the start of the video can be black, white, blurry, etc. This PR uses a set of filters to find a suitable frame. ~~It also includes a fallback to use the first frame in case it can't find a good frame.~~

Note on performance: it will be slower depending on how far it needs to scrub to find a good thumbnail. This could be fast if it's at or close to the start, or it could be slower if it needs to go far. I tuned the parameters to not be overly strict to minimize the chance of the latter and set it to choose the first available frame after 20s of searching.

## How Has This Been Tested?

[Source video](https://www.youtube.com/watch?v=njX2bu-_Vw4)

Before:

img

After:

img2


mertalev:
It looks like the fallback approach here only works for still image formats like PNG and JPEG. I can do it in code instead. However, FFmpeg doesn't actually throw an error if there are no output files, so it would have to be handled by checking that the file exists.

jrasm91 (approved):

danieldietzler (approved):
FFmpeg magic 🎉

PR #9800, opened 2024-05-27
chore: refactor transcode config routing

mertalev:
## Description

The transcode config to use is currently handled in the media service. This PR moves it to the util instead as the routing is somewhat complex and should be considered an implementation detail. Creating a config is now declarative, without the service needing to care about what config it really is.

Also renames `TranscodeOptions` to `TranscodeCommand` and `getOptions` to `getCommand`. This is to better distinguish it from the admin config.

Lastly, it removes some uses of `null` in the service.


danieldietzler (approved):
Nice, I like it!

jrasm91 (approved):
Nice

PR #9802, opened 2024-05-27
fix(server): nvenc not working when there are no filters

mertalev:
## Description

The code currently adds `format=nv12` if there is no filter, but this makes FFmpeg throw an error. Omitting it works for both HDR and SDR inputs, so there's no need for it.

## How Has This Been Tested?

Tested on a video that was throwing an error with `format=nv12`, confirming that it works when omitted. Also tested on an HDR video that still worked without this.


danieldietzler (approved):
🪄

PR #9805, opened 2024-05-27
feat(deployment): add TZ to example.env

mmomjian:
`TZ` is used by exiftool and, apparently, log file timestamps. This needs to be set even if /etc/localtime is passed through. I propose that we add a commented TZ field to example.env to prompt users to select their timezone.


mmomjian:
`TZ` is also used to localize the cron job entries for library scanning

alextran1502 (approved):

PR #9815, opened 2024-05-28
Update backup-and-restore.md - added "-d immich" to psql restore command

bhugh:
The current command to unzip the backup and restore it to the database using the psql command malfunctions, because the database name "immich" needs to be included.

I've made several backups/restores now and none work if the "-d immich" is absent - all work just fine when it is present.

So, we need to update the instructions now.

I backed-up/restored using v1.104.0 - it's possible something works different in v1.105.x.  But regardless of that, many people restoring will be using versions earlier than 1.105.


bo0tzz:
The suggested backup command was changed to use `pg_dumpall` in #8546, which means that you don't need to set a database to restore into. Since Immich is still moving quickly we've opted not to keep information for older versions around in the docs, since they would grow very full if we did that. (We're working on keeping the complete docs for older versions around instead, which should compensate for that somewhat)

PR #9831, opened 2024-05-28
refactor: asset media endpoints

jrasm91:
Update asset media endpoints. All endpoints starting with `/asset` have been replaced with `/assets`. Additionally, the following endpoints have been updated:

| Operation | Old endpoint | New endpoint  | Notes |
| - | - | - | - |
| `uploadAsset` | `POST /asset/upload` | `POST /assets` | Create an asset, optionally with `livePhotoVideoId` |
| `downloadAsset` | `GET /asset/file/:id` | `GET /assets/:id/original` | Download the original (always) |
| `viewAsset` | `GET /asset/thumbnail/:id` | `GET /assets/:id/thumbnail` | View the thumbnail or preview (via query param) |
| `playAssetVideo` | (split off of `downloadAsset`) | `GET /assets/:id/video/playback` | View the encoded video or original |
| `downloadFile` | `POST /download/asset/:id` | (use `downloadAsset`) | Deleted |


server/src/controllers/asset-media.controller.ts
```diff
@@ -42,10 +56,59 @@ export class AssetMediaController {
     private service: AssetMediaService,
   ) {}
```
midzelis, line 58: use getFiles() and call it `createAsset` instead of `uploadFile`
```suggestion
@Post()
  @UseInterceptors(AssetMediaUploadInterceptor, FileUploadInterceptor)
  @ApiConsumes('multipart/form-data')
  @ApiHeader({
    name: ImmichHeader.CHECKSUM,
    description: 'sha1 checksum that can be used for duplicate detection before the file is uploaded',
    required: false,
  })
  @ApiBody({
    type: CreateAssetMediaDto,
  })
  @Authenticated({ sharedLink: true })
  @EndpointLifecycle({ addedAt: 'v1.106.0' })
  async createAsset(
    @Auth() auth: AuthDto,
    @UploadedFiles(new ParseFilePipe({ validators: [new FileNotEmptyValidator(['assetData'])] })) files: UploadFiles,
    @Body() dto: AssetMediaReplaceDto,
    @Res({ passthrough: true }) res: Response,
  ): Promise<AssetMediaResponseDto> {
    const { file, sidecarFile } = getFiles(files);
    const assetMediaResponse = await this.service.createAsset(auth, dto, file, sidecarFile);
    if (assetMediaResponse.status === AssetMediaStatusEnum.DUPLICATE) {
      res.status(HttpStatus.OK);
    }
    return assetMediaResponse;
  }
```

server/src/middleware/file-upload.interceptor.ts
```diff
@@ -32,7 +32,7 @@ export function getFiles(files: UploadFiles) {
 }
 
 export enum Route {
-  ASSET = 'asset',
+  ASSET = 'assets',
```
midzelis, line 35: I think you should also move canUploadFIle getUploadFilename getUploadFolder from asset service to asset-media service.

server/src/controllers/asset-media.controller.ts
jrasm91: I have made most of these changes. I might make the "remove live photo data" in a follow-up PR though.

jrasm91: Ended up doing it all in this PR.

server/src/services/metadata.service.ts
```diff
@@ -411,7 +411,11 @@ export class MetadataService {
       }
       const checksum = this.cryptoRepository.hashSha1(video);
 
-      let motionAsset = await this.assetRepository.getByChecksum(asset.libraryId ?? null, checksum);
+      let motionAsset = await this.assetRepository.getByChecksum({
+        ownerId: asset.ownerId,
+        libraryId: asset.libraryId ?? undefined,
```
danieldietzler, line 416:
```suggestion
        libraryId: asset.libraryId,
```

danieldietzler (approved):
R~IP~ asset v1 🚀

server/src/services/metadata.service.ts
jrasm91: It has a null type but the method doesn't accept null

danieldietzler: Ah shoot I'm stupid, never mind.

mobile/lib/services/backup.service.dart
```diff
@@ -357,6 +336,48 @@ class BackupService {
...
 
+          if (entity.isLivePhoto) {
+            if (livePhotoFile != null) {
+              final livePhotoTitle = p.setExtension(
+                originalFileName,
+                p.extension(livePhotoFile.path),
+              );
+              final fileStream = livePhotoFile.openRead();
+              final livePhotoRawUploadData = http.MultipartFile(
+                "livePhotoData",
```
jrasm91, line 347: I think this needs to be `assetData` now.

mobile/lib/services/backup.service.dart
```diff
@@ -331,31 +349,35 @@ class BackupService {
...
+              livePhotoReq.files.add(livePhotoRawUploadData);
+
+              // Send live photo only if the non-motion part is successful
+              if (response.statusCode == 200 || response.statusCode == 201) {
+                final data = await response.stream.bytesToString();
+                final mediaResponse = jsonDecode(data);
+                final assetId = mediaResponse['id'].toString();
+                livePhotoReq.fields['livePhotoVideoId'] = assetId;
+
+                response = await httpClient.send(
```
midzelis, line 369: I'm not totally sure, as I'm not too familiar with dart code - but I think this is performing the upload of the livePhoto part of the asset. If so, make sure the form-field is correct (i.e. if you choose to drop it and use assetData instead) 

Additionally, you'll need to add 1 more request after this one, to perform the linking of the live-photo to the still-photo. That can be performed using the standard PUT /api/asset/:id call, and just specify the a single property, which is 'livePhotoVideoId'. I had this change in one of my PRs, but I think it got lost in the shuffle. To complete this, you'll need to modify the UpdateAssetBase DTO to allow PUT request to specify the livePhotoVideoId for a given asset. 

Since there's already so many changes - might as well consider making the names consistent.... (is|has|apply|play|stop)motionPhoto is used in a bunch of .dart code, some web code, and some microservice code. In most other places its called 'livePhotoVideo'. 

It appears that Android calls this "Motion Photos" and iOS calls it "Live Photos" which is where the confusion comes from. But LivePhotoVideo is not a thing ;-)

server/src/dtos/asset-media.dto.ts
```diff
@@ -35,6 +46,28 @@ export class AssetMediaReplaceDto {
...
+  @ValidateBoolean({ optional: true })
+  isArchived?: boolean;
+
+  @ValidateBoolean({ optional: true })
+  isVisible?: boolean;
+
+  @ValidateBoolean({ optional: true })
+  isOffline?: boolean;
+
+  @ValidateUUID({ optional: true })
```
midzelis, line 62: I think we want to drop livePhotoVideoId - as the asset/livephoto should be uploaded separately.

server/src/services/asset-media.service.spec.ts
```diff
@@ -103,6 +220,221 @@ describe('AssetMediaService', () => {
...
+      expect(assetMock.getUploadAssetIdByChecksum).toHaveBeenCalledWith(authStub.admin.user.id, file1);
+    });
+  });
+
+  describe('canUpload', () => {
+    it('should require an authenticated user', () => {
+      expect(() => sut.canUploadFile(uploadFile.nullAuth)).toThrowError(UnauthorizedException);
+    });
+
+    for (const { fieldName, valid, invalid } of uploadTests) {
```
midzelis, line 253: describe.each(uploadTests) for bonus points

server/src/services/asset-media.service.ts
```diff
@@ -57,7 +70,118 @@ export class AssetMediaService {
...
+  ): Promise<AssetMediaResponseDto> {
+    try {
+      await this.access.requirePermission(
+        auth,
+        Permission.ASSET_UPLOAD,
+        // do not need an id here, but the interface requires it
+        auth.user.id,
+      );
+
+      if (dto.livePhotoVideoId) {
```
midzelis, line 159: This code doesn't need to be here, as livePhotos should now be uploaded in their own request, and linked with a PUT request on the asset record.

server/src/services/asset-media.service.ts
```diff
@@ -181,54 +409,59 @@ export class AssetMediaService {
...
+      fileCreatedAt: dto.fileCreatedAt,
+      fileModifiedAt: dto.fileModifiedAt,
+      localDateTime: dto.fileCreatedAt,
+
+      type: mimeTypes.assetType(file.originalPath),
+      isFavorite: dto.isFavorite,
+      isArchived: dto.isArchived ?? false,
+      duration: dto.duration || null,
+      isVisible: dto.isVisible ?? true,
+      livePhotoVideoId: dto.livePhotoVideoId,
```
midzelis, line 437: drop the livephoto prop here

jrasm91: No, this is for linking.

server/src/services/asset-media.service.ts
jrasm91: Why not just link on create?

server/src/services/asset-media.service.spec.ts
jrasm91: I personally don't see the benefit of it.

server/src/dtos/asset-media.dto.ts
jrasm91: Yes it is, this is to link it when the second one is created.

server/src/services/asset-media.service.ts
midzelis: Oh right, you could upload the live photo first, get its id, and then create the primary photo and link it at the same time. Ok, this works. However, I think that means the the modifications to the dart client are backwards. If i'm not mistaken, it appears that the primary photo id is being set as the live photo of the of the live photo.

midzelis: I think it should be possible to upload the primary photo first, and then upload the live photo, and then link the two, which requires a change in asset.service's PUT method to allow the new property to be edited.

jrasm91: This is a good idea. We can add this in a follow-up PR.

PR #9841, opened 2024-05-29
fix(web): navigate to the correct URL when unstacking from the asset viewer

Snowknight26:
Before this change, unstacking from the asset viewer dispatched the close event, which did close the asset viewer, but the route was never updated, so you were still on `/photos/[assetId]`.

### DIFF
--- a/web/src/lib/components/asset-viewer/asset-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/asset-viewer.svelte
@@ -272,10 +272,6 @@
     $isShowDetail = !$isShowDetail;
   };
 
-  const handleCloseViewer = async () => {
-    await closeViewer();
-  };
-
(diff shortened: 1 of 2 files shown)


web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -519,7 +515,7 @@
           asset,
         });
       }
-      dispatch('close');
+      await closeViewer();
```
michelheusschen, line 518: This change fixes the described issue while the other changes just remove an unused event, right?

Snowknight26: Yes, exactly!

jrasm91 (approved):

PR #9854, opened 2024-05-29
feat(web): translations

Manic-87:
See Discussion #1688 

Ignore the hooks.server.ts and the changes to vite.config.js.
The en-EN.json I created manually. 
The de-DE.json I created initially manually but it was filled by weblate. 
fr.json was created by weblate. 
What you see in those files is everything I entered into weblate just to test if it works.

Copy of my comment from the discussion:

So I did invest some time on this and I currently could display quite a few texts in different languages and also make use of weblate.
Disregarding the weblate portion, I want to make sure I'm on the right track about translatability of the web.

1. I added svelte-i18n as a dependency as it looked good to me (mind you, I don't have any experience with svelte, so I can't properly asses if this is actually good). This would be the first question: Is the usage of svelte-i18n ok or are there any concerns?
2.  I started with the "page.svelte" files, took the hardcoded texts, moved them into a "en-EN.json" file and used the svelte-i18n functions to get the texts back. Regarding the structure of the file would be the next important question. For Android it's a flat list. For the web I thought about using a hierarchical structure as it, in my opinion at least, improves later readability/extendability.

This means my file currently would look like this for example:
```
page: {
  admin: {
    library_management: {
      error_create: "Unable to create library"
      ...
      ...
    },
    system_settings: {
      auth_settings_title: "..."
      auth_settings_subtitle: "..."
      ...
      ...
    }
  }
}
```
For components the top level would be "component". For stuff which is used more often like "Video" or "User" there would be a top level "common" and so on. In the svelte files you'd then use
$t(page.admin.user_management.error_create)
for example. For stuff with placeholders you'd use
$t(page.admin.user_management.error_create, {values: {placeholdername: placeholdervalue}}).
With the placeholders you can also use plurals, so instead of having two rules strings for ""Matched 1 item" and "Matched 10 items" there would only be one "ICU plural messages" Matched {count, plural, one {# item} other {# items}}

What do you think? Is this the correct way for you or should I go at it some other way?


web/src/lib/i18n/locales/de-DE.json
```diff
@@ -0,0 +1,38 @@
+{
+    "page": {
+        "error": {
+            "message": "🚨 Fehler - Etwas ist schiefgelaufen",
+            "changelog": "Änderungsprotokoll lesen",
```
MickLesk, line 5: Änderungsprotokoll ansehen

web/src/lib/i18n/locales/de-DE.json
```diff
@@ -0,0 +1,38 @@
...
+            "help": "Erhalte Hilfe",
+            "logs": "Logs überprüfen"
+        },
+        "page": {
+            "title": "Willkommen 🎉",
+            "description": "Immich Web Interface"
+        },
+        "auth": {
+            "register": {
+                "admin_registry_info": "Da Sie der erste Benutzer im System sind, werden Sie als Administrator zugewiesen und sind für administrative Aufgaben zuständig. Weitere Benutzer werden von Ihnen erstellt."
```
MickLesk, line 15: Da du der erste Benutzer im System bist, wirst du automatisch als Administrator zugewiesen. Dadurch bist du für alle administrativen Aufgaben, die Benutzeranlage etc. zuständig.

web/src/lib/i18n/locales/de-DE.json
```diff
@@ -0,0 +1,38 @@
...
+    "common": {
+        "name": "Immich",
+        "welcome": {
+            "title": "Willkommen zu Immich",
+            "getting_started": "Erste Schritte"
+        }
+    },
+    "component": {
+        "form": {
+            "admin_email": "Admin Email",
```
MickLesk, line 28: Admin E-Mail

web/src/lib/i18n/locales/de-DE.json
```diff
@@ -0,0 +1,38 @@
...
+            "admin_password": "Admin Passwort",
+            "admin_password_confirm": "Admin Passwort wiederholen",
+            "name": "Name",
+            "sign_up": "Registrieren Sie sich",
```
MickLesk, line 32: Registrieren

web/src/lib/i18n/locales/de-DE.json
```diff
@@ -0,0 +1,38 @@
...
+            "error_password_mismatch": "Passwörter stimmen nicht überein",
```
MickLesk, line 33: Die Passwörter stimmen nicht überein.

web/src/lib/i18n/locales/de-DE.json
danieldietzler: Actually, I prefer "Email".

web/src/lib/i18n/locales/de-DE.json
danieldietzler: I think both options are fine here.

jrasm91:
This PR is about the technical aspects of translating the web, so please refrain from commenting about specific translations. Those can be dealt with independently of these changes. (The translation values will be managed through an external platform).

With that said, I think this looks really good so far. I wonder what would be the best way to move forward. Presumably we want to share translation strings between mobile and web. Personally, I prefer the flat structure as the hierarchy might become coupled with the implementation (code), which might easily change in the future. Also, I think translating the admin settings are a much lower priority than the sidebar, asset viewer, albums page, and the asset grid component, so maybe we start with those pages. I think it would be good to get a solid foundation and then merge it in, we can incrementally externalize translation strings in follow-up PRs, thus avoiding merge conflicts that inevitably come from long standing PRs.

web/src/hooks.server.ts
```diff
@@ -0,0 +1,10 @@
+import type { Handle } from '@sveltejs/kit'
+import { locale } from 'svelte-i18n'
+
+export const handle: Handle = async ({ event, resolve }) => {
```
midzelis, line 4: This is a good default. Using browser locale is a generally close to accurate. However, I do think we should only treat this as a default/fallback. I think there should be a account-setting that allows you to change the language for your user, overriding the browser locale.

web/src/lib/components/forms/admin-registration-form.svelte
```diff
@@ -16,7 +17,7 @@
 
   $: {
     if (password !== confirmPassword && confirmPassword.length > 0) {
-      errorMessage = 'Password does not match';
+      errorMessage = $t('component.form.error_password_mismatch');
```
midzelis, line 20: Just curious - did you run a codemod or regexp to find all the strings?

web/src/hooks.server.ts
Manic-87: I absolutely agree. This is in a early stage and I was just testing around until now so I didn't bother to also look into how I could achieve something like that.

web/src/lib/components/forms/admin-registration-form.svelte
Manic-87: Up until now: Open File -> look around -> Copy/Paste.

As it was just some testing to get a feel if/how that works, that was alright. For further work on this I thought about trying to use a regexp and small script to at least get most of the strings automatically into a file.

Manic-87:
>Presumably we want to share translation strings between mobile and web. 
I did look into that a bit with weblate. AFAIK you can have different file structures for different components. If you want to make it really easy you'd need to define a glossary (weblate can use existing translations as a glossary) and need to use the same keys for the same translations.
Meaning:
My en-EN.json file has the String `share` under `common`. Weblate interprets it as `common.share`. If the mobile translation file also would have a string with this key the translation of the glossary would be applied to both the web and mobile translation.
The translations for web and mobile could still be translated to something different, but apparently would be overwritten by the glossary translation if the glossary changes.
This would either mean the current translationkeys of the mobile would need to be changed to use `.` instead of `_`, or the web translations would need to use the same keys as the mobile version. 
Both options aren't optimal in my opinion. Changing all translation keys on the mobile seems like an unnecessary effort. Using the same keys on the web as the mobile will sometimes be strange, e.g. `control_bottom_app_bar_add_to_album`. On the web the `Add to album` button is on the top and it's obviously not the app. 
Also I think it is preferable to make use of these ICU plural messages and the mobile don't (or perhaps even can't?) use them.

Unfortunately I haven't found a way to tell weblate to connect a translation to a specific glossary item even if the keys differ or at least mark the translation as "To check" at least. But perhaps I just didn't find it.


Most important question how to proceed would be: How should the keys be defined as changing them later on will be a hassle, but I'm afraid this will lead to neverending discussions.
What could be done would be to just do the translations for the web in the manner I already started and when this is done little by little the strings and translations of the web and mobile could be consolidated. For example: Look at all occurences of "Cancel" and decide on a common key for both web and mobile (mobile currently has 8 occurences in the translation file by the way). This could already be taken into account for strings that exist in both.

web/src/hooks.server.ts
bo0tzz: > there should be a account-setting that allows you to change the language

Strong agree, and ideally this setting works for the mobile app as well. @jrasm91 with the new user preferences this should be relatively straightforward right?

jrasm91: Yup.

jrasm91:
Yeah I think this is a good point. We should decide on our optimal format and start using that. The mobile app should migrate to it in the future as well.

I would prefer having the translations be somewhat independent of the code/UI though. We don't need "app bar" or "bottom sheet" in the key. I think a "cancel" key would probably be fine.

Manic-87:
I changed my approach a bit. Wrote a simple script to extract the strings from the files, automatically create keys and replace the occurences. The automatic creation of the keys I only did for strings of at most 4 words and just replaced whitespaces with underscores and made the string lowercased. So "Add to" has the key "add_to".
This worked reasonably well and the result you can see in my latest commit. There are a few hickups, which need to be fixed again, and obviously a lot of strings which weren't touched but in my opinion this should be a good starting point.

I added an automatically translated german file for testing purposes.

grafik

MickLesk:
Looks good! Massive Update on source Code, Ive only briefly skimmed through it (the individual commits) and everything looks clean so far.

PS: the German translation is crappy :D

waclaw66:
A small correction of `constant_rate_factor_(-crf)` to `constant_rate_factor_crf` would be fine. 
However many thanks for you huge effort, this will help to further widespread of Immich. Many users don't want/can use Immich in English.

Manic-87:
I extracted and replaced some more strings. As these are mostly longer sentences I just used "placeholder_X" for the key for now. I suppose these keys should be decided by you. 
I also changed a few of the keys. There were multiple ones with parantheses or a minus. I removed those.


Big Question: How to proceed. 

Do you want to merge the currently available stuff? If so, with or without the autotranslated german file? Personally, I wouldn't include it, because.... it's really really crappy and I only changed those which annoyed me the most (translation for cancel for example).
Do you want to wait until at least all occurences of "placeholder_X" are changed?
Do you want to wait until even more/all strings are translatable?

jrasm91:
I think it looks good so far. We probably could just commit the English values for now and then finish setting up weblate for languages

bo0tzz:
cc @alextran1502, is weblate just ready to go or do we need to prep anything more?

Manic-87:
FYI: I just rebased the branch to the current main branch of immich and also removed the german translation file.

web/src/lib/i18n/locales/en-EN.json
```diff
@@ -0,0 +1,763 @@
+{
+    "thirty_minutes": "30 minutes",
+    "one_hour": "1 hour",
+    "six_hours": "6 hours",
+    "one_day": "1 day",
+    "seven_days": "7 days",
+    "thirty_days": "30 days",
+    "three_months": "3 months",
+    "one_year": "1 year",
```
bo0tzz, line 9: For fields like this, should we use plurals? Per https://formatjs.io/docs/core-concepts/icu-syntax/#plural-format (though I may have made a mistake here):
```
{"hour": "{count, plural, one {count day} other {{count, number} days}}
```
And then using that like
```
{$_('hour', { values: { count: 3 }})}
```

web/src/lib/i18n/locales/en-EN.json
```diff
@@ -0,0 +1,763 @@
...
+    "target_resolution": "Target resolution",
+    "template": "Template",
+    "temporal_aq": "Temporal AQ",
+    "threads": "Threads",
+    "thumbnail_format": "Thumbnail format",
+    "thumbnail_resolution": "Thumbnail resolution",
+    "tone-mapping_npl": "TONE-MAPPING NPL",
+    "tone-mapping": "TONE-MAPPING",
+    "transcode_policy": "Transcode policy",
+    "two-pass_encoding": "TWO-PASS ENCODING",
```
bo0tzz, line 322: I think there's some inconsistency in the use of casing here (and of toUpperCase where these values are used)

waclaw66:
Those placeholders can be named according previous label... `welcome_message`, `placeholder_0` => `welcome_message_desc`

web/src/lib/i18n/locales/en-EN.json
waclaw66: Plurals are very welcomed, especially for non-germanic languages (etc. slavic).

jrasm91:
I'll go ahead and finish cleaning up the translations and then merge this. Once merged, I'll configure weblate with the git integration and we can go from there.

jrasm91 (approved):
Thanks for working on this!

danieldietzler (approved):
Thanks a lot for pushing this forward! :)

web/src/lib/stores/preferences.store.ts
```diff
@@ -42,6 +42,13 @@ export const locale = persisted<string | undefined>('locale', undefined, {
   },
 });
 
+export const lang = persisted<string | undefined>('lang', undefined, {
+  serializer: {
+    parse: (text) => text,
+    stringify: (object) => object ?? '',
+  },
+});
+
```
jrasm91, line 51: Dedicated preference

web/src/lib/constants.ts
```diff
@@ -155,7 +157,7 @@ export const locales = [
   { code: 'en-TT', name: 'English (Trinidad and Tobago)' },
   { code: 'en-VI', name: 'English (U.S. Virgin Islands)' },
   { code: 'en-GB', name: 'English (United Kingdom)' },
-  { code: 'en-US', name: 'English (United States)' },
+  { code: 'en-US', name: 'English (United States)', loader: () => import('$lib/i18n/en-US.json') },
```
jrasm91, line 160: Languages are a subset of locales.

web/src/lib/constants.ts
```diff
@@ -245,3 +245,11 @@ export const locales = [
...
   { code: 'zu-ZA', name: 'Zulu (South Africa)' },
 ];
+
+export const fallbackLang = 'en-US';
+
+type LanguageLoader = { code: string; name: string; loader: Parameters<typeof register>[1] };
+export const langs = [
+  ...locales.filter((item): item is LanguageLoader => !!item.loader),
+  { name: 'Development', code: 'dev', loader: () => Promise.resolve({}) },
+];
```
jrasm91, line 255: Locales with loaders + development language (shows raw translation keys)

web/src/routes/+layout.ts
```diff
@@ -1,15 +1,27 @@
...
   defaults.fetch = fetch;
 
+  for (const { code, loader } of langs) {
+    register(code, loader);
+  }
+
+  const preferenceLang = get(lang);
+
+  await init({ fallbackLocale: fallbackLang, initialLocale: preferenceLang || fallbackLang });
+
```
jrasm91, line 24: Load languages and default to preference or english if unset.

web/src/lib/components/user-settings-page/app-settings.svelte
```diff
@@ -62,6 +65,20 @@
...
+    const previousLang = get(i18nLocale);
+
+    if (newLang === 'dev') {
+      await init({ fallbackLocale: 'dev', initialLocale: 'dev' });
+    } else if (previousLang == 'dev' && newLang !== 'dev') {
+      await init({ fallbackLocale: 'en-US', initialLocale: newLang });
+    }
+    $i18nLocale = newLang;
+  };
+
```
jrasm91, line 81: Handle language change (need to remove the english fallback to see the development keys)

PR #9856, opened 2024-05-29
fix: immich port with external domain

martabal:
Currently, the fallback for the external domain uses the `:2283` port which is the wrong port if you have `IMMICH_PORT` set.


server/src/constants.ts
```diff
@@ -24,6 +24,8 @@ export const envName = (process.env.IMMICH_ENV || 'production').toUpperCase();
 export const isDev = () => process.env.IMMICH_ENV === 'development';
 export const APP_MEDIA_LOCATION = process.env.IMMICH_MEDIA_LOCATION || './upload';
 export const WEB_ROOT = process.env.IMMICH_WEB_ROOT || '/usr/src/app/www';
+const HOST_SERVER_PORT = process.env.IMMICH_PORT || '2283';
+export const FALLBACK_EXTERNAL_DOMAIN = 'http://localhost:' + HOST_SERVER_PORT;
```
jrasm91, line 28: Can you rename this to `DEFAULT_EXTERNAL_DOMAIN`?

martabal: of course

jrasm91 (approved):

PR #9857, opened 2024-05-29
feat(ci): website deployment IaC and github actions

zackpollard:
Deploy website using github actions instead of letting cloudflare do it

Added terraform for managing DNS records so we have all previews etc going through our main domain

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -1,0 +1,68 @@
+name: Docs deploy
+on:
+  workflow_run:
+    workflows: ["Docs build"]
+    types:
+      - completed
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    outputs:
+      parameters: ${{ steps.parameters.outputs.result }}
+    steps:
+      - if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+        run: echo 'The triggering workflow failed' && exit 1
+      
+      - name: Determine deploy parameters
+        id: parameters
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const eventType = context.payload.workflow_run.event;
+            const isFork = context.payload.workflow_run.repository.fork;
+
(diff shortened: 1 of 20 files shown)


.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,159 @@
...
+      - name: Publish to Cloudflare Pages
+        uses: cloudflare/pages-action@v1
+        with:
+          apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN_PAGES_UPLOAD }}
+          accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
+          projectName: "immich-app"
+          workingDirectory: "docs"
+          directory: "build"
+          branch: ${{ steps.parameters.outputs.name }}
+          wranglerVersion: '3'
```
bo0tzz, line 148: Should we publish before we update the domains, so we don't end up in a situation where the publish fails and we're left with the main docs pointing at an empty deployment?

.github/workflows/docs-deploy.yml
```diff
@@ -0,0 +1,159 @@
...
+
+      - name: Comment
+        uses: actions-cool/maintain-one-comment@v3
+        if: ${{ steps.parameters.outputs.event == 'pr' }}
+        with:
+          number: ${{ github.event.workflow_run.pull_requests[0].number }}
+          token: ${{ secrets.GITHUB_TOKEN }}
```
bo0tzz, line 155: We shouldn't need to set this explicitly, the action picks it up by default.

bo0tzz (commented):
Cool!
Since I don't know terraform I'll just have to trust you that it all works ;)

.github/workflows/docs-deploy.yml
zackpollard: Yea good shout

bo0tzz (approved):

PR #9880, opened 2024-05-30
fix(web): set album description textarea height correctly

### LINKED ISSUE #9879: Web: Album description grows too tall when navigating to an album via an asset
### The bug

The album description field grows too large when navigating to the album that an asset is part of when navigating via the asset viewer.
If you refresh the album page afterwards, the textarea is basically the correct (minimal height).

https://github.com/immich-app/immich/assets/5608270/6d795e63-5a9d-4555-b704-b611ae990bb9

Windows 10 22H2 (19045.4291)

### Version of Immich Server

v1.105.1

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1. Add an asset to an album
2. Navigate to Photos
3. Click on the asset that's is now in the album from step 1
4. Open the Info panel if it's not already open
5. Click on the album that the asset was added to under "Appears in"
```

### Relevant log output

_No response_

### Additional information

_No response_

Snowknight26:
Fixes #9879.

Changes:
* Applied the same approach #9765 implemented to the album description textarea
* Deduplicated the detail panel description and the album description sections into a single component
* Set the textarea value correctly since `value` is not a valid textarea property

### DIFF
--- a/web/src/lib/components/shared-components/autogrow-textarea.svelte
+++ b/web/src/lib/components/shared-components/autogrow-textarea.svelte
@@ -4,10 +4,10 @@
   import { shortcut } from '$lib/actions/shortcut';
   import { tick } from 'svelte';
 
-  export let content: string;
-  export let elementClass: string;
-  export let onContentUpdate: (newContent: string) => void;
-  export let placeholder: string;
+  export let content: string = '';
+  export let elementClass: string = '';
+  export let onContentUpdate: (newContent: string) => void = () => null;
+  export let placeholder: string = '';
 
   let textarea: HTMLTextAreaElement;
   $: newContent = content;
--- a/web/src/lib/components/shared-components/autogrow-textarea.svelte
+++ b/web/src/lib/components/shared-components/autogrow-textarea.svelte
@@ -35,5 +35,6 @@
   use:shortcut={{
     shortcut: { key: 'Enter', ctrl: true },
     onShortcut: (e) => e.currentTarget.blur(),
-  }}>{content}</textarea
+  }}
+  {...$$restProps}>{content}</textarea
 >
--- a/web/src/lib/components/shared-components/autogrow-textarea.svelte
+++ b/web/src/lib/components/shared-components/autogrow-textarea.svelte
@@ -35,5 +35,5 @@
     shortcut: { key: 'Enter', ctrl: true },
     onShortcut: (e) => e.currentTarget.blur(),
   }}
-  {...$$restProps}>{content}</textarea
+  data-testid="autogrow-textarea">{content}</textarea
 >
--- a/web/src/lib/components/shared-components/autogrow-textarea.svelte
+++ b/web/src/lib/components/shared-components/autogrow-textarea.svelte
@@ -1,0 +1,39 @@
+<script lang="ts">
+  import { autoGrowHeight } from '$lib/actions/autogrow';
+  import { clickOutside } from '$lib/actions/click-outside';
+  import { shortcut } from '$lib/actions/shortcut';
+  import { tick } from 'svelte';
+
+  export let content: string;
+  export let elementClass: string;
+  export let onContentUpdate: (newContent) => void;
+  export let placeholder: string;
+
+  let textarea: HTMLTextAreaElement;
+  $: newContent = content;
+
+  $: if (textarea) {
+    newContent;
+    void tick().then(() => autoGrowHeight(textarea));
+  }
+
+  const updateContent = () => {
+    if (content === newContent) {
+      return;
+    }
+    onContentUpdate(newContent);
+  };
+</script>
+
+<textarea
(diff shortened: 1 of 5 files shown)


jrasm91 (commented):
Can you add some tests for this functionality?

Snowknight26:
What kind of tests are you looking for? Unit? Integration/e2e test?

I couldn't find any contributing guidelines on tests so I'm not sure what's expected, but will do my best to add what's needed.

If it matters, the existing `e2e/src/web/specs/asset-viewer/detail-panel.e2e-spec.ts` regression tests at least some changes to `detail-panel.svelte` (and in turn, `detail-panel-description.svelte`).

jrasm91:
There have just been a lot of bugs with the description field not working properly. It should change when you navigate to the next asset and it should resize as well. A unit test would probably be fine. I think there are a few examples of this.

Snowknight26:
Admittedly I don't know how to test the resizing functionality (and admittedly, should resizing be tested, or just that the auto-grow function is called?) or navigating.. but I wonder if what's been added is adequate.

jrasm91 (approved):
LGTM, thanks!

web/src/lib/components/shared-components/autogrow-textarea.svelte
```diff
@@ -0,0 +1,40 @@
+<script lang="ts">
+  import { autoGrowHeight } from '$lib/actions/autogrow';
+  import { clickOutside } from '$lib/actions/click-outside';
+  import { shortcut } from '$lib/actions/shortcut';
+  import { tick } from 'svelte';
+
+  export let content: string = '';
+  export let elementClass: string = '';
```
michelheusschen, line 8: This is more consistent with other components:

```suggestion
  let elementClass: string = '';
  export { elementClass as class };
```

web/src/lib/components/shared-components/autogrow-textarea.svelte
```diff
@@ -0,0 +1,40 @@
...
+    if (content === newContent) {
+      return;
+    }
+    onContentUpdate(newContent);
+  };
+</script>
+
+<textarea
+  bind:this={textarea}
+  class={elementClass}
```
michelheusschen, line 30: Some default styles should be applied here like `resize-none`

web/src/lib/components/shared-components/autogrow-textarea.svelte
```diff
@@ -0,0 +1,40 @@
...
+  on:focusout={updateContent}
+  on:input={(e) => (newContent = e.currentTarget.value)}
+  use:clickOutside={{ onOutclick: updateContent }}
```
michelheusschen, line 33: `use:clickOutside` seems redundant and sometimes causes updateContent to be called twice

web/src/lib/components/shared-components/autogrow-textarea.svelte
```diff
@@ -0,0 +1,40 @@
...
+  {placeholder}
+  use:shortcut={{
+    shortcut: { key: 'Enter', ctrl: true },
+    onShortcut: (e) => e.currentTarget.blur(),
+  }}
+  {...$$restProps}>{content}</textarea
```
michelheusschen, line 39: `data-testid` can be defined here and $$restProps wouldn't be needed

web/src/lib/components/shared-components/autogrow-textarea.svelte
Snowknight26: Good catch. I couldn't get a test to trigger the function being called twice even when firing a click event on document.body, but I see in testing in a browser that it did.

web/src/lib/components/shared-components/autogrow-textarea.svelte
Snowknight26: My thought behind doing it this was two-fold:
* in case two instances of the component were rendered at the same time, it would be beneficial to pass the `data-testid` in so you can explicitly target one over another in a test
* futureproofing in case you wanted to be able to pass other props in (`disabled`, `readonly`, etc.) without having to modify the component's props

Happy to go with your suggestion if we don't feel like these are scenarios we'd have to account for.

michelheusschen: Gotcha. The downside of using `$$restProps` is the lack of type safety unless the `$$Props` type is defined. That's why I usually try to avoid using it. Maybe we can just declare the test id as a prop instead?

michelheusschen (approved):

web/src/lib/components/shared-components/autogrow-textarea.svelte
michelheusschen: Awesome, thank you!

PR #9894, opened 2024-05-30
feat(server): enable exiftool largefilesupport

fvdnabee:
Considering LFS in exiftool is only expected to be problematic on uncommon, older systems ([ref](https://exiftool.org/forum/index.php?msg=85278)), let's enable it by default. [My previous PR](https://github.com/immich-app/immich/pull/7183) introduced a system config, but in light of the previous finding I don't think a system config is necessary anymore.


zackpollard (approved):

bo0tzz:
For future reference, you can just force-push and edit a PR, you don't have to open a new one. It's fine now though of course 🙂

aviv926:
Fix #4349

mertalev (approved):

PR #9929, opened 2024-06-01
fix(web): stop slideshow when pressing escape from asset viewer

### LINKED ISSUE #9904: Web: Exiting slideshow with escape key closes the asset viewer
### The bug

Instead of _just_ exiting the slideshow, pressing Escape while a slideshow is running also closes the asset viewer.

This differs from the behavior of the X ("Exit Slideshow") button in the top left of the slideshow bar, which _only_ exits the slideshow.

Windows 10 22H2 (19045.4291)

### Version of Immich Server

v1.105.1

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1. Navigate to Photos
2. Click a photo
3. Click the triple dots in the top right
4. Click Slideshow
5. Press Escape to exit fullscreen mode
6. Press Escape to exit the slideshow
```

### Relevant log output

_No response_

### Additional information

I imagine there's some logic missing in the SlideshowBar that handles the `escape` event. Instead of the AssetViewer component handling escape, the SlideshowBar should?

Snowknight26:
Fixes #9904.

### DIFF
--- a/web/src/lib/components/asset-viewer/asset-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/asset-viewer.svelte
@@ -273,10 +273,12 @@
   };
 
   const closeViewer = async () => {
-    $slideshowState = SlideshowState.StopSlideshow;
-    document.body.style.cursor = '';
-    dispatch('close');
-    await navigate({ targetRoute: 'current', assetId: null });
+    if($slideshowState === SlideshowState.None) {
+      dispatch('close');
+      await navigate({ targetRoute: 'current', assetId: null });
+    } else {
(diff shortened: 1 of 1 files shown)


Snowknight26:
Not sure there's an effective way to test this. Attempting to render the AssetViewer component requires mocking almost everything under the sun. Would extracting `closeViewer` into its own file and testing the two code paths (if, else) be overkill?

jrasm91 (approved):

PR #9932, opened 2024-06-01
chore(web): unique ID generation

ben-basten:
## Description

Added a unique ID generation utility, which removes the need to include IDs as props in many scenarios.

## How Has This Been Tested?

Tested modals, toggles, and comboboxes with a screen reader to make sure that the aria attributes are still announcing as expected. Also inspected the HTML to make sure that the generated IDs are unique.

### DIFF
--- a/web/src/lib/stores/unique-id.store.ts
+++ b/web/src/lib/stores/unique-id.store.ts
@@ -1,0 +1,15 @@
+import { get, writable } from 'svelte/store';
+
+function createIdStore() {
+  const { subscribe, update } = writable(0);
+
+  return {
+    subscribe,
+    generateId: () => {
+      update((value) => value + 1);
+      return `id-${get(uniqueIdStore)}`;
+    },
+  };
+}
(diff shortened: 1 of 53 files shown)


alextran1502 (approved):

web/src/lib/stores/unique-id.store.ts
```diff
@@ -0,0 +1,16 @@
...
+    subscribe,
+    update,
+    generateId: () => {
+      update((value) => value + 1);
+      return `id-${get(uniqueIdStore)}`;
+    },
+  };
+}
+
+export const uniqueIdStore = createIdStore();
```
jrasm91, line 16: This doesn't need to be a store and it doesn't need to be subscribed to.

It could be a pain exported function with local state. Or, a generator.

ben-basten: Good call, that would be much simpler. Working on a fix.

PR #9941, opened 2024-06-02
Update command-line-interface.md

BoBBer446:
Add immich upload background runs for large inputs

### DIFF
--- a/docs/docs/features/command-line-interface.md
+++ b/docs/docs/features/command-line-interface.md
@@ -166,6 +166,83 @@ By default, hidden files are skipped. If you want to include hidden files, use t
 immich upload --include-hidden --recursive directory/
 ```
 
+### run the `immich upload` command in the background
+To run the `immich upload` command in the background. You can use the following approach. This is useful for large imports, as it allows the process to continue even if the SSH session is closed. Here’s how you can start the `immich upload` command in the background:
+
+```sh
+nohup immich upload /path/to/your/files/ -r &
+```
+
+This command starts the upload process in the background, and `nohup` ensures that it continues running even if the terminal is closed. The output will be directed to `nohup.out` by default, but you can still monitor the logs using the `tail` command:
+
+```sh
+tail -f nohup.out
+```
+
+This command allows you to see the latest lines of the log file and updates in real-time, enabling you to monitor the progress of the upload.
+
+To list background jobs, use the following command:
+
+```sh
+jobs
+```
+
+This will display all current background jobs and their statuses.
+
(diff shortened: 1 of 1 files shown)


bo0tzz:
I'm not quite sure how I feel about this; if we want to document how to run the cli in the background at all, maybe describing something like tmux might be better?

BoBBer446:
That's right. It could also be a possibility.
I don't know if it's really better or worse. But I think it's good if the user knows how and that he can do this in the background.

I import over 200,000 images, which takes a while and I don't want to leave the computer on for that long (around 14 hours). Besides, the network could break.

By the way, this is my first request, I have no idea how something like this works :D

bo0tzz:
Sending the CLI process to the background doesn't do anything for those cases

BoBBer446:
Does that mean this way would be better?

1. **Start a new `tmux` session:**
   ```sh
   tmux new -s immich_upload
   ```

2. **Run the `immich upload` command within the `tmux` session:**
   ```sh
   immich upload /path/to/your/files/ -r
   ```

3. **Detach the `tmux` session:**
   Press `Ctrl+b` followed by `d`.
   _This detaches the session, allowing the process to continue running in the background._

4. **Reattach to the `tmux` session:**
   ```sh
   tmux attach -t immich_upload
   ```

The aim should be to be able to switch off and shut down the computer while the import is running

bo0tzz:
> The aim should be to be able to switch off and shut down the computer while the import is running

None of these approaches will be able to do that, and I'm not aware of any way that will.

BoBBer446:
Are we talking at cross purposes?
If I move the process with nohub or tmux to the background or a new temux, then I can shut down the PC, the server continues to run.

Later, I can go back to my the PC and connect to the server and either move the nohub process to the foreground or look at the tmux. So while I have switched off the PC, the import has continued to run.

If I only do “immich upload” in the terminal and close the PC, the process is also terminated.

So tmux and nohub are able to continue a long upload process in the background. Or am I making a mistake?

bo0tzz:
It wasn't clear to me that you were talking about running the cli over ssh. I had understood it as wanting to shut down the system that the CLI was running on. If it's over ssh then tmux is the way to go - in fact, tmux is really nice whenever you're doing anything over ssh cause stuff like connection failures will not mess up your session.

jrasm91:
Yeah, I don't think we need any detailed guides here. Running commands in the background is a common, well-known problem. It is worth mentioning via a note, but we can just refer users to other tools/docs for the details, IMO.

PR #9945, opened 2024-06-02
chore(web): simpler unique ID generation

ben-basten:
## Description

Simpler way of generating unique IDs, without needing to use Svelte stores. Revised according to the feedback from #9932.


danieldietzler (approved):
LGTM :)

jrasm91 (approved):
Ty

PR #9954, opened 2024-06-03
feat(web): allow ctrl-click / cmd-click on photos

GeorgeShao:
https://github.com/immich-app/immich/assets/31512688/4708e937-482c-4875-b59c-e194050202d8


web/src/lib/components/assets/thumbnail/thumbnail.svelte
```diff
@@ -85,11 +88,11 @@
 </script>
 
 <IntersectionObserver once={false} on:intersected let:intersecting>
-  <!-- svelte-ignore a11y-no-noninteractive-tabindex -->
-  <div
+  <a
+    href={isMultiSelectState ? undefined : `${currentUrlReplaceAssetId(asset.id)}`}
```
Snowknight26, line 92: Is the tertiary needed? Can the href attribute always be set to `currentUrlReplaceAssetId(asset.id)` since the click event handler will now prevent the default when isMultiSelectState is true?

GeorgeShao: You're correct. I've just pushed a change to correct this. Thanks!

danieldietzler (approved):

Snowknight26:
If `disabled` is passed to the Thumbnails component, it seems like clicking still navigates you to the asset, which doesn't appear to be expected behavior.
You can reproduce this by navigating to an album that an asset is a part of, then attempting to more photos to the album. The assets already in the album should not be clickable.

Snowknight26:
Thinking about this more, I feel like `clickable` should be removed in favor of `disabled` (since really they're the same), and then the <a> attributes like `class:hover:cursor-pointer` and `role` should be removed, and `tabindex` should always be set. Not 100% sure if this is desirable but I think that would remove some duplication and would make the code more inline with what we expect from an anchor tag.

GeorgeShao:
> Thinking about this more, I feel like `clickable` should be removed in favor of `disabled` (since really they're the same), and then the attributes like `class:hover:cursor-pointer` and `role` should be removed, and `tabindex` should always be set. Not 100% sure if this is desirable but I think that would remove some duplication and would make the code more inline with what we expect from an anchor tag.

@Snowknight26 Yeah that makes sense to me. I've just pushed some commits implementing what you've described in your 2 comments.

web/src/lib/components/assets/thumbnail/thumbnail.svelte
```diff
@@ -85,20 +87,18 @@
 </script>
 
 <IntersectionObserver once={false} on:intersected let:intersecting>
-  <!-- svelte-ignore a11y-no-noninteractive-tabindex -->
-  <div
+  <a
+    href={disabled ? '' : currentUrlReplaceAssetId(asset.id)}
```
Snowknight26, line 91: Only other change I would suggest is removing this ternary once again. With the `href` set to an empty string, the href becomes the current URL. Because `onIconClickedHandler` still handles the disabled state by stopping the default event behavior, we can simply remove it to allow Ctrl/Cmd+click or middle mouse click to take you to the proper URL.

Snowknight26:
Oh and for some reason the permissions on `thumbnail.svelte` changed from 100644 to 100755. Not sure if that was intended.

GeorgeShao:
> Oh and for some reason the permissions on `thumbnail.svelte` changed from 100644 to 100755. Not sure if that was intended.

@Snowknight26  nice catch, thanks, changes pushed. I was messing around with file permissions because my setup had some issues. I forgot to change it back once I got my system working.

alextran1502:
Thank you for the PR; the current implementation breaks some other functionalities.

- Asset selection for an album, you now need to click on the select icon instead of the whole photo to make a selection 
image

- Similarly, it causes the same effect when selecting a person feature photos.

ayykamp:
I would also expect CTRL/CMD + click to start selection. Maybe middle mouse button is a better idea for opening in a new tab?

Snowknight26:
> I would also expect CTRL/CMD + click to start selection. Maybe middle mouse button is a better idea for opening in a new tab?

Probably don't want to override standard browser functionality in my opinion.

pyorot:
> > I would also expect CTRL/CMD + click to start selection. Maybe middle mouse button is a better idea for opening in a new tab?
> 
> Probably don't want to override standard browser functionality in my opinion.

Incidentally, Ctrl+Click causes Google Photos to open a new tab and Google Drive to multi-select. Which one makes sense depends on what clicking does by itself – whether it opens the file (Photos) or selects it (Drive).

alextran1502:
Hello, thank you for the fix. Another broken view is the search result view. Once in multi-selection mode, clicking on the asset will open the asset viewer instead of selecting it.

GeorgeShao:
> Hello, thank you for the fix. Another broken view is the search result view. Once in multi-selection mode, clicking on the asset will open the asset viewer instead of selecting it.

@alextran1502 pushed a fix. I also noticed another related bug with the stack slideshow at the bottom left of the screen, so I fixed that too

Screenshot 2024-06-07 at 4 03 42 PM

web/src/lib/components/assets/thumbnail/thumbnail.svelte
```diff
@@ -98,11 +101,11 @@
     on:mouseenter={onMouseEnter}
     on:mouseleave={onMouseLeave}
     tabindex={0}
-    on:click={isMultiSelectState ? onIconClickedHandler : undefined}
+    on:click={isMultiSelectState ? onIconClickedHandler : isStackSlideshow ? thumbnailClickedHandler : undefined}
```
alextran1502, line 104: This is very hard to read and understand. Can you help refactor this?

web/src/lib/components/assets/thumbnail/thumbnail.svelte
```diff
@@ -98,11 +101,11 @@
...
     use:shortcut={{ shortcut: { key: 'Enter' }, onShortcut: thumbnailClickedHandler }}
   >
     {#if intersecting}
-      <div class="absolute z-20 h-full w-full {className}">
+      <div class="absolute z-20 h-[{height}px] w-[{width}px] {className}">
```
alextran1502, line 108: Why do we need to change this?

web/src/lib/components/assets/thumbnail/thumbnail.svelte
```diff
@@ -128,7 +131,7 @@
       </div>
 
       <div
-        class="absolute h-full w-full select-none bg-gray-100 transition-transform dark:bg-immich-dark-gray"
+        class="absolute h-full w-full select-none bg-transparent transition-transform dark:bg-immich-dark-gray"
```
alextran1502, line 134: Why do we need to change this?

web/src/lib/components/assets/thumbnail/thumbnail.svelte
GeorgeShao: since we changed this from a <div> with `display: block` to an <a> with `display: flex`, leaving h-full and w-full here will overflow the actual image itself
Screenshot 2024-06-07 at 4 21 55 PM

web/src/lib/components/assets/thumbnail/thumbnail.svelte
GeorgeShao: also related to the change from a `div` with `display: block` to an `a` with `display: flex`

Screenshot 2024-06-07 at 4 23 34 PM

GeorgeShao: previously this wasn't visible, so it didn't matter what the bg colour value was

web/src/lib/components/assets/thumbnail/thumbnail.svelte
GeorgeShao: done

web/src/lib/components/assets/thumbnail/thumbnail.svelte
Snowknight26: I haven't fully tested it but if we don't need this one then we also don't need `dark:bg-immich-dark-gray` either.

web/src/lib/components/assets/thumbnail/thumbnail.svelte
Snowknight26: Not sure this'll work because height and width aren't integers, so the classes will become something like `h-[281.298374982px]`, which isn't something tailwind supports. Plus, these classes are only for the wrapper surrounding the selecting button element. I wasn't able to reproduce the overflow with `h-full, w-full`. Where do you see it?

GeorgeShao: do you have the most recent changes pulled? not sure why you aren't able to reproduce it, because I went back, edited only that one line and got this:

**h-full w-full**
Screenshot 2024-06-07 at 6 13 55 PM

**h-[{height}px] w-[{width}px]**
Screenshot 2024-06-07 at 6 14 12 PM

web/src/lib/components/assets/thumbnail/thumbnail.svelte
Snowknight26: Food for thought, but if component is being rendered in the context of a stack slideshow, should the thumbnail ever be a link (and therefore Ctrl+clickable)?

web/src/lib/components/assets/thumbnail/thumbnail.svelte
GeorgeShao: nice catch, I tested it and leaving `dark:bg-immich-dark-gray` results in this:

Screenshot 2024-06-07 at 6 18 29 PM

just pushed a commit removing this

web/src/lib/components/assets/thumbnail/thumbnail.svelte
Snowknight26: I do, I edited my comment to state I didn't realize this was only for the stack slideshow.
But the problem remains with the CSS class names when not in the context of a stack slideshow when using variables.

GeorgeShao: unfortunately looks like h-auto w-auto resizes the white outline to be so much smaller that it only appears in the top left of the image

Screenshot 2024-06-07 at 6 23 36 PM

GeorgeShao: ok I just replaced `h-[{height}px] w-[{width}px]` with `style:width="{width}px" style:height="{height}px"` which works as expected

Screenshot 2024-06-07 at 6 35 06 PM

PR #9990, opened 2024-06-05
feat(web): add an empty placeholder to the explore page

Snowknight26:
No idea what the wording should be so suggestions are greatly appreciated.

image


mertalev:
The second sentence alone is probably fine.

jrasm91 (approved):

danieldietzler (approved):

PR #10008, opened 2024-06-06
fix(web): clear locale setting

michelheusschen:
Clearing the locale setting closes the combobox. Fixed by not setting `$locale` to `undefined`


jrasm91 (approved):

jrasm91:
Should we do the same for language?

michelheusschen:
The problem isn't really that the combobox closes, but that the entire option disappears. That means it's not an issue for language

PR #10013, opened 2024-06-06
fix(server): closed connections

jrasm91:
- `await _sendFile` so the catch blocks works correctly
- don't write to `res` if headers have been sent


zackpollard (approved):

danieldietzler (approved):

Saberian1992 (commented):
Hamed

PR #10017, opened 2024-06-07
feat(web): keyboard accessible context menus

ben-basten:
## Description

Enable arrow key navigation of dropdown menus. Accessibility features are based on this WCAG "actions menu" pattern:

<https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-actions-active-descendant/>

This also simplifies implementing a `ContextMenu` dropdown, because the implementing component no longer needs to keep track of the dropdown visibility state or position. All of that logic is handled internally within the component.

Also in this change:

- rename `AssetSelectContextMenu` to `ButtonContextMenu`, to make the naming more accurately reflect what it does
- albums list dropdown menu now uses the i18n translations
- darken the active option color in the context menu to make it easier to see
- context menu is scrollable on short screen sizes
- maintain context menu position on screen resize
- the right click context menu on the albums list page supports arrow key navigation

## How Has This Been Tested?

- [x] VoiceOver screen reader testing
    - works with Firefox, Chrome, and Safari on MacOS
    - example test scenarios: navigating with arrow keys, hovering over items with a mouse, clicking options with enter/spacebar/mouse, clicking away to dismiss, using the tab key to dismiss, using the escape key to dismiss
- [ ] NVDA testing on Windows
    - I do not have a Windows environment, so if anyone has the tools to try this it would be very helpful :) 

Verified that options in affected context menus behave as expected:

- [x] Asset viewer options
- [x] External library admin settings
- [x] Albums card, in the albums list
- [x] Albums detail, in the control bar
- [x] Albums detail, when a photo is selected
- [x] Album share options modal, to change permissions and remove users
- [x] People card, in the people list
- [x] Memory viewer, when a photo is selected
- [x] Archive, when a photo is selected
- [x] Favorites, when a photo is selected
- [x] Partners, on the album card
- [x] Person detail, in the control bar
- [x] Person detail, when a photo is selected
- [x] Photos, when a photo is selected
- [x] Search, when a photo is selected

## Screenshots

Example dropdown menu, with the darker selected color.

<details><summary>Expand for screenshot</summary>
<p>

dropdown

</p>
</details>


web/src/lib/components/shared-components/context-menu/button-context-menu.svelte
```diff
@@ -0,0 +1,124 @@
...
+      ariaHasPopup={true}
+      class={buttonClass}
+      color={buttonColor}
+      id={buttonId}
+      on:click={handleClick}
+      padding={buttonPadding}
+      size={buttonSize}
+    />
+  </div>
+  {#if usePortal}
```
ben-basten, line 97: I'm wondering if there's a better way to do this without duplicating code? This seems like a use case for Svelte 5 snippets.

web/src/lib/actions/list-navigation.ts
```diff
@@ -1,31 +1,109 @@
...
+   */
+  container: HTMLElement;
+  /**
+   * The id of the currently selected element.
+   */
+  selectedId: string | undefined;
+  /**
+   * A class name to add to the selected element, if any.
+   */
+  selectedClass?: string;
```
ben-basten, line 17: I don't like the `selectedClass` modification happening in the action here, it feels hacky. I will move this over to use Svelte contexts for the parent `ButtonContextMenu` to communicate with the `<MenuOption>` in its slot and select the active one.

web/src/lib/components/elements/buttons/circle-icon-button.svelte
```diff
@@ -1,19 +1,32 @@
...
 <script lang="ts">
   import Icon from '$lib/components/elements/icon.svelte';
-  type Color = 'transparent' | 'light' | 'dark' | 'gray' | 'primary' | 'opaque';
 
   export let type: 'button' | 'submit' | 'reset' = 'button';
   export let icon: string;
   export let color: Color = 'transparent';
   export let title: string;
+  /**
+   * The padding of the button, used by the `p-{padding}` Tailwind CSS class.
```
jrasm91, line 13: From what I understand, tailwind scans these source files and extracts used class names, which it includes in the final build of the app. Using a dynamic class name like this doesn't work. I mean it would be possible to refer to a class that won't actually be used. It probably happens to work because the same classes are already used and detected, but I think this is an anti-pattern.

web/src/lib/components/elements/buttons/circle-icon-button.svelte
```diff
@@ -1,19 +1,32 @@
...
   export let hideMobile = false;
   export let buttonSize: string | undefined = undefined;
   /**
    * viewBox attribute for the SVG icon.
    */
   export let viewBox: string | undefined = undefined;
+  export let id: string | undefined = undefined;
+  export let ariaHasPopup: boolean | undefined = undefined;
+  export let ariaExpanded: boolean | undefined = undefined;
+  export let ariaControls: string | undefined = undefined;
```
jrasm91, line 29: I think there is an export HTML*Type field you can use to get the actual type for these properties.

web/src/lib/components/faces-page/people-card.svelte
```diff
@@ -73,22 +66,16 @@
...
 
-  <div class="absolute right-2 top-2" class:hidden={!showVerticalDots}>
-    <CircleIconButton
-      color="opaque"
+  <div class="absolute top-2 right-2">
+    <ButtonContextMenu
+      buttonClass="icon-white-drop-shadow focus:opacity-100 {showVerticalDots ? 'opacity-100' : 'opacity-0'}"
+      buttonColor="opaque"
+      buttonPadding="2"
+      buttonSize="20"
```
jrasm91, line 74: I think these props should apply to the button by default, so the `button` prefix looks redundant imo.

web/src/lib/components/faces-page/people-card.svelte
```diff
@@ -73,22 +66,16 @@
...
-      class="icon-white-drop-shadow"
-      on:click={showMenu}
-    />
-  </div>
-</div>
-
-{#if showContextMenu}
-  <Portal target="body">
-    <ContextMenu {...contextMenuPosition} onClose={() => onMenuExit()}>
+      usePortal
```
jrasm91, line 77: Is this actually ever required? It seems like the context menu should always show above/on top of the clicked button.

web/src/lib/components/photos-page/actions/asset-job-actions.svelte
```diff
@@ -32,8 +32,13 @@
...
 </script>
 
-{#each jobs as job}
+{#each jobs as job, index}
   {#if isAllVideos || job !== AssetJobName.TranscodeVideo}
-    <MenuOption text={getAssetJobName(job)} icon={getAssetJobIcon(job)} on:click={() => handleRunJob(job)} />
+    <MenuOption
+      text={getAssetJobName(job)}
+      icon={getAssetJobIcon(job)}
+      topBorder={index === 0}
```
jrasm91, line 40: I think a dedicated <MenuDivider> child component would be better than a class like this.

web/src/lib/actions/list-navigation.ts
jrasm91: I'm not a huge fan of reusing code/functionality like this. I'd much prefer encapsulating this type of behavior in components themselves.

jrasm91 (commented):
It is a lot of changes. It seems like the list enhancements could have been separated out into a different change set.

ben-basten:
> It is a lot of changes. It seems like the list enhancements could have been separated out into a different change set.

Thank you for the time and feedback, @jrasm91! 

I agree, the list changes expanded the scope of the PR. I'll revert those and put them into their own Svelte action, which I think would still be helpful to prevent duplicated code between the `ButtonContextMenu` and `RightClickContextMenu`.

web/src/lib/components/faces-page/people-card.svelte
ben-basten: Yes - I added the `Portal` to get the People card and External Library settings dropdown working as expected. They used a portal before this change and the dropdown menu does not render properly without it in both of these scenarios.

web/src/lib/actions/list-navigation.ts
ben-basten: To fix this, I created a new Svelte Action specifically for context menus, so the logic can be shared between the `ButtonContextMenu` and `RightClickContextMenu`. I also changed the mechanism for selecting a menu option to be a Svelte store, because more of a pub/sub model is needed for letting a `MenuOption` know that it has been selected.

web/src/lib/components/elements/buttons/circle-icon-button.svelte
ben-basten: I will take a look at how the tailwind classes are generated in this scenario. If it is an anti-pattern, I'd propose changing that in a following PR to prevent increasing the scope of this change.

web/src/lib/components/elements/buttons/circle-icon-button.svelte
ben-basten: In my opinion, I find the syntax for automatically adding the HTML props a little confusing/cumbersome, especially because this `CircleIconButton` component has so many props. Here's a similar example:

https://github.com/immich-app/immich/blob/08426e14bcdb3bd8bbad5d778df7b78cd4da5857/web/src/lib/components/shared-components/immich-logo.svelte#L13-L19

I'm wondering if this would be an easier change to do once we have the new Svelte 5 prop typing? Happy to add the automatic HTML props now if that's preferred.

https://svelte-5-preview.vercel.app/docs/runes#$props

web/src/lib/components/faces-page/people-card.svelte
ben-basten: I found a way to remove the need for a portal - set the positioning of the `<ContextMenu>` as `fixed` rather than `absolute`.

web/src/lib/components/elements/buttons/circle-icon-button.svelte
ben-basten: I created a derived variable, so the class name ho longer has a class name with a dynamic number like that.

ben-basten:
This is ready for review! I've responded to open feedback and closed the remaining tasks on my list.

alextran1502 (approved):
Good work! Thanks Ben!

waclaw66:
@ben-basten this PR introduces a bug that makes three dot button still visible after menu is closed.

https://github.com/immich-app/immich/assets/15554561/a12ff9cd-70d0-4234-b480-4557eb2e11d2

ben-basten:
@waclaw66 This is working as designed - focus is on the 3 dot button after the context menu closes, so it stays visible.

New in this PR is that focus returns to the button that triggered the context menu when it is closed. This allows keyboard users to regain their previous focus position after interacting with a context menu. The same behavior occurs when a mouse user clicks away from the menu.

Test scenario to check this:

1. Click the 3 dots to open the album context menu
2. Click away to close the menu
3. Press the enter key or spacebar. The context menu should open again, because the button has focus.

I thought this made sense since buttons/links on the screen aren't interactive when clicking away from the context menu. An option we have is to consider disabling this refocus behavior for clicking away.

waclaw66:
@ben-basten That make sense, thank for detailed explanation as usual.

ben-basten:
> Loosing focus and forwarding click event should be consolidated though. Some context menus (asset viewer, face) allow to click outside and forward click event to the target control (e.g. left panel), album menu doesn't allow that. Please check a compare the behaviour.

Yeah, the albums page is unique because it has a context menu that appears when you right click on the album cover. When I right click on a website on my Mac, the default MacOS context menu doesn't let me click anything behind it until the context menu is dismissed. So, it makes sense to imitate that behavior with our custom right click context menu.

However, this behavior doesn't make as much sense when the 3 dots button on the top right of the album cover is clicked, because we already have an established pattern for how that should work everywhere else.

The fix here would be to move the 3 dot menu from using the `RightClickContextMenu` component to the `ButtonContextMenu` component. I determined this change to be out of scope in this PR because this is how it was previously built, but I agree that it would be a better experience for that button to have the same behavior as other menus. For example, a benefit of migrating would include automatically selecting the first item in the menu when it is opened using the enter key or space bar.

Feel free to pitch out a PR for this, I'm not sure when I'll have time to make this change yet. This can certainly be improved, thanks for calling it out!

PR #10021, opened 2024-06-07
fix(web): translations

michelheusschen:
- Fixed translations using an incorrect ID
- Translation for `oauth` was missing and `interval.hours` had a syntax error
- Added unit test to find missing messages

### DIFF
--- a/web/src/lib/components/admin-page/jobs/jobs-panel.svelte
+++ b/web/src/lib/components/admin-page/jobs/jobs-panel.svelte
@@ -61,12 +61,12 @@
     [JobName.ThumbnailGeneration]: {
       icon: mdiFileJpgBox,
       title: getJobName(JobName.ThumbnailGeneration),
-      subtitle: $t('thumbnail_generation_job_description'),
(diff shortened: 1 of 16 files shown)


web/src/lib/components/admin-page/jobs/jobs-panel.svelte
```diff
@@ -61,12 +61,12 @@
...
-      subtitle: $t('thumbnail_generation_job_description'),
+      subtitle: $t('admin.thumbnail_generation_job_description'),
     },
     [JobName.MetadataExtraction]: {
       icon: mdiTable,
       title: getJobName(JobName.MetadataExtraction),
-      subtitle: $t('metadata_extraction_job_description'),
+      subtitle: $t('admin.metadata_extraction_job_description'),
     },
     [JobName.Library]: {
```
zackpollard, line 71: Does this jobname not require admin.? Should these not all be prefixed with that? @danieldietzler ?

zackpollard (approved):

PR #10028, opened 2024-06-07
fix: prevent trashing of trashed assets

roschaefer:
Motivation
----------
This will improve user experience ~by hiding a pointless action~ by changing "move to trash" into "delete permanently" if the asset is already in the trash.

~You can not trash a trashed asset again. It won't get any trashier than it already is.~

How to test
-----------
1. Visit route `/trash`
2. Click on an asset
3. Press "Delete" on your keyboard
4. You will see the confirmation dialog to permanently delete the asset
5. Click on the trash button in the top right
6. Same as 4)

### DIFF
--- a/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.spec.ts
+++ b/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.spec.ts
@@ -3,7 +3,6 @@ import { assetFactory } from '@test-data/factories/asset-factory';
 import { userAdminFactory } from '@test-data/factories/user-factory';
 import '@testing-library/jest-dom';
 import { render } from '@testing-library/svelte';
-import { init } from 'svelte-i18n';
 import AssetViewerNavBar from './asset-viewer-nav-bar.svelte';
 
 describe('AssetViewerNavBar component', () => {
@@ -14,11 +13,9 @@ describe('AssetViewerNavBar component', () => {
     showDownloadButton: false,
     showMotionPlayButton: false,
     showShareButton: false,
+    onZoomImage: () => {},
+    onCopyImage: () => {},
   };
-
-  beforeAll(async () => {
-    await init({ fallbackLocale: 'en-US' });
-  });
 
   afterEach(() => {
     vi.resetAllMocks();
@@ -32,24 +29,13 @@ describe('AssetViewerNavBar component', () => {
   });
 
   describe('if the current user owns the asset', () => {
-    it('shows trash button', () => {
+    it('shows delete button', () => {
       const ownerId = 'id-of-the-user';
       const user = userAdminFactory.build({ id: ownerId });
       const asset = assetFactory.build({ ownerId, isTrashed: false });
       userStore.set(user);
       const { getByTitle } = render(AssetViewerNavBar, { asset, ...additionalProps });
       expect(getByTitle('delete')).toBeInTheDocument();
-    });
-
-    describe('but if the asset is already trashed', () => {
-      it('hides trash button', () => {
-        const ownerId = 'id-of-the-user';
-        const user = userAdminFactory.build({ id: ownerId });
-        const asset = assetFactory.build({ ownerId, isTrashed: true });
-        userStore.set(user);
-        const { queryByTitle } = render(AssetViewerNavBar, { asset, ...additionalProps });
-        expect(queryByTitle('delete')).toBeNull();
-      });
     });
   });
 });
--- a/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.spec.ts
+++ b/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.spec.ts
@@ -1,0 +1,55 @@
+import { resetSavedUser, user as userStore } from '$lib/stores/user.store';
+import { assetFactory } from '@test-data/factories/asset-factory';
+import { userAdminFactory } from '@test-data/factories/user-factory';
+import '@testing-library/jest-dom';
+import { render } from '@testing-library/svelte';
+import { init } from 'svelte-i18n';
+import AssetViewerNavBar from './asset-viewer-nav-bar.svelte';
+
+describe('AssetViewerNavBar component', () => {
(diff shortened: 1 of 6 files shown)


web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -538,21 +540,24 @@
...
     { shortcut: { key: 'Delete', shift: true }, onShortcut: () => trashOrDelete(true) },
     { shortcut: { key: 'Escape' }, onShortcut: closeViewer },
     { shortcut: { key: 'f' }, onShortcut: toggleFavorite },
     { shortcut: { key: 'i' }, onShortcut: toggleDetailPanel },
-  ]}
-/>
+  ];
+
+  if (!asset.isTrashed) {
+    shortcutOptions.push({ shortcut: { key: 'Delete' }, onShortcut: () => trashOrDelete(false) });
```
roschaefer, line 556: This one is not tested. I could - but I would refactor the callbacks to live outside the component, for easier testing. Shall I?

roschaefer:
So how can I add PR labels as a contributor?

web/src/lib/components/asset-viewer/asset-viewer-nav-bar.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    showZoomButton: false,
+    showDetailButton: false,
+    showDownloadButton: false,
+    showMotionPlayButton: false,
+    showShareButton: false,
+  };
+
+  beforeAll(async () => {
+    await init({ fallbackLocale: 'en-US' });
+  });
```
michelheusschen, line 21: No longer be needed after merging main

web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
```diff
@@ -190,12 +190,14 @@
     {/if}
 
     {#if isOwner}
-      <CircleIconButton
-        color="opaque"
-        icon={mdiDeleteOutline}
-        on:click={() => dispatch('delete')}
-        title={$t('delete')}
-      />
+      {#if !asset.isTrashed}
```
michelheusschen, line 193: I would still expect a delete button for a trashed asset to permanently delete it

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -538,21 +539,24 @@
...
-    { shortcut: { key: 'Delete' }, onShortcut: () => trashOrDelete(false) },
     { shortcut: { key: 'Delete', shift: true }, onShortcut: () => trashOrDelete(true) },
     { shortcut: { key: 'Escape' }, onShortcut: closeViewer },
     { shortcut: { key: 'f' }, onShortcut: toggleFavorite },
     { shortcut: { key: 'i' }, onShortcut: toggleDetailPanel },
-  ]}
-/>
+  ];
+
+  if (!asset.isTrashed) {
```
michelheusschen, line 554: `asset` is reactive, but this statement is not. This has the potential to get out of sync.

web/src/lib/components/asset-viewer/asset-viewer-nav-bar.spec.ts
roschaefer: Thanks! Good to know.

web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
roschaefer: Yes, that's a good idea. I updated the code.

web/src/lib/components/asset-viewer/asset-viewer.svelte
roschaefer: Understandable. I updated the code. Not sure how I can test it or ensure that it is reactive indeed.

roschaefer:
@michelheusschen thanks for the review. I hope I could address all the requested changes?

web/src/lib/components/asset-viewer/asset-viewer.svelte
michelheusschen: Thank you for the changes. Looks all good to me, apart from the `Delete` shortcut not working for trashed assets. I would suggest to keep the shortcuts as they were and only make the following change:

```diff
- { shortcut: { key: 'Delete' }, onShortcut: () => trashOrDelete(false) },
+ { shortcut: { key: 'Delete' }, onShortcut: () => trashOrDelete(asset.isTrashed) },
```

web/src/lib/components/asset-viewer/delete-button.spec.ts
```diff
@@ -0,0 +1,34 @@
+import { type AssetResponseDto } from '@immich/sdk';
+
+import { assetFactory } from '@test-data/factories/asset-factory';
+import '@testing-library/jest-dom';
+import { render } from '@testing-library/svelte';
+import DeleteButton from './delete-button.svelte';
+
+export let asset: AssetResponseDto;
```
michelheusschen, line 8: No need to export this

roschaefer: Are you sure? I want to declare a component prop. I get `TypeError: ctx[1] is undefined` if I remove the `export`.

michelheusschen: This is in the test file, not the component itself

web/src/lib/components/asset-viewer/asset-viewer.svelte
roschaefer: Alright! I followed these suggestions. Initially I thought it might be better to disable a shortcut instead of changing the behaviour. I was afraid of accidental deletions because of muscle memory. However, as we have a confirmation dialog, I think it's fine to assign both `Delete` and `Shift+Delete` to "permanently delete" in this situation.

michelheusschen (approved):

alextran1502 (approved):

PR #10064, opened 2024-06-09
ci: automatically apply PR labels

roschaefer:
Motivation
----------
For me as a new contributor it is frustrating to submit a PR and it will always fail. Even worse: I have to wait for another contributor with more power to assign the label for me.

This will improve developer experience, as some of the labels can be assigned automatically based on changed files.

How to test
-----------
1. Merge this PR
2. Submit a couple of PRs with changes in the respective directories
3. Labels should be automatically applied
4. "Enforce PR labels" github workflow will re-run when "Pull Request Labeler" completes

You can check out my fork: https://github.com/roschaefer/immich/pull/4

image

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,23 @@
+cli:
+- changed-files:
+  - any-glob-to-any-file: cli/**
+
+documentation:
+- changed-files:
+  - any-glob-to-any-file: docs/**
+
+🖥️web:
+- changed-files:
+  - any-glob-to-any-file: web/**
+
+📱mobile:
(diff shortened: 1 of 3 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,23 @@
+cli:
```
roschaefer, line 1: Some labels have emojis, some don't?

Release notes use a different spelling, too: https://github.com/immich-app/immich/releases

bo0tzz (approved):
I'd been meaning to do this, thanks!

PR #10078, opened 2024-06-10
fix(server): otel not working due to port conflicts after combining containers

zackpollard:


docker/prometheus.yml
```diff
@@ -6,7 +6,7 @@ scrape_configs:
   - job_name: immich_server
```
bo0tzz, line 6:
```suggestion
  - job_name: immich_api
```

Maybe?

zackpollard: Done :smile:

server/src/utils/instrumentation.ts
```diff
@@ -33,23 +33,37 @@ const aggregation = new metrics.ExplicitBucketHistogramAggregation(
...
+    instrumentations: [
+      new HttpInstrumentation(),
+      new IORedisInstrumentation(),
+      new NestInstrumentation(),
+      new PgInstrumentation(),
+    ],
+    views: [new metrics.View({ aggregation, instrumentName: '*', instrumentUnit: 'ms' })],
+  });
+  otelSingleton.start();
+  return otelSingleton;
```
jrasm91, line 58: just don't return it imo

jrasm91 (approved):

PR #10131, opened 2024-06-11
fix(server): cache-control header missing from / requests

michelheusschen:
Just ran into the same issue @alextran1502 had a couple of days ago where requests are being made to the old user endpoint. Turns out requests to `/` are missing the cache-control header, because `sirv` runs first and will serve `index.html` that browsers can cache.

Fixed by disabling extension fallback (defaults to `['html', 'htm']`)

### DIFF
--- a/server/src/constants.ts
+++ b/server/src/constants.ts
@@ -43,7 +43,7 @@ export enum AuthType {
   OAUTH = 'oauth',
 }
 
-export const excludePaths = ['/', '/.well-known/immich', '/custom.css', '/favicon.ico'];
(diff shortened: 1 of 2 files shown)


jrasm91:
I am pretty sure this won't work. The static file serving needs to run first so we know if we need to fall back to the index.html page or not.

jrasm91:
This now basically returns the index.html page for every web request.

michelheusschen:
Updated to disable extension fallback (defaults to `['html', 'htm']`)

jrasm91 (approved):

PR #10178, opened 2024-06-12
fix: postgres health check reporting any db without checksums as unhealthy

### LINKED ISSUE #10134: immich_postgres healthcheck not working
### The bug

In the latest dockercompose was added a healthcheck but in immich_postgres
`SELECT SUM(checksum_failures) FROM pg_stat_database` returns NULL not 0, so the healthcheck fails every time
For a temporary solution i just changed to `SELECT COALESCE(SUM(checksum_failures), 0) FROM pg_stat_database`

Docker/Non Relatable

### Version of Immich Server

v1.106.1

v1.106.1

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
database:
    image: ${IMMICH_POSTGRES}
    env_file:
      - .env
    networks:
      - non_exposed
    environment:
      POSTGRES_USER: ${DB_USERNAME}
    volumes:
    healthcheck:
      test: pg_isready --dbname='${DB_DATABASE_NAME}' || exit 1; Chksum="$$(psql --dbname='${DB_DATABASE_NAME}' --username='${DB_USERNAME}' --tuples-only --no-align --command='SELECT SUM(checksum_failures) FROM pg_stat_database')"; echo "checksum failure count is $$Chksum"; [ "$$Chksum" = '0' ] || exit 1
      interval: 5m
      start_interval: 30s
      start_period: 5m
    restart: always
    labels:
      - 'wud.watch=false'
```

### Your .env content

```Shell
NON RELATABLE
```

### Reproduction steps

```bash
1. Run the container
2. Wait 5 minutes
3. Healthcheck fails
```

### Relevant log output

_No response_

### Additional information

_No response_

mertalev:
## Description

This is a small change to make the health check respond as healthy if checksums are disabled and it gets a successful response from Postgres. It currently bogusly reports existing databases as unhealthy.

Fixes #10134

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -103,7 +103,7 @@ services:
     ports:
       - 5432:5432
     healthcheck:
-      test: pg_isready --dbname='${DB_DATABASE_NAME}' || exit 1; Chksum="$$(psql --dbname='${DB_DATABASE_NAME}' --username='${DB_USERNAME}' --tuples-only --no-align --command='SELECT SUM(checksum_failures) FROM pg_stat_database')"; echo "checksum failure count is $$Chksum"; [ "$$Chksum" = '0' ] || exit 1
+      test: pg_isready --dbname='${DB_DATABASE_NAME}' || exit 1; Chksum="$$(psql --dbname='${DB_DATABASE_NAME}' --username='${DB_USERNAME}' --tuples-only --no-align --command='SELECT COALESCE(SUM(checksum_failures), 0) FROM pg_stat_database')"; echo "checksum failure count is $$Chksum"; [ "$$Chksum" = '0' ] || exit 1
       interval: 5m
       start_interval: 30s
       start_period: 5m
--- a/docker/docker-compose.prod.yml
+++ b/docker/docker-compose.prod.yml
@@ -61,7 +61,7 @@ services:
     ports:
       - 5432:5432
     healthcheck:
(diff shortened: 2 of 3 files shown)


zackpollard (approved):

mmomjian:
Ty. I should have known everyone would just copy paste the new changes haha

reini12345:
I still have the issue with v1.106.4

image

zackpollard:
> I still have the issue with v1.106.4
> 
> image
> 

That is actually reporting checksum errors it seems

reini12345:
did not have this before the 1.106 update

zackpollard:
> did not have this before the 1.106 update

We didn't have health checks before the 1.106 update, so these issues wouldn't have been visible before then. This is not a bug, the recently added checksums are reporting an inconsistency in your database data

mmomjian:
> I still have the issue with v1.106.4
> 
> image

This looks fine to me. Please use the updated healthcheck command found in the docker-compose that contains the COALESCE command

The failure_count is a docker internal counter that increments with successive failed health checks. If it was a checksum error, it should appear in the message itself. In this case the result is NULL which appears as empty (checksums disabled).

zackpollard:
> > I still have the issue with v1.106.4
> > 
> > image
> 
> This looks fine to me. Please use the updated healthcheck command found in this PR that contains the COALESCE command
> 
> The failure_count is a docker internal counter that increments with successive failed health checks. If it was a checksum error, it should appear in the message itself. In this case the result is NULL which appears as empty (checksums disabled). 

Ah, well that's not confusing at all 🤣 It did make me think though, do we have docs to help people recover if they do end up with a legitimate checksum error? Or at least somewhere to point them at in the postgres docs?

mmomjian:
Practically speaking, the answer is to restore from backup. The checksums are only sufficient to detect corruption, not to restore it. I can add that to the FAQ entry about it though.

reini12345:
thx. With the updated healthcheck command the conatiner is now "healthy"

image

For people like me who have no idea about postgres, more documentation would not be bad. I had already started looking for a way to repair the database.

PR #10189, opened 2024-06-12
feat(web): select all duplicates

danielhep:
Small feature addition to the web UI allowing users to select all or none when deleting duplicates. 

I enjoyed using the new feature myself recently but found it annoying to select a lot of duplicates when they were part of a burst and I wanted to keep all of them. 

image


alextran1502:
Thanks! I modified the button UI a bit

image


image

alextran1502 (approved):

PR #10218, opened 2024-06-12
fix: no floats (replace with doubles)

jrasm91:
Mobile generated client dies when using floats when json sends a whole number.


zackpollard (approved):
LGTM

PR #10237, opened 2024-06-12
fix(web): Update prompt

aviv926:
Add information about irreversible action when merging people to the prompt


alextran1502 (approved):
Can you help make this into a translatable string as well? Thanks!

aviv926:
> Can you help make this into a translatable string as well? Thanks!



> Can you help make this into a translatable string as well? Thanks!

I have updated.
I also tried to do
message: `Merged ${count} ${count === 1 ? 'person' : 'people'}`,

But I didn't really succeed, so I left it as a source

PR #10254, opened 2024-06-13
chore(web): use "immich-form-label" class for combobox label

flesser:
On the "Edit date and time" dialog I noticed that the font sizes between the two input labels were different:
before

This PR changes the combobox label to use the existing `immich-form-label` CSS class instead of tailwind's `text-sm`:
after

The spacing is still off, but this may be fixed in a different PR (or in a later edit of this one).

As far as I can tell, the "Edit date and time" dialog is currently the only place where a combobox with visible label is used. In the settings pages, all the comboboxes have their label hidden.

_(Kudos for having such a smooth and simple https://immich.app/docs/developer/setup/ which makes it really low-threshold to contribute! ❤️ )_

### DIFF
--- a/web/src/lib/components/shared-components/combobox.svelte
+++ b/web/src/lib/components/shared-components/combobox.svelte
@@ -112,7 +112,7 @@
   };
 </script>
 
-<label class="text-sm text-black dark:text-white" class:sr-only={hideLabel} for={inputId}>{label}</label>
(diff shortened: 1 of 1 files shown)


michelheusschen:
Thanks for the PR. The combobox is also used in the search bar where the label color changes slightly, but I don't think that's an issue:

Before
search-options-before

After
search-options-after

If you could also adjust the spacing of `Date and Time` to match, that would be great

jrasm91 (approved):

PR #10271, opened 2024-06-13
fix(web): more language selector nits

zackpollard:


bo0tzz (approved):

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -60,6 +60,7 @@
   const activate = () => {
     isActive = true;
     input?.select();
+    searchQuery = '';
```
michelheusschen, line 63: Why clear the value?

bo0tzz: In order to get the dropdown to show the full list. When removing focus, the `deactivate()` fn below restores the query.

michelheusschen: Ideally the search query remains in the input field while also showing all options, but there doesn't seem to be an easy way to do this. `input?.select();` can be removed though as it no longer serves a purpose

PR #10274, opened 2024-06-13
chore(web): another missing translations

waclaw66:
Add another missing translations.


web/src/lib/components/shared-components/upload-panel.svelte
```diff
@@ -65,12 +64,22 @@
...
+                  processed: $successCounter + $errorCounter,
+                  total: $totalUploadCounter,
+                },
+              })}
             </p>
             <p class="immich-form-label text-xs">
-              Uploaded <span class="text-immich-success">{$successCounter}</span> - Error
-              <span class="text-immich-error">{$errorCounter}</span>
-              - Duplicates <span class="text-immich-warning">{$duplicateCounter}</span>
+              {@html $t('upload_status', {
```
michelheusschen, line 76: This leaves us vulnerable to XSS attacks. I don't have a solution for this yet, but it's also an issue in other places.

waclaw66: Working on a solution, currently seems to be simplest named parametr replace within translated string.

web/src/lib/components/album-page/albums-controls.svelte
```diff
@@ -118,8 +118,8 @@
   options={Object.values(sortOptionsMetadata)}
   selectedOption={selectedSortOption}
   on:select={({ detail }) => handleChangeSortBy(detail)}
-  render={({ text }) => ({
-    title: text,
+  render={({ key }) => ({
```
jrasm91, line 121: I don't think we should ever dynamically translate values. Just translate them in directly in the options list.

web/src/lib/components/album-page/albums-table-header.svelte
```diff
@@ -27,6 +28,6 @@
         &#8593;
       {/if}
     {/if}
-    {option.text}
+    {$t(option.key)}
```
jrasm91, line 31: Same here.

jrasm91 (changes requested):
IMO every translation usage should use this format `$t('string_key_value')`. Please don't dynamically translate variables and please always follow this syntax. If it isn't in a svelte component then you can do this `const $t = get(t);` and then `$t('string_key_value')`.

web/src/lib/components/album-page/albums-controls.svelte
waclaw66: Ok, I will fix it...

waclaw66: There is a problem with translating text within `web/src/lib/utils/album-utils.ts`...
```ts
export const sortOptionsMetadata: AlbumSortOptionMetadata[] = [
  {
    id: AlbumSortBy.Title,
    text: $t('sort_title'),
    defaultOrder: SortOrder.Asc,
    columnStyle: 'text-left w-8/12 sm:w-4/12 md:w-4/12 md:w-4/12 xl:w-[30%] 2xl:w-[40%]',
  },
```
` Cannot format a message without first setting the initial locale.`

Therefore I used on the fly translation before.
Could you please help to fix that?

waclaw66: A possible solution would be to remove `text` attribute and create mapping table in svelte with `AlbumSortBy` as a key and translated text as value and then use texts from this within `render`.

waclaw66: It should be fixed, please check if this solution is ok.

web/src/lib/components/album-page/albums-controls.svelte
```diff
@@ -88,12 +89,33 @@
...
+    [AlbumGroupBy.Owner]: $t('group_owner'),
+    [AlbumGroupBy.Year]: $t('group_year'),
+  };
 </script>
 
 <!-- Filter Albums by Sharing Status (All, Owned, Shared) -->
 <div class="hidden xl:block h-10">
   <GroupTab
-    filters={Object.keys(AlbumFilter)}
+    filters={Object.values(albumFilterNames)}
```
michelheusschen, line 118: Does this work for languages other than English? And does it remember the correct state after switching languages?

web/src/lib/i18n/en.json
```diff
@@ -877,8 +960,16 @@
   "updated_password": "Updated password",
   "upload": "Upload",
   "upload_concurrency": "Upload concurrency",
+  "upload_errors": "Upload completed with {count, plural, one {# error} other {# errors}}, refresh the page to see new upload assets.",
+  "upload_progress": "Remaining {remaining} - Processed {processed}/{total}",
+  "upload_skipped_duplicates": "Skipped {count, plural, one {duplicate} other {duplicates}}",
```
michelheusschen, line 965: Missing the count value and the message changed. Original message is `Skipped ${$duplicateCounter} duplicate asset${s($duplicateCounter)}`

web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte
```diff
@@ -173,10 +173,10 @@
       });
       notificationController.show({
         type: NotificationType.Info,
-        message: `Activity is ${album.isActivityEnabled ? 'enabled' : 'disabled'}`,
+        message: $t('activity_changed', { values: { enabled: album.isActivityEnabled } }),
       });
     } catch (error) {
-      handleError(error, `Can't ${album.isActivityEnabled ? 'disable' : 'enable'} activity`);
+      handleError(error, $t('errors.cant_change_activity'));
```
Comment (michelheusschen) [inline web/src/routes/(user)/albums/[albumId=id]/[[photos=photos]]/[[assetId=id]]/+page.svelte:179]: Missing `enabled` variable

web/src/lib/components/album-page/albums-controls.svelte
waclaw66: Yes, it works for others, unfortunatelly it doesn't keep setting after language is switched, it saves dropdown text value, it should be fixed.

michelheusschen: Just tested it using development language and the buttons don't work for me. Remembering the selected state is also broken. This could be fixed by passing both the enum value and a translated label to the `GroupTab` component, and then only storing the enum as selected value.

waclaw66: Fixed

web/src/lib/utils.ts
```diff
@@ -265,7 +269,7 @@ export const oauth = {
       window.location.href = url;
       return true;
     } catch (error) {
-      handleError(error, 'Unable to login with OAuth');
+      handleError(error, get(t)('errors.unable_to_login_with_oauth'));
```
jrasm91, line 272: Can you change this to use $t?

web/src/routes/auth/login/+page.ts
```diff
@@ -13,7 +15,7 @@ export const load = (async ({ fetch }) => {
 
   return {
     meta: {
-      title: 'Login',
+      title: get(t)('login'),
```
jrasm91, line 18: And this one

jrasm91 (commented):
Looks a lot better, thanks!

web/src/lib/utils.ts
waclaw66: Done

web/src/lib/i18n/en.json
```diff
@@ -557,6 +608,7 @@
   "immich_web_interface": "Immich Web Interface",
   "import_from_json": "Import from JSON",
   "import_path": "Import path",
+  "in_albums": "In {count} {count, plural, one {album} other {albums}}",
```
michelheusschen, line 611: Just to get proper number formatting

```suggestion
  "in_albums": "In {count, plural, one {# album} other {# albums}}",
```

web/src/lib/components/album-page/albums-controls.svelte
```diff
@@ -88,14 +97,35 @@
...
+  const albumSortByNames: { [key in AlbumSortBy]: string } = {
+    [AlbumSortBy.Title]: $t('sort_title'),
+    [AlbumSortBy.ItemCount]: $t('sort_items'),
+    [AlbumSortBy.DateModified]: $t('sort_modified'),
+    [AlbumSortBy.DateCreated]: $t('sort_created'),
+    [AlbumSortBy.MostRecentPhoto]: $t('sort_recent'),
+    [AlbumSortBy.OldestPhoto]: $t('sort_oldest'),
+  };
+
+  const albumGroupByNames: { [key in AlbumGroupBy]: string } = {
```
michelheusschen, line 116: All these objects should be reactive

web/src/lib/components/album-page/albums-table-header.svelte
```diff
@@ -12,6 +13,15 @@
       $albumViewSettings.sortOrder = option.defaultOrder;
     }
   };
+
+  const albumSortByNames: { [key in AlbumSortBy]: string } = {
```
michelheusschen, line 17: Also missing reactivity

web/src/lib/components/album-page/share-info-modal.svelte
```diff
@@ -65,7 +68,7 @@
   const handleSetReadonly = async (user: UserResponseDto, role: AlbumUserRole) => {
     try {
       await updateAlbumUser({ id: album.id, userId: user.id, updateAlbumUserDto: { role } });
-      const message = `Set ${user.name} as ${role}`;
+      const message = $t('user_role_set', { values: { user: user.name, role: role } });
```
michelheusschen, line 71: `role` isn't translated

web/src/lib/components/album-page/user-selection-modal.svelte
```diff
@@ -24,9 +24,9 @@
   let selectedUsers: Record<string, { user: UserResponseDto; role: AlbumUserRole }> = {};
 
   const roleOptions: Array<{ title: string; value: AlbumUserRole | 'none'; icon?: string }> = [
-    { title: $t('editor'), value: AlbumUserRole.Editor, icon: mdiPencil },
-    { title: $t('viewer'), value: AlbumUserRole.Viewer, icon: mdiEye },
-    { title: $t('remove'), value: 'none' },
+    { title: $t('role_editor'), value: AlbumUserRole.Editor, icon: mdiPencil },
+    { title: $t('role_viewer'), value: AlbumUserRole.Viewer, icon: mdiEye },
+    { title: $t('remove_user'), value: 'none' },
```
michelheusschen, line 29: Why not reuse the `remove` key?

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -395,10 +395,10 @@
 
       notificationController.show({
         type: NotificationType.Info,
-        message: asset.isFavorite ? `Added to favorites` : `Removed from favorites`,
+        message: $t(asset.isFavorite ? 'added_to_favorites' : 'removed_from_favorites'),
```
michelheusschen, line 398: The automated test for missing translation doesn't work for dynamic translation keys, so I think we should avoid using them.

```suggestion
        message: asset.isFavorite ? $t('added_to_favorites') : $t('removed_from_favorites'),
```

web/src/lib/i18n/en.json
```diff
@@ -717,20 +792,37 @@
   "previous_or_next_photo": "Previous or next photo",
   "primary": "Primary",
   "profile_picture_set": "Profile picture set.",
+  "public_album": "Public album",
   "public_share": "Public Share",
+  "reaction_deleted": "{type, select, comment {Comment} like {Like} other {Reaction}} deleted",
```
michelheusschen, line 797: I don't think we should include mappings to an enum in the translation file. It has no type safety and it forces us to include `other` which often doesn't make sense for an enum.

web/src/lib/components/album-page/user-selection-modal.svelte
waclaw66: "Remove a user" sounds odd in other languages(e.g. cz).

michelheusschen: Ah that makes sense, we could also make it clearer by changing it to "Remove user"

web/src/lib/i18n/en.json
waclaw66: Understood, how would you fix that?
`reaction.type == Type.Comment ? $t('comment_deleted') : $t('like_deleted')`
Unfortunatelly there is always "else/other".

web/src/lib/components/album-page/albums-controls.svelte
waclaw66: Can you please help here, not sure how to fix it, thanks.

web/src/lib/i18n/en.json
michelheusschen: Something like this should work

```ts
const messages: Record<Type, string> = {
  [Type.Comment]: $t('comment_deleted'),
  [Type.Like]: $t('like_deleted'),
};

messages[reaction.type];
```

web/src/lib/components/album-page/albums-controls.svelte
michelheusschen: It's a bit annoying the type can't be set directly, but you could use an IIFE

```ts
$: albumGroupByNames = ((): Record<AlbumGroupBy, string> => {
  return {
    [AlbumGroupBy.None]: $t('group_no'),
    [AlbumGroupBy.Owner]: $t('group_owner'),
    [AlbumGroupBy.Year]: $t('group_year'),
  };
})();
```

waclaw66: Thanks, fixing...

web/src/lib/i18n/en.json
```diff
@@ -64,8 +66,8 @@
     "job_settings": "Job Settings",
     "job_settings_description": "Manage job concurrency",
     "job_status": "Job Status",
-    "jobs_delayed": "{jobCount} delayed",
-    "jobs_failed": "{jobCount} failed",
+    "jobs_delayed": "{jobCount, plural, other {#}} delayed",
```
michelheusschen, line 69: Why use the plural form over `{jobCount, number}`? I've also seen this with a few other translations.

waclaw66: When plural is not defined in the source language and it's used in target language Weblate reports item translation as forced ICU MessageFormat. Other languages can inflect "delayed" depending on job count.

michelheusschen: Understood. I think using `{# delayed}` in that case would communicate the intention a bit better, but that's just a minor thing

michelheusschen (approved):
Thanks a lot for the effort!

web/src/lib/i18n/en.json
waclaw66: You're right, `{# delayed}` is much better.

waclaw66:
> Thanks a lot for the effort!

Happy to contribute.

web/src/lib/components/shared-components/profile-image-cropper.svelte
```diff
@@ -28,13 +28,13 @@
...
     if (!context) {
-      throw new Error('Could not get canvas context.');
+      throw new Error($t('errors.could_not_get_canvas_context'));
     }
     context.drawImage(img, 0, 0);
     const imageData = context.getImageData(0, 0, canvas.width, canvas.height);
     const data = imageData?.data;
     if (!data) {
-      throw new Error('Could not get image data.');
+      throw new Error($t('errors.could_not_get_image_data'));
```
michelheusschen, line 37: Messages that are only logged to console shouldn't be translated

waclaw66: You're right, fixed. Thanks!

jrasm91 (approved):

PR #10276, opened 2024-06-13
chore(doc): update quick-start.mdx

richard-salam:
The following changes are made:

- Changed the headings to capital case.
- Changed a few sentences to sound more clear.
- Removed '?' from the heading as per the standards.


bo0tzz (approved):

jrasm91:
Personally I think we should migrate to _not_ using capital case for all the headings.

PR #10321, opened 2024-06-14
refactor(server): partner ids

jrasm91:
Consistently handle deleted users, only selecting user ids in the right direction, etc.


danieldietzler (approved):
LGTM

PR #10332, opened 2024-06-14
refactor(web): byte unit utils

danieldietzler:
Simplify byte utils using ✨ Math ✨


mertalev (approved):

web/src/lib/utils/byte-units.ts
```diff
@@ -8,23 +18,10 @@
...
-      break;
-    }
-  }
-
-  remainder = Number.parseFloat(remainder.toFixed(maxPrecision));
+export function getBytesWithUnit(bytes: number, maxPrecision = 1): [number, ByteUnit] {
+  const magnitude = Math.floor(Math.log2(bytes === 0 ? 1 : bytes) / 10);
 
-  return [remainder, units[magnitude]];
+  return [Number.parseFloat((bytes / 2 ** (magnitude * 10)).toFixed(maxPrecision)), magnitude];
```
michelheusschen, line 24: Maybe we can use base 1024 for these functions? But this isn't wrong at all.

```suggestion
  const magnitude = Math.floor(Math.log(bytes === 0 ? 1 : bytes) / Math.log(1024));

  return [Number.parseFloat((bytes / 1024 ** magnitude).toFixed(maxPrecision)), magnitude];
```

michelheusschen (approved):
Nice!

web/src/lib/utils/byte-units.ts
danieldietzler: Oh yeah we could, good idea!

jrasm91 (approved):

aviv926:
Could this change affect https://github.com/immich-app/immich/issues/10927?

PR #10335, opened 2024-06-14
feat(web): short live SSR

jrasm91:


jrasm91:
Looks like it is not _too_ crazy to hook the sveltekit handler into the API express service, however... I have run into _a lot_ of runtime problems with code that's not compatible in a nodejs envrionment:
- navigator references
- window references
- location/url references
- cjs vs esm module conflicts

In short, while it would be nice to have SSR, specifically for the shared link page, it becomes too easy to introduce runtime errors so it's just not worth it IMO.

PR #10346, opened 2024-06-15
feat(web): full screen view for duplicates

michelheusschen:
Adds a button to view duplicate assets in full screen. Also changed:
- Use relative units instead of pixels
- More descriptive buttons for permanent delete confirmation

Preview where the left button opens the asset viewer
image

Closes #10173


web/src/lib/i18n/en.json
```diff
@@ -317,6 +317,7 @@
   "archived": "Archived",
   "asset_offline": "Asset offline",
   "assets": "Assets",
+  "assets_moved_to_trash": "Moved {count, plural, one {# asset} other {# assets}} to trash",
```
danieldietzler, line 320: Does `#` work similar to `{count, number}`?

danieldietzler (approved):
Code looks great to me!

web/src/lib/i18n/en.json
michelheusschen: Yes exactly, we're already using the shorter `#` syntax for some of our existing translations

danieldietzler: Ah ok, that's cool (and easier)! :D

alextran1502 (approved):
Thanks so much!

PR #10371, opened 2024-06-15
feat(server): separate face search relation

### LINKED ISSUE #10277: Unassigned faces start showing up later
### The bug

First, I'm sorry because I don't really understand what "Unassigned faces" is all about. I couldn't find anything in the documentation, but I just realized it is related to the following problem:

- Face Detection and Face Clustering is performed
- I start tagging faces, merging people, hiding others, etc.
- There are some pics where not everyone is available for tagging: either because it is labelled as "Unassigned face" or just because it was not detected (i guess)
- When I'm done with tagging, I run Face Detection and Face Clustering again
- Then new people appear that are available for tagging: either from the unassigned faces or from the pool that was missed before.

I feel like this is a bug: I can never predict the amount of work it's going to take me to tag everyone, since it keeps adding up.
Why is a face unassigned in the first place? Why are some faces detected afterwards that were not detected the first time?
Are there some threshold that I meet only after I reduced the number of unknown people?

I played a little bit with the settings to understand things better, but even using min-cluster-size of 1 and asking immich to display all available people in the people tab (by editing `repositories/person.repository.js` ), I still have some new people showing up after I have spent time tagging people.
I also tried with a smaller library (18GB instead of 500GB), but the behavior is the same, to a smaller extent, or course, but still.

Thanks for the help as I am very confused.

Debian 12

### Version of Immich Server

v1.106.3

v1.106.3

### Platform with the issue

- [X] Server
- [X] Web
- [ ] Mobile

```YAML
Not relevant
```

### Your .env content

```Shell
Not relevant. no modification of the settings
```

### Reproduction steps

```bash
- Face Detection and Face Clustering is performed
- I start tagging faces, merging people, hiding others, etc.
- There are some pics where not everyone is available for tagging: either because it is labelled as "Unassigned face" or just because it was not detected (i guess)
- When I'm done with tagging, I run Face Detection and Face Clustering again
- Then new people appear that are available for tagging: either from the unassigned faces or from the pool that was missed before.
```

### Relevant log output

_No response_

### Additional information

_No response_

mertalev:
### Description

This PR addresses a subtle issue with the current facial recognition. Each time a face is assigned or reassigned a person, for both the initial and later facial recognition runs, an additional duplicate embedding is inserted into the face vector index. This can lead to index degradation as the majority of the index is duplicated, in turn leading to faces sometimes not being recognized when they should.

This PR changes the embedding to be in a separate table as a one-one relation, similar to how smart search is handled. This means changes to the face, such as which person it's assigned to, have no effect on the index. It has a smaller but notable benefit of making these changes faster and producing less WAL.

A notable benefit of this change is also that it makes supporting manually added faces easier as an embedding is no longer required.

Also sets storage to external so Postgres doesn't try to compress the embeddings, following the finding [here](https://github.com/tensorchord/pgvecto.rs/pull/321).

Fixes #10277

### How Has This Been Tested?

Tested that the migration is successful without loss of data and that both face detection and facial recognition jobs continue to work.
- The number of unassigned faces decreased by 38% after running with this change
- Facial recognition is twice as fast as before
- Face detection is 20% faster (I'm not really sure why)
- `SELECT idx_tuples FROM pg_vector_index_stat WHERE indexname = 'face_index';` is identical to the number of faces, i.e. no duplicate embeddings


server/src/entities/face-search.entity.ts
```diff
@@ -0,0 +1,21 @@
+import { AssetFaceEntity } from 'src/entities/asset-face.entity';
+import { asVector } from 'src/utils/database';
+import { Column, Entity, Index, JoinColumn, OneToOne, PrimaryColumn } from 'typeorm';
+
+@Entity('face_search', { synchronize: false })
```
jrasm91, line 5: Can we synchronize this table?

jrasm91 (approved):
Can we also mark the clip embeddings as external?

mertalev:
> Can we also mark the clip embeddings as external?

Yep, this makes the clip embeddings external too.

server/src/entities/face-search.entity.ts
mertalev: I assumed it would try to change the column type, but I can try it.

mertalev: Yeah, it tries to change it to `real array`.

server/src/migrations/1718486162779-AddFaceSearchRelation.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    if (getVectorExtension() === DatabaseExtension.VECTORS) {
+      await queryRunner.query(`SET search_path TO "$user", public, vectors`);
+      await queryRunner.query(`SET vectors.pgvector_compatibility=on`);
+    }
+
+    await queryRunner.query(`ALTER TABLE asset_faces ADD COLUMN "embedding" vector(512)`);
+    await queryRunner.query(`ALTER TABLE face_search ALTER COLUMN embedding SET STORAGE DEFAULT`);
+    await queryRunner.query(`ALTER TABLE smart_search ALTER COLUMN embedding SET STORAGE DEFAULT`);
```
jrasm91, line 41: Ah, I missed this line. Nice!

PR #10389, opened 2024-06-16
fix(server): include trashed assets in forced thumbnail generation

michelheusschen:
Fixes #10387

PR #10409, opened 2024-06-16
fix(server): include archived assets in forced thumbnail generation

### LINKED ISSUE #10387: Generating all thumbnails do not generate thumbs of archived images
### The bug

After changing the format from jpeg to webp and regenerating all thumbnails, archived images are still using jpeg thumbnails and previews, no webp thumbs were generated

Arch

### Version of Immich Server

v1.106.4

v1.106.4

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
default
```

### Your .env content

```Shell
default
```

### Reproduction steps

```bash
1. Have an archived image
2. Regenerate all thumbnails, e.g. in different format to see a difference (or delete the thumbs from disk)
3. No new thumbnail is generated for the archived image
```

### Relevant log output

_No response_

### Additional information

_No response_

michelheusschen:
Actually fixes #10387. In my previous PR #10389 trashed assets got included, but not archived assets as mentioned in the issue.


danieldietzler (approved):
LGTM

server/src/services/media.service.ts
```diff
@@ -70,7 +70,7 @@ export class MediaService {
   async handleQueueGenerateThumbnails({ force }: IBaseJob): Promise<JobStatus> {
     const assetPagination = usePagination(JOBS_ASSET_PAGINATION_SIZE, (pagination) => {
       return force
-        ? this.assetRepository.getAll(pagination, { isVisible: true, withDeleted: true })
```
mertalev, line 73: Do you think trashed assets should still get queued? Presumably the user doesn't care about them since they're trashed.

danieldietzler: I think it does make sense, as if a user later restores an asset from trash it could be confusing if the thumbnail for that is broken. Or even just looking at assets in the trash for that purpose

mertalev: Oh that's true, it could be good if they need to figure out what should be restored from the trash and the thumbnail is broken.

mertalev (approved):

server/src/services/media.service.ts
michelheusschen: Yes, I think so. You could argue that they'll likely be permanently deleted, which might seem wasteful. However, if they are restored you'll end up with assets that aren't in the correct format.

PR #10413, opened 2024-06-16
feat(web): mark all faces

oezgueremir:
When watching photos and the info tab (detail-panel) is open, all the…

… faces in the photo are marked.

tasks:
- [x] the "mark all faces" function can be de-/activated on the info tab.
- [x] CircleIconButton mark_all_faces with mdiLabelMultiple icon as controller.
- [x] Each person/face has a different, static border color given for the pictogramm and for the drawn mark rect.
      This is provided with a little quick'n'dirty solution. A better solution could/would require changes in the DB model...
- [x] The mouseover event will override that functionallity and mark only the faces of this single person.
      More over, it will change the visualization of the marking rectangle to border=white, filling=person/face color, opacity=0.6
- [x] If mark_all_faces is active, the CircleIconButton show_hidden_people will always be available.
      So you are able to see even hidden and unassigned faces.
      Hidden and unassigned faces will be visualized, as if the mouseover event is triggered for them.
      It can be further evaluated, if these two states have to be controlled individually and presented with unique designs.


alextran1502:
Hello, can you include a screen capture or screenshot for the UI changes?

web/package.json
```diff
@@ -74,6 +74,7 @@
     "justified-layout": "^4.1.0",
     "lodash-es": "^4.17.21",
     "luxon": "^3.4.4",
+    "sharp": "^0.33.4",
```
danieldietzler, line 77: Why did you install sharp in the web?

danieldietzler (changes requested):

oezgueremir:
Before this PR:
original

After this PR
Option deactivated:
off

Option activated:
on

Mouse hovering over "Demo02"
Demo02

web/package.json
oezgueremir: Anything, that is outside of the real source files, was not intentionally.


I saw, that this has changed, yes. But I assumed, that the setup steps have done that.
I just followed these steps.
https://github.com/immich-app/immich/blob/main/docs/docs/developer/setup.md#connect-web-to-a-remote-backend

Initially I had problems with the setup, Ubuntu LTS has way too old nodeJS and npm delivered in its package... took me some time to find that out...
See https://discord.com/channels/979116623879368755/1251343704975868005

After updating to the latest versions, all was fine.

vboxuser@ubuntults:~$ npm -v
10.7.0
vboxuser@ubuntults:~$ node -v
v20.14.0


Long story short, tell me, how I have done that, and I will revert it :)

oezgueremir:
Continuing from this scene, feature is in use, option is off
340455657-2f3e4c56-c35d-43e1-829c-881f3b98a262


Hide Demo03
Demo03Hidden

Activate option
on

Activate "show hidden people" (mouse not hovering over anything)
grafik

Mouse hovering over "Demo02"
grafik

Mouse hovering over "Demo03"
grafik

web/package-lock.json
danieldietzler, line 1: Your package lock is still screwed up. If you us to, me or someone else can just fix it before it gets merged

oezgueremir:
As discussed, I will send a new PR.
Again, thanks

PR #10430, opened 2024-06-17
fix(cli): handle patterns correctly on Windows

fky2015:
[`fast-glob` requires slash (instead of backslash) for matching path. ](https://www.npmjs.com/package/fast-glob#how-to-write-patterns-on-windows)

So before this change, `--recursive` didn't work on WIndows.

Before: 
image

After:
image

## Some changes

- On Windows, use `convertPathToPattern` from `fast-glob`.
- Add test CI for CLI on the Windows platform.
- To unify the test cases (a file may have multiple valid path formats), compare file contents instead of the paths.

### DIFF
--- a/cli/src/utils.spec.ts
+++ b/cli/src/utils.spec.ts
@@ -1,4 +1,5 @@
 import mockfs from 'mock-fs';
+import { readFileSync } from 'node:fs';
 import { CrawlOptions, crawl } from 'src/utils';
 
 interface Test {
@@ -8,6 +9,10 @@ interface Test {
 }
 
 const cwd = process.cwd();
+
+const readContent = (path: string) => {
+  return readFileSync(path).toString();
+};
 
 const extensions = [
   '.jpg',
@@ -256,7 +261,8 @@ const tests: Test[] = [
   {
     test: 'should support ignoring absolute paths',
     options: {
-      pathsToCrawl: ['/'],
+      // Currently, fast-glob has some caveat when dealing with `/`.
+      pathsToCrawl: ['/*s'],
       recursive: true,
       exclusionPattern: '/images/**',
     },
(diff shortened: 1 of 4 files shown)


alextran1502 (approved):

alextran1502:
Can you please update the unit test as well?

fky2015:
Sure, I'm working on that.

cli/src/utils.ts
```diff
@@ -106,6 +107,11 @@ export interface CrawlOptions {
   exclusionPattern?: string;
   extensions: string[];
 }
+
+const convertPathToPatternOnWin = (path: string) => {
+  return platform() === 'win32' ? convertPathToPattern(path) : path;
+};
```
bo0tzz, line 113: Does this need to be conditional to the platform, would there be any issues with just always converting?

cli/vite.config.ts
```diff
@@ -2,6 +2,7 @@ import { defineConfig } from 'vite';
 import tsconfigPaths from 'vite-tsconfig-paths';
 
 export default defineConfig({
+  resolve: { alias: { src: '/src' } },
```
bo0tzz, line 5: Why is this needed?

cli/src/utils.spec.ts
bo0tzz, line 1: There's a few changes in this test that look odd to me, can you explain what's going on here?

cli/src/utils.ts
fky2015: Yes, `convertPathToPattern` will change the matching behavior on *nix platform:

[`/photo*` will become `/photo\*`](https://www.npmjs.com/package/fast-glob#convertpathtopatternpath) thus the asterisk will lose its effect.

cli/vite.config.ts
fky2015: I followed `cli/README.md` and filed to build the project without this config (on Windows).

image

Generally speaking, I think this change will improve the compatibility of dev configs.

cli/src/utils.spec.ts
fky2015: Line `#265`: On the Windows platform, `/` cannot be treated as expected in `fast-glob`. I've filed an issue here https://github.com/mrmlnc/fast-glob/issues/447.

Line `#286 & #294`: Comparing path strings directly will lead to some **false positive** cases. For example, on Windows, a path can be written as `D:\Something\Like\This` while `\something\like\this` on Linux. So It should be better to compare what they've actually read in the file system. To achieve this, I'm letting each file contain unique contents (their paths) so that it can be used for comparison later. A good thing to note is that we are still comparing their paths and this is friendly when debugging.

cli/src/utils.ts
bo0tzz: But it doesn't change the behaviour on windows? If so that sounds weird, do you know why it is like that?

fky2015: No, it doesn't change the behaviour on Windows.

From my understanding, `convertPathToPattern` is a function especially used for converting a Windows path to `fast-glob` search pattern. It is meaningless to use this function if your inputs are a valid POSIX path.

PR #10448, opened 2024-06-18
feat(web): bulk deduplicate

mertalev:
## Description

Manual deduplication is impractical with a high number of duplicates. This PR adds two new buttons in the top right to either deduplicate all groups (keeping the largest asset from each) or simply keep all duplicates.

Implemented client-side to be consistent with how normal deduplication works. A server-side implementation makes sense as well since it would reduce the amount of data passed around and be part of the API for scripts, etc. I can change it if requested.

## How Has This Been Tested?

- Both buttons take no action until a confirmation
- Keep All resolves all duplicates without trashing anything
- Keep All displays a notification
- Deduplicate All results in assets being trashed with no duplicate groups remaining afterwards
- Deduplicate All displays a notification
- The trashed assets are the smaller versions as expected
- Running Duplicate Detection after Deduplicate All does not result in any duplicates being detected


web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte
Comment (michelheusschen) [inline web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte:1]: Passing variables to `$t` or using the ternary operator makes it harder to track which translation keys are used and the automated test won't be able to detect missing translations. I think we should try to avoid dynamic translation keys as much as possible.

web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte
```diff
@@ -11,42 +11,124 @@
...
     }
   };
+
+  const deletedNotification = (trashedCount: number) => {
+    if (!trashedCount) {
+      return;
+    }
+
+    notificationController.show({
+      message: $t($featureFlags.trash ? 'assets_moved_to_trash' : 'assets_deleted', {
```
Comment (michelheusschen) [inline web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte:48]: `assets_deleted` is missing

web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte
```diff
@@ -11,42 +11,124 @@
...
+            ids: [...idsToDelete, ...idsToKeep.filter((id): id is string => !!id)],
+            duplicateId: null,
+          },
+        });
+
+        data.duplicates = [];
+
+        deletedNotification(idsToDelete.length);
+      },
+      $featureFlags.trash ? 'bulk_trash_duplicates_confirmation' : 'bulk_delete_duplicates_confirmation',
```
Comment (michelheusschen) [inline web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte:92]: Maybe we can provide the exact count to the user

web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte
Comment (mertalev) [reply web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte:1]: I suppose there's no way to do that and avoid duplication. I can move the logic in `withConfirmation` into each handler in that case.

web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte
Comment (mertalev) [reply web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte:48]: Good catch! I renamed it to `permanently_deleted_assets` in the JSON but forgot to change it here.

web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte
Comment (mertalev) [reply web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte:92]: Good point, I can see that being helpful.

web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte
Comment (mertalev) [reply web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte:1]: Last comment was a brain fart haha. I changed the logic so `$t` always has static inputs.

web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte
Comment (mertalev) [reply web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.svelte]: Yeah, definitely much better now.

web/src/lib/utils.ts
```diff
@@ -310,3 +312,7 @@ export const withError = async <T>(fn: () => Promise<T>): Promise<[undefined, T]
     return [error, undefined];
   }
 };
+
+export const suggestDuplicateByFileSize = (assets: AssetResponseDto[]): AssetResponseDto | undefined => {
```
jrasm91, line 316: This isn't bad thing to make a query param that is implemented on the server instead.

jrasm91 (approved):

web/src/lib/utils.ts
mertalev: Sure, I'll make a follow-up PR for that.

aviv926:
@mertalev  Does the marking of everything also include videos? If so, is it user selectable?

mertalev:
It does include videos, with no way to change that. You can make a feature request to make it more selective.

PR #10469, opened 2024-06-18
feat(web): show face's border on hovered

oezgueremir:
A new PR for a new branch...the other branch had some unrecoverable issues.

For Screenshots, see  #10413

When watching photos and the info tab (detail-panel) is open, all the faces in the photo are marked.

tasks:
- [x] the "mark all faces" function can be de-/activated on the info tab.
- [x] CircleIconButton mark_all_faces with mdiLabelMultiple icon as controller.
- [x] Each person/face has a different, static border color given. This is provided with a little quick'n'dirty solution. A better solution could/would require changes in the DB model...
- [x] The mouseover event will override that functionallity and mark only the faces of this single person. More over, it will change the visualization of the marking rectangle to border=white, filling=person/face color, opacity=0.6
- [x] If mark_all_faces is active, the CircleIconButton show_hidden_people will always be available. So you are able to see even hidden and unassigned faces. Hidden and unassigned faces will be visualized, as if the mouseover event is triggered for them. It can be further evaluated, if these two states have to be controlled individually and presented with unique designs.


mertalev:
I think it'd look cleaner if the boxes were all white without the different colors. It's not like there's usually multiple faces of a single person in an image; it's generally safe to assume each box is a different person.

oezgueremir:
Usually and ideally, yes.
But, and this is where this use case comes from, in many group photos in my library (especially pictures of my children), some people have multiple faces assigned to them.
This feature is not meant for every-day usage. This feature is intended to help you find and fix such problems.

waclaw66:
Nice feature, colored rectangles in people list should be deactivated by default though. It would be cleaner to show them when Mark all faces is activated by clicking that button. Filling face background is quite distracting when a light background color is used under white text person name, I would preffer to make colored border only (first 3 faces).

obrazek

oezgueremir:
> Nice feature, colored rectangles in people list should be deactivated by default though. It would be cleaner to show them when Mark all faces is activated by clicking that button. Filling face background is quite distracting when a light background color is used under white text person name, I would preffer to make colored border only (first 3 faces).

Ok, this is if the option is not active:
grafik

After activation the option (to emphesize my use case, I mapped two pairs of different faces on a single person):
grafik



How do you like the mouse-over for the visualization in the actual picture be handled?

Currently the style is,
border=white, filling=person/face color, opacity=0.6
grafik

This style is used, even if the option is not active. The intention is, to make it always easy to spot all markings in the picture.
grafik

This would be the style in the released version, same two faces marked as the screenshot before
grafik


The style could be bound to the same switch, or to another (global) setting, or (my preference) just always use the new style. With no additional switches to consider.

kingvuke:
Will this be taken over? What is the plan?

alextran1502 (approved):

jrasm91:
This looks pretty good, but it's a bit outdated. Can you help rebase it?

Also, I was thinking that we should...
- add the option to  "associate a color to a person" (#15937)
- add the toggle button to show/hide all faces (override the on:hover functionality - this PR)
- update the hover border to be the person's color (future PR)

For now, can you have the color always be white, and we can implement the per-color portion in a separate PR?

PR #10477, opened 2024-06-19
fix(server): video thumbnail generation failing with single i-frame

mertalev:
## Description

The sole i-frame of a video can get dropped by the `fps` filter. This PR sets `eof_action=pass`, meaning that it will always pass the last (and in this case first) frame.

Fixes #10468

## How Has This Been Tested?

Tested the sample video in #10468 and ran on a few hundred videos without error.


danieldietzler (approved):
I just trust your wizardry skills :)

jrasm91 (approved):

PR #10491, opened 2024-06-19
feat(web): translations containing html

michelheusschen:
Adds the ability to translate messages that contain HTML by building on [Rich Text Formatting](https://formatjs.io/docs/core-concepts/icu-syntax#rich-text-formatting) provided by `formatjs`. Because `svelte-i18n` doesn't export the needed function, the libraries are used directly.

The implementation considers the following:
 - tags in a translation can change position
 - translators can easily see the context and the syntax should be easy to reduce errors
 - alternative solutions are often susceptible to xss by using `@html`


waclaw66:
Looks good! One question, would it be possible to define idividual parameters like tags without if conditions?

```svelte
          <FormatTags message={$json('admin.transcoding_codecs_learn_more')} let:message>
            <h264Link>
              <a href="https://trac.ffmpeg.org/wiki/Encode/H.264" class="underline" target="_blank" rel="noreferrer">
                {message}
              </a>
            </h264Link>
            <hevcLink>
              <a href="https://trac.ffmpeg.org/wiki/Encode/H.265" class="underline" target="_blank" rel="noreferrer">
                {message}
              </a>
            </hevcLink>
            <vp9Link>
              <a href="https://trac.ffmpeg.org/wiki/Encode/VP9" class="underline" target="_blank" rel="noreferrer">
                {message}
              </a>
            </vp9Link>
          </FormatTags>
```

michelheusschen:
I don't think that's easily achievable. Using Custom Elements would come with too many caveats. I tried using dynamic slots, but that isn't possible either:

```svelte
<FormatTags>
  <!-- not possible -->
  <a href="..." slot="h264Link"></a>
</FormatTags>
```

waclaw66:
> I don't think that's easily achievable. Using Custom Elements would come with too many caveats. I tried using dynamic slots, but that isn't possible either:

Nevermind, this way is totaly usable for those few multi-tag html formated string. Thanks!

waclaw66:
If you're lookign for more translations with links, there are few within onbording pages, I havn't translated those yet, waiting for such solution like yours.

jrasm91 (approved):
Maybe "FormatMessage" or "MessageFormat" would be a better component name

web/src/lib/components/admin-page/delete-confirm-dialogue.svelte
```diff
@@ -54,12 +55,19 @@
     <div class="flex flex-col gap-4">
       {#if forceDelete}
         <p>
-          <b>{user.name}</b>'s account and assets will be queued for permanent deletion <b>immediately</b>.
+          <FormatMessage message={$json('admin.user_delete_immediately')} values={{ user: user.name }} let:message>
+            <b>{message}</b>
```
danieldietzler, line 59: So now the entire message will be bold? Is there a way to keep the existing formatting?

web/src/lib/components/admin-page/restore-dialogue.svelte
```diff
@@ -36,6 +37,10 @@
   onCancel={() => dispatch('cancel')}
 >
   <svelte:fragment slot="prompt">
-    <p><b>{user.name}</b>'s account will be restored.</p>
+    <p>
+      <FormatMessage message={$json('admin.user_restore_description')} values={{ user: user.name }} let:message>
+        <b>{message}</b>
```
danieldietzler, line 42: In this case the translation string also has `<b></b>`. Do we need both?

web/src/lib/i18n/en.json
```diff
@@ -902,6 +915,7 @@
   "validate": "Validate",
   "variables": "Variables",
   "version": "Version",
+  "version_announcement_message": "Hi friend, there is a new version of the application please take your time to visit the <link>release notes</link> and ensure your <code>docker-compose</code>, and <code>.env</code> setup is up-to-date to prevent any misconfigurations, especially if you use WatchTower or any mechanism that handles updating your application automatically.",
```
danieldietzler, line 918:
```suggestion
  "version_announcement_message": "Hi friend, there is a new version of the application please take your time to visit the <link>release notes</link> and ensure your <code>docker compose</code>, and <code>.env</code> setup is up-to-date to prevent any misconfigurations, especially if you use WatchTower or any mechanism that handles updating your application automatically.",
```
(unrelated to this PR)

web/src/lib/components/admin-page/delete-confirm-dialogue.svelte
michelheusschen: The slot is only used for message parts that have a tag. In this case it will be used for `<b>{user}</b>` and `<b>immediately</b>`. Only those parts will be bold.

web/src/lib/components/admin-page/restore-dialogue.svelte
michelheusschen: HTML from translations is never rendered directly to mitigate xss risk. So yes, we need both.

web/src/lib/i18n/en.json
michelheusschen: I changed it to `docker-compose.yml` because `.env` is also a filename

web/src/lib/components/admin-page/delete-confirm-dialogue.svelte
danieldietzler: Oh got it, thanks for the explanation! :)

danieldietzler (approved):
Awesome! :)

alextran1502 (approved):

waclaw66:
I've discovered that it doesn't show translation keys  in developer mode, it shows `undefined` instead.

PR #10512, opened 2024-06-20
fix(server): allow library id to be null in metadata search

danieldietzler:


mobile/openapi/lib/model/metadata_search_dto.dart
```diff
@@ -192,12 +192,6 @@ class MetadataSearchDto {
   ///
   String? lensModel;
 
-  ///
-  /// Please note: This property should have been non-nullable! Since the specification file
-  /// does not include a default value (using the "default:" property), however, the generated
-  /// source code must fall back to having a nullable type.
-  /// Consider adding a "default:" property in the specification file to hide this note.
```
danieldietzler, line 199: LMAO

jrasm91 (approved):
🎉

PR #10514, opened 2024-06-21
fix(server): `/places` entries sometimes not ordered alphabetically

mertalev:
## Description

Whether this query returns a sorted result depends on the query plan. This PR makes it consistently sort them before returning.

## How Has This Been Tested?

Tested on an instance where the results were not sorted and confirmed that the revised query sorts them as expected.

### DIFF
--- a/server/src/queries/search.repository.sql
+++ b/server/src/queries/search.repository.sql
@@ -405,3 +405,5 @@ FROM
   "assets" "asset"
   INNER JOIN "exif" "exif" ON "exif"."assetId" = "asset"."id"
   INNER JOIN cte ON asset.id = cte."assetId"
+ORDER BY
+  exif.city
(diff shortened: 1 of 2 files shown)


jrasm91 (approved):

pyorot:
@mertalev Is alphabet sort what we want here? Could it be easier for users to "explore" if it worked similarly to /people, that is, the user doesn't have a idea of who ey might find in advance, and so is most likely to find the most useful people by having them sorted by number of photos they appear in. So sorting places by number of contained assets comes to mind. Down the line, there could be clustering done, sorting by distance from a centre and making nearby places contiguous.

Lmk if I should make this a discussion; cheers.

mertalev:
You can check if there's no existing feature request to change it and make one if so.

I think the challenge is that counting etc. makes the query (and hence the page) much slower. With >1 million assets, a naive query can take seconds to run, whereas right now it still loads instantly.

PR #10528, opened 2024-06-21
feat(web): Adds sort order from album to shared album.

erathmus:
Pulls the sort order set in the album through to the shared album. 
Feature discussion: https://github.com/immich-app/immich/discussions/8565


michelheusschen (approved):

danieldietzler (approved):

createchange:
Thanks for listening to your users. I just returned from a trip and now have an album that makes chronological sense when providing it to family and friends.

PR #10559, opened 2024-06-23
refactor(ml): model sessions

mertalev:
### Description

The ML service has some coupling with ONNX Runtime: much of the logic in the base inference class is only relevant to this backend, while handling for other backends is treated as a special case. This PR abstracts ONNX Runtime into a `ModelSession` wrapper similar to how ARM NN is handled.

The catalyst for this change is that facial recognition recently broke for ARM NN due to a change that only applied to ONNX Runtime. As part of this, it also adds relevant tests for facial recognition, as well as `AnnSession` in general as its test coverage was quite low.

### How Has This Been Tested?

Tested that the endpoint still works for each model task using ONNX Runtime. I'm unable to test ARM NN.

### DIFF
--- a/machine-learning/ann/ann.py
+++ b/machine-learning/ann/ann.py
@@ -52,8 +52,6 @@ class Ann(metaclass=_Singleton):
     def __init__(self, log_level: int = 3, tuning_level: int = 1, tuning_file: str | None = None) -> None:
         if not is_available:
             raise RuntimeError("libann is not available!")
-        if tuning_file and not exists(tuning_file):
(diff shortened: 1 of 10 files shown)


fyfrey:
Great refactoring, I'll give this a spin on ARM in few days

jrasm91 (approved):

fyfrey (approved):
Tested on my prod instance, ARM-NN face detection&recognition models work fine again with this change :)

zackpollard (approved):

PR #10601, opened 2024-06-24
chore(web): translation finetuning

waclaw66:
Translation finetuning.


web/src/lib/i18n/en.json
```diff
@@ -900,7 +899,7 @@
   "removed_api_key": "Removed API Key: {name}",
   "removed_from_archive": "Removed from archive",
   "removed_from_favorites": "Removed from favorites",
-  "removed_from_favorites_count": "Removed {count} from favorites",
+  "removed_from_favorites_count": "{count, plural, other {Removed #}} from favorites",
```
danieldietzler, line 902: Why would this be better?

waclaw66: "Removed" can be inflected in other languages.

danieldietzler: Oh, interesting. TIL

danieldietzler (approved):

jrasm91 (approved):

PR #10612, opened 2024-06-25
chore: build metadata

jrasm91:
- Add a new endpoint `GET /server-info/about`, which returns detailed version info
- Change the behavior of clicking on the version number in the server status box to instead open a new "about" modal
- Hopefully this helps with debugging and verifying (also hopefully useful for soon-to-come preview environments) 
 
| Light | Dark |
| - | - |
| image | image |


martabal:
Looks great ! But I don't know if adding the build URL is relevant: GitHub deletes builds after a certain time.

server/src/services/server-info.service.ts
```diff
@@ -42,6 +47,35 @@ export class ServerInfoService {
...
+      buildUrl: buildId ? `https://github.com/immich-app/immich/actions/runs/${buildId}` : '',
+
+      gitRef,
+      gitCommit,
+      gitCommitUrl: gitCommit ? `https://github.com/immich-app/immich/commit/${gitCommit}` : '',
+
+      imageTag,
+      imageUrl: imageTag
+        ? `https://github.com/immich-app/immich/pkgs/container/immich-server/234402559?tag=${imageTag}`
+        : '',
```
martabal, line 69: Can URLs be changed? + it doesn't make sense to have an imageUrl for unofficial deployments like the Immich package in the AUR

jrasm91: I could just make them configurable like everything else and pass them in as build args. That way you could link to your own releases.

danieldietzler (approved):

PR #10613, opened 2024-06-25
chore(deps): use full semver docker tag for node images

jrasm91:
Migrate to tags with full semver (with the idea that renovate PRs should update the tags with the actual version bumps, instead of just changing the hash).

If this works, I can make a separate PR in the base-images repo.

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -5,15 +5,15 @@ RUN apt-get install --no-install-recommends -yqq tini
 WORKDIR /usr/src/app
 COPY server/package.json server/package-lock.json ./
 RUN npm ci && \
-    # exiftool-vendored.pl, sharp-linux-x64 and sharp-linux-arm64 are the only ones we need
-    # they're marked as optional dependencies, so we need to copy them manually after pruning
-    rm -rf node_modules/@img/sharp-libvips* && \
-    rm -rf node_modules/@img/sharp-linuxmusl-x64
+  # exiftool-vendored.pl, sharp-linux-x64 and sharp-linux-arm64 are the only ones we need
+  # they're marked as optional dependencies, so we need to copy them manually after pruning
+  rm -rf node_modules/@img/sharp-libvips* && \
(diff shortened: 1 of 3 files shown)


danieldietzler (approved):
Why haven't we had this already? lmao

bo0tzz (approved):
Definitely better to set the full tag, yeah.

Note that pinDigests is enabled so renovate will create PRs to add shas again:
https://github.com/immich-app/immich/blob/d7a33c8ec24bb66bd82472bf58045442349bca05/renovate.json#L3

danieldietzler:
> Definitely better to set the full tag, yeah.
> 
> Note that pinDigests is enabled so renovate will create PRs to add shas again:
> 
> https://github.com/immich-app/immich/blob/d7a33c8ec24bb66bd82472bf58045442349bca05/renovate.json#L3

That probably was intentionally left out as it's just easier to let renovate add the hashs later than having to look them up yourself :D

bo0tzz:
I thought the intention was to remove the hash, that's why I mentioned it. If we wanna keep it then that's fine of course.

server/Dockerfile
```diff
@@ -25,7 +25,7 @@ COPY --from=dev /usr/src/app/node_modules/@img ./node_modules/@img
 COPY --from=dev /usr/src/app/node_modules/exiftool-vendored.pl ./node_modules/exiftool-vendored.pl
 
 # web build
-FROM node:iron-alpine3.18@sha256:53108f67824964a573ea435fed258f6cee4d88343e9859a99d356883e71b490c as web
+FROM node:20.13.1-alpine3.18
```
bo0tzz, line 28:
```suggestion
FROM node:20.13.1-alpine3.18 as web
```

bo0tzz (changes requested):
Oopsie

zackpollard (approved):

bo0tzz (approved):

PR #10622, opened 2024-06-25
fix(server): keep full datetime precision from database

fyfrey:
we have had these sync issues due to JavaScript reducing precision of DateTime when converting to JS Date. This still affects changes for albums, users etc. in the app and can result in not doing an update while it would be needed etc. This draft PR changes all related fields (`createdAt`, `updatedAt`, `deletedAt`) in the server entities and DTOs to type `string` instead of `Date`. DateTime information is passed through the JavaScript code as full precision ISO Strings.


bo0tzz:
Breaking change, right?

fyfrey:
> Breaking change, right?

no API changes at all.

bo0tzz:
Ah, the API type was already string of course. Nice :D

jrasm91 (commented):
LGTM

jrasm91:
With kysely we should be able to move to using luxon DateTime objects everywhere, at that point we can migrate the dto parsing to convert incoming requests to those objects as well, and this should also solve the problems/limitations of javascript Date objects.

PR #10642, opened 2024-06-26
fix(server): user delete with stacked assets

jrasm91:
Fixes #1046

Delete stacks before assets are deleted (during user delete)


danieldietzler (approved):
LGTM

PR #10644, opened 2024-06-26
feat(web): link router

jrasm91:
- Add a common entrypoint for external links
- Fix fallback error message for non http errors (they were being suppressed before)

Examples:
- http://localhost:2283/link?target=view_asset&id=107cd029-07c3-44b9-af12-f6770d56122f
- http://localhost:2283/link?target=unsubscribe

The goal of this component is to make it possible to change internal routes without breaking existing links in emails, docs, etc.

### DIFF
--- a/web/src/hooks.client.ts
+++ b/web/src/hooks.client.ts
@@ -25,13 +25,13 @@ const parseError = (error: unknown, status: number, message: string) => {
   }
 
   return {
-    message: message || DEFAULT_MESSAGE,
+    message: (error as Error)?.message || message || DEFAULT_MESSAGE,
     code: status,
   };
 };
 
 export const handleError: HandleClientError = ({ error, status, message }) => {
   const result = parseError(error, status, message);
-  console.error(`[hooks.client.ts]:handleError ${result.message}`);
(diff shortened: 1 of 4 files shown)


danieldietzler (approved):
Awesome! 🚀

alextran1502 (approved):

web/src/routes/link/+page.svelte
michelheusschen, line 1: Why not redirect inside the `+page.ts` load function?

web/src/lib/i18n/en.json
```diff
@@ -880,6 +880,7 @@
   "reassing_hint": "Assign selected assets to an existing person",
   "recent": "Recent",
   "recent_searches": "Recent searches",
+  "redirecting": "Redirecting...",
```
michelheusschen, line 883: This seems unused

jrasm91: Oh yeah you are right. I had it at one point.

web/src/routes/link/+page.svelte
jrasm91: I have no idea lol.

PR #8757, opened 2024-04-12
fix(web): ui tweaks

mgabor3141:
# Changes

## remove height limit from admin page user management list for better scrolling

Screenshot 2024-04-13 at 09 12 05 Screenshot 2024-04-13 at 09 10 36
Before / After

## move slideshow button out from menu so that non-owners can see it
Closes #8383

Screenshot 2024-04-13 at 09 13 37

## fix activity covering up video player controls
Closes #6191

### Before

Screenshot 2024-04-13 at 09 15 02

### After
Screenshot 2024-04-13 at 00 36 42
Screenshot 2024-04-13 at 00 37 08
Screenshot 2024-04-13 at 00 36 52

PR #10667, opened 2024-06-27
fix(web): remove black bezels + better integrate ActivityStatus

### LINKED ISSUE #10526: Web: black bezels in video viewer
### The bug

A change was made in #8757 to close #6191, with a Google-derived like/comment UI control overlapping HTML video player volume/settings/fullscreen controls. Unfortunately, the change simply added a bottom bezel and so shrank the video player, which has made the UI inconsistent with photos, most notably when clicking to play a Live Photo (which then shrinks).

> style="height: calc(100% - 64px)" \[see [original](https://github.com/immich-app/immich/blob/42f3b50422a4f3793480b25989b57956a57c222f/web/src/lib/components/asset-viewer/video-native-viewer.svelte#L57)\]

image

The [discussion](https://github.com/immich-app/immich/issues/6191#issuecomment-2052647784) mentions several better ways of handling the problem, most notably shifting the Google controls up on videos, which would be inconsistent with their position on photos, but not annoyingly so. The bezeling change is justified by "This would have different effects on different screen sizes, but most of the time it would be an improvement.", which I think is a dubious claim. The Google controls don't even show up on my client and I don't want or know how to make them do so.

Google Photos itself doesn't even bother lol (which is still preferable to the bezels imo).

image

Debian 12

### Version of Immich Server

v1.106.3

n/a

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
-
```

### Your .env content

```Shell
-
```

### Reproduction steps

```bash
Always happens on web video player.
```

### Relevant log output

_No response_

### Additional information

_No response_

pyorot:
My entire description was deleted so I'll keep it brief I guess.

Fixes #10526 

image

Couldn't test properly cos the headers-already-sent bug broke the video and changed the appearance of the controls but think it's OK.

Tried to right-align the buttons and make ActivityControl as-if a kind of second-row from the bottom, symmetrical to the top row. **TODO**: It should be first-row from bottom on PhotoViewer but idk how to condition in Svelte macros for PhotoViewer vs VideoViewer appearing.

Tried to also make ActivityControl less obtrusive, so removed the "Say Something" text (because it does not fade out nor disappear if the viewport width is decreased). I left the code in a comment and noted these two conditions for maybe bringing it back, tho it's a bit subjective whether it's useful or not.

### DIFF
--- a/web/src/lib/components/asset-viewer/activity-status.svelte
+++ b/web/src/lib/components/asset-viewer/activity-status.svelte
@@ -3,7 +3,6 @@
   import { mdiCommentOutline, mdiHeart, mdiHeartOutline } from '@mdi/js';
   import { createEventDispatcher } from 'svelte';
   import Icon from '../elements/icon.svelte';
-  import { t } from 'svelte-i18n';
 
   export let isLiked: ActivityResponseDto | null;
   export let numberOfComments: number | undefined;
--- a/web/src/lib/components/asset-viewer/activity-status.svelte
+++ b/web/src/lib/components/asset-viewer/activity-status.svelte
@@ -17,7 +17,7 @@
 </script>
 
 <div
-  class="w-full h-14 flex p-4 text-white items-center justify-center rounded-full gap-4 bg-immich-dark-bg bg-opacity-60"
+  class="w-full h-10 flex p-4 text-white items-center justify-center rounded-full gap-4 bg-immich-dark-bg bg-opacity-60"
 >
   <button type="button" class={disabled ? 'cursor-not-allowed' : ''} on:click={() => dispatch('favorite')} {disabled}>
     <div class="items-center justify-center">
@@ -29,8 +29,9 @@
       <Icon path={mdiCommentOutline} class="scale-x-[-1]" size={24} />
       {#if numberOfComments}
         <div class="text-xl">{numberOfComments}</div>
-      {:else if !isShowActivity}
-        <div class="text-lg">{$t('say_something')}</div>
+      <!-- the say_something text is removed for being obtrusive but could be put back in -->
+      <!-- if it disappears on narrowing the viewport or once user control auto fadeout is implemented -->
+      <!-- {:else if !isShowActivity} <div class="text-lg">{$t('say_something')}</div> -->
       {/if}
     </div>
   </button>
--- a/web/src/lib/components/asset-viewer/activity-status.svelte
+++ b/web/src/lib/components/asset-viewer/activity-status.svelte
@@ -7,7 +7,6 @@
 
   export let isLiked: ActivityResponseDto | null;
   export let numberOfComments: number | undefined;
-  export let isShowActivity: boolean | undefined;
   export let disabled: boolean;
 
   const dispatch = createEventDispatcher<{
@@ -16,9 +15,7 @@
   }>();
 </script>
 
-<div
-  class="w-full h-10 flex p-4 text-white items-center justify-center rounded-full gap-4 bg-immich-dark-bg bg-opacity-60"
->
+<div class="w-full flex text-white items-center justify-center rounded-full gap-4 bg-immich-dark-bg bg-opacity-60">
   <button type="button" class={disabled ? 'cursor-not-allowed' : ''} on:click={() => dispatch('favorite')} {disabled}>
     <div class="items-center justify-center">
       <Icon path={isLiked ? mdiHeart : mdiHeartOutline} size={24} />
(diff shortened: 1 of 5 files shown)


pyorot:
An extra comment – I wasn’t sure how to handle removing the text so left the checks, unused variables etc for review. My thoughts were that it made the most sense now to do it, also because no other controls have text next to them, but that it could debatably be useful in future.

web/src/lib/components/asset-viewer/activity-status.svelte
```diff
@@ -29,8 +29,9 @@
       <Icon path={mdiCommentOutline} class="scale-x-[-1]" size={24} />
       {#if numberOfComments}
         <div class="text-xl">{numberOfComments}</div>
-      {:else if !isShowActivity}
-        <div class="text-lg">{$t('say_something')}</div>
+      <!-- the say_something text is removed for being obtrusive but could be put back in -->
+      <!-- if it disappears on narrowing the viewport or once user control auto fadeout is implemented -->
+      <!-- {:else if !isShowActivity} <div class="text-lg">{$t('say_something')}</div> -->
```
michelheusschen, line 34: How about only hiding it on mobile?

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -682,7 +682,7 @@
             />
           {/if}
           {#if $slideshowState === SlideshowState.None && isShared && ((album && album.isActivityEnabled) || numberOfComments > 0)}
-            <div class="z-[9999] absolute bottom-0 right-0 mb-4 mr-6">
+            <div class="z-[9999] absolute bottom-0 right-0 mb-12 mr-4 justify-self-end">
```
michelheusschen, line 685: I don't think `justify-self-end` works here. What browser is your screenshot from? Getting the bottom spacing right is pretty difficult, because there's a lot of variability between browsers and even within the same browser when zooming.

web/src/lib/components/asset-viewer/activity-status.svelte
pyorot: Maybe the way forward is lots of screenshots. Fortunately, the usual-looking video player is back today on my setup lmao.

pyorot: First idea: disable the text whenever the info pane is showing (existing behaviour: only if activity pane is showing.
(I will always be pro–always disable personally because no other user buttons in the asset viewer have text).

**immich (release):**
image

**immich (pr branch):**
image

**google photos:**
image

pyorot: Second idea: <600px viewport width (examples are 550×400px)

**immich (release):**
image

**immich (pr branch):**
image

**google photos:**
image

google photos' logic is basically:
- side pane showing → activity status: no text + bottom row
- side pane hidden + <600px width → activity status: no text + third row ish
- side pane hidden + >600px width → activity status: yes text + bottom row

pyorot: On google photos, this overlap is annoying, but let's say they fixed it by raising the activity status to third row as they do on narrow width viewport. Then the text is still unobtrusive, but only because it fades out after a few seconds, which immich doesn't currently do (is probably fairly important to work on for release?).

image

web/src/lib/components/asset-viewer/asset-viewer.svelte
pyorot: The browser is windows chrome, same as today's screenshots (the video controls yesterday changed owing to videos consistently failing to load). `justify-self-end` is imported from reverting the change that introduced the black bezels (#8757), but I can look further into it.

pyorot: I am not noticing a difference from removing `justify-self-end` using dev tools so it was possibly a trivial fix in the aforementioned change.

pyorot: Zooming seems to keep things consistent on windows chrome, though things overlap more in all cases since the usual video control style came back to my setup, so maybe I would advise switching to row 2.5 rather than 2 from the bottom (where row 1 is by definition symmetric to the top row).

image

image

pyorot: Windows firefox looks superb on pr branch at all zoom levels (within reason haha).
image
image
My iPhone won't connect to the Vite server unfort.

pyorot:
r.e. mr-4 → mr-3: this is mr-3. Obviously these are all approximations since the styling is implemened differently to the top control bar.

image

pyorot:
With this PR, I'm personally happy to merge any change that gets rid of the black bottom margin, even if it's a straight reversion to where the controls overlap again. So I'm happy to take any design cues for where to put the controls and what text to show when.

alextran1502 (approved):

PR #10672, opened 2024-06-28
chore(ml): target arm64 cuda

mertalev:
## Description

There are cases where CUDA is used with an arm64 architecture, most notably with Jetson devices.

Fixes #10647


mertalev:
Hmm, looks like `onnxruntime-gpu` itself doesn't have arm64 wheels.

PR #10688, opened 2024-06-28
feat(web): show favorite icon in duplicate asset

Pascal-So:
This show the favorite-heart on the duplicate review page:

screenshot of the updated duplicate review UI

Note that here it is shown in the top left rather than the bottom left, because the view button is already located in the bottom right. I also experimented with moving the heart up above the view button, but that didn't look as nice in my opinion:

alternative design where the heart is located in the bottom left

Feel free to object if you think otherwise.

I think this change would be beneficial for the following reason: When reviewing duplicates, I find it useful to know if one of the assets is favorited, because in that case I would then want to ensure afterwards that the asset that I keep is favorited. That way I avoid losing information during the cleanup, i.e. the "isFavorite" information.

What do you think about this suggested change?

### DIFF
--- a/web/src/lib/components/utilities-page/duplicates/duplicate-asset.svelte
+++ b/web/src/lib/components/utilities-page/duplicates/duplicate-asset.svelte
@@ -1,10 +1,10 @@
 <script lang="ts">
   import Icon from '$lib/components/elements/icon.svelte';
-  import { getAssetThumbnailUrl } from '$lib/utils';
+  import { getAssetThumbnailUrl, isSharedLink } from '$lib/utils';
   import { getAssetResolution, getFileSize } from '$lib/utils/asset-utils';
   import { getAltText } from '$lib/utils/thumbnail-util';
   import { getAllAlbums, type AssetResponseDto } from '@immich/sdk';
-  import { mdiMagnifyPlus } from '@mdi/js';
+  import { mdiHeart, mdiMagnifyPlus } from '@mdi/js';
   import { t } from 'svelte-i18n';
 
   export let asset: AssetResponseDto;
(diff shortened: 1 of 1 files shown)


web/src/lib/components/utilities-page/duplicates/duplicate-asset.svelte
```diff
@@ -38,6 +38,13 @@
         draggable="false"
       />
 
+      <!-- FAVORITE ICON -->
+      {#if !isSharedLink() && asset.isFavorite}
```
Pascal-So, line 42: I don't actually understand why the `!isSharedLink` is here. I copied this snippet from `thumbnail.svelte` so I'm guessing we also need this condition here? Let me know if I should take it out :smile:

jrasm91: You don't need it. We hide those types of icons when you are viewing stuff as an anonymous user.

Pascal-So: okay, I've removed the condition

alextran1502 (approved):

alextran1502:
Nice job, thanks!

bo0tzz:
Nice! Imo having the favorite icon always be in a consistent place would be nice, feel free to move the view button out of the way to another corner if you want.

Pascal-So:
@bo0tzz okay I moved the heart back to `bottom-2 left-2` just like in `thumbnail.svelte`.

As for the view button, the only thing in immich that i could kinda consider a precedent to guide its placement is the select button in the asset grid, so I chose to place it in the top left just like the select button.

image

It's not really the same functionality (select vs view) but overall this feels the least inconsistent to me. What do you think?

bo0tzz (approved):

PR #10696, opened 2024-06-29
feat(web): html tags inside plural and select messages

michelheusschen:
Adds support for using html tags in messages like `{count, plural, other {<b>#</b> items}}` which would previously result in an error. The implementation isn't perfect, because it parses/formats the plural and select parts twice. Suggestions to improve that are welcome.


danieldietzler (approved):
After going through it many times I think I have a good grasp of what's happening. The code is as clean as it could be imo and I don't really have an idea of how to approach this differently, let alone a good one.

alextran1502:
On top of the FormatMessage component, can we add documentation of how this element can be used, the variation, and the message it will output?

michelheusschen:
I've added some documentation to the component

image

alextran1502:
Amazing! Thanks!

PR #10750, opened 2024-07-02
feat(cli): use a queue for duplicate and upload

### LINKED ISSUE #5839: CLI retry failed request and don't cancel entire flow on error
Currently the CLI will error and cancel the whole process if one request fails. We should probably instead retry a couple of times and then log an error that the asset failed, but continue on to other assets.

masterT:
Using a queue to process the files makes the file duplicate detection and asset upload more stable and tolerant of network errors. If an error occurs, the whole command will not stop; the task will be retried (3 times) before logging the error and moving to the next step.

The new queue abstraction is using [fastq](https://www.npmjs.com/package/fastq) internally.

Fixes: #5839

### DIFF
--- a/cli/src/commands/asset.ts
+++ b/cli/src/commands/asset.ts
@@ -16,6 +16,7 @@ import { chunk } from 'lodash-es';
 import { Stats, createReadStream } from 'node:fs';
 import { stat, unlink } from 'node:fs/promises';
 import path, { basename } from 'node:path';
+import { Queue } from 'src/queue';
 import { BaseOptions, authenticate, crawl, sha1 } from 'src/utils';
 
 const s = (count: number) => (count === 1 ? '' : 's');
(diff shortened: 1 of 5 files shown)


cli/src/commands/asset.ts
```diff
@@ -99,28 +100,45 @@ const checkForDuplicates = async (files: string[], { concurrency, skipHash }: Up
...
-    }
-  } finally {
-    progressBar.stop();
+      progressBar.increment(filepaths.length);
+      return results;
+    },
+    { concurrency, retry: 3 },
+  );
+
+  for (const items of chunk(files, concurrency)) {
```
masterT, line 124: I reproduced the same behaviour, but I'm unsure if the file should be batched by `concurrency`; I tested it with 100, and it worked fine. Maybe the batch size should be hardcoded to a greater number than the default `concurrency` value.

masterT:
I tested those changes with a 62.7 GB upload (12k files), but I did not write any tests; let me know if you would like some tests. 🙂

zackpollard:
> I tested those changes with a 62.7 GB upload (12k files), but I did not write any tests; let me know if you would like some tests. 🙂

If you would be willing to write some tests, that would be awesome! We're always trying to increase our coverage to decrease the chance of regressions in future releases :slightly_smiling_face:

masterT:
> > I tested those changes with a 62.7 GB upload (12k files), but I did not write any tests; let me know if you would like some tests. 🙂
> 
> If you would be willing to write some tests, that would be awesome! We're always trying to increase our coverage to decrease the chance of regressions in future releases 🙂

I see no setup for e2e/integration test (testing a command). Would you prefer unit tests or an e2e/integration test?

jrasm91:
There are e2e tests for the cli inside of the `e2e/` folder

masterT:
@zackpollard @jrasm91 I'm not sure what the best strategy for testing the upload retry could be. This test may not be an e2e test. I thought of mocking the HTTP request in the e2e tests with something like [vitest-fetch-mock](https://www.npmjs.com/package/vitest-fetch-mock), but since we are executing a shell command to run the CLI, it's not possible. 

What do you suggest?

jrasm91:
Yeah, good point. I don't see any great way to accomplish that in our actual e2e environment. The only thing i can think of is doing something like executing `docker restart immich_server` or something. Not sure if that is a good, reliable option or not.

masterT:
> Yeah, good point. I don't see any great way to accomplish that in our actual e2e environment. The only thing i can think of is doing something like executing `docker restart immich_server` or something. Not sure if that is a good, reliable option or not.

It feels like it could introduce flaky tests as the request retry will exhaust until the server has restarted.

jrasm91:
Yeah, probably. You could always add some unit tests that run with vitest where you mock the sdk methods and have one reject with a network error.

cli/src/commands/asset.ts
```diff
@@ -158,37 +177,52 @@ const uploadFiles = async (files: string[], { dryRun, concurrency }: UploadOptio
...
+      } else {
+        successCount++;
+        successSize += stats.size ?? 0;
+      }
+
+      uploadProgress.update(successSize, { value_formatted: byteSize(successSize + duplicateSize) });
+
+      return response;
+    },
+    { concurrency, retry: 3 },
```
masterT, line 201: Maybe this could be passed as an option through the CLI.

cli/src/commands/asset.ts
```diff
@@ -99,32 +100,50 @@ const checkForDuplicates = async (files: string[], { concurrency, skipHash }: Up
...
         }
-        progressBar.increment();
       }
-    }
-  } finally {
-    progressBar.stop();
+      progressBar.increment(filepaths.length);
+      return results;
+    },
+    { concurrency, retry: 3 },
```
masterT, line 121: Maybe this could be passed as an option through the CLI.

masterT:
@zackpollard @jrasm91 This is ready for review.

I did my best to mock with accuracy, inspecting the requests made in the e2e tests. 

I'm waiting for your feedback! 🙂

alextran1502 (approved):

PR #10752, opened 2024-07-02
fix(web): scroll jank on memories page

### LINKED ISSUE #10731: Jank when folding/unfolding memories gallery
https://github.com/immich-app/immich/assets/11602424/2397b514-5c9e-4da8-b6a3-243335a1fe65

michelheusschen:
Fixes #10731 by using fixed positioning to prevent changes in page height. That stops the intersection observer from firing rapidly, which was causing the page to jump up and down. Also changed margin to padding to ensure the button scrolls all the way to the bottom.


zackpollard (approved):
Tested and LGTM, thanks!

PR #10809, opened 2024-07-03
docs: Update Nginx Reverse Proxy Configuration for Improved iOS Compatibility

MFYDev:
## Introduction
I encountered an issue with the current Nginx reverse proxy configuration in the documentation, which prevented videos from playing on iOS clients. After troubleshooting, I found a solution and would like to propose an update to improve the documentation.

## Issue Description

Related issue: #10777 

The initial configuration provided in the documentation did not support video playback on iOS devices. The videos failed to load, and I did not receive any specific error messages in the logs.

## Solution
The following changes resolved the issue:

- Disabled proxy_buffering to avoid issues with video streaming.
- Adjusted buffer sizes with proxy_buffer_size, proxy_busy_buffers_size, and proxy_buffers.
- Enabled proxy_force_ranges to ensure proper handling of byte-range requests, which is crucial for video playback.
- Extended timeouts (proxy_send_timeout, proxy_read_timeout, client_body_timeout, keepalive_timeout) to prevent interruptions during large file uploads or long video streams.

## Conclusion
This updated configuration resolves the issue of video playback on iOS clients and improves overall performance. I encourage the maintainers to review and test these changes and consider updating the documentation to assist others who might face similar challenges.

Thank you for your consideration.


docs/docs/administration/reverse-proxy.md
```diff
@@ -30,9 +30,21 @@ server {
...
+    proxy_read_timeout  600s;
+    proxy_send_timeout  600s;
+    send_timeout        600s;
+    client_body_timeout 120s;
+    keepalive_timeout   150s;
+
+    # improve video streaming for mobile client
+    gzip_static on;
+    gzip_min_length 1000;
+    gzip_comp_level 2;
```
bo0tzz, line 42: I think this is unnecessary since nginx is not serving static files.

docs/docs/administration/reverse-proxy.md
```diff
@@ -30,9 +30,21 @@ server {
...
+    proxy_buffering off;
+    proxy_buffer_size 16k;
+    proxy_busy_buffers_size 24k;
+    proxy_buffers 64 4k;
+    proxy_force_ranges on;
```
bo0tzz, line 47: afaik with `proxy_buffering off;` none of the other proxy_ directives have any effect.

MFYDev: Please take a look on the issue #10777 and after I added this part for proxy buffering, the ios client video playback issue got solved directly.

The original configuration did not specify any buffering settings. By default, Nginx enables proxy buffering, which can cause issues when streaming video content as shown in #10777 

The new configuration explicitly disables proxy buffering with proxy_buffering off; and sets various buffer sizes (proxy_buffer_size, proxy_busy_buffers_size, proxy_buffers). Disabling buffering allows Nginx to pass the data directly to the client as it arrives, which is crucial for streaming videos.

bo0tzz: What I'm saying is that with buffering off, there's no need to set any buffer sizes since those aren't used.

MFYDev: You are right, but what happened to me is also weird. I just tested on my own webserver, with only `proxy_buffering off;` The video content directly went blank on my IOS client, and adding the rest of it make the video can play again.

MFYDev: I tried to investigated in my nginx, none of my nginx config set proxy buffers anywhere else. So I myself don't know why it is behaving like this, I will follow your advice. If it has too much confusion I can drop the PR

PR #10849, opened 2024-07-04
feat(web): use browser language by default

michelheusschen:
The default language is set using the first match in `navigator.languages`, with english as fallback. Handling languages like `ar-AE` which should use the less specific `ar`, adds some complexity. Although `svelte-i18n` already uses the less specific language internally, it's not exposed so we can't use it. This means some extra logic is needed to determine which language is selected by default.

Less specific languages such as `zh` are ignored, because you can't safely choose between zh-Hans or zh-Hant.

### DIFF
--- a/web/src/lib/components/user-settings-page/app-settings.svelte
+++ b/web/src/lib/components/user-settings-page/app-settings.svelte
@@ -14,9 +14,10 @@
     sidebarSettings,
   } from '$lib/stores/preferences.store';
   import { findLocale } from '$lib/utils';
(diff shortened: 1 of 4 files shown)


waclaw66:
Fixes https://github.com/immich-app/immich/discussions/10720.

danieldietzler (commented):
Am I blind or do we not support `dev` as a language anymore (to show the keys)

michelheusschen:
It's still there for me

image

danieldietzler (approved):
Oh now it's there, nvm. Thank you! :)

PR #10866, opened 2024-07-04
feat(server): search unknown place

jforseth210:
This is a rough draft for allowing users to search for photos that specifically don't have a location (or at least a country) as requested in #8362. 

It updates the server to accept `null` as a value for country, which tells the database to search for assets with no country `exifInfo`. 

The frontend is updated to automatically include an "Unknown" option in the country combo box, which is passed to the server as `{"country":null}`. The filter list in the results is also updated to display "Unknown" if the value of a filter is `null`. 

I don't have a ton of experience contributing to FOSS, so this probably is a ways from merge ready. Let me know what I need to do differently. 

- I suspect there's something I need to update in `/open-api` but I'm not sure what.
- Removing the validators in `search.dto.ts` makes me a bit nervous. Is there anything that's expecting a string that I missed?
- I'd love some advice to make `selectedOption` easier to read
- I haven't touched the mobile app at all


jforseth210:
I'm not really sure how to correctly update the types for `@immich/sdk`. Tried reading through the docs but still not getting it... Where's the best place to get a hand? Discord?

danieldietzler:
> I'm not really sure how to correctly update the types for `@immich/sdk`. Tried reading through the docs but still not getting it... Where's the best place to get a hand? Discord?

`make open-api`. But yeah, generally joining our discord (https://discord.immich.app) is a good idea

server/src/dtos/search.dto.ts
```diff
@@ -84,8 +84,6 @@ class BaseSearchDto {
   @Optional()
   state?: string;
 
-  @IsString()
-  @IsNotEmpty()
```
jrasm91, line 88: You need to add these back. The next line needs to be `@Optional({ nullable: true })` instead.

jforseth210:
I think that last commit should fix those api issues. Thanks @danieldietzler and @jrasm91 for the help! What's next?

web/src/lib/components/shared-components/search-bar/search-location-section.svelte
```diff
@@ -45,7 +49,16 @@
...
+  async function updateCities(country?: string | null, state?: string | null) {
+    if (country === null) {
+      cities = [];
+      return;
+    }
+    if (state == null) {
+      cities = [];
+      return;
+    }
+
```
jrasm91, line 61: Combine these into one statement

jforseth210: No problem!

danieldietzler (approved):

jrasm91 (approved):
Thanks for the PR! I made a few changes, but I think it is good to go now:
- Moved the logic to return "null" options to the server. This means they are only shown in the UI if there are actually valid results (given other criteria) with null values.
- Updated the api to accept empty string or null. (The mobile app can't actually send null, and empty string makes the combobox types a lot simpler)
- Added some e2e tests for searching with empty/null
- Expanded the scope to include camera make, camera model, and camera lens.

danieldietzler (approved):
LGTM

server/src/repositories/metadata.repository.ts
```diff
@@ -60,49 +60,42 @@ export class MetadataRepository implements IMetadataRepository {
...
   @GenerateSql({ params: [DummyValue.UUID] })
   async getCountries(userId: string): Promise<string[]> {
-    const entity = await this.exifRepository
+    const results = await this.exifRepository
       .createQueryBuilder('exif')
       .leftJoin('exif.asset', 'asset')
       .where('asset.ownerId = :userId', { userId })
-      .andWhere('exif.country IS NOT NULL')
-      .select('exif.country')
+      .select('exif.country', 'exif_country')
```
mertalev, line 67: Is there a particular reason you set an alias here and make it a raw query? You can remove the alias and use `getMany` like the original, just removing the `filter` at the end.

server/src/utils/database.ts
```diff
@@ -48,7 +48,13 @@ export function searchAssetBuilder(
       ? builder.leftJoinAndSelect(`${builder.alias}.exifInfo`, 'exifInfo')
       : builder.leftJoin(`${builder.alias}.exifInfo`, 'exifInfo');
 
-    builder.andWhere({ exifInfo });
+    for (const [key, value] of Object.entries(exifInfo)) {
+      if (value === null || value === '') {
+        builder.andWhere(`exifInfo.${key} IS NULL`);
```
mertalev, line 53: An empty string is not the same as null and shouldn't have this behavior.

web/src/lib/components/shared-components/search-bar/search-filter-box.svelte
```diff
@@ -42,18 +42,22 @@
   const toStartOfDayDate = (dateString: string) => parseUtcDate(dateString)?.startOf('day').toISODate() || undefined;
   const dispatch = createEventDispatcher<{ search: SmartSearchDto | MetadataSearchDto }>();
 
+  function withoutNull<T>(value: T | null) {
```
mertalev, line 45: I'm not a huge fan of this. It adds a level of indirection just for a || or ?? operator. Just use the operator.

web/src/routes/(user)/search/[[photos=photos]]/[[assetId=id]]/+page.svelte
```diff
@@ -259,6 +259,8 @@
             {#await getPersonName(value) then personName}
               {personName}
             {/await}
+          {:else if value === null || value === ''}
```
Comment (mertalev) [inline web/src/routes/(user)/search/[[photos=photos]]/[[assetId=id]]/+page.svelte:262]:
```suggestion
          {:else if !value}
```

web/src/lib/components/shared-components/search-bar/search-camera-section.svelte
```diff
@@ -26,13 +26,19 @@
       $type: SearchSuggestionType.CameraMake,
       model,
     });
+    if (filters.make && !makes.includes(filters.make)) {
+      filters.make = undefined;
+    }
```
mertalev, line 31: Why is this necessary?

server/src/dtos/search.dto.ts
```diff
@@ -75,34 +75,28 @@ class BaseSearchDto {
...
 
   @IsString()
-  @IsNotEmpty()
-  @Optional()
-  model?: string;
+  @Optional({ nullable: true })
+  model?: string | null;
 
   @IsString()
-  @IsNotEmpty()
```
mertalev, line 103: I think the request should explicitly be null and not an empty string, so this doesn't need to be removed.

mertalev (commented):
Thanks for working on this! I think some of the backend behavior is strange and surprising for the sake of letting the frontend send '' instead of null. I'd much prefer if you could make the frontend send null instead.

mertalev:
> The mobile app can't actually send null

How can it not send null? That's so weird. In that case I'd rather handle the '' => null conversion as a transformer without polluting other code.

server/src/repositories/metadata.repository.ts
jrasm91: Right, when I did that it only returned one result (United States) instead of two (United States, null).

web/src/lib/components/shared-components/search-bar/search-filter-box.svelte
jrasm91: How do you propose doing it instead? `country: searchQuery.country || undefined` doesn't work since both empty string, 0 and null map to `undefined`. `country: searchQuery.country ?? undefined` also doesn't work, since it maps `null` to `undefined`.

web/src/lib/components/shared-components/search-bar/search-camera-section.svelte
jrasm91: This just matches what was in the location filter file. If the current set filter isn't in the list of valid options it is discarded.

web/src/routes/(user)/search/[[photos=photos]]/[[assetId=id]]/+page.svelte
Comment (jrasm91) [reply web/src/routes/(user)/search/[[photos=photos]]/[[assetId=id]]/+page.svelte:262]: That might work right now, it's not actually correct. It would make searching for a numeric value of 0 or `false` show the wrong value. `!value` impies "falsy", but what we actually want is "null or empty string".

server/src/dtos/search.dto.ts
jrasm91: I could do that, at least for the web. But the mobile app can't actually send `null` values through the generated client. So either the mobile app doesn't get this functionality or we need to use an empty string in the api to indicate querying for null.

web/src/routes/(user)/search/[[photos=photos]]/[[assetId=id]]/+page.svelte
Comment (mertalev) [reply web/src/routes/(user)/search/[[photos=photos]]/[[assetId=id]]/+page.svelte:262]: That's fair, it could lead to wrong behavior down the line.

web/src/lib/components/shared-components/search-bar/search-camera-section.svelte
mertalev: Sure, but when can that actually happen?

server/src/repositories/metadata.repository.ts
mertalev: Oh, maybe because you don't select the `id` field. That's fine, it can be a raw query. Maybe just make the alias camel case (or remove it).

web/src/lib/components/shared-components/search-bar/search-camera-section.svelte
jrasm91: I think only by directly navigating with those params. Just seemed good to make them consistent, but maybe it isn't necessary

jforseth210:
Is there anything I still need to do on this?

mertalev:
The empty string / null handling is the main outstanding issue. It'd be great if you could add a '' => null transformer in the request DTO and remove the special-casing of empty strings for this elsewhere.

web/src/lib/components/shared-components/search-bar/search-camera-section.svelte
mertalev: It can be removed IMO, but it's fine if you prefer to keep it.

e2e/src/api/specs/search.e2e-spec.ts
```diff
@@ -108,7 +105,7 @@ describe('/search', () => {
     );
 
     await Promise.all(updates);
-    for (let i = 0; i < coordinates.length; i++) {
+    for (const [i] of coordinates.entries()) {
```
mertalev, line 108: Interesting usage of `.entries()`.

mertalev (approved):

maxime1992:
Hey, was there a regression on this?

Trying to filter my pics to find some with missing GPS metadata, I'm not able to:

image

And I've got some that have a location that are showing up :thinking: 

Do note though that the pics showing up had no GPS metadata initially. I added them on Immich but I'm not looking for pictures which don't have metadata attached onto the picture itself, I'm looking for pictures that don't have any GPS coordinate in Immich.

jforseth210:
I'm not able to reproduce on my instance (1.124.2). Searched for photos with unknown location, added location to one, searched again, and that photo was no longer in the results.

maxime1992:
Hi, I checked again and same issue for me. Also on `1.124.2`. 

I filter for `Unknown` country and for a certain camera model, I get some results with a location (on pics which did not have the location initally but it was manually added).

Is there a way for me to help debug this, maybe share logs or something else?

duncaninnes:
Can I just point out that filtering on Unknown Country is not the same as filtering on Unknown Location.

I have many images which were taken at sea, and however Immich is calculating country, they are deemed to have Unknown Country.  So most of my boating/diving pictures come up when I search by Unknown Country.

Not sure if this means it would be better to provide a real search by Unknown Location, or provide better calculation of Country to include maritime water boundaries?  Or both?

But there is a gap here which leaves a lot of photos mis-attributed.

PR #10891, opened 2024-07-05
refactor(server): remove has asset

jrasm91:
Migrate album repository from `hasAsset` (removed) to `getAssetIds`.

### DIFF
--- a/server/src/services/album.service.ts
+++ b/server/src/services/album.service.ts
@@ -146,8 +146,8 @@ export class AlbumService {
     const album = await this.findOrFail(id, { withAssets: true });
 
     if (dto.albumThumbnailAssetId) {
-      const valid = await this.albumRepository.hasAsset({ albumId: id, assetId: dto.albumThumbnailAssetId });
-      if (!valid) {
+      const results = await this.albumRepository.getAssetIds(id, [dto.albumThumbnailAssetId]);
+      if (results.size === 0) {
         throw new BadRequestException('Invalid album thumbnail');
       }
     }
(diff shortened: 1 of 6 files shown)


server/src/services/album.service.ts
```diff
@@ -146,8 +146,8 @@ export class AlbumService {
     const album = await this.findOrFail(id, { withAssets: true });
 
     if (dto.albumThumbnailAssetId) {
-      const valid = await this.albumRepository.hasAsset({ albumId: id, assetId: dto.albumThumbnailAssetId });
-      if (!valid) {
+      const results = await this.albumRepository.getAssetIds(id, [dto.albumThumbnailAssetId]);
```
michelheusschen, line 149: `hasAsset` excluded deleted entities, while `getAssetIds` includes them. Will this still behave as expected?

jrasm91: I guess that would technically let you set the thumbnail to an id that is in the album but deleted?

michelheusschen: While that might not be ideal, web and mobile hide deleted assets in albums so you won't end up in this situation. The invalid thumbnail check for albums also doesn't exclude deleted assets, so it should be fine.

jrasm91: That is a whole separate class of problems IMO. We might want to re-asses how deleted images are handled in albums, etc.

michelheusschen: Yeah I agree, that's beyond the scope of this PR

danieldietzler (approved):

PR #10900, opened 2024-07-05
refactor(server): stack owner

jrasm91:
Add `ownerId` to stacks, which makes it possible to cascade delete them.


danieldietzler (approved):
Awesome! :)

PR #10904, opened 2024-07-05
refactor(server): use .toSorted

jrasm91:
Use es2023, which has .toSorted.


danieldietzler (approved):

mertalev:
There are two instances you can change in [mime-types.spec.ts](https://github.com/immich-app/immich/blob/main/server/src/utils/mime-types.spec.ts#L148) as well.

mertalev (approved):

PR #10916, opened 2024-07-06
ci: Provide geonames database as GitHub release asset

jvanbruegge:
As discussed on Discord, this adds an archive containing the geonames database at the time of release as release asset.
This is useful for reproducible builds such as on NixOS.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -14,6 +14,7 @@ concurrency:
   cancel-in-progress: true
 
 permissions:
+  contents: write
   packages: write
 
 jobs:
@@ -119,6 +120,7 @@ jobs:
         with:
           context: ${{ matrix.context }}
           file: ${{ matrix.file }}
+          load: ${{ github.event_name == 'release' }}
           platforms: ${{ matrix.platforms }}
           # Skip pushing when PR from a fork
           push: ${{ !github.event.pull_request.head.repo.fork }}
@@ -132,3 +134,16 @@ jobs:
             BUILD_IMAGE=${{ github.event_name == 'release' && github.ref_name || steps.metadata.outputs.tags }}
             BUILD_SOURCE_REF=${{ github.ref_name }}
             BUILD_SOURCE_COMMIT=${{ github.sha }}
+
+      - name: Extract geodata
+        if: ${{ github.event_name == 'release' && matrix.image == 'immich-server' }}
+        run: |
+          id="$(docker create ghcr.io/${{ github.repository_owner }}/${{ matrix.image }}:${{ github.ref_name }})"
+          docker cp "$id":/usr/bin/resources geodata
+          zip -r geodata.zip geodata
+
+      - name: Upload geodata as release asset
+        if: ${{ github.event_name == 'release' && matrix.image == 'immich-server' }}
+        uses: softprops/action-gh-release@v2
+        with:
+          files: geodata.zip


bo0tzz (approved):
Strictly speaking this won't necessarily be the same version of the data as is in our release image. If that matters to you, an alternative option is to load up that image here and copy the data out of it (not as big of an issue here as it is for the nix build, since here the image will be in a very nearby cache).

jvanbruegge:
It does not quite matter if it is the same version as in the weekly container build, more that there is one fixed version per release

zackpollard:
I didn't realise we wanted to do this in this repo, I don't like the idea of duplicating this script over here really

jvanbruegge:
I could make it a script in the base-image repo and clone that in this workflow instead

jvanbruegge:
I fixed that now. The script that downloads the geodata is in this PR: https://github.com/immich-app/base-images/pull/79
Here, the CI just downloads and executes that scripts and zips the folder. No more code duplication

jrasm91 (commented):
This isn't even right though. This is probably not what is actually included in the release.

zackpollard:
> This isn't even right though. This is probably not what is actually included in the release. 

Yea, I think I have similar feelings, if we are going to publish these in the release they should be the actual ones we bundle in the release.

jrasm91:
If the goal is to just have a reproducible download, they can be added to git in a separate repository outside of immich. If we're going to attach them to the release, I would recommend just adding them to the base image release itself.

jvanbruegge:
ok, now this is indeed saving the exact version that is also in the release docker container.

zackpollard:
Will take a look at this on Monday and merge it in if all looks good! 🙂

.github/workflows/docker.yml
```diff
@@ -132,3 +134,16 @@ jobs:
...
+        run: |
+          id="$(docker create ghcr.io/${{ github.repository_owner }}/${{ matrix.image }}:${{ github.ref_name }})"
+          docker cp "$id":/usr/bin/resources geodata
+          zip -r geodata.zip geodata
+
+      - name: Upload geodata as release asset
+        if: ${{ github.event_name == 'release' && matrix.image == 'immich-server' }}
+        uses: softprops/action-gh-release@v2
+        with:
+          files: geodata.zip
```
bo0tzz, line 149: I assume calling it this way means that the zip file gets added to the existing release? Is that behaviour documented anywhere?

jvanbruegge: I think I found a hint to that on Stackoverflow? The action README is a bit vague.
I verified this behavior in a fork though: https://github.com/jvanbruegge/immich/releases/tag/v1.108.0
The zip was attached by the action

jvanbruegge: I've [opened a PR](https://github.com/softprops/action-gh-release/pull/474) to document this behavior

jvanbruegge:
Is there anything still to do here for me?

zackpollard:
Hey, apologies for the silence on this. I'll give you as brief of an update as I can for now.

Essentially there are two issues here.
1.) We don't really want to have this linking between the repos that this introduces, we would prefer that this artifact is included as part of the base-image releases, though we don't currently create github releases for these.
2.) We are currently looking into overhauling the reverse geocoding and providing the option of a system with much improved accuracy. This will require a bunch more files to be created and stored, so it may completely overhaul the way we currently include the geonames data too, as the systems are related.

Basically, if we were to merge something like this, it would need to exist entirely within the base-images solution, but we are hesitant on that too right now due to the fact this may all change and potentially live in a separate repository pretty soon anyway, so don't want to add something that may immediately be changed. If that happens there will likely be automated releases for all this stuff included within that repo anyway.

jvanbruegge:
Ok, for the nix build I now used the internet archive as a stopgap solution until your new system is in place

PR #10922, opened 2024-07-06
chore(web): translate alt text

ben-basten:
## Description

Converting the alt text for assets into a translated Weblate string.

Getting the translation Svelte store was also moved into a util function, to make unit testing as easy as mocking a function, rather than having to mock a Svelte store.

## How Has This Been Tested?

Navigating through the asset grid and asset viewer using VoiceOver on Mac.


web/src/lib/utils/thumbnail-util.ts
```diff
@@ -40,24 +41,33 @@ export function getAltText(asset: AssetResponseDto) {
     return asset.exifInfo.description;
   }
 
-  let altText = 'Image taken';
-  if (asset.exifInfo?.city && asset.exifInfo.country) {
-    altText += ` in ${asset.exifInfo.city}, ${asset.exifInfo.country}`;
+  const $t = getFormatterSync();
```
michelheusschen, line 44: Can this be a derived store instead?

web/src/lib/utils/thumbnail-util.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+      },
+    });
+    expect(mockFormatter).toHaveBeenNthCalledWith(4, 'image_alt_text_date', {
+      values: { date: 'January 1, 2024' },
+    });
+    expect(result).toBe('formatted formatted formatted formatted');
+  });
+});
+
+function getMockFormatter() {
```
michelheusschen, line 140: Might be better to use the actual formatter for these tests

ben-basten: Makes sense - I wasn't sure how coupled we want to be to the actual translation strings. I've made this change!

web/src/lib/utils/thumbnail-util.ts
ben-basten: This util function isn't needed if tests are being run against the actual translations - I've removed it.

michelheusschen: Thanks for the changes. This function should still use a derived to preserve reactivity when the language changes:

```ts
const getAltText = derived(t, ($t) => {
  return (asset: AssetResponseDto) => {
    
  };
});
```

ben-basten: Ok, I can make that change. I'm not sure how to test this - what is the scenario that this derived statement is needed for? I'm seeing the alt text changing when I go to the settings and back.

ben-basten: And thanks for the feedback! This is good to know going forward.

michelheusschen: This scenario would occur when the language changes while you're on a page where `getAltText` is used. I don't think it's currently an issue, because going to settings and back will cause a re-render.

michelheusschen (approved):

PR #10929, opened 2024-07-06
fix(ml): armnn not being used

mertalev:
## Description

We check if an ARM NN model exists and fall back to ONNX if it does not. However, we do this check before the model is downloaded at all, so it will always appear missing unless it has already been downloaded. Additionally, since we have already fallen back to ONNX when downloading, we avoid downloading the ARM NN model. This catch-22 prevents ARM NN from being used in practice without manually moving the model into the model cache.

This PR fixes this behavior and additionally moves the fallback handling outside of the model class itself to being in the main `load` function. This makes the model's behavior more predictable: it's unexpected for the model to choose a different format than what the caller specified.

### DIFF
--- a/machine-learning/app/conftest.py
+++ b/machine-learning/app/conftest.py
@@ -169,6 +169,12 @@ def warning() -> Iterator[mock.Mock]:
 
 
 @pytest.fixture(scope="function")
+def exception() -> Iterator[mock.Mock]:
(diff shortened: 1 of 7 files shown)


fyfrey (approved):
Wonderful!
I've tested this on my RK3588 device and ML with ARMNN now works out-of-the-box for CLIP vision model :-)

PR #10950, opened 2024-07-07
feat(server): country geocoding for remote locations

pokjay:
**Add country reverse geocoding in case city geocoding is null. Uses [country boundaries from Natural Earth]( https://github.com/nvkelso/natural-earth-vector/blob/ca96624a56bd078437bca8184e78163e5039ad19/geojson/ne_10m_admin_0_countries.geojson)**

### Dependencies
base-images PR: https://github.com/immich-app/base-images/pull/80

### Database changes:
1. Create new table naturalearth_countries with country polygons. Each country will have a row for each polygon.

### Server changes:
1. Load geojson file to database
2. When city geocoding fails, fallback to geocode using country polygon

Note - this uses the current geocoding state metadata for DB import

### Web UI changes:
1. Allow showing partial location info in asset panel in cases where there is no city:
Screenshot 2024-07-07 at 22 54 03


server/src/repositories/map.repository.ts
```diff
@@ -134,7 +138,29 @@ export class MapRepository implements IMapRepository {
       this.logger.warn(
         `Response from database for reverse geocoding latitude: ${point.latitude}, longitude: ${point.longitude} was null`,
       );
-      return null;
+
+      const ne_response = await this.naturalEarthCountriesRepository
```
jrasm91, line 142: We should change the above condition to return if it has a response. Then, de-indent all of this.

pokjay: I updated the code. Definitely makes it clearer!

danieldietzler (approved):

mschiff:
What needs to be done after an upgrade to 1.109 in order to make use of this for existing images?

zackpollard:
Re-running the extract metadata job for all images should do it 🙂

zackpollard:
> What needs to be done after an upgrade to 1.109 in order to make use of this for existing images?

Re-running the extract metadata job for all images should do it 🙂

mschiff:
> > What needs to be done after an upgrade to 1.109 in order to make use of this for existing images?
> 
> Re-running the extract metadata job for all images should do it 🙂

Thanks! I think this would be worth mentioning in the changelog/release notes for such kind of change, thanks for helping! Much appreciated!

PR #10978, opened 2024-07-09
fix(server): avoid server error for invalid email data type

michelheusschen:
Calling `/api/auth/login` with `{"email":[]}` causes a 500 error, because the transformer function incorrectly assumes the value is a string. This results in an error being thrown when `value.toLowerCase()` is called.


jrasm91 (approved):
Can you actually add an e2e for this?

jrasm91 (approved):
Tyvm

PR #10991, opened 2024-07-10
fix(server): skip smtp validation when there are no changes

michelheusschen:
A connection is made to the smtp server each time the system config is saved, causing unnecessary delays ranging from 100ms to 10s. Fixed by only verifying the smtp transport when changes are made to the smtp config.

### DIFF
--- a/server/src/services/notification.service.spec.ts
+++ b/server/src/services/notification.service.spec.ts
@@ -1,0 +1,118 @@
+import { defaults, SystemConfig } from 'src/config';
+import { IAlbumRepository } from 'src/interfaces/album.interface';
+import { IAssetRepository } from 'src/interfaces/asset.interface';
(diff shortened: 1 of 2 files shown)


danieldietzler (approved):
LGTM

server/src/services/notification.service.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+          port: 587,
+          username: 'test',
+          password: 'test',
+        },
+      },
+    },
+  }),
+};
+
+describe(PartnerService.name, () => {
```
jrasm91, line 57: Change this name

jrasm91 (approved):

server/src/services/notification.service.spec.ts
michelheusschen: You are sharp, changed!

PR #10999, opened 2024-07-10
chore(server): remove unused code

jrasm91:


danieldietzler (approved):
Can we have a linter rule for unused class variables?

alextran1502 (approved):

PR #11060, opened 2024-07-12
feat(web): change the lang attribute based on the user language

martabal:
Currently, the lang attribute is `en` even if the user has changed the language


martabal:
Mmmh, it works now, I don't know why it didn't work :thinking:

PR #10989, opened 2024-07-10
feat(mobile): adds crop and rotate to mobile

Yuvi-raj-P:
Added an image editor and it has a crop feature only (for now)

PR #11074, opened 2024-07-13
feat(web): image editor - panel and cropping

ilyaChuk:
Another attempt to create a photo editor. I hope it's the final one.  #10883

image
[Video](https://um.ichuk.ru/s/6HxcJDbcN3KAe43)

I made a sidebar with tools. so far, there is a cropping and adjustment tool (placeholder). 
Cropping works - you can freely choose the size or use the preset.
I checked on different pictures: horizontal, square, vertical. I didn't notice any problems.
Only the desktop version works. I haven't adapted it to mobile screens.
This is only a web interface, there is no server part of this functionality yet.

TODO:
- ~~[ ] Adaptation to mobile screens~~.
- ~~[x] Canvas performance optimization~~.


opbod:
Related: https://github.com/immich-app/immich/pull/10989 (Mobile); https://github.com/immich-app/immich/pull/3271 (Web); https://github.com/immich-app/immich/pull/9575 (Web)

ilyaChuk:
I feel like I'm ready. The cropping interface is ready. 
image
There is a editor tools sidebar on desktop screens. So far, there's only cropping, but she's ready to work with an array of tools.
Since last time, I added a "save" button. It is active only when changes have been made in the editor. So far, clicking on it does nothing except close the editor and the panel.
I've double-checked everything several times, different aspect ratios of the images, it works stably for me.
This is the first time I have had such an experience, but I tried to make the code convenient for other developers as well.
[video](https://um.ichuk.ru/s/C9txzeTYaHJ2yr5)

opbod:
https://github.com/immich-app/immich/pull/10989 just got accepted into main few hours ago, mobile-only. 
@martyfuhry does this web-only image editor PR conflict with or enrich the overall setup?

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -667,6 +682,8 @@
                 .toLowerCase()
                 .endsWith('.insp'))}
             <PanoramaViewer {asset} />
+          {:else if $isShowEditor && selectedEditType == 'crop'}
```
michelheusschen, line 685: Please use strict equality by default

```suggestion
          {:else if $isShowEditor && selectedEditType === 'crop'}
```

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -715,6 +732,21 @@
     </div>
   {/if}
 
+  {#if $isShowEditor}
```
michelheusschen, line 735: When the editor is opened the info panel should be hidden. Also worth considering:
- Pressing escape closes the editor instead of the asset viewer
- Hide the navbar actions while the editor is open and disable their shortcuts
- Disable shortcuts for previous/next asset

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
michelheusschen, line 1: This component is very big which makes it difficult to read and understand how data flows. I think it's better to make a separate component for the crop functionality, because more editing modes will most likely be added in the future.

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
```diff
@@ -0,0 +1,740 @@
...
+    cropImageSize,
+    cropSettings,
+    type CropSettings,
+    type CropAspectRatio,
+    cropSettingsChanged,
+  } from '$lib/stores/asset-editor.store';
+  import { onMount, afterUpdate } from 'svelte';
+  import { get } from 'svelte/store';
+
+  export let crop = get(cropSettings);
```
michelheusschen, line 14: Use the dollar sign syntax for stores inside components
- `$store` to read the value instead of get() or .subscribe()
- `$store = newValue` to update the value instead of .set()

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
```diff
@@ -0,0 +1,740 @@
...
+
+  let canvas: HTMLCanvasElement;
+  let ctx: CanvasRenderingContext2D;
+  let img: HTMLImageElement;
+
+  let darkenLevel = 0.65; // Initial darkening level
+  let animationFrame: ReturnType<typeof requestAnimationFrame>;
+  let canvasCursor: string;
+  const getAssetUrl = (id: string, checksum: string) => {
+    return `http://localhost:2283/api/assets/${id}/original?c=${checksum}`;
```
michelheusschen, line 29: What if the original asset can't be displayed by the browser? Also this URL should be changed

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
```diff
@@ -0,0 +1,740 @@
...
+    ctx.strokeStyle = 'white';
+    ctx.lineWidth = 2;
+    ctx.strokeRect(crop.x, crop.y, crop.width, crop.height);
+
+    // Set blend mode
+    ctx.strokeStyle = 'white';
+    ctx.lineWidth = 1;
+
+    // Vertical lines
+    const thirdWidth = crop.width / 3;
```
michelheusschen, line 62: Not a huge fan of the grid lines personally, maybe only show them while changing the crop area?

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
```diff
@@ -0,0 +1,740 @@
...
+    }, 1);
+  };
+
+  onMount(() => {
+    img = new Image();
+    img.src = getAssetUrl(asset.id, asset.checksum);
+
+    img.addEventListener('load', onImageLoad);
+    img.addEventListener('error', () => {
+      console.error('Failed to load image');
```
michelheusschen, line 701: This error should be more obvious, consider using `handleError()`

web/src/lib/components/asset-viewer/editor/editor-panel.svelte
```diff
@@ -0,0 +1,95 @@
...
+  }
+</script>
+
+<section class="relative p-2 dark:bg-immich-dark-bg dark:text-immich-dark-fg">
+  <div class="flex place-items-center gap-2">
+    <CircleIconButton icon={mdiClose} title={$t('close')} on:click={() => dispatch('close')} />
+    <p class="text-lg text-immich-fg dark:text-immich-dark-fg">{$t('editor')}</p>
+  </div>
+  <section class="px-4 py-4">
+    <ul class="editorul">
```
michelheusschen, line 66: Just use tailwind classes directly, same for some other components

```suggestion
    <ul class="flex w-full justify-around">
```

web/src/lib/stores/preferences.store.ts
```diff
@@ -77,6 +77,7 @@ export const videoViewerVolume = persisted<number>('video-viewer-volume', 1, {})
 export const videoViewerMuted = persisted<boolean>('video-viewer-muted', false, {});
 
 export const isShowDetail = persisted<boolean>('info-opened', false, {});
+export const isShowEditor = persisted<boolean>('editor-opened', false, {});
```
michelheusschen, line 80: I don't think this should be persisted or even be a store

michelheusschen (commented):
First off, a big thank you for the PR. Having an editor would be a great addition to Immich!

ilyaChuk:
@opbod 
> does this web-only image editor PR conflict with or enrich the overall setup?

We try to work together by discussing our roles. But the difference so far is that I cannot rotate the image, and it is not saved.

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
ilyaChuk: Thanks, I'll remember. I just started getting acquainted with svelte in this project, before that I worked with other frameworks

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
ilyaChuk: I made the lines visible only when the cropping rectangle is changed. But then, when the cropping rectangle is the same size as the image, it is poorly visible
image
I can't imagine how to solve this in a simple way. I added small circles on the corners for more visibility. 
image
But maybe as long as the rectangle is the same size as the image, draw these lines?
Although if this is just a problem for me, then everything is ok.

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
ilyaChuk: how should I reduce it more correctly? should I put functions in separate modules in `src\lib\components\asset-viewer\editor\crop-tool`?

web/src/lib/components/asset-viewer/asset-viewer.svelte
ilyaChuk: is this the right way to disable shortcuts?
```js
<svelte:window
  use:shortcuts={[
    { shortcut: { key: 'a', shift: true }, onShortcut: toggleAssetArchive },
    { shortcut: { key: 'ArrowLeft' }, onShortcut: () => isShowEditor?null:navigateAsset('previous') },
    ...etc
```

ilyaChuk:
@michelheusschen Thank you very much for the review. I fixed all the problems. 
But I postponed the decomposition stage to the very end. Therefore, there is a possibility that I broke something in the component. I still try to check all the work and as if everything is ok.

ilyaChuk:
there is Test / Web (pull_request)  error
```
 src/lib/i18n.spec.ts  (50 tests | 1 failed) 6739ms
   ❯ src/lib/i18n.spec.ts > i18n > no missing messages
     → expected { errors: { …(121) }, …(801) } to deeply equal { about: '', account: '', …(799) }
     
     ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  src/lib/i18n.spec.ts > i18n > no missing messages
AssertionError: expected { errors: { …(121) }, …(801) } to deeply equal { about: '', account: '', …(799) }
```
is it my fault? I didn't seem to touch anything there.

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -103,13 +105,15 @@
...
   let isShowActivity = false;
+  let isShowEditor = false;
   let isLiked: ActivityResponseDto | null = null;
   let numberOfComments: number;
   let fullscreenElement: Element;
   let unsubscribe: () => void;
   let zoomToggle = () => void 0;
   let copyImage: () => Promise<void>;
   $: isFullScreen = fullscreenElement !== null;
+  $: showNavbar = ![isShowEditor].some(Boolean); // can add another variables
```
alextran1502, line 116: Why is `isShowEditor` put into an array?

web/src/lib/components/asset-viewer/asset-viewer.svelte
```diff
@@ -334,8 +346,17 @@
...
   };
+  const showEditorHandler = () => {
+    if (isShowActivity) {
+      isShowActivity = false;
+    }
+    isShowEditor = !isShowEditor;
+  };
 
   const trashOrDelete = async (force: boolean = false) => {
+    if (!showNavbar) {
```
alextran1502, line 357: If navbar isn't shown, can you actually click on these button?

web/src/lib/components/asset-viewer/asset-viewer.svelte
ilyaChuk: I suspect that in the future there will be several conditions for hiding the navbar. So I prepared the code to easily add new variables to this condition.

web/src/lib/components/asset-viewer/asset-viewer.svelte
ilyaChuk: no, navbar is hidden when the editor is open. and their shortcuts are disabled. only while the editor is open
image

alextran1502: Then we don't really need the logic here right? Because there is no way this action is triggered if all the buttons and shortcut are hidden/disabled

ilyaChuk:
```js
const trashOrDelete = async (force: boolean = false) => {
  if (!showNavbar) {
```
it is needed to prevent a call through a shortcut. the `trashOrDelete` function is called in the shortcut
```js
{ shortcut: { key: 'Delete' }, onShortcut: () => trashOrDelete(asset.isTrashed) },
{ shortcut: { key: 'Delete', shift: true }, onShortcut: () => trashOrDelete(true) },
```

alextran1502: This is debatable, but do you think it would be more precise to handle it in the `onShortcut` action? :thinking: 

My thinking is that we keep the `trashOrDelete` and other actions to just do what it is meant to do when it gets invoked.

web/src/lib/components/asset-viewer/editor/editor-panel.svelte
```diff
@@ -0,0 +1,87 @@
...
+
+  onMount(() => {
+    return websocketEvents.on('on_asset_update', (assetUpdate) => {
+      if (assetUpdate.id === asset.id) {
+        asset = assetUpdate;
+      }
+    });
+  });
+
+  const dispatch = createEventDispatcher<{
```
alextran1502, line 23: We are moving away from `createEventDispatcher` in favor of exposing callback function as props. You can see an example here https://github.com/immich-app/immich/blob/21d3f248dab47a3a7f835e8eb2c576a45aa37de9/web/src/lib/components/shared-components/context-menu/menu-option.svelte#L11

alextran1502:
Just try the PR, when we hit `Save` it doesn't do anything yet, correct?

ilyaChuk:
@alextran1502 
> `Save` it doesn't do anything yet, correct?

Yes, clicking on it does nothing except close the editor and the panel. In the future, this button should send a request to the server. And it will already save the photo somehow. Most likely, this will be my next PR

alextran1502:
I think it makes sense to include the save functionality in this PR as well, if we merge it to main and have to push out a release, there will be a features that is not usable. Or maybe we can hide the button?

ilyaChuk:
In the discord, they told me "Definitely a separate PR. Keep things as small as possible". I'm not sure what's the right thing to do.
Because the saving functionality is, as far as I can imagine, quite a lot of functionality. 
Maybe we can not send it to the release yet, but keep it in some other branch? Together with the photo editor for the mobile application #10989.

alextran1502:
I think we want to keep different editing features in a separate PR, but each should have a finished state. I am fine with hiding the button in this PR and getting it merged. 

The mobile app editor is fine, since it produces the local file that can be uploaded/shared

ilyaChuk:
Ok, then I'll remove this button now. but I don't really understand what you mean by "merged". You're not going to release this, are you? because it's now a bare interface with no real save

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
michelheusschen: This is only partially resolved. There is still the possibility that a browser can't display the original asset. This will also be an issue for videos, live photos and panoramas.

web/src/lib/components/asset-viewer/asset-viewer.svelte
ilyaChuk: Yes, I also had doubts about the right place to do it.
But this method, in my opinion, looks very cumbersome:
```js
<svelte:window
  use:shortcuts={[
    { shortcut: { key: 'a', shift: true }, onShortcut: ()=>showNavbar?toggleAssetArchive:null },
    { shortcut: { key: 'ArrowLeft' }, onShortcut: () => showNavbar?navigateAsset('previous'):null },
    { shortcut: { key: 'ArrowRight' }, onShortcut: () => showNavbar?navigateAsset('next'):null },
    { shortcut: { key: 'd', shift: true }, onShortcut: () => (showNavbar ? downloadFile(asset) : null) },
    { shortcut: { key: 'Delete' }, onShortcut: () => showNavbar?trashOrDelete(asset.isTrashed):null },
    { shortcut: { key: 'Delete', shift: true }, onShortcut: () => showNavbar?trashOrDelete(true):null },
```

Should I make a wrapper?
```js
function handleShortcut(condition, action) => {
    return () => (condition ? action() : null);
  };
```

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
michelheusschen: Yeah I think this is better. The dots only seem to get cut off at the edges, maybe there's a way to extend them? I also noticed that cropping immediately stops when the cursor leaves the canvas. Can we allow cropping to continue when that happens?

web/src/lib/components/asset-viewer/asset-viewer.svelte
michelheusschen: #11449 should improve this situation, I think you can also do `onShortcut: () => showNavbar && toggleAssetArchive` to keep is a bit shorter, but both aren't an ideal solution

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
ilyaChuk: Should we even allow the editor to be opened for live photos, panoramas, and gifs?

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
ilyaChuk: It will require some reworking, but I will try

web/src/lib/components/asset-viewer/editor/editor-panel.svelte
ilyaChuk: done

web/src/lib/components/asset-viewer/asset-viewer.svelte
ilyaChuk: done

alextran1502:
Merge means getting this into `main` so that you don't have to keep working around merge conflicts, which happens often since many PRs get merged into `main` every day.

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
michelheusschen: Right now it makes sense to disable the editor for these types of assets. If there are tools added in the future that can be used for more asset types (like rotate), we can always reconsider this

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
ilyaChuk: It wasn't easy, but it's done. [video](https://um.ichuk.ru/s/miEkaAbmzmTJ2Fs)

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
ilyaChuk: @michelheusschen 
>  There is still the possibility that a browser can't display the original asset

And then what to show? now there is an error message when opening the editor and an error loading the image. An error message, a black editor screen, and a toolbar on the right. 
Should I close the editor automatically in this case?

ilyaChuk:
how do I understand in the `web\src\lib\components\asset-viewer\asset-viewer-nav-bar.svelte` if the image is gif || panoramic || live?

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
ilyaChuk: > Right now it makes sense to disable the editor for these types of assets

done

ilyaChuk:
Should we have parity with the mobile version in terms of the corners of the cropping rectangle?
On the one hand, I want the editor to look the same in both places. But on the other hand, the circles on the corners look more convenient for me to use the mouse. In fact, this is just a visual, the capture boundaries are the same, but it looks less convenient for me.
image
image

ilyaChuk:
if I have not forgotten anything, then I have fulfilled all the wishes from your review. 
but tomorrow I want to add a pop-up question when trying to close the editor, "are you sure you want to quit? the changes will not be saved" as in google photos.
and presets for vertical aspect ratio.
and then the PR will be ready

alextran1502:
> if I have not forgotten anything, then I have fulfilled all the wishes from your review. but tomorrow I want to add a pop-up question when trying to close the editor, "are you sure you want to quit? the changes will not be saved" as in google photos. and presets for vertical aspect ratio. and then the PR will be ready

Sounds good to me!

web/src/lib/components/asset-viewer/editor/crop-canvas.svelte
jrasm91: Another thing that might be worth looking into down the road is using sharp to do the actual cropping on the original asset. So always show the crop tool on the jpeg/webp thumbnail and use it to get the cropping coordinates, but when send those to the server which can actually perform the operation. I believe using sharp it would be possible to keep the metadata, etc.

ilyaChuk:
merge conflict was not easy to resolve. I hope I didn't break anything. But otherwise, everything is ready @alextran1502

ilyaChuk:
image
image

web/src/lib/components/asset-viewer/editor/crop-tool/canvas-drawing.ts
michelheusschen, line 1: I'm starting to think that using regular HTML elements for the controls might be a better choice. I can think of a couple advantages:
1. Can be styled with CSS
2. Event handling is easier with essentially built-in hit detection and likely better support for touch devices
3. Accessibility features can be added

ilyaChuk: I tried to make a rotation tool yesterday. I had to repeat the course of linear algebra to remember how to work with matrices and vectors 💀.
Yes, the canvas creates difficulties. After yesterday, I also think about switching to html elements

ilyaChuk:
I'm a little tilted because I've done so much, and it really needs to be rewritten from canvas to html elements.
It won't take super long. but it's mentally hard.
It's ok, I had to come to this in a couple of days. and it will be ready in a few days.

jrasm91:
I'm happy to help work on this if you want some help finishing it. I also started work on a server API for creating assets from edits that might be possible to plug this into (#11658). I think most of what you have now is good to go, just needs a bit of cleanup and polish.

ilyaChuk:
@jrasm91, I'm glad you've joined the editor's work too, and thanks for your willingness to help. But there's not much left to do in this PR, I'll finish it up

ilyaChuk:
I think it's ready. Moved from canvas to html elements.
image

[video](https://um.ichuk.ru/s/4YZPnKpskNFHmFf)

ilyaChuk:
Thanks to the move away from canvas, it's easier to make rotations. 
you_spin_me_right_round
But there is one point of contention for me. Aspect ratios. Should it be relative to the image or relative to the window? I checked, in the mobile version it is now relative to the window. 
But look at the 1st video. Here I can see the advantage of the aspect ratio being relative to the image. I'll be honest, it's a lot easier for me. Secondly, trying to change the aspect ratio while trying to stay relative to the window results in awkward behaviour:
https://github.com/user-attachments/assets/7859c57a-ec89-49ab-bf88-51c78a34656b
Yeah, it's got an extra bug. But you get the point. If you want to select the lettering "poison". When you rotate and try to save the relative ratio, the inscription will be selected wrong (perpendicular).
vote 👍 - relative to image
👎 - relative to window

ilyaChuk:
Feels ready 😴

ilyaChuk:
Ok, now I don't see any errors. Yet

ilyaChuk:
but I still don't understand the error I'm getting here https://github.com/immich-app/immich/actions/runs/10328229151/job/28597174488

alextran1502 (approved):

eygraber:
I saw this in the release notes, but it doesn't look like it is available. Is it disabled, or hidden?

ilyaChuk:
@eygraber Yeah, it's hidden for now. I think we'll launch it when the server side is ready.

dharapvj:
Yes -which github issue to track and vote on for server side of this?

NicholasFlamy:
#11658 seems to be a potential server-side implementation.

RichardMawdsley:
Any update on this??

RazgrizHsu:
any update on this?

bo0tzz:
https://justinmayer.com/posts/any-updates/

RazgrizHsu:
I apologize if my previous comment came across as inappropriate. English is not my native language, and I sincerely regret if my message was misinterpreted. I was simply inquiring about whether there had been any progress on this issue. Thank you for your understanding.

PR #11080, opened 2024-07-14
feat(server): conditionally run facial recognition nightly

mertalev:
## Description

The facial recognition nightly currently runs every night, but there's no need when no new face has been detected since the last run.

This PR stores the last run date in the system metadata table and compares it with the newest `facesRecognizedAt` value in the job status table. It only does this in the nightly; manual runs still behave like before.

The query for getting the latest face detection date is a full table scan, but it isn't worth indexing the column just for one non-interactive query a day. It at least avoids sorting.

Resolves #11078

### DIFF
--- a/server/src/entities/system-metadata.entity.ts
+++ b/server/src/entities/system-metadata.entity.ts
@@ -12,6 +12,7 @@ export class SystemMetadataEntity<T extends keyof SystemMetadata = SystemMetadat
 
 export enum SystemMetadataKey {
   REVERSE_GEOCODING_STATE = 'reverse-geocoding-state',
+  FACIAL_RECOGNITION_STATE = 'facial-recognition-state',
   ADMIN_ONBOARDING = 'admin-onboarding',
   SYSTEM_CONFIG = 'system-config',
   VERSION_CHECK_STATE = 'version-check-state',
@@ -22,6 +23,7 @@ export type VersionCheckMetadata = { checkedAt: string; releaseVersion: string }
 
 export interface SystemMetadata extends Record<SystemMetadataKey, Record<string, any>> {
   [SystemMetadataKey.REVERSE_GEOCODING_STATE]: { lastUpdate?: string; lastImportFileName?: string };
+  [SystemMetadataKey.FACIAL_RECOGNITION_STATE]: { lastRun?: Date };
   [SystemMetadataKey.ADMIN_ONBOARDING]: { isOnboarded: boolean };
   [SystemMetadataKey.SYSTEM_CONFIG]: DeepPartial<SystemConfig>;
   [SystemMetadataKey.VERSION_CHECK_STATE]: VersionCheckMetadata;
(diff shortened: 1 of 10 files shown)


server/src/entities/system-metadata.entity.ts
```diff
@@ -22,6 +23,7 @@ export type VersionCheckMetadata = { checkedAt: string; releaseVersion: string }
 
 export interface SystemMetadata extends Record<SystemMetadataKey, Record<string, any>> {
   [SystemMetadataKey.REVERSE_GEOCODING_STATE]: { lastUpdate?: string; lastImportFileName?: string };
+  [SystemMetadataKey.FACIAL_RECOGNITION_STATE]: { lastRun?: Date };
```
jrasm91, line 26: I think using a date here is just going to cause problems. This is serialized when it is written, and probably deserializes to a string when it is read afterwards. Did you check this behavior?

mertalev: I didn't check that it's a date - I think you're right that it's a string in practice.

server/src/services/person.service.ts
```diff
@@ -376,13 +378,26 @@ export class PersonService {
...
 
     await this.jobRepository.waitForQueueCompletion(QueueName.THUMBNAIL_GENERATION, QueueName.FACE_DETECTION);
+
+    if (nightly) {
+      const [state, latestFaceDate] = await Promise.all([
+        this.systemMetadataRepository.get(SystemMetadataKey.FACIAL_RECOGNITION_STATE),
+        this.repository.getLatestFaceDate(),
+      ]);
+
+      if (state && state.lastRun && latestFaceDate && state.lastRun > latestFaceDate) {
```
danieldietzler, line 395: You could merge the first two: `state?.lastRun &&`

danieldietzler (approved):
LGTM

server/src/services/person.service.ts
```diff
@@ -376,13 +378,26 @@ export class PersonService {
...
 
     await this.jobRepository.waitForQueueCompletion(QueueName.THUMBNAIL_GENERATION, QueueName.FACE_DETECTION);
+
+    if (nightly) {
+      const [state, latestFaceDate] = await Promise.all([
+        this.systemMetadataRepository.get(SystemMetadataKey.FACIAL_RECOGNITION_STATE),
+        this.repository.getLatestFaceDate(),
+      ]);
+
+      if (state?.lastRun && latestFaceDate && state.lastRun > latestFaceDate) {
```
jrasm91, line 395: Is this a string compare? Does that even work?

mertalev: Yes, since they're both ISO strings.

mertalev: I tested by hardcoding nightly: true. The first time it ran (since there was no system metadata entry) and the second time it skipped.

jrasm91: Ok. Maybe we should add a comment or something. I think it isn't very obvious, which makes it an easy place to introduce a bug.

PR #11082, opened 2024-07-14
fix(server): use fqdn for og:image meta tag value

### LINKED ISSUE #11081: open graph image urls are relative and may not work for non-relaxed implementations
### The bug

sometimes, the open graph image from a shared link is not found, resulting in no image preview (ex: signal app on android)

n/a

### Version of Immich Server

latest

latest

### Platform with the issue

- [X] Web

```YAML
n/a
```

### Your .env content

```Shell
n/a
```

### Reproduction steps

```bash
1. generate a share link from immich
2. share the link on a social network that implements a strict form of the open graph protocol
3. note that the image preview is not found / is broken
```

### Relevant log output

```shell
n/a
```

### Additional information

when sharing a link from immich on social networks / messaging apps / etc, a preview is grabbed from the `og:image` value in the headers of the sharing page.

the value is currently a relative path to a thumbnail of the first asset being shared. many implementations will correctly concat this path to the base URL of the share link.

however, not all implementations will do this correctly (ex: signal on android), so many opengraph checkers will not show any broken behavior.

the [opengraph protocol itself](https://ogp.me/) doesn't make this immediately clear, but later in the document (and the turtle spec), they do declare that url values are defined as: 

_All valid URLs that utilize the https:// or https:// protocols_

all examples in the document also make use of fqdn form of the URLs, so in general, it seems safe and more transparent to include this form of the URL.

eleith:
opengraph image specifies that the url contains http or https, thus implying a fqdn.

this change uses the external domain from the server config to attempt to make the og:image have both the existing path to the thumbnail along with the desired domain

please note, some og implementations do work with relative paths, so not all og image checkers may still pass, but not all implementations have this fallback and thus will not find the image otherwise

fix #11081

### DIFF
--- a/server/src/services/api.service.ts
+++ b/server/src/services/api.service.ts
@@ -3,7 +3,9 @@ import { Cron, CronExpression, Interval } from '@nestjs/schedule';
 import { NextFunction, Request, Response } from 'express';
 import { readFileSync } from 'node:fs';
 import { ONE_HOUR, resourcePaths } from 'src/constants';
+import { SystemConfigCore } from 'src/cores/system-config.core';
 import { ILoggerRepository } from 'src/interfaces/logger.interface';
+import { ISystemMetadataRepository } from 'src/interfaces/system-metadata.interface';
 import { AuthService } from 'src/services/auth.service';
 import { JobService } from 'src/services/job.service';
 import { SharedLinkService } from 'src/services/shared-link.service';
@@ -32,14 +34,18 @@ const render = (index: string, meta: OpenGraphTags) => {
 
 @Injectable()
 export class ApiService {
+  private configCore: SystemConfigCore;
+
   constructor(
     private authService: AuthService,
     private jobService: JobService,
     private sharedLinkService: SharedLinkService,
     private versionService: VersionService,
     @Inject(ILoggerRepository) private logger: ILoggerRepository,
+    @Inject(ISystemMetadataRepository) systemMetadataRepository: ISystemMetadataRepository,
   ) {
     this.logger.setContext(ApiService.name);
+    this.configCore = SystemConfigCore.create(systemMetadataRepository, logger);
   }
 
   @Interval(ONE_HOUR.as('milliseconds'))
@@ -69,13 +75,14 @@ export class ApiService {
         return next();
       }
 
+      const config = await this.configCore.getConfig({withCache: true})
       const targets = [
         {
           regex: /^\/share\/(.+)$/,
           onMatch: async (matches: RegExpMatchArray) => {
             const key = matches[1];
             const auth = await this.authService.validateSharedLink(key);
-            return this.sharedLinkService.getMetadataTags(auth);
+            return this.sharedLinkService.getMetadataTags(auth, config.server.externalDomain);
           },
         },
       ];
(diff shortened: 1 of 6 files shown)


web/src/routes/(user)/share/[key]/[[photos=photos]]/[[assetId=id]]/+page.ts
```diff
@@ -2,19 +2,23 @@ import { getAssetThumbnailUrl, setSharedLink } from '$lib/utils';
...
-import { getMySharedLink, isHttpError } from '@immich/sdk';
+import { getMySharedLink, isHttpError, getConfig } from '@immich/sdk';
 import type { PageLoad } from './$types';
 
 export const load = (async ({ params }) => {
   const { key } = params;
   await authenticate({ public: true });
   const asset = await getAssetInfoFromParam(params);
+  const config = await getConfig()
+  const domain = config.server.externalDomain;
```
Comment (mertalev) [inline web/src/routes/(user)/share/[key]/[[photos=photos]]/[[assetId=id]]/+page.ts:13]: You can move these two in the try/catch and put `getAssetInfoFromParam`, `getConfig` and `getMySharedLink` in a `Promise.all` call. It's better to do this in parallel instead of waiting for each to finish.

jrasm91 (dismissed):
This is a bit pointless since the change isn't server side rendered. You should take a look at api.service.ts in the server repository instead. There is also an e2e test for shared links that checks this specific functionality.

danieldietzler (approved):

mertalev (approved):

eleith:
i've addressed the comments, lints and tests. there is likely some opportunity to improve the approach (SSR vs web) to improve future DX concerns, but they could be decoupled from this PR as well. i am open to either. 

thanks for the feedback so far.

server/src/services/api.service.ts
```diff
@@ -69,13 +75,14 @@ export class ApiService {
         return next();
       }
 
+      const config = await this.configCore.getConfig({ withCache: true });
```
jrasm91, line 78: The code that loads the config should be inside of the getMetadataTags method in the shared link service.

jrasm91 (commented):
Looks a lot better.

server/src/services/api.service.ts
eleith: implemented

jrasm91 (approved):
This looks good, the only issue iM thinking about now is if the extension domain has a trailing slash or not? Did you test both situations? Alternatively new URL(path, domain).href would probably fix the issue as well

eleith:
> This looks good, the only issue iM thinking about now is if the extension domain has a trailing slash or not? Did you test both situations? Alternatively new URL(path, domain).href would probably fix the issue as well

i didn't test that! 

i just dove in assuming `externalDomain` was a trimmed and validated host and lazily joined them. 

i've pushed updates now to better assemble the host and pathname/query params and have successfully tested with trailing slashes.

jrasm91:
I mean, we probably should validate/normalize it lol. Maybe in a follow up PR we'll fix it all?

jrasm91 (approved):

eleith:
I'll fix the tests

eleith:
locally, lint, e2e and svelte check now all pass for me

eleith:
@jrasm91 I assume I don't have to keep this PR updated until further notice.

let me know if there are additional requirements or actions I can take.

jrasm91:
> @jrasm91 I assume I don't have to keep this PR updated until further notice.
> 
> let me know if there are additional requirements or actions I can take.

Correct. It has just been a busy week, but I will try to get it merged in next week.

eleith:
thanks for all the feedback and i appreciate understanding the next steps.

I mostly wanted clarification that I don't need to monitor the PR anymore.

all sounds good to me, it will land when it lands!

PR #11083, opened 2024-07-14
memory - clear timeout after setTimeout complete

iamleson98:
I notice the the asyncTimeout creates timeout value using setTimeout. The timeout values are not properly cleared. Not sure if the garbage collector does this job for us but I think we better should clear timeout after done.

### DIFF
--- a/web/src/lib/utils.ts
+++ b/web/src/lib/utils.ts
@@ -308,8 +308,11 @@ export const findLocale = (code: string | undefined) => {
 };
 
 export const asyncTimeout = (ms: number) => {
-  return new Promise((resolve) => {
-    setTimeout(resolve, ms);
+  return new Promise<void>((resolve) => {
+    const timeout = setTimeout(() => {
+      clearTimeout(timeout);
+      resolve();
+    }, ms);
   });
 };
 


jrasm91 (commented):
I don't think this is necessary. clearTimeout is only used to cancel a timeout, not to free resources.

iamleson98:
@jrasm91, Hi
https://github.com/immich-app/immich/blob/d23aa5e8e21bc3049fa65102582105025469a24a/web/src/routes/admin/jobs-status/%2Bpage.svelte#L20

Please have a look at this code.
the setTimeout is seems to be called every 5 seconds.
I think it's best practice to add the clearTimeout, to prevent any potential problems

web/src/lib/utils.ts
```diff
@@ -308,8 +308,11 @@ export const findLocale = (code: string | undefined) => {
 };
 
 export const asyncTimeout = (ms: number) => {
-  return new Promise((resolve) => {
-    setTimeout(resolve, ms);
+  return new Promise<void>((resolve) => {
+    const timeout = setTimeout(() => {
+      clearTimeout(timeout);
```
michelheusschen, line 313: Calling `clearTimeout` is pointless because the timer has already fired by the time this callback is executed.

In the code you linked, the timeout can fire up to 5 seconds after leaving the page, but it won't make an API call. I think it's fine to leave it as is for simplicity's sake.

PR #11089, opened 2024-07-14
feat(server): add more prometheus metrics

joaopedrocg27:
Implementation of #11069

Added the following metrics:

- immich_users_deleted_total
- immich_host_version
- immich_host_version_latest

in progress...

### DIFF
--- a/server/src/services/server.service.ts
+++ b/server/src/services/server.service.ts
@@ -18,11 +18,11 @@ import { SystemMetadataKey } from 'src/entities/system-metadata.entity';
 import { ICryptoRepository } from 'src/interfaces/crypto.interface';
 import { OnEvents } from 'src/interfaces/event.interface';
 import { ILoggerRepository } from 'src/interfaces/logger.interface';
+import { IMetricRepository } from 'src/interfaces/metric.interface';
 import { IServerInfoRepository } from 'src/interfaces/server-info.interface';
 import { IStorageRepository } from 'src/interfaces/storage.interface';
 import { ISystemMetadataRepository } from 'src/interfaces/system-metadata.interface';
 import { IUserRepository, UserStatsQueryResponse } from 'src/interfaces/user.interface';
-import {IMetricRepository} from "src/interfaces/metric.interface";
 
 import { asHumanReadable } from 'src/utils/bytes';
 import { mimeTypes } from 'src/utils/mime-types';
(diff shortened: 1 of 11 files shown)


joaopedrocg27:
Where could I get the user created metric?

bo0tzz:
Probably https://github.com/immich-app/immich/blob/d23aa5e8e21bc3049fa65102582105025469a24a/server/src/cores/user.core.ts#L28

joaopedrocg27:
> Probably
> 
> https://github.com/immich-app/immich/blob/d23aa5e8e21bc3049fa65102582105025469a24a/server/src/cores/user.core.ts#L28

Ok, I thought there was a better option. Ty

server/src/utils/instrumentation.ts
```diff
@@ -19,11 +19,13 @@ export const hostMetrics =
 export const apiMetrics =
   process.env.IMMICH_API_METRICS == null ? metricsEnabled : process.env.IMMICH_API_METRICS === 'true';
 export const repoMetrics =
-  process.env.IMMICH_IO_METRICS == null ? metricsEnabled : process.env.IMMICH_IO_METRICS === 'true';
+  process.env.IMMICH_IO_METRICS == null ? metricsEnabled : process.env.IMMICH_IO_METRICS === 'true'; // check usage of repoMetrics. If it is IO related why not call it IO?
```
mertalev, line 22: The wording was repo everywhere at first, then I changed the env to IO since repo doesn't mean anything to an immich user without context. You can change the variable here and `MetricRepository` to also use `io` for consistency.

joaopedrocg27: Done! I just commented that there so I didn't forget to ask you guys.

server/src/services/server.service.ts
```diff
@@ -147,7 +150,14 @@ export class ServerService implements OnEvents {
...
       serverStats.usage += usage.usage;
       serverStats.usageByUser.push(usage);
+
+      this.metricRepository.user.addToGauge(`immich.user.photos`, usage.photos, {},{"userName": user.userName})
+      this.metricRepository.user.addToGauge(`immich.user.videos`, usage.videos, {},{"userName": user.userName})
+      this.metricRepository.user.addToGauge(`immich.user.usage`, usage.usage, {},{"userName": user.userName})
     }
+    this.metricRepository.host.addToGauge(`immich.host.photos`, serverStats.photos)
+    this.metricRepository.host.addToGauge(`immich.host.videos`, serverStats.videos)
+    this.metricRepository.host.addToGauge(`immich.host.usage`, serverStats.usage)
```
joaopedrocg27, line 160: Btw, am I being stupid or you cannot send metrics without creating "sums"

Currently this keeps adding up the one photo I have. I can't use a Counter because it can decrease and the Gauge is adding stuff. 

I am very noob in Otel, I am used to send prometheus data with bash or python directly :)

joaopedrocg27, line 160: Btw, am I being stupid or you cannot send metrics without creating "sums"

Currently this keeps adding up the one photo I have. I can't use a Counter because it can decrease and the Gauge is adding stuff.

I am very noob in Otel, I am used to send prometheus data with bash or python directly :)

mertalev: It's not as easy to do it for persistent data like that. Options:
 - Increment whenever an asset gets created (might be difficult to keep in sync with actual)
 - Query Postgres for the data (accurate, but counting in Postgres is very slow since it's a full table scan, so doing it all the time is a very bad idea).
 - Have a trigger in Postgres that increments values in a separate counting table when assets are added or removed (fast retrieval for metrics, but makes insertions slower, not sure by how much).

joaopedrocg27: Exactly, I'll think about it! It is missing that on my grafana

mertalev: But to just set it in this handler and not automatically, I'm sure you can set the gauges to explicitly be those values. The metric repo doesn't expose everything from the underlying otel gauges.

mertalev: Ah, I think the up/down counter is different from a gauge. [This](https://opentelemetry.io/docs/languages/js/instrumentation/#using-observable-async-gauges) should be what you're looking for.

joaopedrocg27: I tried that but then it got stuck on 1... I need to try again. Thank you

joaopedrocg27: I also need to find an endpoint that gets called more often (that one is just when the admin opens that panel) or create a cron for that

mertalev: You can add it to the nightly jobs [here](https://github.com/immich-app/immich/blob/d23aa5e8e21bc3049fa65102582105025469a24a/server/src/services/job.service.ts#L205).

joaopedrocg27: > You can add it to the nightly jobs [here](https://github.com/immich-app/immich/blob/d23aa5e8e21bc3049fa65102582105025469a24a/server/src/services/job.service.ts#L205).

That might not be needed with the Observable Gauge since it triggers the function on scrape

I attempted with it but the value is still stuck at 0. Do you have any other ideas?

mertalev: It shouldn't make a new observable gauge with each call. The gauge should be called when the service is created and reference variables in the service. These variables should be updated when the user statistics handler is called.

*Definitely don't let this run every time the endpoint is scraped*.

PR #11099, opened 2024-07-14
docs: cleanup external libraries

mmomjian:
Getting a lot of help tickets with multiple EXTERNAL_PATH, which isn't needed anymore with the monolith server container.

### DIFF
--- a/docs/docs/features/libraries.md
+++ b/docs/docs/features/libraries.md
@@ -112,7 +112,7 @@ The `immich-server` container will need access to the gallery. Modify your docke
 ```
 
 :::tip
-The `ro` flag at the end only gives read-only access to the volumes. This will disallow the images from being delete in the web UI.
+The `ro` flag at the end only gives read-only access to the volumes. This will disallow the images from being deleted in the web UI.
 :::
 
 :::info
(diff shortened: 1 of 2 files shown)


docs/docs/features/libraries.md
```diff
@@ -104,18 +104,19 @@ The `immich-server` container will need access to the gallery. Modify your docke
...
++     - /mnt/nas/christmas-trip:/mnt/nas/christmas-trip:ro
++     - /home/user/old-pics:/home/user/old-pics:ro
 +     - /mnt/media/videos:/mnt/media/videos:ro
++     - /mnt/media/videos2:/mnt/media/videos2 # the files in this folder can be deleted, as it does not end with :ro
 +     - "C:/Users/user_name/Desktop/my media:/mnt/media/my-media:ro" # import path in Windows system.
 ```
 
 :::tip
-The `ro` flag at the end only gives read-only access to the volumes. While Immich does not modify files, it's a good practice to mount read-only.
+The `ro` flag at the end only gives read-only access to the volumes. This will disallow the images from being delete in the web UI.
```
mertalev, line 115:
```suggestion
The `ro` flag at the end only gives read-only access to the volumes. This will disallow the images from being deleted in the web UI.
```

mertalev (approved):

PR #11124, opened 2024-07-15
fix(web): easier alt text translation for other languages

ben-basten:
## Description

Response to [the Discord feedback](https://discord.com/channels/979116623879368755/1250427404976132138/1260046995553259684) that the current alt text implementation is too rigid for languages with different sentence structures than English.

I rewrote the string constants to have all the alt text variations in full sentences, to allow differing sentence structures.

## How Has This Been Tested?

Created a unit test matrix to assert the correct output is being generated for different variations of the alt text.


web/src/lib/utils/thumbnail-util.ts
```diff
@@ -43,33 +43,44 @@ export const getAltText = derived(t, ($t) => {
...
+  if (count === 1) {
+    return '_1_person';
+  }
+
+  if (count === 2 || count === 3) {
+    return `_${count}_people`;
+  }
+
+  return '_4_or_more_people';
+};
```
jrasm91, line 86: I am not a fan of dynamically computed translation keys. It makes it really hard to find in the code and for other tools to integrate with.

ben-basten: Yeah I search by translation keys all the time, thanks for the feedback! I'll get this fixed.

web/src/lib/utils/thumbnail-util.ts
```diff
@@ -43,33 +43,49 @@ export const getAltText = derived(t, ($t) => {
...
-          count: names.length,
-          person1: names[0],
-          person2: names[1],
-          person3: names[2],
-          others: names.length > 3 ? names.length - 2 : 0,
-        },
-      });
-      altText += ` ${namesText}`;
+    if (peopleCount > 0) {
+      const key = getPlacePersonKey(peopleCount, hasPlace);
```
jrasm91, line 64: This looks better. Similarly it is easier to find translations if they use $t('actual_key'). Can you just move the `$t` part to be inside the method, along with the keys? I don't think there is any benefit to separating them.

ben-basten: Ok, this has been fixed. 

The `svelte-i18n` library doesn't export a type for the `$t` translation function, so there wasn't an easy way to pass a typed function closure into the method. Instead, I put the logic directly into the `getAltText` method.

jrasm91 (approved):

PR #11145, opened 2024-07-16
fix: downgrade exiftool-vendored to fix motion photo extraction

zackpollard:


danieldietzler (approved):

alextran1502 (approved):

server/package.json
```diff
@@ -60,7 +60,7 @@
     "class-transformer": "^0.5.1",
     "class-validator": "^0.14.0",
     "cookie-parser": "^1.4.6",
-    "exiftool-vendored": "~27.0.0",
+    "exiftool-vendored": "26.0.0",
```
jrasm91, line 63: I think we want to keep the ~ right?

zackpollard: Maybe, although it doesn't really matter as renovate will ignore that and try to bump anyway

PR #11161, opened 2024-07-17
fix(web): byte units enum

michelheusschen:
Returning the byte unit as a numeric enum in `getBytesWithUnit` is error-prone:
1. Manual mapping to a string can be forgotten
2. `ByteUnit.B` is zero and thus falsy, causing issues with `if (unit)` checks

Switching to a string enum solves these problems. This also fixes an issue I saw on discord for small viewports (the 3 should be GiB):

image


danieldietzler (approved):
Nice, thank you!

jrasm91 (approved):

PR #11174, opened 2024-07-18
chore(ml): set higher worker timeout for openvino

mertalev:
## Description

The first time a model is loaded into OpenVINO, it must be compiled into OpenVINO format. This process is intensive and can take over 120s. Since it also takes up the GIL and hence doesn't allow the server to respond to requests, it can exceed gunicorn's worker timeout and cause crashes. 

Increasing the default timeout for OpenVINO will prevent the service from crashing here.

### DIFF
--- a/docs/docs/install/environment-variables.md
+++ b/docs/docs/install/environment-variables.md
@@ -155,18 +155,18 @@ Redis (Sentinel) URL example JSON before encoding:
 
 ## Machine Learning
 
-| Variable                                         | Description                                                          |       Default       | Containers       |
-| :----------------------------------------------- | :------------------------------------------------------------------- | :-----------------: | :--------------- |
-| `MACHINE_LEARNING_MODEL_TTL`                     | Inactivity time (s) before a model is unloaded (disabled if \<= 0)   |        `300`        | machine learning |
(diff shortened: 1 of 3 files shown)


bo0tzz (commented):
LGTM

Can we do this ahead of time before pushing the model to HF?

bo0tzz (approved):

mertalev:
> Can we do this ahead of time before pushing the model to HF?

OpenVINO does device-specific optimizations so I'm not sure, but we can definitely look into it.

kszyd:
你好，我现在也遇到了同样的问题。设备是树莓派p4，通过omv-compose插件部署的immich，请问从哪个地方可以修改超时时间？我是一个新手。谢谢。

mertalev:
> 你好，我现在也遇到了同样的问题。设备是树莓派p4，通过omv-compose插件部署的immich，请问从哪个地方可以修改超时时间？能给出具体路径吗？我是一个新手。谢谢。

You can set `MACHINE_LEARNING_WORKER_TIMEOUT=3000` in the `.env` file and redeploy Immich to give it more time. But the Raspberry Pi 4 might not be powerful enough to run machine learning. I recommend using [remote machine learning](https://immich.app/docs/guides/remote-machine-learning) if possible to process the data on a more powerful device.

PR #11184, opened 2024-07-18
feat(mobile): remove asset from album in gallery view

fyfrey:
This PR adds a nice convenience feature to the album view of the mobile app:
You can now easily remove an asset from the album when viewing it in the gallery/detail view.
Previously, you would need go back to the asset grid, select it and remove it from the album (very error prone when comparing highly similar assets and you need to remove the correct one...)

This PR also changes the icon in the asset grid for "remove from album" to match the icon in the gallery view (and match what Google photos uses resp. what the material icon guidelines suggest for a remove action)

Before:

image

image

After:

image

image


alextran1502 (approved):
Thank you

PR #11206, opened 2024-07-19
feat(server): Allow activating non-admin user with server license

stephen304:
Here's a small suggestion to allow non-admin users to activate their user account with a server license. This would allow a user to move between a self-hosted server to server managed by someone else and not need to buy an individual user license if they already have a server license.

Only a few changes:
* Change user endpoint to accept both server/client licenses
* Change web logic so that the behavior for admins stays the same (server license -> server service, invididual license -> user service), but for non-admins both individual and server licenses go to the user service (currently entering a server license as a non-admin causes web to try to activate the server through the server endpoint, which of course fails due to not being an admin)
* Added test for user endpoint to accept server licenses

### DIFF
--- a/server/src/services/user.service.spec.ts
+++ b/server/src/services/user.service.spec.ts
@@ -274,10 +274,22 @@ describe(UserService.name, () => {
   });
 
   describe('setLicense', () => {
-    it('should save license if valid', async () => {
+    it('should save client license if valid', async () => {
       userMock.upsertMetadata.mockResolvedValue();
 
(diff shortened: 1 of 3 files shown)


jrasm91 (approved):

danieldietzler (approved):
This sounds logical to me, thanks!

stephen304:
Hmmm it appears I made an oversight, the user.isAdmin is not working right so it only activates the user even when you are admin and use a server key. It appears `isAdmin` is on the object (`UserAdminResponseDto`) on the definition but linting says otherwise, so I'm not sure how to make this work.

PR #11233, opened 2024-07-19
feat(ml): improved ARM-NN support

fyfrey:
- updates ARM-NN from version 23.11 to 24.05
- allow loading both fp16 and fp32 models with optional fp16 turbo
- new ML env config `ann_fp16_turbo` (default False) 
- new ML env config `ann_tuning_level` (default 2)
- add documentation


mertalev (approved):
LGTM. Thanks for fixing this!

mertalev:
Could you add the new env to the env docs? A mention of it in the hardware acceleration doc would be nice too for visibility.

mertalev (approved):

PR #11249, opened 2024-07-20
fix(server): exif description whitespace handling

### LINKED ISSUE #11238: Can't set image description to blank
### The bug

When trying to set the image description to blank, it defaults to what's stored in the photo's exif tag. There doesn't seem to be a way to clear this without going and editing the image file itself.

Also somewhat related I guess, I have a camera that produces image descriptions of just 31 blank spaces. Would it be possible to treat whitespace-only descriptions as non-existant? It bugs be when browsing photos and some have the "Add a description" text (because they are empty) and some are just missing it (because they are full of spaces).

TrueNAS Scale

### Version of Immich Server

v1.108.0

N/A

### Platform with the issue

- [X] Server
- [X] Web
- [ ] Mobile

michelheusschen:
Fixes #11238 which consists of two issues:
1. Not possible to clear description of an asset with the `Description` exif tag
2. Whitespace in exif description tags is included in the asset description

### DIFF
--- a/server/src/services/metadata.service.spec.ts
+++ b/server/src/services/metadata.service.spec.ts
@@ -710,6 +710,26 @@ describe(MetadataService.name, () => {
         }),
       );
     });
+
+    it('trims whitespace from description', async () => {
+      assetMock.getByIds.mockResolvedValue([assetStub.image]);
+      metadataMock.readTags.mockResolvedValue({ Description: '\t \v \f \n \r' });
+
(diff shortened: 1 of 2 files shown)


jrasm91 (approved):
Maybe just or it with '' and then trim it without the ?.

danieldietzler (approved):
Yeah, since there are already two ORs just adding `|| ''` at the end there makes sense to me as well

Gon128 (commented):
Fo9image

PR #11261, opened 2024-07-21
fix(server): correct openapi response type for getServerLicense()

michelheusschen:
Set schema manually to make sure the openapi spec has the correct response type

### DIFF
--- a/server/src/controllers/server.controller.ts
+++ b/server/src/controllers/server.controller.ts
@@ -1,5 +1,5 @@
 import { Body, Controller, Delete, Get, Put } from '@nestjs/common';
-import { ApiExcludeEndpoint, ApiOkResponse, ApiTags, getSchemaPath } from '@nestjs/swagger';
+import { ApiExcludeEndpoint, ApiNotFoundResponse, ApiTags } from '@nestjs/swagger';
 import { LicenseKeyDto, LicenseResponseDto } from 'src/dtos/license.dto';
 import {
   ServerAboutResponseDto,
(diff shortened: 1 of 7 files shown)


danieldietzler (approved):

server/src/controllers/server.controller.ts
```diff
@@ -95,6 +95,18 @@ export class ServerController {
...
+      oneOf: [
+        {
+          $ref: getSchemaPath(LicenseResponseDto),
+        },
+        {
+          type: 'null',
+        },
+      ],
+    },
+  })
```
jrasm91, line 109: I understand this works, but idk why we even return null in the first place. I'm fine with just throwing an error if you are asking for a resource that doesn't exist. You can always check the about response to know if it is licensed in the just place. We have no other endpoints that return a dto or null and I see no reason to start doing that here. 

(I understand you are just making the typing correct)

michelheusschen: Good point. I've updated it to return a 404 error for non-existing licenses, consistent with the user license endpoint.

jrasm91 (commented):
I think there is a debate between 404 meaning the url is wrong/doesn't exist vs 404 meaning the url is right and the resource didn't exist. 

In general I think the server codebase uses a 400 in this situation - it is the right url, but the client shouldn't request the resource without first creating it. 

This is part of a bigger discussion though. If we want to use 404 for this situation moving forward that is also fine with me. What do you think?

jrasm91 (approved):

PR #11286, opened 2024-07-22
feat(server): add IP trust list for reverse proxy

hitech95:
Added support for configurable trusted proxy IPs using environment variables. This enhancement allows you to specify trusted proxy IPs directly through environment settings, improving flexibility and security in various deployment environments.

It follow the rules of express `trust proxy` option. See [Behind Proxies](https://expressjs.com/en/guide/behind-proxies.html)

New Environment Variable:

    IMMICH_TRUSTED_PROXIES: A comma-separated list of IP addresses that should be trusted as proxies.

Example Usage in Docker Compose:

```yaml

version: '3.8'

immich-server:
    container_name: immich_server
    volumes:
    env_file:
      - .env
   environment:
      - TRUSTED_PROXY_IPS=192.168.0.1,192.168.0.2
    ports:
      - 2283:3001
    depends_on:
      - redis
      - database
    restart: always
```

With this fix, you can now manage trusted proxy IPs without modifying the application code, making it easier to adjust settings in different environments.

### DIFF
--- a/server/src/workers/api.ts
+++ b/server/src/workers/api.ts
@@ -14,9 +14,18 @@ import { useSwagger } from 'src/utils/misc';
 
 const host = process.env.HOST;
 
+function parseTrustedProxy(input?: string) {
+  if (!input) {
+    return [];
+  }
+  // Split on ',' char to allow multiple IPs
+  return input.split(',');
+}
+
 async function bootstrap() {
   process.title = 'immich-api';
   const otelPort = Number.parseInt(process.env.IMMICH_API_METRICS_PORT ?? '8081');
+  const trustedProxies = parseTrustedProxy(process.env.IMMICH_TRUSTED_PROXIES ?? '');
 
   otelStart(otelPort);
 
@@ -27,7 +36,7 @@ async function bootstrap() {
   logger.setAppName('Api');
   logger.setContext('Bootstrap');
   app.useLogger(logger);
-  app.set('trust proxy', ['loopback', 'linklocal', 'uniquelocal']);
+  app.set('trust proxy', ['loopback', 'linklocal', 'uniquelocal', ...trustedProxies]);
   app.set('etag', 'strong');
   app.use(cookieParser());
   app.use(json({ limit: '10mb' }));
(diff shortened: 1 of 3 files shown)


bo0tzz:
Please also add this to the [env var validation](https://github.com/immich-app/immich/blob/b53bd8c525c9760871529334b26571f672eabaf0/server/src/config.ts#L374) and [the docs](https://github.com/immich-app/immich/blob/b53bd8c525c9760871529334b26571f672eabaf0/docs/docs/install/environment-variables.md?plain=1#L39).

server/src/workers/api.ts
```diff
@@ -14,9 +14,18 @@ import { useSwagger } from 'src/utils/misc';
 
 const host = process.env.HOST;
 
+function parseTrustedProxy(input?: string) {
+  if (input === undefined || input === '') {
```
danieldietzler, line 18:
```suggestion
  if (!input) {
```

hitech95:
> Please also add this to the [env var validation].

I cannot validate the IPs without creating a huge regexp, can I just validate the string?
accepting suggestions for that.

Those should all be valid inputs and validating them precisely is a bit of a mess.

```
console.log(validateAndSplit("192.168.1.1;2001:db8::ff00:42:8329;10.0.0.0/24")); // ["192.168.1.1", "2001:db8::ff00:42:8329", "10.0.0.0/24"]
console.log(validateAndSplit("")); // []
console.log(validateAndSplit(undefined)); // []
```

server/src/workers/api.ts
hitech95: Ok, you prefere a less explicit check :D

danieldietzler: I think it's equally explicit :P
If you prefer keeping it I'm also fine with that though

hitech95: The end result is the same, I think that the explicit check is easier to see as not everyone knows that empty string is false.
I've already applied the change anyway 😄

danieldietzler (approved):
LGTM

bo0tzz (approved):

PR #11306, opened 2024-07-23
fix(server): DateTimeOriginal overwrite issue with sidecar file

### LINKED ISSUE #11263: Change of date/time for local asset fails silently
### The bug

Trying to change the date/time of an asset via the WebUI fails silently. I don't see anything in the container logs - even with verbose option activated. All assets are uploaded to Immich - no external libraries.

The UI allows the update, but no sidecar .xmp is created, and the UI does not reflect the change either.

Also trying to changing from the app on iOS leads to the same outcome.

Changing the location works perfectly fine, this makes me think it's not a permission issue at all.

Example video of the bug:

https://github.com/user-attachments/assets/0d723b47-ba0d-4845-8ff4-7319faf7ba6d

Ubuntu 22.04.4 LTS

### Version of Immich Server

v1.109.2

v1.109.0

### Platform with the issue

- [X] Server
- [X] Web
- [X] Mobile

```YAML
#
#
#
#

services:
  immich-server:
    container_name: immich_server
    # extends:
    volumes:
    devices:
      - /dev/dri:/dev/dri
    env_file:
      - .env
    ports:
      - 2283:3001
    depends_on:
      - redis
      - database
    restart: always
  immich-machine-learning:
    #   file: hwaccel.ml.yml
    volumes:
      - /data/dockerdata/immich/model-cache:/cache
    env_file:
      - .env
    restart: always
  redis:
    container_name: immich_redis
    image: docker.io/redis:6.2-alpine@sha256:328fe6a5822256d065debb36617a8169dbfbd77b797c525288e465f56c1d392b
    healthcheck:
    restart: always
  database:
    environment:
      POSTGRES_USER: ${DB_USERNAME}
      POSTGRES_INITDB_ARGS: --data-checksums
    volumes:
    healthcheck:
      test: pg_isready --dbname='${DB_DATABASE_NAME}' --username='${DB_USERNAME}' ||
        exit 1; Chksum="$$(psql --dbname='${DB_DATABASE_NAME}'
        --username='${DB_USERNAME}' --tuples-only --no-align --command='SELECT
        COALESCE(SUM(checksum_failures), 0) FROM pg_stat_database')"; echo
        "checksum failure count is $$Chksum"; [ "$$Chksum" = '0' ] || exit 1
      interval: 5m
      start_interval: 30s
      start_period: 5m
    command:
      - postgres
      - -c
      - shared_preload_libraries=vectors.so
      - -c
      - search_path="$$user", public, vectors
      - -c
      - logging_collector=on
      - -c
      - max_wal_size=2GB
      - -c
      - shared_buffers=512MB
      - -c
      - wal_compression=on
    restart: always
networks: {}
```

### Your .env content

```Shell

UPLOAD_LOCATION=/nfs/immich/library
DB_DATA_LOCATION=/data/dockerdata/immich/postgres

TZ=Europe/Rome

IMMICH_VERSION=release

DB_PASSWORD=REDACTED

# Set logs to verbose
IMMICH_LOG_LEVEL=verbose
DB_USERNAME=postgres
DB_DATABASE_NAME=immich
```

### Reproduction steps

```bash
1. Try to change date or time of an asset via WebUI or mobile app
2. No error output/no log output and no change in the date/time of the asset
```

### Relevant log output

_No response_

### Additional information

This has been reported in Discord by multiple users, relevant support tickets open here:
https://discord.com/channels/979116623879368755/1263448032893210716
https://discord.com/channels/979116623879368755/1264305376522801244
https://discord.com/channels/979116623879368755/1261312326544658432

michelheusschen:
When the sidecar file contains DateTimeOriginal and the dateTimeOriginal property of an asset is updated through the API, the updated date is returned as expected. However, metadata extraction runs immediately after and overwrites it with the value from the sidecar file. This happens because changes are written to CreationDate which has lower priority than DateTimeOriginal.

Fixes #11263 by writing changes to DateTimeOriginal in the sidecar file, which has the highest priority: https://github.com/immich-app/immich/blob/b53bd8c525c9760871529334b26571f672eabaf0/server/src/services/metadata.service.ts#L39-L49

### DIFF
--- a/e2e/src/api/specs/asset.e2e-spec.ts
+++ b/e2e/src/api/specs/asset.e2e-spec.ts
@@ -472,6 +472,44 @@ describe('/asset', () => {
       expect(status).toEqual(200);
     });
 
+    it('should update date time original when sidecar file contains DateTimeOriginal', async () => {
+      const sidecarData = `<?xpacket begin='?' id='W5M0MpCehiHzreSzNTczkc9d'?>
+<x:xmpmeta xmlns:x='adobe:ns:meta/' x:xmptk='Image::ExifTool 12.40'>
+<rdf:RDF xmlns:rdf='http://www.w3.org/1999/02/22-rdf-syntax-ns#'>
+ <rdf:Description rdf:about=''
+  xmlns:exif='http://ns.adobe.com/exif/1.0/'>
+  <exif:ExifVersion>0220</exif:ExifVersion>  <exif:DateTimeOriginal>2024-07-11T10:32:52Z</exif:DateTimeOriginal>
+  <exif:GPSVersionID>2.3.0.0</exif:GPSVersionID>
+ </rdf:Description>
+</rdf:RDF>
+</x:xmpmeta>
+<?xpacket end='w'?>`;
+
+      const { id } = await utils.createAsset(user1.accessToken, {
+        sidecarData: {
+          bytes: Buffer.from(sidecarData),
+          filename: 'example.xmp',
+        },
+      });
+      await utils.waitForQueueFinish(admin.accessToken, 'metadataExtraction');
+
+      const assetInfo = await utils.getAssetInfo(user1.accessToken, id);
+      expect(assetInfo.exifInfo?.dateTimeOriginal).toBe('2024-07-11T10:32:52.000Z');
+
+      const { status, body } = await request(app)
+        .put(`/assets/${id}`)
+        .set('Authorization', `Bearer ${user1.accessToken}`)
+        .send({ dateTimeOriginal: '2023-11-19T18:11:00.000-07:00' });
+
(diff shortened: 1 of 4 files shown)


bo0tzz (approved):
If someone writes SubSecDateTimeOriginal then the issue would still appear, but with the input we have it doesn't make sense for us to write edits to that field. Should we change the ordering of EXIF_DATE_TAGS? cc @alextran1502

michelheusschen:
SubSecDateTimeOriginal is a [composite tag](https://exiftool.org/TagNames/Composite.html) derived from DateTimeOriginal, so it shouldn't cause any issues

danieldietzler (approved):
LGTM

jrasm91 (approved):

PR #11326, opened 2024-07-24
feat: people infinite scroll

### LINKED ISSUE #11070: Number of people shown is limited to 500 (after update to v1.108.0)
### The bug

After updating from v1.107.2 to v1.108.0 the number of people shown in Immich Web is limited to 500.

v1.107.2 Explore -> People -> View all -> (15,721)
v1.108.0 Explore -> People -> View all -> (500)

Ubuntu 22.04.4 LTS

### Version of Immich Server

v1.108.0

1.108.0 build:148

### Platform with the issue

- [x] Server
- [x] Web
- [ ] Mobile

```YAML
#
#
#
#

name: immich

services:
  immich-server:
    container_name: immich_server
    volumes:
      - /daten/archive/Fotos:/mnt/media/archive:ro
    env_file:
      - .env
    ports:
      - 2283:3001
    depends_on:
      - redis
      - database
    restart: always

  immich-machine-learning:
    #   file: hwaccel.ml.yml
    volumes:
      - model-cache:/cache
    env_file:
      - .env
    restart: always

  redis:
    container_name: immich_redis
    image: registry.hub.docker.com/library/redis:6.2-alpine@sha256:84882e87b54734154586e5f8abd4dce69fe7311315e2fc6d67c29614c8de2672
    restart: always

  database:
    environment:
      POSTGRES_USER: ${DB_USERNAME}
    volumes:
    restart: always

  backup:
   image: prodrigestivill/postgres-backup-local
   env_file:
     - .env
   environment:
     POSTGRES_HOST: database
     POSTGRES_USER: ${DB_USERNAME}
     SCHEDULE: "@daily"
     BACKUP_DIR: /db_dumps
   volumes:
     - ${DB_BACKUP_LOCATION}:/db_dumps
   depends_on:
     - database

volumes:
  pgdata:
  model-cache:
```

### Your .env content

```Shell

DB_BACKUP_LOCATION=/daten/immich.app/backup
UPLOAD_LOCATION=/daten/immich.app/upload

IMMICH_VERSION=release

DB_PASSWORD=***************

DB_HOSTNAME=immich_postgres
DB_USERNAME=postgres
DB_DATABASE_NAME=immich

REDIS_HOSTNAME=immich_redis
```

### Reproduction steps

```bash
1. With version v1.107.2 go to Explore -> People -> View all
2. 15,721 people are shown
3. After update to v1.108.0 only 500 people are shown

The endpoint https://immich.some.domain/api/people?withHidden=true returns only 500 entries under people.
{
    "people": [ ... ],  <--- max 500 entries
    "total": 15778,
    "hidden": 43
}

Expected behavior:
- Show all 15778-43 people either bei introducing pagination, or by returning all
- Currently it is not possible to access people 500-15778
```

### Relevant log output

_No response_

### Additional information

_No response_

michelheusschen:
Adds pagination for people to the server and infinite scroll to the people page on web for viewing more then 500 people.

- Order by `person.createdAt` to have results in the same order every time
- `hasNextPage` is optional to keep backwards compatibility with the mobile app
- Web will see some performance issues when loading a lot of people. I plan to address this in a future PR using virtual scroll.

### DIFF
--- a/server/src/dtos/person.dto.ts
+++ b/server/src/dtos/person.dto.ts
@@ -1,4 +1,4 @@
-import { ApiProperty } from '@nestjs/swagger';
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 import { Type } from 'class-transformer';
 import { IsArray, IsInt, IsNotEmpty, IsString, Max, MaxDate, Min, ValidateNested } from 'class-validator';
 import { PropertyLifecycle } from 'src/decorators';
@@ -64,13 +64,15 @@ export class PersonSearchDto {
   @ValidateBoolean({ optional: true })
   withHidden?: boolean;
 
-  @ApiProperty({ description: 'Page number for pagination', default: 1, required: false })
+  /** Page number for pagination */
+  @ApiPropertyOptional()
   @IsInt()
   @Min(1)
   @Type(() => Number)
   page: number = 1;
 
-  @ApiProperty({ description: 'Number of items per page', default: 500, required: false })
+  /** Number of items per page */
+  @ApiPropertyOptional()
   @IsInt()
   @Min(1)
   @Max(1000)
--- a/server/src/dtos/person.dto.ts
+++ b/server/src/dtos/person.dto.ts
@@ -1,6 +1,6 @@
 import { ApiProperty } from '@nestjs/swagger';
 import { Type } from 'class-transformer';
-import { IsArray, IsNotEmpty, IsString, MaxDate, ValidateNested } from 'class-validator';
+import { IsArray, IsInt, IsNotEmpty, IsString, Max, MaxDate, Min, ValidateNested } from 'class-validator';
 import { PropertyLifecycle } from 'src/decorators';
 import { AuthDto } from 'src/dtos/auth.dto';
 import { AssetFaceEntity } from 'src/entities/asset-face.entity';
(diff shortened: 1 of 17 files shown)


server/src/dtos/person.dto.ts
```diff
@@ -63,6 +63,19 @@ export class MergePersonDto {
...
+  @IsInt()
+  @Min(1)
+  @Type(() => Number)
+  page: number = 1;
+
+  @ApiProperty({ description: 'Number of items per page', default: 500, required: false })
+  @IsInt()
+  @Min(1)
+  @Max(1000)
+  @Type(() => Number)
```
danieldietzler, line 77: I don't think you need the type annotation, do you?

danieldietzler (approved):
With the pagination not being optional this is a breaking change, correct? Besides, that, the code looks really good! You're on a rollllll

server/src/dtos/person.dto.ts
michelheusschen: The conversion is necessary because implicit conversions are disabled.  Pagination is optional with the same defaults as before, so there shouldn't be any breaking changes.

bo0tzz:
> Order by person.createdAt to have results in the same order every time

This means the person page is no longer ordered by how many photos there are of a person, right?

michelheusschen:
No, the current order is kept. If you have multiple unnamed people with the same photo count the order would previously be undefined which is a problem for pagination. Now it additionally sorts by `person.createdAt` to have a consistent sort order.

server/src/dtos/person.dto.ts
danieldietzler: Ah ok cool!

jrasm91: We only do because it is a query param so it would be a string otherwise.

server/src/dtos/person.dto.ts
```diff
@@ -63,6 +63,19 @@ export class MergePersonDto {
 export class PersonSearchDto {
   @ValidateBoolean({ optional: true })
   withHidden?: boolean;
+
+  @ApiProperty({ description: 'Page number for pagination', default: 1, required: false })
```
jrasm91, line 67: Pretty sure you can just use jsdoc comment and also default should come automatically from types

jrasm91 (approved):

server/src/dtos/person.dto.ts
michelheusschen: Yep that works

mertalev:
The only thing with this is that it will scan and sort all the faces in the table to count them, then do that again for the next page. But that’s still much better than not being able to see the next page at all.

PR #11332, opened 2024-07-24
feat: use immich hosted map tiles

zackpollard:
We are moving away from our third party provider and hosting our own vector map tile server as well as hosting the sprites and glyphs required for the maps

### DIFF
--- a/server/resources/style-light.json
+++ b/server/resources/style-light.json
@@ -3175,6 +3175,6 @@
       }
     }
   ],
-  "sprite": "https://protomaps.github.io/basemaps-assets/sprites/v3/dark",
-  "glyphs": "https://protomaps.github.io/basemaps-assets/fonts/{fontstack}/{range}.pbf"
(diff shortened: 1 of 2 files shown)


server/resources/style-light.json
```diff
@@ -1,1999 +1,3180 @@
...
       "paint": {
-        "text-color": "#633",
-        "text-halo-color": "rgba(255,255,255,0.7)",
-        "text-halo-width": 1
+        "text-color": "#a3a3a3"
       }
     }
   ],
-  "id": "immich-map-light"
+  "sprite": "https://protomaps.github.io/basemaps-assets/sprites/v3/dark",
```
michelheusschen, line 3178: Does it matter that the light style uses dark sprites?

alextran1502 (approved):

server/resources/style-light.json
zackpollard: Ah yes this should be changed, good spot

PR #11338, opened 2024-07-24
fix(deps): exiftool-vendored

jrasm91:
Follow up to #11145

### DIFF
--- a/e2e/src/api/specs/asset.e2e-spec.ts
+++ b/e2e/src/api/specs/asset.e2e-spec.ts
@@ -1137,7 +1137,7 @@ describe('/asset', () => {
           type: AssetTypeEnum.Image,
           originalFileName: '14bit-uncompressed-(3_2).arw',
           resized: true,
-          fileCreatedAt: '2016-01-08T15:08:01.000Z',
+          fileCreatedAt: '2016-01-08T14:08:01.000Z',
           exifInfo: {
             make: 'SONY',
             model: 'ILCE-7M2',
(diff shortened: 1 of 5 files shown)


danieldietzler (approved):

e2e/src/api/specs/asset.e2e-spec.ts
```diff
@@ -1137,7 +1137,7 @@ describe('/asset', () => {
           type: AssetTypeEnum.Image,
           originalFileName: '14bit-uncompressed-(3_2).arw',
           resized: true,
-          fileCreatedAt: '2016-01-08T15:08:01.000Z',
+          fileCreatedAt: '2016-01-08T14:08:01.000Z',
```
zackpollard, line 1140: Why the timestamp change?

server/package.json
```diff
@@ -60,7 +60,7 @@
     "class-transformer": "^0.5.1",
     "class-validator": "^0.14.0",
     "cookie-parser": "^1.4.6",
-    "exiftool-vendored": "26.0.0",
+    "exiftool-vendored": "^28.1.0",
```
zackpollard, line 63: Could we please fix and re-enable the e2e test for the motion photos before this gets merged in?

zackpollard (changes requested):

e2e/src/api/specs/asset.e2e-spec.ts
jrasm91: https://github.com/photostructure/exiftool-vendored.js/issues/187

zackpollard (approved):

PR #11347, opened 2024-07-25
chore(docs): compress homepage screenshots, 10x smaller

zackpollard:


benmccann:
I bet it'll be even better if you use avif. You could make it a picture tag and serve both. And ideally you could leave the original image and have a plugin transform it so that we also have the images everywhere else on the site optimized as well and don't accidentally add any non-optimized images in the future. Maybe you could use something like @docusaurus/plugin-ideal-image

zackpollard:
> I bet it'll be even better if you use avif. You could make it a picture tag and serve both. And ideally you could leave the original image and have a plugin transform it so that we also have the images everywhere else on the site optimized as well and don't accidentally add any non-optimized images in the future. Maybe you could use something like @docusaurus/plugin-ideal-image

Happy for a plugin to be added but not going to do that here. Also didn't want to deal with multiple formats unless we did have an automatic plugin for it. This was just a quick fix as the current 2MB file was a bit obnoxious

benmccann:
I actually can't find that there's any docusaurus plugin for this. Maybe we could use a general webpack plugin like image-minimizer-webpack-plugin

My webpack knowledge isn't great. I'm pretty sure I could do it if we were using vitepress 😆

jrasm91 (approved):

PR #11350, opened 2024-07-25
docs: update custom map style guide

pcouy:
Following my conversation with @bo0tzz and @zackpollard on discord, I'm proposing this PR that adds details to the instance settings page to clarify that enabling maps makes you rely on third parties.

I also came-up with a Nginx-based solution to build a simple caching proxy in front of a tile server, for which I included a guide in the docs.

There are still some things that I would like to address, but for these I need to know what is okay with you :

- I still think that because of this reliance on third parties, the map feature should be disabled by default, or at the very least a screen about this (featuring a toggle) should be shown during onboarding.
- It would be really nice to have a way to edit the `style.json` files directly from a textarea on the settings page
- The ideal would be to allow users to pick a `style.json` from a default list of pre-made ones, using different tile providers. Though this probably requires server admins to fill an API key for their chosen service, which may in turn require proxying requests to the tile servers in order not to leak the API key to every clients

In any case, I'm still worried about the current default tile provider and by the lack of transparency about it.


mmomjian:
Thanks for your PR. I was following this conversation. 

my recommendation is that this be hosted on a separate repo and a link provided under Community Guides. This level of specific instructions provides a large documentation burden to keep up to date that the current team isn’t familiar with.

jrasm91 (changes requested):
Everything looks good to me, with the exception of the nginx caching part. We should move that to an external gist/repo and add it the the community guides list instead.

jrasm91 (approved):

PR #11362, opened 2024-07-25
Enable `getAllAlbums` to map with assets by default to match documentation

aphex3k:
Enable `getAllAlbums` to be mapped with assets.

In the current code, the `getAllAlbums` endpoint will never return any assets with each album in the response.

Please:

1. Update the API documentation to reflect this fact or
2. Add argument `withoutAssets` to the endpoint or
3. Make the endpoint return assets with each album by default as documented (this PR)

Thank you


jrasm91 (changes requested):
Some users have thousands of albums and those each might have have thousands or tens of thousands of assets. This is specifically not supposed to return assets, otherwise it becomes unable for certain users.

aphex3k:
> This is specifically not supposed to return assets, otherwise it becomes unable for certain users.

Please update the documentation here: https://immich.app/docs/api/get-all-albums

According to the documentation, the second part of the response is a populated list of assets for each album.

PR #11369, opened 2024-07-26
fix(server): correct person birth date across timezones

### LINKED ISSUE #11354: [Bug] Date of birth in set incorrectly
### The bug

Sets the date of birth one day earlier if the current time zone is less than UTC (American).

TrueNAS-SCALE-23.10.2

### Version of Immich Server

v1.109.2

v1.109.2

### Platform with the issue

- [X] Server
- [X] Web
- [X] Mobile

```YAML
https://github.com/truenas/charts/tree/master/community/immich
```

### Your .env content

```Shell
https://github.com/truenas/charts/tree/master/community/immich
```

### Reproduction steps

```bash
1. TrueNAS - Application - Immich Configuration - Timezone: America/Winnipeg timezone
2. Go to the People, set date of birth a person, for example 1980/07/22
3. Then, if open "Set date of birth" again, I see 1980/07/21
```

### Relevant log output

_No response_

### Additional information

I tried to figure out.
Why is date of birth has format as a datetime on the client, but as a date on the server?

/mobile/openapi/lib/model/person_update_dto.dart
`
class PersonUpdateDto {
`
...
`
  DateTime? birthDate;
`
...
`
  @override
  String toString() => 'PersonUpdateDto[birthDate=$birthDate, featureFaceAssetId=$featureFaceAssetId, isHidden=$isHidden, name=$name]';
  Map<String, dynamic> toJson() {
    final json = <String, dynamic>{};
    if (this.birthDate != null) {
      json[r'birthDate'] = _dateFormatter.format(this.birthDate!.toUtc());
    } else {
`
...
`
  static PersonUpdateDto? fromJson(dynamic value) {
    if (value is Map) {
      final json = value.cast<String, dynamic>();

      return PersonUpdateDto(
        birthDate: mapDateTime(json, r'birthDate', r''),
`

And the same code in /mobile/openapi/lib/model/people_update_item.dart

But in /server/src/dtos/person.dto.ts
`
export class PersonResponseDto {
`
...
`
  @ApiProperty({ format: 'date' })
  birthDate!: Date | null;
`
and in /server/src/entities/person.entity.ts
`
export class PersonEntity {
`
...
`
  @Column({ type: 'date', nullable: true })
  birthDate!: Date | null;
`

michelheusschen:
The person's birth date is being set to the previous day when the server has a timezone with a negative UTC offset. Fixes #11354 by keeping `birthDate` as a string instead of converting it to a `Date`.

Testing this is difficult, because there doesn't seem to be a way to set the timezone during tests.

### DIFF
--- a/e2e/src/api/specs/person.e2e-spec.ts
+++ b/e2e/src/api/specs/person.e2e-spec.ts
@@ -6,10 +6,19 @@ import request from 'supertest';
 import { beforeAll, beforeEach, describe, expect, it } from 'vitest';
 
 const invalidBirthday = [
-  { birthDate: 'false', response: 'birthDate must be a date string' },
-  { birthDate: '123567', response: 'birthDate must be a date string' },
-  { birthDate: 123_567, response: 'birthDate must be a date string' },
-  { birthDate: new Date(9999, 0, 0).toISOString(), response: ['Birth date cannot be in the future'] },
+  {
+    birthDate: 'false',
+    response: ['birthDate must be a string in the format yyyy-MM-dd', 'Birth date cannot be in the future'],
+  },
+  {
+    birthDate: '123567',
+    response: ['birthDate must be a string in the format yyyy-MM-dd', 'Birth date cannot be in the future'],
+  },
+  {
+    birthDate: 123_567,
+    response: ['birthDate must be a string in the format yyyy-MM-dd', 'Birth date cannot be in the future'],
+  },
+  { birthDate: '9999-01-01', response: ['Birth date cannot be in the future'] },
 ];
 
 describe('/people', () => {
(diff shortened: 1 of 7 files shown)


e2e/src/api/specs/person.e2e-spec.ts
```diff
@@ -185,13 +194,13 @@ describe('/people', () => {
         .set('Authorization', `Bearer ${admin.accessToken}`)
         .send({
           name: 'New Person',
-          birthDate: '1990-01-01T05:00:00.000Z',
+          birthDate: '1990-01-01',
```
michelheusschen, line 197: Not sure if dates with a time should be accepted. I've removed the time part for now.

server/src/dtos/person.dto.ts
```diff
@@ -19,9 +19,11 @@ export class PersonCreateDto {
    * Person date of birth.
    * Note: the mobile app cannot currently set the birth date to null.
    */
-  @MaxDate(() => new Date(), { message: 'Birth date cannot be in the future' })
-  @ValidateDate({ optional: true, nullable: true, format: 'date' })
-  birthDate?: Date | null;
+  @ApiProperty({ format: 'date' })
+  @MaxDateString(() => new Date(), { message: 'Birth date cannot be in the future' })
+  @IsDateStringFormat('yyyy-MM-dd')
+  @IsOptional()
```
jrasm91, line 25: Can you use `@Optional({ nullable: true })` instead?

jrasm91 (approved):

danieldietzler (approved):
LGTM

PR #11388, opened 2024-07-27
feat(web): persist scroll position on navigation back to album

### LINKED ISSUE #11346: Returning to Albums ends at the top
### The bug

I have a lot of albums in the library "Album".

When I scroll down and select an album and then go back to albums, I end up at the top of the page again.
This only happens in web and not in the app.

I think it´s be better to stay at the scrollingpoint as it is in the app.

Ubuntu 22.10

### Version of Immich Server

v1.109.2

v1.109.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
name: immich

services:
  immich-server:
    container_name: immich_server
    volumes:
      - pictures:/usr/src/app/upload
    env_file:
      - .env
    ports:
      - 2283:3001
    depends_on:
      - redis
      - database
    restart: always

  immich-machine-learning:
    #   file: hwaccel.ml.yml
    volumes:
      - model-cache:/cache
    env_file:
      - .env
    restart: always

  redis:
    container_name: immich_redis
    image: registry.hub.docker.com/library/redis:6.2-alpine@sha256:84882e87b54734154586e5f8abd4dce69fe7311315e2fc6d67c29614c8de2672
    restart: always

  database:
    environment:
      POSTGRES_USER: ${DB_USERNAME}
    volumes:
    restart: always

volumes:
  model-cache:
  pictures:
    driver_opts:
      type: nfs
      o: addr=,nfsvers=3
      device: :/mnt/Backup/Immich_pictures
```

### Your .env content

```Shell

# UPLOAD_LOCATION=./library

IMMICH_VERSION=release
TZ=Europe/Stockholm

DB_PASSWORD=***

DB_HOSTNAME=immich_postgres
DB_USERNAME=postgres
DB_DATABASE_NAME=immich
DB_DATA_LOCATION=./postgres

REDIS_HOSTNAME=immich_redis
```

### Reproduction steps

```bash
1. Enter Albums
2. Scroll down and choose an ablum
3. Choose back
```

### Relevant log output

_No response_

### Additional information

_No response_

caburum:
I was going to use `history.state` to store the data originally, but due to the way the asset viewer writes the state and back buttons navigate directly to the page instead of using the history, I had to implement it using `sessionStorage` to save the scroll location (which is cleared when leaving the group of pages).

Search page doesn't return to the exact position as I couldn't figure out where to hook the navigation properly, but now seems to at least center the clicked item instead of going to the top.

Fixes #11346


mertalev:
@midzelis can correct me if I'm wrong, but I think this is covered by #10646, which makes some *huge* changes to how the timeline works.

caburum:
I don't believe it does the same thing. This PR fixes navigation back from a specific album page (`/albums/:id`) to the list of albums (`/albums`) by scrolling to the proper position in the list.

It doesn't affect anything related to the timeline except for a slight fix for the timeline on the search page that uses the GalleryViewer. If that conflicts with the timeline PR, I can just remove that as it's a different bug from the albums and people fixes.

jrasm91 (commented):
I'm not a huge fan of this because the browser has built in stuff to keep track of scroll position between page navigation. Is there a way we could use that instead?

caburum:
I believe the reason the browser can't do it is because the main scrollable area is nested inside of multiple elements, only the page's scroll position is saved. Doing it this way handles both the browser's back/forward navigation, clicking the in-app back button (which can't use the browser's history as the asset viewer messes up the state), and also for when merging a face navigates back.

Internally, SvelteKit does basically the same thing (but using the history API to store a scroll key that can be looked up in session storage, which doesn't cover all 3 of our cases).
image

jrasm91:
It seems like we're basically grabbing and saving the scroll position for an element before navigate and setting it back after navigate. Is there a way we can make this re-usable? Like using an action that binds to an element and saves it's scroll position automatically when enabled?

caburum:
That's probably doable, one concern is current for the infinite loading on the people page it will load the pages up to the correct one before scrolling, but that could probably be added easily as well. I can take a look at implementing that tomorrow/this weekend.

caburum:
@jrasm91 I've moved the repeated code into a set of actions which should make it a lot cleaner.

eacunha:
Keep doing good work. I love immich. I really dislike that it always go to the top. S2

zackpollard:
@caburum Hey! I'm going through old PRs to try to get them moved along or closed. What's the state of this PR, was it awaiting review or was this still more work to be done on it?

caburum:
@zachpollard It was awaiting review, but I should probably check up on it and might have to make some changes since the Svelte 5 upgrade.

zackpollard:
> @zackpollard It was awaiting review, but I should probably check up on it and might have to make some changes since the Svelte 5 upgrade.

Alright, ping me when you've had a chance to do that and then I'll take a look! Sorry that your PR was left for so long 😅

caburum:
@zackpollard Just checked it and everything still seems to be working, I don't think the Svelte 5 upgrade actually affected anything that was changed.

zackpollard:
> @zackpollard Just checked it and everything still seems to be working, I don't think the Svelte 5 upgrade actually affected anything that was changed.

That's good, I think you may have checked slightly too early though :sweat_smile: There were two PRs for svelte 5, one was done a little while back, and one was merged yesterday to mostly move to the new syntax so there are some more merge conflicts, sorry about that...

caburum:
@zackpollard I've merged in what should be the latest svelte 5 changes & updated the props I changed so it should be good now.

zackpollard:
> @zackpollard I've merged in what should be the latest svelte 5 changes & updated the props I changed so it should be good now.

That's great, thanks!  @jrasm91 you had opinions on this being a reusable module which it now is, do you want to re-review it now?

jrasm91 (approved):
LGTM

zackpollard (approved):
Thanks for the contribution! LGTM!

Issue #9777, opened 2024-05-26
External Libraries not finding files


ddshd:
### The bug

External libraries not finding files - even if using some of the force update options.

Here are all the logs. Contains entries after hitting any of the Scan Library options but nothing happen. No change in CPU usage either.
```
[Nest] 8  - 05/26/2024, 3:23:03 PM VERBOSE [ImmichServer] [LoggingInterceptor] [u3e32tun] GET /api/user/info/8f9c3933-1533-4d20-bb28-7e5475af550b 200 1.12ms ::ffff:192.168.176.1
[Nest] 8  - 05/26/2024, 3:23:05 PM VERBOSE [ImmichServer] [LoggingInterceptor] [qiih1asb] POST /api/library/748f0c0a-98d9-47a3-9794-973088ced227/validate 200 0.72ms ::ffff:192.168.176.1
[Nest] 8  - 05/26/2024, 3:25:57 PM VERBOSE [ImmichServer] [LoggingInterceptor] [lge1fweh] POST /api/library/748f0c0a-98d9-47a3-9794-973088ced227/scan 204 12.26ms ::ffff:192.168.176.1
[Nest] 8  - 05/26/2024, 3:25:58 PM VERBOSE [ImmichServer] [LoggingInterceptor] [r60akxtv] POST /api/library/748f0c0a-98d9-47a3-9794-973088ced227/scan 204 7.32ms ::ffff:192.168.176.1
[Nest] 8  - 05/26/2024, 3:25:59 PM VERBOSE [ImmichServer] [LoggingInterceptor] [3phemmzn] POST /api/library/748f0c0a-98d9-47a3-9794-973088ced227/scan 204 10.61ms ::ffff:192.168.176.1
[Nest] 8  - 05/26/2024, 3:26:02 PM VERBOSE [ImmichServer] [LoggingInterceptor] [411pcg5l] POST /api/library/90bc7ca9-7fbc-4629-bc89-e2cd3b74a079/scan 204 11.73ms ::ffff:192.168.176.1
[Nest] 8  - 05/26/2024, 3:26:02 PM VERBOSE [ImmichServer] [LoggingInterceptor] [mxssdzel] POST /api/library/6d17b379-1c1b-40ae-85f7-a42dfaaf165e/scan 204 11.58ms ::ffff:192.168.176.1
[...]

PR #11389, opened 2024-07-27
Fix/9777 documentation

iplaughlin:
Updated instructions on how to properly implement and use external libraries. Relates to #9777.

This fixes two problems that were identified when I was going through the documentation and setting up external libraries.

- The original documentation states to update the docker-compose.yml in the `immich-server` section only. This causes the path to be validated in the UI, but nothing is ran because the microservices container does not have access to the path. Updated the documentation to show that you need to add it in both places for this to work.
- The original documentation shows `/home/user/photos1` in the docker-compose.yml, but uses `/usr/src/app/external` in the demonstration. The documentation was updated to show that this path should be the same path as found in the docker-compose.yml

### DIFF
--- a/docs/docs/guides/external-library.md
+++ b/docs/docs/guides/external-library.md
@@ -6,11 +6,19 @@ in a directory on the same machine.
 
 # Mount the directory into the containers.
 
-Edit `docker-compose.yml` to add one or more new mount points in the section `immich-server:` under `volumes:`.
-If you want Immich to be able to delete the images in the external library, remove `:ro` from the end of the mount point.
+Edit `docker-compose.yml` to add one or more new mount points in two sections:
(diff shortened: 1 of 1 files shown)


jcbevns:
This is still an issue
```
immich-server:
    volumes:
+       - /home/user/photos1:/home/user/photos1:ro
+       - /mnt/photos2:/mnt/photos2:ro # you can delete this line if you only have one mount point, or you can add more lines if you have more than two
```
https://github.com/immich-app/immich/blob/main/docs/docs/guides/external-library.md?plain=1#L16

then...

`Enter /usr/src/app/external as the path and click Add`
https://github.com/immich-app/immich/blob/main/docs/docs/guides/external-library.md?plain=1#L44

PR #11413, opened 2024-07-28
docs: file custom location

mmomjian:
Update file custom locations guide for monolith container


bo0tzz (approved):

alextran1502 (approved):

psla:
QQ: Currently, `thumbs` include both thumbnails and previews.

When using these custom locations, is that still the case? I was thinking of putting `THUMB_LOCATION` on an SSD, but don't need the previews in the SSD.

PR #11430, opened 2024-07-29
feat(server): reverse geocoding endpoint

jrasm91:
Added e2e tests for #11370 via a new reverse geocoding endpoint (`GET /map/reverse-geocode`)

### DIFF
--- a/e2e/src/api/specs/map.e2e-spec.ts
+++ b/e2e/src/api/specs/map.e2e-spec.ts
@@ -172,7 +172,7 @@ describe('/map', () => {
         .get('/map/reverse-geocode?lon=123')
         .set('Authorization', `Bearer ${admin.accessToken}`);
       expect(status).toBe(400);
-      expect(body).toEqual(errorDto.badRequest(['lat must be a latitude string or number']));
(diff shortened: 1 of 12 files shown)


e2e/src/api/specs/map.e2e-spec.ts
```diff
@@ -159,4 +159,75 @@ describe('/map', () => {
...
+      expect(status).toBe(400);
+      expect(body).toEqual(errorDto.badRequest(['lat must be a latitude string or number']));
+    });
+
+    it('should throw an error if a lat is out of range', async () => {
+      const { status, body } = await request(app)
+        .get('/map/reverse-geocode?lat=91&lon=123.456')
+        .set('Authorization', `Bearer ${admin.accessToken}`);
+      expect(status).toBe(400);
+      expect(body).toEqual(errorDto.badRequest(['lat must be a latitude string or number']));
```
danieldietzler, line 191: Why is this repo method always throwing the same error? That's kinda annoying imo and makes the error a bit useless

danieldietzler (approved):

PR #11453, opened 2024-07-30
refactor(server): stacks

jrasm91:
Refactor stacks to use a dedicated controller and set of endpoints.

TODO:
- Mobile
- e2e tests


server/src/repositories/stack.repository.ts
```diff
@@ -1,23 +1,95 @@
...
+      },
+      relations: {
+        assets: {
+          exifInfo: true,
+        },
+      },
+    });
+  }
+
+  async create(entity: { ownerId: string; assetIds: string[] }): Promise<StackEntity> {
```
danieldietzler, line 31: Do you think it's performance-wise fine to just do that in (well-readable) TS and not on the DB directly? Presumably you can assume that stacks are always "small", so it's even less of an issue?

server/src/services/stack.service.ts
```diff
@@ -0,0 +1,77 @@
...
+    const stacks = await this.stackRepository.search({
+      ownerId: auth.user.id,
+      primaryAssetId: dto.primaryAssetId,
+    });
+
+    return stacks.map((stack) => mapStack(stack, { auth }));
+  }
+
+  async create(auth: AuthDto, dto: StackCreateDto): Promise<StackResponseDto> {
+    await this.access.requirePermission(auth, Permission.ASSET_READ, dto.assetIds);
```
danieldietzler, line 32: I'm not sure if read permissions should be sufficient here. In my mind `ASSET_UPDATE` would make more sense

server/src/services/stack.service.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  async get(auth: AuthDto, id: string): Promise<StackResponseDto> {
+    await this.access.requirePermission(auth, Permission.STACK_READ, id);
+    const stack = await this.findOrFail(id);
+    return mapStack(stack, { auth });
+  }
+
+  async update(auth: AuthDto, id: string, dto: StackUpdateDto): Promise<StackResponseDto> {
+    await this.access.requirePermission(auth, Permission.STACK_WRITE, id);
```
danieldietzler, line 48: Assuming you add new assets to the stack, you'd probably also want to enforce the same permissions for them you'd enforce when creating a stack

web/src/lib/utils/asset-utils.ts
```diff
@@ -329,77 +332,52 @@ export const stackAssets = async (assets: AssetResponseDto[]) => {
...
-        stackParentId: parent.id,
+    const stack = await createStack({ stackCreateDto: { assetIds: assets.map(({ id }) => id) } });
+    notificationController.show({
+      message: $t('stacked_assets_count', { values: { count: stack.assets.length } }),
+      type: NotificationType.Info,
+      button: {
+        text: $t('view_stack'),
+        onClick() {
+          return assetViewingStore.setAssetId(stack.primaryAssetId);
+        },
```
danieldietzler, line 346: I was not aware this is valid syntax lmao

danieldietzler (commented):
Looking great so far!

server/src/dtos/stack.dto.ts
```diff
@@ -1,9 +1,38 @@
...
 }
+
+export class StackResponseDto {
+  id!: string;
+  primaryAssetId!: string;
+  assets!: AssetResponseDto[];
+}
+
+export const mapStack = (stack: StackEntity, { auth }: { auth?: AuthDto }) => {
+  const primary = stack.assets.filter((asset) => asset.id === stack.primaryAssetId);
```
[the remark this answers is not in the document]
jrasm91: This just made it easier to combine down below if it was always a list.

server/src/services/stack.service.ts
jrasm91: Yeah, good point.

server/src/repositories/stack.repository.ts
jrasm91: The issue is we need to delete the stacks before creating new ones. I need the list of children from the stacks that are deleted. It seems a lot better to do that in a transaction so we don't delete them and then it fails and now the old stacks are gone.

server/src/services/stack.service.ts
jrasm91: I believe update only supports changing the primary asset id, which already has to be in the stack, so I don't think any checks are necessary about asset visibility.

server/src/repositories/stack.repository.ts
danieldietzler: That makes a lot of sense, fair. Making it a fair bit more resilient

zackpollard: Yea, this was what we decided in our initial talks, this all needs to happen in a transaction 😃

alextran1502 (approved):

akostadinov:
Could this cause issues on android 1.114.0 https://github.com/immich-app/immich/issues/7286#issuecomment-2344562996 ?

PR #11465, opened 2024-07-31
fix: make HTML valid

benmccann:
You can't have a `tbody` inside a `button`. Needs to be fixed in order to migrate to Svelte 5


jrasm91 (approved):
Is there a way to lint for these types of issues or not until svelte 5?

benmccann:
I'm not sure there is now, but Svelte 5 will address it shortly

PR #11559, opened 2024-08-04
refactor(server): startup checks for vector extension

mertalev:
- Fixes the 0.1.1 `extversion` not being corrected to 0.1.11 in the DB, preventing an upgrade from versions 1.91.0-1.94.2 to 1.106.0 or later
- Fixes the reindexing check assuming the existence of the `face_search` table, preventing an upgrade from versions 1.91.0-1.94.2 to 1.107.0 or later
- Throws an error if the available version is outside of the allowed version range since this means the wrong version is installed. Creating or updating to this just makes downgrading more difficult
- Does not recommend running `ALTER EXTENSION UPDATE` unless the update query itself failed
- Throws immediately if the extension isn't created and isn't available
- Checks if extension is nightly *before* creating it
- Runs `pgvectors_upgrade()` on minor or major updates even if there's no schema change
- Throws an error if the Postgres image was downgraded to have a lower version of an extension than what's activated in the DB
  - This will just throw an uglier and less helpful error later
  - Could benefit from official downgrading instructions, but it's enough steps to warrant its own documentation PR
- Gets both current and available versions in one query
- Refactors tests to be less coupled to current version constraints and avoid manually duplicating the test suite for each extension
- Adds a few helper methods in the database service for readability

### How Has This Been Tested?
- Tested upgrading directly from 1.90.2 to main
  - Confirmed a successful startup with a correct DB layout containing the expected vector data
  - Search and face detection both worked
- Tested on an existing instance with no issues during startup
- Tested upgrading from 0.2.0 to 0.2.1, confirming it did update and started up successfully
- Tested that switching to the pgvecto.rs 0.3.0 image raised the correct error

Also tested that 0.3.0 works following these changes, but updating can be a separate PR.

### DIFF
--- a/server/src/repositories/database.repository.ts
+++ b/server/src/repositories/database.repository.ts
@@ -7,6 +7,7 @@ import {
   DatabaseExtension,
   DatabaseLock,
   EXTENSION_NAMES,
+  ExtensionVersion,
   IDatabaseRepository,
   VectorExtension,
   VectorIndex,
@@ -29,20 +30,14 @@ export class DatabaseRepository implements IDatabaseRepository {
     this.logger.setContext(DatabaseRepository.name);
   }
 
-  async getExtensionVersion(extension: DatabaseExtension): Promise<string | undefined> {
-    const res = await this.dataSource.query(`SELECT extversion FROM pg_extension WHERE extname = $1`, [extension]);
-    return res[0]?.['extversion'];
-  }
-
-  async getAvailableExtensionVersion(extension: DatabaseExtension): Promise<string | undefined> {
-    const res = await this.dataSource.query(
-      `
-    SELECT version FROM pg_available_extension_versions
-    WHERE name = $1 AND installed = false
-    ORDER BY version DESC`,
+  async getExtensionVersion(extension: DatabaseExtension): Promise<ExtensionVersion> {
+    const [res]: ExtensionVersion[] = await this.dataSource.query(
+      `SELECT default_version as "availableVersion", installed_version as "installedVersion"
+      FROM pg_available_extensions
+      WHERE name = $1`,
       [extension],
     );
-    return res[0]?.['version'];
+    return res ?? { availableVersion: null, installedVersion: null };
   }
 
   async getPostgresVersion(): Promise<string> {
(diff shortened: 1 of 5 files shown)


danieldietzler (approved):
Nice!

server/src/services/database.service.spec.ts
```diff
@@ -9,287 +11,254 @@ describe(DatabaseService.name, () => {
...
+      }
+      versionBelowRange = `${minVersion.major}.${minVersion.minor - 1}.${minVersion.patch}`;
+      minVersionInRange = minVersion.version;
+      patchUpgradeInRange = minVersion.inc('patch')!.version;
+      versionAboveRange = minVersion.inc('major')!.version;
+      databaseMock.getExtensionVersion.mockResolvedValue({
+        installedVersion: minVersionInRange,
+        availableVersion: minVersionInRange,
+      });
+    });
```
jrasm91, line 46: This defeats the whole point of the unit test in my opinion. You aren't actually testing with specific values you are programmatically writing tests to test the code. You could just as easily introduce a bug in _this_ code. Tests should have less programming and more static assertions otherwise they lose a lot of value. I previously changed the tests to be less like that and now you have just reverted everything again. 

Like we should not be using semver to generate tests for our code that _relies_ on semver.

jrasm91 (changes requested):

server/src/services/database.service.spec.ts
mertalev: The existing unit tests are a huge pain honestly. They're brittle because they stop testing the same thing as soon as a version constraint changes, they're repetitive because you have the same test for both extensions, and there were several mistakes in the unit tests that mixed up the two because they're hard to spot. It's just all very manual and error-prone.

jrasm91: Maybe we need to move getting the expected range to a repository call, because then the tests could validate everything without relying on a specific version that will change over time. Programming the brittleness away also isn't the answer.

jrasm91: Unless we have logic that is different/specific for a particular extension we don't need to be unit trying everything twice.

mertalev: > Maybe we need to move getting the expected range to a repository call, because then the tests could validate everything without relying on a specific version that will change over time.

I can add a repo method for that, mock it in the `beforeEach` with a specific range and set fixed values for the other variables. Does that sound good?

> Unless we have logic that is different/specific for a particular extension we don't need to be unit testing everything twice.

I think it's easier and less error-prone to just run them on both. I originally tried working with the existing unit tests, but going back and forth to see if this behavior is being tested for this extension is no fun. It's also easy to forget to add it for one, make a change in one test and not apply it to the other, etc.

jrasm91: Mocking the version and using static variables for the rest is fine. 

Testing both extensions all the time is also fine. The problem for me is when your test cases start having stuff like "if the extension is this, then do this one off assertions". Basically it is the same problem again. 

You could run a bunch of tests for both with the same assertions, but if there is some specific logic for a specific scenario that should be in a dedicated test imo (without if statements).

mertalev: That makes sense. I think the "suggest the other extension" check falls in that category, so I can move that out into two dedicated tests.

server/src/repositories/database.repository.ts
```diff
@@ -29,20 +30,14 @@ export class DatabaseRepository implements IDatabaseRepository {
...
-  }
-
-  async getAvailableExtensionVersion(extension: DatabaseExtension): Promise<string | undefined> {
-    const res = await this.dataSource.query(
-      `
-    SELECT version FROM pg_available_extension_versions
-    WHERE name = $1 AND installed = false
-    ORDER BY version DESC`,
+  async getExtensionVersion(extension: DatabaseExtension): Promise<ExtensionVersion> {
+    const [res]: ExtensionVersion[] = await this.dataSource.query(
```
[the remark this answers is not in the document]
mertalev: `query` returns `any`, so this is just to cast it to a more helpful type.

server/src/services/database.service.spec.ts
jrasm91: They would be perfect, thanks.

jrasm91 (approved):

PR #11560, opened 2024-08-04
feat(mobile): add support for material themes

pruthvi-21:
Added support for custom theming.

</br>
</br>
<p>With color interface turned on, primary color will be applied to the backgrounds.</p>

</br>
</br>
<p>In Android, we can use system color</p>


moritzbindewald:
❤️

alextran1502:
Hello, thank you for the PR. Can you please have the default color as the current color used? It is the branding color of Immich

pruthvi-21:
Hi @alextran1502, actually the ColorScheme class adjusted the contrast of those colors based on light/dark mode. Now, I've overridden those with Immich brand logo colors fetched from the website.

And, what font weights are you referring to?

Also, when turning of the colorful interface do we want dark background or a black background?

alextran1502:
@pruthvi-21 thank you for the quick addressing of the comments.

I am seeing the font weight's changes of the text from the Settings page

alextran1502:
Can you preserve the shading of all the components? For example, the bottom app bar and the top bar background color now have a light gray shade, which creates a more confined feeling of the actual screen space of the timeline.

pruthvi-21:
I believe the bottom navigation bar has always been gray. As for the top bar, it dynamically changes color when the user scrolls through the timeline, which is the default behavior and gives the app a more modern feel. 

Do you want a static color instead of this ?

alextran1502:
> I believe the bottom navigation bar has always been gray. As for the top bar, it dynamically changes color when the user scrolls through the timeline, which is the default behavior and gives the app a more modern feel.

This is not correct. You can see how it works in the release version. I understand that it dynamically changes the color, and we don't want it to change the color because we don't want to create a separation feeling of the top/bottom bar, which will narrow the perceived field of view of the actual available content.

Please change the color and leave all the other behavior/styling as is.

pruthvi-21:
@alextran1502 like this?

alextran1502:
Yes, is the stop bar still changing color when scroll?

alextran1502 (approved):
This PR has a lot of good changes, thank you very much!

pruthvi-21:
@alextran1502 did you reset the immich logo in immich_app_bar.dart to only use the brand colors? if so, we also need to reset in login page and app_bar_dialog for consistency.

PR #11580, opened 2024-08-05
feat(web): add Exif-Rating

stumpigit:
This PR enables the rating tag to be read from the Exif data and saves it in the database. The rating can be displayed in the photo viewer using a star display and can also be customised with the appropriate rights.

Rating

This PR contains

- Database migration
- Adaptation to the DTOs
- Adjustments to detail-panel.svelte with integration of the Svelte component [svelte-star-rating ](https://github.com/ErnaneJ/svelte-star-rating)

This PR fulfils parts of the following discussions:
[#7100 ](https://github.com/immich-app/immich/discussions/7100)
[#3619](https://github.com/immich-app/immich/discussions/3619)

As a further development, the option of filtering by rating will be implemented.


web/package.json
```diff
@@ -23,6 +23,7 @@
     "prepare": "svelte-kit sync"
   },
   "devDependencies": {
+    "@ernane/svelte-star-rating": "^1.1.7",
```
bo0tzz, line 26: I don't think we should need a whole separate dependency to show a few stars.

bo0tzz (dismissed):
Nice!

web/package.json
stumpigit: Ok, I understand. The star library is really very small, I could easily include it in the code as a component, if that makes sense?

bo0tzz: I'll let Alex have the final call on the looks, but to me the styling of this star-rating component looks pretty out of place in Immich, so it might be better to just make our own thing.

alextran1502:
Thank you for the PR. I have difficulty convincing myself to display the star count because it is tailored for photo editing rather than viewing/browsing software. 

My thought process is that if you only show the star count in the Info panel, it doesn't help you find more "likes" when browsing the timeline, and if we show that information on the timeline, it gets cluttered.

My main concern is that it gets the UI clutter while doesn't serve a lot of purpose

stumpigit:
> Thank you for the PR. I have difficulty convincing myself to display the star count because it is tailored for photo editing rather than viewing/browsing software.
> 
> My thought process is that if you only show the star count in the Info panel, it doesn't help you find more "likes" when browsing the timeline, and if we show that information on the timeline, it gets cluttered.
> 
> My main concern is that it gets the UI clutter while doesn't serve a lot of purpose

Thank you Alex for your feedback and for your work in general. From my point of view, the discussions in #3619 and also general discussions on the internet about the large cloud provider show the need for this function. As a photographer, it is very important to me that I can assign stars in my photo workflow. I want to have all photos online, but quickly filter only those that have more than 3 stars, for example. I then use these to create an album. A lot of image editing tools and many cameras support the rating and are based on this system.

In my opinion, Synology Photo is a good example of how this has been implemented (you can try it out online at https://demo.synology.com/de-de/dsm). The rating appears in the info panel and then in the filter for the photos. I think this way the user interface is not overloaded, is intuitive and gives many photographers a powerful tool at hand.

syno rating
syno rating2

server/src/dtos/asset.dto.ts
```diff
@@ -46,6 +46,9 @@ export class UpdateAssetBase {
   @IsLongitude()
   @IsNotEmpty()
   longitude?: number;
+
+  @Optional()
```
jrasm91, line 50: should this assert the value is bigger than or equal to 0? also we need a validator to enforce it is a number.

also, I think this should allow "clearing" the rating by setting it to `null`.

server/src/entities/exif.entity.ts
```diff
@@ -95,6 +95,9 @@ export class ExifEntity {
   @Column({ type: 'integer', nullable: true })
   bitsPerSample!: number | null;
 
+  @Column({ type: 'integer', nullable: true })
```
jrasm91, line 98: should this to a decimal and allow values like 1.5, 2.5, etc?

web/package.json
```diff
@@ -85,4 +86,4 @@
   "volta": {
     "node": "20.16.0"
   }
-}
+}
```
jrasm91, line 89: revert this and add the newline back please

jrasm91 (commented):
Looks pretty good! Can you add an e2e test for setting the rating on an asset as well? (`e2e/` folder)

stumpigit:
I have just comitted the following adjustments:

- The Svelte-Star-Rating dependency has been removed
- Developed my own component "detail-panel-rating.svelte", which creates the stars with SVG and Tailwind. These are also somewhat finer than the first proposal. So I hope that they fit better into the design.
- Proposal to place the rating just below the description (modelled on Synology Photos)

Screenshot 2024-08-05 203845

mertalev:
Just to offer another perspective: I never use this feature in photo apps, personally. It generally adds visual clutter and the way the stars tend to change color when you hover over them is a nuisance. This last version you shared doesn't look too intrusive, though.

server/src/dtos/asset.dto.ts
stumpigit: Thanks. Correct, I added three validators:   @IsInt()  @Max(5)  @Min(0)
The Exif library doesn't allow to "clear" the rating. Just set it to 0.

server/src/entities/exif.entity.ts
stumpigit: I looked at the exif-specs. It should be Integer.

web/package.json
stumpigit: Sorry for that...

jrasm91:
Another option is to have this be a setting the use can turn on our off

alextran1502:
> Another option is to have this be a setting the use can turn on our off

yep, user's preference to the rescue

web/src/lib/components/asset-viewer/detail-panel-rating.svelte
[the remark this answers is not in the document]
stumpigit: OK, my idea was that it is the same as detail-panel-description. It is part of the detail panel. Or would you put the star rating in the Shared Components directory?

stumpigit:
So should I try to add it in the user settings? Under app settings? I'm welcome to try that.

alextran1502:
@stumpigit yeah, let do that. Please use the user preference setting mechanism for this, so that the user preference will be used across the clients

server/src/dtos/asset.dto.ts
danieldietzler: You should definitely wrap those in code blocks `` (using backticks) to not ping random people lmao.

danieldietzler: Fixed for you.

web/src/lib/components/asset-viewer/detail-panel-rating.svelte
```diff
@@ -0,0 +1,102 @@
...
+  import { updateAsset, type AssetResponseDto } from '@immich/sdk';
+  import { t } from 'svelte-i18n';
+
+  export let asset: AssetResponseDto;
+  export let isOwner: boolean;
+
+  const countStars = 5;
+
+  $: rating = asset.exifInfo?.rating || 0;
+  $: currentRating = rating;
```
danieldietzler, line 12: Why are `currentRating` and `rating` reactive?

web/src/lib/components/asset-viewer/detail-panel-rating.svelte
```diff
@@ -0,0 +1,102 @@
...
+  const handleChangeRating = async (newRating: number) => {
+    try {
+      await updateAsset({ id: asset.id, updateAssetDto: { rating: newRating } });
+    } catch (error) {
+      handleError(error, $t('cant_apply_changes'));
+    }
+    rating = currentRating;
+  };
+
+  const onChangeRating = () => {
```
danieldietzler, line 23: Pretty sure you can also pass an async function to `on:change`, so you could `await handleChangeRating`.

danieldietzler (commented):
Looking pretty good so far! :)

web/src/lib/components/asset-viewer/detail-panel-rating.svelte
stumpigit: You're right, it only needs one reactive variable. Has been corrected.

web/src/lib/components/asset-viewer/detail-panel-rating.svelte
stumpigit: That's right, now onChangeRating is no longer needed

web/src/lib/components/asset-viewer/detail-panel-rating.svelte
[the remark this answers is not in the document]
stumpigit: Habe es unbenannt

stumpigit:
Now you can set in the User Settings under App Settings whether the star rating should be displayed.
Screenshot 2024-08-07 095958
Unfortunately, I'm not quite sure how the process works with the translations. I've comitted en.json and de.json.

web/src/lib/components/asset-viewer/detail-panel-star-rating.svelte
```diff
@@ -0,0 +1,98 @@
...
+  let currentRating = rating;
+
+  const handleChangeRating = async () => {
+    if (currentRating === asset.exifInfo?.rating) {
+      return;
+    }
+    try {
+      await updateAsset({ id: asset.id, updateAssetDto: { rating: currentRating } });
+    } catch (error) {
+      handleError(error, $t('cant_apply_changes'));
```
michelheusschen, line 21: Message doesn't exist

```suggestion
      handleError(error, $t('errors.cant_apply_changes'));
```

web/src/lib/components/asset-viewer/detail-panel-star-rating.svelte
```diff
@@ -0,0 +1,98 @@
...
+    <div class="flex">
+      {#each { length: countStars } as _, id}
+        {#if rating > id}
+          <span
+            ><svg
+              xmlns="http://www.w3.org/2000/svg"
+              viewBox="0 0 24 24"
+              fill="currentColor"
+              transform="scale(0.9,0.9)"
+              class="w-6 h-6 text-yellow-400 cursor-pointer"
```
michelheusschen, line 38: I don't think we need two separate star icons. Why not add styling such as `text-yellow-400` conditionally when `rating > id`?

web/src/lib/components/asset-viewer/detail-panel.svelte
```diff
@@ -161,6 +163,9 @@
   {/if}
 
   <DetailPanelDescription {asset} {isOwner} />
+  {#if $preferences?.app_settings?.rating}
+    <DetailPanelRating {asset} {isOwner} />
```
michelheusschen, line 167: This won't be visible for shared links

web/src/lib/i18n/en.json
```diff
@@ -956,6 +956,7 @@
   "purchase_server_description_2": "Supporter status",
   "purchase_server_title": "Server",
   "purchase_settings_server_activated": "The server product key is managed by the admin",
+  "rating": "Rating",
```
michelheusschen, line 959: `Star rating` would be clearer

web/src/lib/components/asset-viewer/detail-panel-star-rating.svelte
```diff
@@ -0,0 +1,98 @@
...
+        name="starrating"
+        class="opacity-0
+        cursor-pointer
+        absolute top-[0] -left-[30px] w-[120%] right-[0] h-full"
+        type="range"
+        min="0"
+        max={countStars}
+        step="1"
+        on:change={handleChangeRating}
+        on:click
```
michelheusschen, line 81: `on:click` should be removed since the parent component doesn't use it

web/src/lib/components/asset-viewer/detail-panel-star-rating.svelte
```diff
@@ -0,0 +1,98 @@
...
+      {/each}
+    </div>
+
+    {#if isOwner}
+      <input
+        name="starrating"
+        class="opacity-0
+        cursor-pointer
+        absolute top-[0] -left-[30px] w-[120%] right-[0] h-full"
+        type="range"
```
michelheusschen, line 76: Pretty creative way to use a range input, but I see a few potential issues.
1. Positioning for the fifth star will be a bit off when using 70% zoom or less
2. It's not really obvious how to remove all stars
3. Scrolling while hovering the stars will change them after receiving focus
4. The `range` type should be added here to allow keyboard interaction with the input https://github.com/immich-app/immich/blob/5b64456f4821b87d3a4b11076a2803953e918994/web/src/lib/actions/shortcut.ts#L23
5. A label should be added for accessibility

web/src/lib/components/asset-viewer/detail-panel-star-rating.svelte
stumpigit: Thank you Michel, just changed it

web/src/lib/components/asset-viewer/detail-panel.svelte
stumpigit: Yes, I think that is ok? Just added !isSharedLink() to the if statement for safety

web/src/lib/components/asset-viewer/detail-panel-star-rating.svelte
stumpigit: Thank you. Yes, I took this from a library and saw that this is also done in other places. But I didn't like it either. That's why I changed it. The stars are now buttons (can also be selected with Tab). There is also a crossed-out star to set the rating to 0. I hope this corresponds to your feedback?

alextran1502:
Sorry put in the wrong number for the issue :P

mertalev:
Wait, so a PR with the `fixes` keyword can close other PRs?

alextran1502:
> Wait, so a PR with the `fixes` keyword can close other PRs?

YEP, TIL

stumpigit:
I have adapted the behaviour of the stars to the usual behaviour patterns. The design has also become a little more refined.

https://github.com/user-attachments/assets/5c68e58d-6f6b-4455-bd1c-88947714d67c

web/src/lib/components/asset-viewer/detail-panel-star-rating.svelte
```diff
@@ -0,0 +1,40 @@
+<script lang="ts">
+  import { handleError } from '$lib/utils/handle-error';
+  import { updateAsset, type AssetResponseDto } from '@immich/sdk';
+  import { t } from 'svelte-i18n';
+  import Star from '$lib/components/shared-components/settings/star-rating.svelte';
```
jrasm91, line 5: Can you call this `StarRating`?

web/src/lib/components/user-settings-page/app-settings.svelte
```diff
@@ -185,6 +207,14 @@
           bind:checked={$sidebarSettings.sharing}
         />
       </div>
+      <div class="ml-4">
+        <SettingSwitch
+          title={$t('rating')}
+          subtitle={$t('show_rating_details_panel')}
```
jrasm91, line 213: Can you change these keys to be `rating` and `rating_description`?

jrasm91 (approved):
Looks good to me! Thanks for working on this

web/src/lib/components/asset-viewer/detail-panel-star-rating.svelte
```diff
@@ -0,0 +1,40 @@
...
+      }
+      return;
+    }
+    try {
+      await updateAsset({ id: asset.id, updateAssetDto: { rating: currentRating } });
+    } catch (error) {
+      handleError(error, $t('errors.cant_apply_changes'));
+    }
+    rating = currentRating;
+  };
```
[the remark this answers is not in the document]
stumpigit: thanks for the hint. Indeed, it looks much nicer now

server/src/entities/user-metadata.entity.ts
```diff
@@ -31,6 +31,9 @@ export enum UserAvatarColor {
 }
 
 export interface UserPreferences {
+  app_settings: {
```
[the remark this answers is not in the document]
stumpigit: I would like to keep it that way. The parent "app_settings" corresponds to the accordion in the UI. This is also how it works with the other settings in the json. The proposed change would not correspond to the rest.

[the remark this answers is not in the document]
jrasm91: This is wrong though. App settings in general is for client-side only settings. This is a user preference that is being synced with the database and will be the same across all settings, so it isn't an "app setting". Also, the app setting accordion might not even be the right place for this setting, but I don't know exactly where else to put it, so it's probably fine for now. Basing the naming convention in the server codebase on a tentative UI location (at best) is not a justifiable reason to keep it there IMO.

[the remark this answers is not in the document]
jrasm91: The convention we are using is for everything else in user settings is
```ts
{  
  [feature: string]: { 
    enabled: boolean, 
    <other feature settings> }
  }
}
```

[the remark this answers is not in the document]
stumpigit: Ok, thanks for the comments. Then I misunderstood. I will gladly make the change.
I wasn't sure about the location of the setting either, so I chatted with Alex about it via Discord. He thought the App Settings location was ok there.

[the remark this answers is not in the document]
jrasm91: I just did final testing and cleaned up a few things. I can push the rename if you want, as I've already done it.

jrasm91 (approved):
I think this is good to go now. I changed a few things with the code:
- StarRating wasn't actually a rating component, it was just a single star icon. I changed it to be the full component with the interface of 
```
<StarRating rating={3} count={5} onRating={(rating => ... } readOnly={true}/>
```
- Swapped out the svg + path elements for `<Icon/>` which are use everywhere else for icons
- Changed the color to immich-purple, which is obviously subjective we can change it to yellow if there is a consensus on the preferred color
- Changed the hover functionality to only revert back to the set value when the mouse leaves the _entire component_ - this prevents jank
- Changed the conditional rendering logic to be inside of the Component instead of in the asset viewer, which is already big enough as it is
- Moved the click/blur/tab handlers to the _button_ instead of the svg path element
- Fixed e2e formatting
- Renamed app_settings to rating, etc.

danieldietzler (approved):
LGTM

Orbitally:
> I have adapted the behaviour of the stars to the usual behaviour patterns. The design has also become a little more refined.
> 
> 
> 
> https://github.com/user-attachments/assets/5c68e58d-6f6b-4455-bd1c-88947714d67c
> 
> 
> 
> 

I know this has already been merged but just figured I'd chime in quickly anyways with a tiny suggestion. The outline of the star you're hovering holding is a really clean way of displaying the interaction but my initial thought when I saw it was that it may make more sense if all the stars leading up to the one being hovered bolded as well. I wasn't sure if this would just help with continuity since they all become coloured once the star gets clicked as well as just a visual indicator that you're about to apply all the stars up to and including the one you click.

stumpigit:
Thank you for your feedback @Orbitally. @jrasm91 has already made some adjustments to the behaviour of the stars. In the merged version it now looks like this:

https://github.com/user-attachments/assets/d9d2caaf-4ed3-416c-8f83-87999fae64af

From my point of view it makes sense. I have seen very different behaviours of the stars in my research. But it certainly fits in well with the further operation of Immich.

Orbitally:
Nice! I didn't catch that change but that looks great too!

PR #11603, opened 2024-08-06
fix(server): avoid transcoding thumbnail streams

mincrmatt12:
Currently the only property of video streams used when picking one to transcode is their lengths (always trying to select the longest one). Certain video formats embed thumbnails as "video" streams but then don't report back their lengths in frames via `ffprobe`.

For example, some WMVs from an ancient camera of mine have an embedded thumbnail as stream 0 and the actual video as stream 1, but `ffprobe` doesn't report an `nb_frames` for either so they both get treated as length 0 and the first one is selected, which causes `ffmpeg` to explode.

This patch adds a second layer to the sort for "main video streams", where if two streams are of the same length but one was marked by `ffprobe` as having the disposition `attached_pic` (which is used for video thumbnails), the one _not_ marked as `attached_pic` is preferred. In this way, `ffmpeg` is given the correct stream and successfully generates a transcoded version of my ancient WMVs on a test instance. I wouldn't be surprised if this is the problem behind #11100 as well, but there isn't enough information in that report to check.

Currently the only consumer of this new `holdsThumbnail` property is the transcoding logic; there is probably some more intelligent way to use it when generating our own video thumbnails but I have not explored that here.

### DIFF
--- a/server/src/repositories/media.repository.ts
+++ b/server/src/repositories/media.repository.ts
@@ -76,6 +76,7 @@ export class MediaRepository implements IMediaRepository {
       },
       videoStreams: results.streams
         .filter((stream) => stream.codec_type === 'video')
+        .filter((stream) => !stream.disposition?.attached_pic)
         .map((stream) => ({
           index: stream.index,
           height: stream.height || 0,


server/src/repositories/media.repository.ts
```diff
@@ -86,6 +86,7 @@ export class MediaRepository implements IMediaRepository {
           rotation: Number.parseInt(`${stream.rotation ?? 0}`),
           isHDR: stream.color_transfer === 'smpte2084' || stream.color_transfer === 'arib-std-b67',
           bitrate: Number.parseInt(stream.bit_rate ?? '0'),
+          holdsThumbnail: (stream.disposition?.attached_pic ?? 0) != 0,
```
mertalev, line 89: Wouldn't it be simpler to just filter for streams where this expression is `false`?

mincrmatt12: You could, I was just going for the minimal change that would fix my files, and filtering streams that previously _might_ have worked seemed like there was a remote chance it would break some esoteric video.

Also, if we keep the streams around in our metadata it might make it easier to change the video thumbnail generation later to prioritize using these streams for that purpose.

mertalev: I don't think we need to worry about the former: a video thumbnail stream isn't a video stream, so if that's the only stream it's not going to end well anyway.

To the latter point, that could be interesting, but even then I'd prefer for the thumbnail stream to be separate from the video streams. `probe` could return a `thumbnailStreams` object down the line if we need it.

mincrmatt12: Alright, I've updated the PR to just do stream filtering, which does simplify the changes significantly.

I looked around but I couldn't find any tests of the `probe()` implementation itself (just ones using mocked results from it) so there's no longer an associated unit test for this.

mertalev (approved):

server/src/repositories/media.repository.ts
```diff
@@ -76,6 +76,7 @@ export class MediaRepository implements IMediaRepository {
       },
       videoStreams: results.streams
         .filter((stream) => stream.codec_type === 'video')
+        .filter((stream) => (stream.disposition?.attached_pic ?? 0) == 0)
```
mertalev, line 79:
```suggestion
        .filter((stream) => !stream.disposition?.attached_pic)
```

mincrmatt12:
Just checking, am I supposed to do something about the CI failures on this? I don't see any useful error on the erroring job.

mertalev:
The GH runner ran out of storage apparently 😅

PR #11610, opened 2024-08-06
fix(server): video duration extraction

### LINKED ISSUE #7312: Video duration is 0:00
### The bug

I have two relatively large (6gb) 4k videos that were uploaded from the Android app. In the timeline their duration displays as 0:00 but when opening them, the duration is correct.

Ubuntu 22.04

### Version of Immich Server

v1.95.1

v1.95.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1. Open the website and note the duration showing as 0:00
```

### Additional information

_No response_

### LINKED ISSUE #8244: Video length is always incorrect
### The bug

image
It was displayed as 78 min. After putting the mouse hovering on the play button, the correct length is shown (2 min)
image

Arch

### Version of Immich Server

1.99.0

N/A

### Platform with the issue

- [X] Server
- [X] Web
- [ ] Mobile

```YAML
https://github.com/immich-app/immich/releases/download/v1.99.0/docker-compose.yml
```

### Your .env content

```Shell
https://github.com/immich-app/immich/releases/download/v1.99.0/example.env
```

### Reproduction steps

```bash
1. Upload videos
2. Many videos got incorrect length
```

### Additional information

_No response_

michelheusschen:
Extracting the video duration based on EXIF data has proven to be unreliable. Exiftool has [7 different ways](https://exiftool.org/TagNames/Composite.html) for calculating the `Duration` value, and if any of those are missing or invalid the duration will also be wrong. Using `ffprobe` is a more reliable solution to determine the video duration.

Using an [MKV test file](https://github.com/ietf-wg-cellar/matroska-test-files/blob/e6965e5ca666322ed93e2748a10a4f132309e005/test_files/test1.mkv) the duration is now correctly extracted as `00:01:27.336`, whereas it was previously `24:15:36.000`. This should fix #7312 and fix #8244, though I can't verify that myself

### DIFF
--- a/server/src/services/metadata.service.ts
+++ b/server/src/services/metadata.service.ts
@@ -214,28 +214,7 @@ export class MetadataService implements OnEvents {
     const { exifData, tags } = await this.exifData(asset);
 
     if (asset.type === AssetType.VIDEO) {
-      const { videoStreams } = await this.mediaRepository.probe(asset.originalPath);
-
-      if (videoStreams[0]) {
-        switch (videoStreams[0].rotation) {
-          case -90: {
-            exifData.orientation = Orientation.Rotate90CW;
-            break;
-          }
-          case 0: {
-            exifData.orientation = Orientation.Horizontal;
-            break;
-          }
-          case 90: {
-            exifData.orientation = Orientation.Rotate270CW;
-            break;
-          }
-          case 180: {
-            exifData.orientation = Orientation.Rotate180;
-            break;
-          }
-        }
-      }
+      await this.applyVideoMetadata(asset, exifData);
     }
 
     await this.applyMotionPhotos(asset, tags);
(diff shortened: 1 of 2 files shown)


server/src/services/metadata.service.ts
```diff
@@ -567,16 +546,33 @@ export class MetadataService implements OnEvents {
...
-    let _seconds = seconds as number;
+  private async applyVideoMetadata(asset: AssetEntity, exifData: ExifEntityWithoutGeocodeAndTypeOrm) {
+    const { videoStreams, format } = await this.mediaRepository.probe(asset.originalPath);
 
-    if (typeof seconds === 'object') {
-      _seconds = seconds.Value * (seconds?.Scale || 1);
-    } else if (typeof seconds === 'string') {
-      _seconds = Duration.fromISOTime(seconds).as('seconds');
+    if (videoStreams[0]) {
+      switch (videoStreams[0].rotation) {
```
mertalev, line 553: Not blocking since this is existing code, but only checking the first stream isn't reliable. The media service uses the stream with the highest frame count, which may or may not be the first stream.

mertalev (approved):

alextran1502 (approved):

mio-19:
How do I apply this to existing videos?

jrasm91 (approved):

PR #11624, opened 2024-08-06
chore(server): support pgvecto.rs 0.3.0

mertalev:
## Description

There were no visible issues when upgrading or creating a new instance with 0.3.0 during my testing. Smart search jobs, search, face detection, facial recognition, and changing search models all worked.

Upgrading just required a restart of the database. For non-superuser instances upgrading from 0.2.x to 0.3.0, I believe these two commands are all that's needed, followed by restarting the database:

```sql
ALTER EXTENSION vectors UPDATE;
SELECT pgvectors_upgrade();
```

I'm fine with bumping the default to 0.3.0, but can also wait a bit in case any issues pop up.

### DIFF
--- a/server/src/constants.ts
+++ b/server/src/constants.ts
@@ -4,7 +4,7 @@ import { join } from 'node:path';
 import { SemVer } from 'semver';
 
 export const POSTGRES_VERSION_RANGE = '>=14.0.0';
-export const VECTORS_VERSION_RANGE = '0.2.x';
+export const VECTORS_VERSION_RANGE = '>=0.2 <0.4';
 export const VECTOR_VERSION_RANGE = '>=0.5 <1';
 
 export const NEXT_RELEASE = 'NEXT_RELEASE';


jrasm91 (approved):

masterwishx:
> ## Description
> There were no visible issues when upgrading or creating a new instance with 0.3.0 during my testing. Smart search jobs, search, face detection, facial recognition, and changing search models all worked.
> 
> Upgrading just required a restart of the database. For non-superuser instances upgrading from 0.2.x to 0.3.0, I believe these two commands are all that's needed, followed by restarting the database:
> 
> ```sql
> ALTER EXTENSION vectors UPDATE;
> SELECT pgvectors_upgrade();
> ```
> 
> I'm fine with bumping the default to 0.3.0, but can also wait a bit in case any issues pop up.

So we need only restart if using Docker image `tensorchord/pgvecto-rs:pg16-v0.2.0`?

mertalev:
Basically yes. Immich will handle the rest.

masterwishx:
> Basically yes. Immich will handle the rest.

Have error after updated : 

image

masterwishx:
Using `ghcr.io/imagegenius/immich:latest` image , working after `SELECT pgvectors_upgrade();` and restart Postgres

masterwishx:
or i was need to restart container only?

jsimcina:
> or i was need to restart container only?

I had to fully restart postgres (or if using cnpg, kubectl cnpg restart) then restart the container

mertalev:
> or i was need to restart container only?

Immich already runs `SELECT pgvectors_upgrade();`, so restarting the container would have been enough. This is fine too, though.

PR #11643, opened 2024-08-07
feat: change web asset detail map to zoom level 12.5

zackpollard:
Zoom level right now is pretty much useless, we should start the asset viewer map zoom a bit further out to give more context on the location. You're still able to zoom in more if you want more detail. This will also help with higher cache-hit percentage and faster responses, as zoom level 12 is more likely to be cached

### DIFF
--- a/web/src/lib/components/asset-viewer/detail-panel.svelte
+++ b/web/src/lib/components/asset-viewer/detail-panel.svelte
@@ -439,17 +439,17 @@
           },
         ]}
         center={latlng}
-        zoom={15}
+        zoom={12.5}
         simplified
         useLocationPin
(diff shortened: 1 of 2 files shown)


danieldietzler (approved):

alextran1502 (approved):

jrasm91 (approved):
I'm not sure if all tile providers support 12.5. I know some users use other providers and this might break for them if their provider doesn't.

zackpollard:
> I'm not sure if all tile providers support 12.5. I know some users use other providers and this might break for them if their provider doesn't. 

This just sets the map scaling, but the underlying request is still for zoom level 12, so it should work with any z/x/y provider

PR #11658, opened 2024-08-08
feat: editor endpoints

jrasm91:
Add a server endpoint to create assets from edits. For example,

```ts
{
  id: asset.id,
  edits: [
    { action: 'adjust', brightness, saturation, hue, lightness },
    { action: 'crop', region: { top, left, width, height },
    { action 'rotate', angle: 90 },
  ]
}
  ```

(basically a wrapper around sharp, with the output file being added as a new asset for the caller)


foobarbecue:
I would really love to be able to use this feature. Is there anything I can help with? I'm new to immich and have coded in lots of languages but not dart. Am willing to test.

aronsky:
I'd just like to add that another use-case would be for transferring backups from Google Photos. When an image is edited in Google Photos, only the edited version is shown in the library (at least, in certain conditions), though the original is still available for download. In the takeout archive, both the original and the edited image are present. I would like to import such pairs into Immich and be able to somehow instruct Immich to only show the edited copy, but to provide the option to get the original, if I so desire.

Not sure this is the right place to suggest this, but I came here from #11074 - and the server-side part of image editing sounded most relevant.

ecw74:
I would like to have this feature too. After the migration of my photo albums (~60k pictures and videos) there are a lot of old pictures with a wrong orientation. It would be nice if I could rotate this in the web frontend (maybe als as batch operation).

seraasch:
Upvote

Howy75:
This basic feature to rotate photos and save the changed image on the server (external library with write privilegs) would be highly appreciated.

jagfirerwalker:
How can I help get the pending check resolved?

jrasm91:
Closing in favor of #24155

PR #11678, opened 2024-08-09
docs: Update docs + example.env for DB_PASSWORD

mmomjian:
Getting a lot of support tickets for this, especially passwords with spaces or `$`

### DIFF
--- a/docker/example.env
+++ b/docker/example.env
@@ -12,6 +12,7 @@ DB_DATA_LOCATION=./postgres
 IMMICH_VERSION=release
 
 # Connection secret for postgres. You should change it to a random password
+# Please use only the characters `A-Za-z0-9`, without special characters or spaces
 DB_PASSWORD=postgres
 
 # The values below this line do not need to be changed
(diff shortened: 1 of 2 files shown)


docker/example.env
```diff
@@ -12,6 +12,7 @@ DB_DATA_LOCATION=./postgres
 IMMICH_VERSION=release
 
 # Connection secret for postgres. You should change it to a random password
+# Please use only the characters `A-Za-z0-9`, without special characters or spaces
```
danieldietzler, line 15: Shouldn't we rather recommend just quoting the password?

mmomjian: We could. If we did, we should add single quotes to the default docker-compose.yml and specify that single quotes are an unallowed value.

This seemed easier to me, because I helped someone on Discord earlier this week who had the password set to (snippet of PW): `\æ¨Þèùdïîû`

I don't really know if everything downstream will work with such a password tbh

danieldietzler: OMG lmao. Ok I'm definitely happy with this then lmao
It's not like the DB should be exposed anyways

danieldietzler (approved):

jrasm91 (approved):

PR #11685, opened 2024-08-10
fix(server): follow symlinks when zipping assets

### LINKED ISSUE #9335: Symlinked photos in external library download as a zip of symlinks
### The bug

I have some photos that are presented to immich in an "external library".  That external library happens to be a git-annex repository, so all of the photos are actually symlinks to somewhere in its `.git` directory.  Immich displays these files with no problems.  However, if one tries to download a set of them, one ends up with a zip file full of symlinks and no actual images.

For example, here's what I get from unzipping a zip that immich generated that contained two images:

```
 ls -l
total 8
lrwxrwxrwx. 1 rob rob 207 May  8 20:17 IMG_6611.JPG -> ../../../.git/annex/objects/kp/74/SHA256E-s6007094--24cb5836a3634c8c663ea53846cf05d8c1471dbfbe6d61e346db0c944d002e42.JPG/SHA256E-s6007094--24cb5836a3634c8c663ea53846cf05d8c1471dbfbe6d61e346db0c944d002e42.JPG
lrwxrwxrwx. 1 rob rob 207 May  8 20:17 IMG_6613.JPG -> ../../../.git/annex/objects/vx/jw/SHA256E-s5979551--0895fe458d242ba5ea5fbefc3e7816c7621ed54e0bd8f06471f2431f30e64b5c.JPG/SHA256E-s5979551--0895fe458d242ba5ea5fbefc3e7816c7621ed54e0bd8f06471f2431f30e64b5c.JPG
```

As you can see, both of those files are symlinks, and not actually image files.

Downloading the images individually works fine.

nixos

### Version of Immich Server

1.103.1

n/a

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
n/a: Not using docker-compose -- immich is running in podman.
```

### Your .env content

```Shell
n/a
```

### Reproduction steps

```bash
1. Have an external library that has images that are symlinks to somewhere internally within the library.
2. Select two of those images in the immich web interface
3. Observe that the resulting zip file does not actually contain the images
...
```

### Relevant log output

_No response_

### Additional information

_No response_

C-Otto:
fixes #9335

### DIFF
--- a/server/src/services/download.service.ts
+++ b/server/src/services/download.service.ts
@@ -1,4 +1,5 @@
 import { BadRequestException, Inject, Injectable } from '@nestjs/common';
+import fs from 'node:fs/promises';
 import { parse } from 'node:path';
 import { AccessCore, Permission } from 'src/cores/access.core';
 import { AssetIdsDto } from 'src/dtos/asset.dto';
@@ -7,7 +8,7 @@ import { DownloadArchiveInfo, DownloadInfoDto, DownloadResponseDto } from 'src/d
 import { AssetEntity } from 'src/entities/asset.entity';
 import { IAccessRepository } from 'src/interfaces/access.interface';
 import { IAssetRepository } from 'src/interfaces/asset.interface';
-import { IStorageRepository, ImmichReadStream } from 'src/interfaces/storage.interface';
+import { ImmichReadStream, IStorageRepository } from 'src/interfaces/storage.interface';
 import { HumanReadableSize } from 'src/utils/bytes';
 import { usePagination } from 'src/utils/pagination';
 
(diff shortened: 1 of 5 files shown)


server/src/services/download.service.ts
```diff
@@ -83,10 +84,10 @@ export class DownloadService {
         filename = `${parsedFilename.name}+${count}${parsedFilename.ext}`;
       }
 
-      zip.addFile(originalPath, filename);
+      promises.push(zip.addFile(originalPath, filename));
```
jrasm91, line 87: Why don't you just get the real path here and revert every other change?

C-Otto:
This works fine on my instance, but I'm unsure of how the test code works on filesystems that don't support symlinks (Windows?). Furthermore, as `git annex` protects files by marking them as read-only, the corresponding attributes are also used for the ZIP file returned by immich. It'd be a bit more convenient to have those files with `rw-r-r` flags instead - but that might be an unrelated fix.

jrasm91 (approved):
I changed the `realpath` to a repo call, which made the test a lot easier to write. I also added `mode: 644`, which sets the mode on the file in the archive. Do you want to re-test it on your library?

C-Otto:
Thanks, that works like a charm! :rocket:

PR #11700, opened 2024-08-11
feat(server): crop and rotate

zdimension:
This PR adds support for using the Orientation EXIF field from the sidecar (which can be modified without touching the original file). 

It also adds two actions to the web viewer that "demo" that feature:
image

Issues:
- the web viewer displays images in two ways: as thumbnails, and through the zooming component. The zooming component doesn't seem to take the EXIF orientation in account, so the image displays in its "raw" orientation when one zooms.
  - people on the Discord server have suggested to also render the "edited" version of the original image, in original resolution. This would mean images with EXIF orientation would have an original resolution "thumbnail" render that would be used for viewing. This would make it easy to display images without having to constantly process the rotation client-side.
- after changing the rotation, the thumbnail needs to be refreshed. This is currently done by triggering a thumbnail regeneration task and refreshing the `img` tag's source field after 500 milliseconds.


server/src/dtos/asset.dto.ts
```diff
@@ -74,6 +75,10 @@ export class UpdateAssetDto extends UpdateAssetBase {
   @Optional()
   @IsString()
   description?: string;
+
+  @Optional()
+  @IsIn(['1', '2', '3', '4', '5', '6', '7', '8'])
```
jrasm91, line 80: Might be worth using the enum here

server/src/services/media.service.ts
```diff
@@ -206,8 +207,40 @@ export class MediaService {
...
+            if ([
+              ExifOrientation.MirrorHorizontal,
+              ExifOrientation.MirrorVertical,
+              ExifOrientation.MirrorHorizontalRotate90CW,
+              ExifOrientation.MirrorHorizontalRotate270CW,
+            ].includes(asset.exifInfo.orientation as ExifOrientation)) {
+              imageOptions.mirror = true;
+            }
+          }
+
```
jrasm91, line 243: Let's put this in a method which returns { angle, mirror }

zdimension: Should I put it in media.service.ts? Or in some kind of ExifOrientationUtil class defined near ExifOrientation?

jrasm91: The service is fine

server/src/dtos/asset.dto.ts
zdimension: done

server/src/services/media.service.ts
zdimension: done

server/src/dtos/asset.dto.ts
jrasm91: You should use `@IsEnum` and @ApiProperty` for the enum actually. You can look at other usages in the codebase.

zdimension: done

web/src/lib/components/asset-viewer/actions/rotate-action.svelte
```diff
@@ -0,0 +1,82 @@
...
+  import { photoViewer } from '$lib/stores/assets.store';
+
+  export let asset: AssetResponseDto;
+  export let onAction: OnAction;
+  export let to: 'left' | 'right';
+
+  const handleRotateAsset = async () => {
+    let newOrientation = '1';
+    try {
+      switch (asset.exifInfo?.orientation) {
```
jrasm91, line 22: Can we use the enums here? And also can we just have a list and either find the next or previous element?

zdimension: We can. I was unsure whether using lists and stuff would be too "clever" and hinder readability so I wrote a dumb switch-case but I'll change it.

jrasm91: I'd rather see a map or a list than a long case statement

zdimension: is this readable enough?
```ts
// every rotation, in order
// interleaved normal / mirrored
const orientations = ['1', '2', '6', '7', '3', '4', '8', '5'];
let index = orientations.indexOf(asset.exifInfo?.orientation ?? '1');
if (index === -1) {
  index = 0;
}
if (to === 'right') {
  index = (index + 2) % orientations.length;
} else {
  index = (index - 2 + orientations.length) % orientations.length;
}
```

jrasm91: Seems fine to me. I probably would have preferred two lists and to use the enum instead of plain strings.

zdimension: I'm not sure how to use the enum, actually, since it's defined in the server package and this is the web app.

e2e/test-assets
michelheusschen, line 1: Please revert this change

web/src/lib/components/asset-viewer/actions/rotate-action.svelte
```diff
@@ -0,0 +1,59 @@
...
+        type: NotificationType.Info,
+        message: $t('edited_asset'),
+      });
+
+      onAction({ type: AssetAction.ROTATE, asset });
+      setTimeout(() => {
+        // force the image to refresh the thumbnail
+        const oldSrc = new URL($photoViewer!.src);
+        oldSrc.searchParams.set('t', Date.now().toString());
+        $photoViewer!.src = oldSrc.toString();
```
michelheusschen, line 47: We rely on the checksum field to detect changes in the thumbnail and bypass the cache. Refreshing based on the time might work in this component, but it will cause caching issues everywhere else.

web/src/lib/components/asset-viewer/actions/rotate-action.svelte
```diff
@@ -0,0 +1,59 @@
...
+    } catch (error) {
+      handleError(error, $t('errors.unable_to_edit_asset'));
+    }
+  };
+</script>
+
+<MenuOption
+  icon={to === 'left' ? mdiRotateLeft : mdiRotateRight}
+  onClick={handleRotateAsset}
+  text={$t(`rotate_${to}`)}
```
michelheusschen, line 58: Please avoid dynamic translation keys, as the tests can't detect missing translations in that case

```suggestion
  text={to === 'left' ? $t('rotate_left') : $t('rotate_right')}
```

web/src/lib/components/asset-viewer/photo-viewer.svelte
```diff
@@ -132,7 +132,11 @@
       <LoadingSpinner />
     </div>
   {:else if !imageError}
-    <div use:zoomImageAction class="h-full w-full" transition:fade={{ duration: haveFadeTransition ? 150 : 0 }}>
+    <div
+      use:zoomImageAction
+      class="h-full w-full duration-500"
```
michelheusschen, line 137: `duration-500` changes the duration of a transition, but there is no transition?

zdimension: Ah, missed that, sorry. It was part of something else I tried to change but reverted, and I forgot to revert that bit.

web/src/lib/components/asset-viewer/actions/rotate-action.svelte
zdimension: I agree that it's not pretty; but @jrasm91 told me on Discord the ideal way would through a web socket event but that since we don't currently have one for thumbnail generation, refreshing like that was a good enough hack until it's done. 

How would you do it?

e2e/test-assets
zdimension: it seems like on main this points to an outdated commit: e2e/test-assets points to https://github.com/immich-app/test-assets/tree/39f25a96f13f743c96cdb7c6d93b031fcb61b83c but the commit hash has changed when the PR was merged

web/src/lib/components/asset-viewer/photo-viewer.svelte
zdimension: fixed

web/src/lib/components/asset-viewer/actions/rotate-action.svelte
zdimension: done (but without the enum since it doesn't export the names)

danieldietzler (approved):
The code looks very good to me

web/src/lib/components/asset-viewer/actions/rotate-action.svelte
michelheusschen: I don't see how a websocket event will solve caching issues in other components/pages. We include the `checksum` for thumbnail requests, but the checksum doesn't change in this case so the cached thumbnail is still used.

jrasm91: Maybe we should use a counter/nonce instead of the checksum

danieldietzler: Oh sorry now I get your point. You're absolutely right, yes.

zdimension: so, uh, sorry, but what should I change here?

zackpollard:
#16089 superceeds this PR and will add this functionality

zdimension:
>  #16089 superceeds this PR and will add this functionality

Should I make another PR for the crop feature, since this PR was doing both but yours only handles rotation?

zackpollard:
> > #16089 superceeds this PR and will add this functionality
> 
> Should I make another PR for the crop feature, since this PR was doing both but yours only handles rotation?

Sorry, missed that. For crop, it's more complicated. The web already has some editor functionality built-in including crop I think, however it's currently disabled pending us figuring out how best to save these and apply the edits.

zdimension:
Yes, the current crop UI doesn't save the crop options, my PR was using the standard EXIF crop fields to store that. The current prototype pretty much works modulo some thumbnail-related stuff. I stopped working on my PR after people on Discord said changes would be made to the way Immich stores attached media so I was waiting on that to continue.

PR #11704, opened 2024-08-11
fix(web): use fallback image if shared asset isn't resized

michelheusschen:
share-cover-before
share-cover-after

### DIFF
--- a/web/src/test-data/index.ts
+++ b/web/src/test-data/index.ts
@@ -1,5 +1,0 @@
-export * from './factories/album-factory';
-export * from './factories/asset-factory';
-export * from './factories/person-factory';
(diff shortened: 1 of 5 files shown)


web/src/test-data/index.ts
```diff
@@ -1 +1,5 @@
 export * from './factories/album-factory';
```
jrasm91, line 1: We used to have these types of files in the server and it caused several problems. I'd rather avoid them.

jrasm91 (approved):

web/src/test-data/index.ts
michelheusschen: Alright I've removed this file

PR #11729, opened 2024-08-12
fix(web): single row of items

michelheusschen:
Single rows of items (such as on the explore page) have several issues:
- Requires manual size calculation
- Doesn't fill the entire row
- Shows too few or too many items when zooming

Fixed by using grid instead of flexbox and using an action that automatically calculates the amount of items that fit. The width of the first item will be used and it assumes all items have equal width.

explore-page-items


web/src/lib/actions/single-grid-row.ts
```diff
@@ -0,0 +1,40 @@
...
+  const style = getComputedStyle(element);
+
+  return {
+    columnGap: parsePixels(style.columnGap),
+  };
+}
+
+const parsePixels = (style: string) => Number.parseInt(style, 10) || 0;
+
+export function singleGridRow(container: HTMLElement, { onChange }: Options) {
```
jrasm91, line 15: Why is this better than just using a wrapper component instead? In general I don't really like the `use:action`. It seems harder to debug, troubleshoot, and understand, but that may just be my lack of understanding. I can easily understand `<SingleGridRow let:container>` though. Is it just a matter of data accessibility? Accessing and controlling the width/visibility of dynamic children?

michelheusschen: I generally don't mind actions, but after rewriting it as a component it seems the better option. You loose the flexibility of applying actions to arbitrary elements, but that's not really an issue in this case.

jrasm91 (approved):

PR #11780, opened 2024-08-14
refactor: server emit events

jrasm91:
Use decorator `@OnEmit({ event: 'onBootstrap' })` and utility type `(app: ArgOf<'onBootstrap'>)` for easier maintenance of emit events and their types.

### DIFF
--- a/server/src/services/library.service.ts
+++ b/server/src/services/library.service.ts
@@ -6,6 +6,7 @@ import path, { basename, parse } from 'node:path';
 import picomatch from 'picomatch';
 import { StorageCore } from 'src/cores/storage.core';
 import { SystemConfigCore } from 'src/cores/system-config.core';
+import { OnEmit } from 'src/decorators';
 import {
   CreateLibraryDto,
   LibraryResponseDto,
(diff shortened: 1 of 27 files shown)


server/src/services/library.service.ts
```diff
@@ -102,7 +104,7 @@ export class LibraryService implements OnEvents {
     });
   }
 
-  onConfigValidateEvent({ newConfig }: SystemConfigUpdateEvent) {
+  onConfigValidate({ newConfig }: ArgOf<'onConfigValidate'>) {
```
danieldietzler, line 107: I think you're missing the `@OnEmit({ event: 'onConfigValidate' })` here?

server/src/utils/events.ts
```diff
@@ -1,33 +1,57 @@
...
     }
   }
 
+  const handlers = _.orderBy(items, ['priority'], ['asc']);
+
   // register by priority
-  for (const { event, handler } of _.orderBy(handlers, ['priority'], ['asc'])) {
-    repository.on(event, handler);
+  for (const { event, handler } of handlers) {
+    repository.on(event as EmitEvent, handler);
```
danieldietzler, line 53:
```suggestion
    repository.on(event, handler);
```
Tested it and you don't need the cast here.

danieldietzler (approved):
Noice!

PR #11823, opened 2024-08-15
fix(docs): read-only affects XMP writing

C-Otto:
Immich cannot add XMP files to a read-only external library. This isn't obvious, which is why this mentions the issue in the documentation.

### DIFF
--- a/docs/docs/features/libraries.md
+++ b/docs/docs/features/libraries.md
@@ -112,7 +112,8 @@ The `immich-server` container will need access to the gallery. Modify your docke
 ```
 
 :::tip
-The `ro` flag at the end only gives read-only access to the volumes. This will disallow the images from being deleted in the web UI.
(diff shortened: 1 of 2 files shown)


stumpigit:
How funny: I was just about to open an issue about it. If you add a description, for example, it says that everything has been saved successfully, which it doesn't do.

C-Otto:
Yeah, there's no error handling, as far as I'm aware. It's "fun" to delete an image, remove it from trash, scan the external library - just to see it appear again!

stumpigit:
I think we should still open an issue about this, as I think it is definitely a bug. If nobody else cares, I'll be happy to have a look to see if I can fix it.

pyorot:
> I think we should still open an issue about this, as I think it is definitely a bug. If nobody else cares, I'll be happy to have a look to see if I can fix it.

See #10538

alextran1502 (approved):

PR #11824, opened 2024-08-15
feat(server): granular permissions for api keys

jrasm91:
- There is a new required field to make API keys (`permissions`)
- To retain previous behavior, api keys can should be created with `{ name: 'My api key', permissions: ['all'] }`
- Add initial support for granular permission for **api keys**. 
- Permission list is subject to change


zackpollard (approved):
Love it!

PR #11828, opened 2024-08-15
fix(cli): Concurrency not fully using queue potential

Tiefseetauchner:
Hey!

Tl;dr the cli was chunking the concurrency which lead to a lot of needless wait time. With this change, the cli will correctly hash `concurrency` number of files at the same time.

Long story: 

In asset.ts there was code handling the concurrency parameter but it was waiting for each file to complete hashing before tackling the next. The "solution" seems to have been to chunk files and process those concurrently in the queue. Which ever only has a single entry. So it's... Not reaaaaally a queue.

The solution is to throw all the items in the queue, wait for all of them to complete before moving on. Another improvement I made was to collect all items that were checked and only after they've been hashed send them to the server in one large mega request. Reducing chatter is my specialty.

I did the same thing to the upload queue.

Hope this is fine to just make a PR for ya'll without a prior issue!

Best, Lena <3


Tiefseetauchner:
Side effect I just noticed with a larger test dataset: the ETA is now completely useless, as opposed to nearly always incorrect

alextran1502:
Hello, thank you for the PR. Do you have any metrics to show before and after for this improvement?

Tiefseetauchner:
> Hello, thank you for the PR. Do you have any metrics to show before and after for this improvement?

I can make one, I'll fix the tests first though

cli/src/commands/asset.ts
```diff
@@ -97,36 +97,41 @@ export const checkForDuplicates = async (files: string[], { concurrency, skipHas
...
+      await queue.push(item);
+    }),
+  );
 
   await queue.drained();
 
+  const newFiles: string[] = [];
+  const duplicates: Asset[] = [];
+
+  const response = await checkBulkUpload({ assetBulkUploadCheckDto: { assets: results } });
```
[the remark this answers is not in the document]
Tiefseetauchner: I'll generate some images for the metrics anyways, I'll make stats on something like 2000000 images then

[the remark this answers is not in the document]
Tiefseetauchner: Yeah it throws a 413
I'll chunk it

Tiefseetauchner:
> Hello, thank you for the PR. Do you have any metrics to show before and after for this improvement?

  | Old Time | New Time |   |   | CMD | time ./dist/index.js upload -nrc 64 ../testdata_…
-- | -- | -- | -- | -- | -- | --
Real Data | 03:36.28 | 01:50.80 |   |   | Real Data | The real data test includes 2330 DNG, 353 JPG files 9 MP4 and 13 MOV files. The largest files are MOV with ca. 30GB in size. The dataset totals to 124GB
PNG Data | 00:41.39 | 00:18.42 |   |   | PNG Data | The PNG data test includes 31,000 PNG files totaling 12.6GB of image Data

cli/src/commands/asset.ts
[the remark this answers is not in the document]
jrasm91: Regardless, some machines probably can't handle that many assets at once. I think it makes sense to loop through the results by chunks to keep memory usage down in both the cli and the server.

cli/src/commands/asset.ts
```diff
@@ -97,36 +97,60 @@ export const checkForDuplicates = async (files: string[], { concurrency, skipHas
...
+        }
+      }
+
+      break;
+    } catch (error: any) {
+      retries++;
+      if (retries >= maxRetries) {
+        throw new Error(`An error occurred while checking for duplicates: ${error.message}`);
+      }
+    }
```
jrasm91, line 151: this should probably be after the while loop, since you want the while loop condition to cause it to abort imo

cli/src/commands/asset.ts
```diff
@@ -97,36 +97,60 @@ export const checkForDuplicates = async (files: string[], { concurrency, skipHas
...
+  const duplicates: Asset[] = [];
+
+  let retries = 0;
+  const maxRetries = 3;
+
+  while (retries < maxRetries) {
+    try {
+      const chunks = chunk(results, 1000);
+
+      for (const chunk of chunks) {
```
mertalev, line 132: It'd be cool to process these chunks while the queue is active so it can do the hashing and requests in parallel. After `results.push(dto);`, you could check if the list is 1000 elements and either do the `checkBulkUpload` in the same job, or add it to a separate queue. In the former, you'd have to be a little more careful about the retries since you don't want an error in `checkBulkUpload` to cause it to add the same asset to `results` twice.

cli/src/commands/asset.ts
Tiefseetauchner: NOTE: Not applicable anymore

Tiefseetauchner:
I'm gonna test a little and fix the Test Builds, and I'll do another performance test to check that there wasn't a regression in the latest changes, and then I think this is good to go from my part

Tiefseetauchner:
New data is in! The most surprising result to me was the 3rd one, as I didn't expect it to make as much a difference with files of same size

  | Old Time | New Time |   |   | CMD | time ./dist/index.js upload -nrc 64 ../testdata_…
-- | -- | -- | -- | -- | -- | --
Real Data | 03:09.35 | 01:43.37 |   |   | Real Data | The real data test includes 2330 DNG, 353 JPG files 9 MP4 and 13 MOV files. The largest files are MOV with ca. 30GB in size. The dataset totals to 124GB
PNG Data | 00:26.92 | 00:12.55 |   |   | PNG Data | The PNG data test includes 31,000 PNG files totaling 12.6GB of image Data
Many Data | 15:44.53 | 08:24.93 |   |   | Many Data | The Many Data test includes 2,000,000 files with random content at 1kb each

cli/src/commands/asset.ts
```diff
@@ -115,19 +115,43 @@ export const checkForDuplicates = async (files: string[], { concurrency, skipHas
...
-    await queue.push(items);
-  }
+  files.map((item) => {
+    void queue.push(item);
+  });
 
   await queue.drained();
 
-  progressBar.stop();
+  await checkBulkUploadQueue.push(checkBulkUploadRequests);
```
mertalev, line 150: It should check if the list is empty before queueing here.

mertalev (commented):
Thanks for sharing benchmark results! A few minor comments, but looks good overall.

cli/src/commands/asset.ts
```diff
@@ -201,9 +227,9 @@ export const uploadFiles = async (files: string[], { dryRun, concurrency }: Uplo
     { concurrency, retry: 3 },
   );
 
-  for (const filepath of files) {
-    await queue.push(filepath);
-  }
+  files.map((item) => {
+    void queue.push(item);
+  });
```
jrasm91, line 232: Can you just use a for loop here please?

jrasm91 (approved):
LGTM

cli/src/commands/asset.ts
Tiefseetauchner: Oops, remainder of a previous piece of code, sry

alextran1502 (approved):
Thank you! Let's fix the test and have this merged. It will be a great improvement

Tiefseetauchner:
I am struggling to figure out why the e2e tests are doing what they're doing, could someone more knowledgeable have a peek? Thanks :3

alextran1502:
@Tiefseetauchner From my understanding of the logs from the e2e test, maybe it was expecting the queue to be filled before executing. And now we have the process executed as soon as an item is in the queue?

Tiefseetauchner:
Hey again! Only took me like

Three months....

Finally thought of a not completely unacceptable way to do the stderr test, sorry it took so long. Please please please someone check that last commit, I don't know whether you like what I did there

cli/src/commands/asset.ts
```diff
@@ -115,19 +116,45 @@ export const checkForDuplicates = async (files: string[], { concurrency, skipHas
...
+
+  const queue = new Queue<string, AssetBulkUploadCheckItem[]>(
+    async (filepath: string): Promise<AssetBulkUploadCheckItem[]> => {
+      const dto = { id: filepath, checksum: await sha1(filepath) };
+
+      results.push(dto);
+      checkBulkUploadRequests.push(dto);
+      if (checkBulkUploadRequests.length > 5000) {
+        void checkBulkUploadQueue.push([...checkBulkUploadRequests]);
+        checkBulkUploadRequests = [];
```
mertalev, line 136:
```suggestion
        const batch = checkBulkUploadRequests;
        checkBulkUploadRequests = [];
        void checkBulkUploadQueue.push(batch);
```
I haven't tested this, but it avoids the need to copy the list. Incidentally, it would also fix the batching race condition.

cli/src/commands/asset.ts
```diff
@@ -115,19 +116,45 @@ export const checkForDuplicates = async (files: string[], { concurrency, skipHas
...
-    await queue.push(items);
+  for (const item of files) {
+    void queue.push(item);
   }
 
   await queue.drained();
 
-  progressBar.stop();
+  if (checkBulkUploadRequests.length > 0) {
+    void checkBulkUploadQueue.push([...checkBulkUploadRequests]);
```
mertalev, line 152:
```suggestion
    const batch = checkBulkUploadRequests;
    checkBulkUploadRequests = [];
    void checkBulkUploadQueue.push(batch);
```
Same here.

Tiefseetauchner: Setting checkBulkUploadRequests here isn't needed

cli/src/commands/asset.ts
Tiefseetauchner: Seems to work (tested with 10422 assets)

mertalev (approved):

Tiefseetauchner:
Thank ya'll for your patience, this contributor badge will be a nice addition to my collection

Sorry this took so long to complete, got busy

PR #11861, opened 2024-08-17
refactor(server): move files to separate table

jrasm91:
Move thumbnail and preview paths from the `assets` table to a new table, `asset_files`.


server/src/utils/asset.util.ts
```diff
@@ -11,6 +12,15 @@ export interface IBulkAsset {
   removeAssetIds: (id: string, assetIds: string[]) => Promise<void>;
 }
 
+const getFileByType = (files: AssetFileEntity[] | undefined, type: AssetFileType) => {
+  return (files || []).find((file) => file.type === type);
```
danieldietzler, line 16: I'd prefer changing the order of the arguments here and then making files optional with `files: AssetFileEntity[] = []`. Very minor though

danieldietzler (approved):
Looks great to me!

server/src/utils/asset.util.ts
jrasm91: Changing the order would allow the code to be slightly simpler, but I think it makes more sense for the arguments to be in this order.

danieldietzler: Hm yeah. Alternatively your favorite way of passing arguments, by using an object, is always an option :P

PR #11880, opened 2024-08-18
feat: folder view

davidakerr:
A quick exploration into the possibilities of folder view, full thread here https://github.com/immich-app/immich/discussions/5409#discussioncomment-10365869

### Todo's
- Design
  - [x] Designs mockups
  - [x] Designs Implemented in tailwind/svelte
  
- Core Functionality
  - [x] Inline Folder Thumbnail Icons (and back to previous folder icon)
  - [x] Reuse Thumbnail Display Components
  - [x] On-Click Image Display (like `photos` and `photos/:id`)
  - [ ] Image/Folder Multi Select Actions (Create Album, Delete, etc)
  - [x] `getAssetsByOriginalPath` function should return only images directly in that path, not all images from that path down
  - [ ] Gather External Library root folders, remove unnecessary absolute pathing from begining [[1]](https://github.com/immich-app/immich/blob/main/server/src/entities/library.entity.ts#L34) [[2]](https://github.com/immich-app/immich/blob/e093218f5010a73cae10f2f1f549514e05052793/server/src/repositories/asset.repository.ts#L826)
  - [ ] Move server logic to `folder.*.ts` rather than `asset.*.ts` ?
  - Redis caching **_?_** _(currently mocked out as svelte store with caching)_
    - [ ] Folder Structure
    - [ ] Path Asset Searches
    - [ ] Cache invalidation on relevant library modifications

_Any more todo items I can add them in here, if you add them to the comments._

### Screenshots

| Light mode | Dark mode |
| - | - |
| light | dark |


alextran1502:
FYI, we use tailwindcss, so you don't have to create custom CSS classes

web/src/lib/utils/folder-utils.ts
```diff
@@ -0,0 +1,14 @@
+export function buildFolderTree(paths: string[]): any {
```
benmccann, line 1: could we avoid the `any` return type here? maybe we need to create a `Folder` type to be used in place of the `any` or something like that?

web/src/lib/components/shared-components/side-bar/folder-view-sidebar.svelte
```diff
@@ -0,0 +1,23 @@
+<script lang="ts">
+  import { onMount } from 'svelte';
+  import Folder from '$lib/components/folder-browser/Folder.svelte';
+  import { getUniqueOriginalPaths } from '@immich/sdk';
+  import { buildFolderTree } from '$lib/utils/folder-utils';
+
+
+  let folderTree: { [key: string]: any } = {};
```
benmccann, line 8: I think this could be updated to use the new type rather than `any`

xttlegendapi:
It looks very promising! 
External libraries where not showing up at all first. but now after our latest commits external libraries are showing in the sidebar but not in the main view
screenshot

davidakerr:
> It looks very promising! External libraries where not showing up at all first. but now after our latest commits external libraries are showing in the sidebar but not in the main view...

Good catch, I should have tested this. Try it now. Thanks @xttlegendapi  ❤️

etnoy:
Haven't had the opportunity to review this very much but YES I want this feature

benmccann:
In terms of UI, I think one thing that could make it a bit cleaner is not having so many folder icons. Maybe we could look at Synology Photos for inspiration which uses arrow characters - those also helps differentiate which folders are expanded vs not. They also use a non-bold font for the folder names, which I think would be worth testing as well

https://www.synology.com/img/beta/dsm70/photos/all_in_one_1@2x.png

alextran1502:
@davidakerr It's alive!!

SCR-20240819-szgs

server/src/controllers/asset.controller.ts
```diff
@@ -22,6 +23,17 @@ import { UUIDParamDto } from 'src/validation';
...
+  @Authenticated()
+  getUniqueOriginalPaths(@Auth() auth: AuthDto): Promise<string[]> {
+    return this.service.getUniqueOriginalPaths(auth);
+  }
+
+  @Get('folder')
+  @Authenticated()
+  getAssetsByOriginalPath(@Auth() auth: AuthDto, @Query('path') path: string): Promise<AssetResponseDto[]> {
+    return this.service.getAssetsByOriginalPath(auth, path);
+  }
```
jrasm91, line 36: These endpoints should either be in their own controller, or in the search controller. If we are ever going to allow file/folder level actions, it would make sense to have a dedicated controller IMO. Something like `FolderViewController` at `/views/folder/...`

server/src/services/asset.service.ts
```diff
@@ -328,4 +328,15 @@ export class AssetService {
...
   }
+
+  async getUniqueOriginalPaths(auth: AuthDto): Promise<string[]> {
+    return this.assetRepository.getUniqueOriginalPaths(auth.user.id);
+  }
+
+  async getAssetsByOriginalPath(auth: AuthDto, path: string): Promise<AssetResponseDto[]> {
+    const assets = await this.assetRepository.getAssetsByOriginalPath(auth.user.id, path);
+    const ids = assets.map((a) => a.id);
+    await this.access.requirePermission(auth, Permission.ASSET_READ, ids);
```
jrasm91, line 339: If the query itself is filtered by the user id, I don't see any reason to do an access check.

web/src/lib/components/folder-tree/folder-tree.svelte
```diff
@@ -0,0 +1,65 @@
...
+  </a>
+  <button on:click={toggleExpand} type="button">
+    <p class="text-nowrap overflow-clip">{folderName}</p>
+  </button>
+</button>
+
+{#if isExpanded}
+  <ul class="list-none ml-2">
+    {#each Object.entries(content) as [subFolderName, subContent]}
+      <li class="my-1">
```
danieldietzler, line 60: The list items should probably be keyed I think

web/src/lib/utils/folder-utils.ts
```diff
@@ -0,0 +1,18 @@
...
+  [key: string]: RecursiveObject;
+}
+
+export function buildFolderTree(paths: string[]) {
+  const root: RecursiveObject = {};
+  for (const path of paths) {
+    const parts = path.split('/');
+    let current = root;
+    for (const part of parts) {
+      if (!current[part]) {
```
danieldietzler, line 11: `!{}` is also true, which makes this a bit weird imo. I think we should be more explicit here by writing `current[part] === undefined`

server/src/controllers/view.controller.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Auth, Authenticated } from 'src/middleware/auth.guard';
+import { ViewService } from 'src/services/view.service';
+
+@ApiTags('View')
+@Controller('view')
+export class ViewController {
+  constructor(private service: ViewService) {}
+
+  @Get('folder/unique-paths')
+  @Authenticated()
```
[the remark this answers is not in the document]
jrasm91: I think `asset.read` is fine. The data is just `asset.originalPath` which is covered by `asset.read`

server/src/repositories/asset.repository.ts
```diff
@@ -820,6 +820,50 @@ export class AssetRepository implements IAssetRepository {
...
+    const builder = this.getBuilder({
+      userIds: [userId],
+      exifInfo: false,
+      withStacked: false,
+      isArchived: false,
+      isTrashed: false,
+    });
+
+    const results = await builder
+      .select("DISTINCT substring(asset.originalPath FROM '^(.*/)[^/]*$')", 'directoryPath')
```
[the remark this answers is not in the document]
davidakerr: I can't speak to the performance. But you're right, it matches assets within a specific directory and not those in subdirectories below it.  It's not prohibitively slow in my tests, potentially the result can be cached in redis against the file path and invalidate it upon asset updates.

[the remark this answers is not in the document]
davidakerr: I presumed it would be quicker to do it in the DB query rather than fetching all assets and producing a Set with JS

web/src/lib/utils/folder-utils.ts
davidakerr:
```
> !{}
false
```

server/src/repositories/asset.repository.ts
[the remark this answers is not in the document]
jrasm91: I think we just need to make sure this is using an index and not doing a full table scan.

docker/docker-compose.dev.yml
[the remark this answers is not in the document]
alextran1502: I purposely do this so we can mount external for dev environment easily without the need of tracking the docker-compose.dev.yml changes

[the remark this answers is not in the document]
danieldietzler: Hm ok, I wouldn't mount them at `/home/user` then though, that feels weird to me

web/src/lib/utils/folder-utils.ts
danieldietzler: Oh huh, that's weird.

server/src/services/view.service.ts
```diff
@@ -0,0 +1,18 @@
+import { Inject } from '@nestjs/common';
+import { AssetResponseDto, mapAsset } from 'src/dtos/asset-response.dto';
+import { AuthDto } from 'src/dtos/auth.dto';
+import { IAssetRepository } from 'src/interfaces/asset.interface';
+
+export class ViewService {
+  constructor(@Inject(IAssetRepository) private assetRepository: IAssetRepository) {}
+
+  async getUniqueOriginalPaths(auth: AuthDto): Promise<string[]> {
```
jrasm91, line 9:
```suggestion
  getUniqueOriginalPaths(auth: AuthDto): Promise<string[]> {
```

server/src/services/view.service.ts
```diff
@@ -0,0 +1,18 @@
...
+    return this.assetRepository.getUniqueOriginalPaths(auth.user.id);
+  }
+
+  async getAssetsByOriginalPath(auth: AuthDto, path: string): Promise<AssetResponseDto[]> {
+    const assets = await this.assetRepository.getAssetsByOriginalPath(auth.user.id, path);
+
+    return assets.map((a) => mapAsset(a, { auth }));
```
jrasm91, line 16:
```suggestion
    return assets.map((asset) => mapAsset(asset, { auth }));
```

jrasm91 (approved):

cole21771:
This is such a great feature and helps so much with being able to view photos in external libraries! Is there any plans to add this to the mobile version?

bo0tzz:
Yes, mobile UI for this and the tags view is being worked on this week.

el-tiuri:
Has this been implemented in the mobile app? I checked the app itself as well as the change log for the past few months, but I couldn’t find it in either.

etnoy:
> Has this been implemented in the mobile app? I checked the app itself as well as the change log for the past few months, but I couldn’t find it in either.

No, this isn't done yet

bfentor:
This feature would be perfect for me if I was able to share it to my users, which I am unable to do at the moment. I have a vast library of imported photos through the "import external library" feature, but I have not found a way to allow anyone other than the admin user to access folder view for the external library. Am I missing something or is this not implemented yet?
And yes, the external library is shared with my users. They are able to see the imported photos on their timeline but not in folder view. 
Apologies if I'm posting this in the wrong place.

etnoy:
> This feature would be perfect for me if I was able to share it to my users, which I am unable to do at the moment. I have a vast library of imported photos through the "import external library" feature, but I have not found a way to allow anyone other than the admin user to access folder view for the external library. Am I missing something or is this not implemented yet? And yes, the external library is shared with my users. They are able to see the imported photos on their timeline but not in folder view. Apologies if I'm posting this in the wrong place.

It must be enabled by each user

image

bfentor:
Thank you for your reply. What I meant was that the folders themselves don't show up in the folder view
Screenshot From 2025-01-10 00-22-23
This is on one of the other users. On the admin user this view is populated with the folder structure I expect.

PR #11890, opened 2024-08-18
Add funding button on GitHub

lpil:
Hello! Thank you for this software.

I thought it would be nice if folks could more easily find your GitHub sponsors page. I had to go looking for it and wasn't sure you had one at all from the project.

Thanks,
Louis

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+github: alextran1502


bo0tzz:
That is Alex's personal sponsor page. We don't take donations as a project anymore, but you can support us by buying a key on https://buy.immich.app/.

PR #11891, opened 2024-08-18
fix(docs): adds "immich-machine-learning:/cache" to non-root FAQ

YusefOuda:
Right now, the FAQ for non-root docker only specifies you need bind mounts for:

- immich-machine-learning:/.config
- immich-machine-learning:/.cache
- redis:/data

This PR adds `immich-machine-learning:/cache` since that seems to be required too


mmomjian:
`cache` is contained in the default docker-compose :) So it may make sense to add a *note* that /cache must be retained from the default, but it's confusing to add this since we don't include DB_DATA_LOCATION or UPLOAD_LOCATION either.

```
    volumes:
      - model-cache:/cache
```

PR #11923, opened 2024-08-20
refactor(server): controller cleanup

jrasm91:
- Ship new `/server/*` endpoint paths in the SDKs (previously used `/server-info/*`
 - Add a few more permissions
 - Clean up decorator order
 - Merge "restore" permission into "delete"


danieldietzler (approved):
Good stuff

PR #11932, opened 2024-08-20
fix: random e2e test

jrasm91:


danieldietzler (approved):
That was a tricky race condition, good catch!

PR #11951, opened 2024-08-20
fix: align camera model drop down behavior with other drop downs on web and mobile

x24git:
Current search filterbox behavior for the camera make and model is not consistent with other linked search criteria (like location). 
For the web, when selecting a make and model, trying to re-select a make only shows the make linked to the currently selected model. 
image
_^In the above image, the library has many different makes, but this is not shown since a model is currently selected._

This PR changes the behavior (to align with locations) to show all makes no matter which model is selected, and when a make is selected, the model is cleared (if the make selection actually changes).

A similar behavior is observed in the mobile, where when changing a parented linked selection (make->model and country->state->city) the child is not cleared, allowing for a mismatch of parent and child. This PR changes that behavior (resolving a regression in #10385 that did not set the dropdown values correctly [initial, or when changing linked pairs]) to always clear the child when a parent linked item is changed. It also sets the dropdown's initial values to match an existing value.

_^In the above image, the location drop downs are not set despite having an existing filter set. ._


alextran1502 (approved):
Thank you

PR #11980, opened 2024-08-22
feat: tags

jrasm91:
> [!WARNING]
> **NOTE: deleting an entire tag will not sync tags change to impacted asset sidecars at the moment. Re-running metadata extract will most likely recreate the tags again. Untagging the assets and then deleting that tag should be fine though.**

### Add support for tags
- Read `TagsList` and `Keywords` from exif
- Any `/` character in a tag in interpreted to mean nested tags
- Any parents that don't exist will be created
- Uniqueness is determined by absolute path (`Parent1/Child`, `Parent2/Child`, and `Child` are all unique tags)
- When assets are tagged, the tags and written to `TagList` in the associated XMP sidecar file
- When assets are untragged the updated tag list is written as per the above statement (with the exception of deleting an entire tag).
- You can give a tag a color using hex code `#123456`, although this only is used in the explorer currently. Future work can expand this to the asset detail panel.

<details>
<summary>Screenshots</summary>

| Description | Image |
| - | - |
| Sidebar | image |
| Explorer | image |
| Tag assets action | image |
| Create tag | image |

</details>


bo0tzz:
Closes #2168, #3518, #3950.

briandipalma:
Are tag edits applied to the picture file? It's nicer if the file itself is modified instead of putting metadata in a sidecar file (which you then have to delete once you apply the metadata to the file with Digikam).

C-Otto:
The tags are stored in the database, not sidecars, nor in the assets themselves.

mtojay:
> The tags are stored in the database, not sidecars, nor in the assets themselves.

oh really? so we cant edit them outside of immich at all with digikam for example? what is with xmp tags created with other software before import? will it import those to immich?

bo0tzz:
Tags that exist in the file's metadata are imported. This is just a first pass at the feature (and it's still in draft!), writing tags can be added later.

> It's nicer if the file itself is modified instead of putting metadata in a sidecar file 

We fundamentally never make changes to the original file.

jrasm91:
Please don't bring over to this PR the debate of sidecar vs originals, which has already been answered several times elsewhere. This discussion should be centered around the technical implementation of the feature. The major design decision have already been made:
- Read tags from xmp
- Write tags to xmp
- Dedicated tag table
- Many to many between assets and tags
- Tag explorer
- Viewing parent tag implies seeing assets linked to all children
- View tags in asset detail panel
- Add new tags
- Delete existing tags
- Add/remove tags in bulk via mult-select option
- Eventually add multiple tags at once to a set of assets

alextran1502 (approved):

web/src/lib/components/forms/tag-asset-form.svelte
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  const handleSelect = (option?: ComboBoxOption) => {
+    if (!option) {
+      selectedTags = new Set();
+      return;
+    }
+
+    selectedTags.add(option.value);
+    selectedTags = new Set(selectedTags);
```
danieldietzler, line 33: Why do we create a copy here? To trigger reactivity we can also write `selectedTags = selectedTags`, right?

web/src/lib/components/photos-page/actions/tag-action.svelte
```diff
@@ -0,0 +1,47 @@
...
+    }
+    loading = false;
+  };
+</script>
+
+{#if menuItem}
+  <MenuOption {text} {icon} onClick={handleOpen} />
+{/if}
+
+{#if !menuItem}
```
danieldietzler, line 37: This should just be an `else`

web/src/lib/components/shared-components/settings/setting-input-field.svelte
```diff
@@ -69,22 +82,46 @@
...
+    <div class="flex place-items-center place-content-center">
+      {#if inputType === SettingInputFieldType.COLOR}
+        <input
+          bind:this={input}
+          class="immich-form-input w-full pb-2 rounded-none mr-1"
+          aria-describedby={desc ? `${label}-desc` : undefined}
+          aria-labelledby="{label}-label"
+          id={label}
+          name={label}
+          type="text"
```
danieldietzler, line 94: Do you plan on changing this to `type="color"`?

web/src/lib/utils/asset-utils.ts
```diff
@@ -61,6 +63,46 @@ export const addAssetsToAlbum = async (albumId: string, assetIds: string[], show
...
+  assetIds: string[];
+  tagIds: string[];
+  showNotification?: boolean;
+}) => {
+  for (const tagId of tagIds) {
+    await tagAllAssets({ id: tagId, bulkIdsDto: { ids: assetIds } });
+  }
+
+  if (showNotification) {
+    notificationController.show({ message: `Tagged ${assetIds.length} assets`, type: NotificationType.Info });
```
danieldietzler, line 80: This message should be an i18n key.

web/src/lib/utils/asset-utils.ts
```diff
@@ -61,6 +63,46 @@ export const addAssetsToAlbum = async (albumId: string, assetIds: string[], show
...
+  assetIds: string[];
+  tagIds: string[];
+  showNotification?: boolean;
+}) => {
+  for (const tagId of tagIds) {
+    await untagAssets({ id: tagId, bulkIdsDto: { ids: assetIds } });
+  }
+
+  if (showNotification) {
+    notificationController.show({ message: `Remove tag for ${assetIds.length} assets`, type: NotificationType.Info });
```
danieldietzler, line 100: Same here

web/src/routes/(user)/tags/[[photos=photos]]/[[assetId=id]]/+page.svelte
```diff
@@ -0,0 +1,252 @@
...
+  };
+
+  const handleCancel = () => {
+    isNewOpen = false;
+    isEditOpen = false;
+  };
+
+  const handleSubmit = async () => {
+    if (tag && isEditOpen && newTagColor) {
+      console.log('update tag', tag.id, newTagColor);
```
Comment (danieldietzler) [inline web/src/routes/(user)/tags/[[photos=photos]]/[[assetId=id]]/+page.svelte:85]: Just flagging this to be removed before merging

web/src/routes/(user)/tags/[[photos=photos]]/[[assetId=id]]/+page.svelte
```diff
@@ -0,0 +1,252 @@
...
+          <p class="hidden md:block">{$t('delete_tag')}</p>
+        </div>
+      </LinkButton>
+    {/if}
+  </section>
+
+  <section
+    class="flex place-items-center gap-2 mt-2 text-immich-primary dark:text-immich-dark-primary rounded-2xl bg-gray-50 dark:bg-immich-dark-gray/50 w-full py-2 px-4 border border-gray-100 dark:border-gray-900"
+  >
+    <a href={`${AppRoute.TAGS}`} title="To root">
```
Comment (danieldietzler) [inline web/src/routes/(user)/tags/[[photos=photos]]/[[assetId=id]]/+page.svelte:173]: This link title should also be internationalized

web/src/routes/(user)/tags/[[photos=photos]]/[[assetId=id]]/+page.ts
```diff
@@ -0,0 +1,38 @@
...
+export const load = (async ({ params, url }) => {
+  await authenticate();
+  const asset = await getAssetInfoFromParam(params);
+  const $t = await getFormatter();
+
+  let pathAssets: AssetResponseDto[] = [];
+  const path = url.searchParams.get(QueryParameter.PATH);
+  if (path) {
+    pathAssets = [];
+  }
```
Comment (danieldietzler) [inline web/src/routes/(user)/tags/[[photos=photos]]/[[assetId=id]]/+page.ts:18]: What is happening here? `pathAssets` is already an empty array at this point

Comment (jrasm91) [reply web/src/routes/(user)/tags/[[photos=photos]]/[[assetId=id]]/+page.ts:18]: Probably not needed anymore.

server/src/migrations/1724790460210-NestedTagTable.ts
```diff
@@ -0,0 +1,56 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class NestedTagTable1724790460210 implements MigrationInterface {
+    name = 'NestedTagTable1724790460210'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "tags" DROP CONSTRAINT "FK_92e67dc508c705dd66c94615576"`);
```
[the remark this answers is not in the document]
jrasm91: I'm using the generated migration which made it easier to roll back. I can add a line to truncate the table though.

web/src/lib/components/shared-components/settings/setting-input-field.svelte
jrasm91: Not at the moment, no.

web/src/lib/components/photos-page/actions/tag-action.svelte
jrasm91: It condenses it into an if else loading which makes it harder to read imo.

waclaw66:
Would it be possible to add option item for Tags and Folders to hide those from Library section same as for People and Sharing? They take up space at the expense of the storage and version info below on small screens.

alextran1502:
@waclaw66 yep, see #12099

danieldietzler (commented):
Great job!

ilyaChuk:
Please pay attention #12165

Knuppel1983:
Does this feature write the tags to the original files? If not can this be added as an option? Or an option to export the tags with the original files?

I want my photo library to be vendor-independant, so i can import / export without having to tag everything all over again...

danieldietzler:
> Does this feature write the tags to the original files? If not can this be added as an option? Or an option to export the tags with the original files?
> 
> I want my photo library to be vendor-independant, so i can import / export without having to tag everything all over again...

It writes the tags to an XMP sidecar file, just like any other metadata you change within immich

Knuppel1983:
> > Does this feature write the tags to the original files? If not can this be added as an option? Or an option to export the tags with the original files?
> > I want my photo library to be vendor-independant, so i can import / export without having to tag everything all over again...
> 
> It writes the tags to an XMP sidecar file, just like any other metadata you change within immich

Ok that will do. Please also add an option to disable smart search, and search by tag only. Some of us don't need machine learning, and it puts a lot of strain on low powered devices.

jrasm91:
@Knuppel1983 you can already do this from the admin settings.

image

PR #11988, opened 2024-08-23
fix(web): show a clearer confirmation message when deleting an unnamed album

Snowknight26:


web/src/lib/utils/album-utils.ts
```diff
@@ -199,3 +200,16 @@ export const collapseAllAlbumGroups = (groupIds: string[]) => {
 export const expandAllAlbumGroups = () => {
   collapseAllAlbumGroups([]);
 };
+
+export const getDeleteAlbumConfirmationDialog = async (album: AlbumResponseDto) => {
```
jrasm91, line 204: Let's rename this to `confirmAlbumDelete()`

jrasm91 (approved):
Noice

pikapower9080:
Should it be clarified that assets in the album won't be deleted or is that obvious?

jrasm91:
Can you fix the formatting?

Snowknight26:
Doh, thanks for the reminder. Feels like this could be a great opportunity to have a commit hook that auto-formats.

alextran1502 (approved):

PR #12054, opened 2024-08-26
fix: do not code format repro steps in issue template

C-Otto:
When describing the reproduction steps, the template currently uses bash highlighting/colors. This is confusing and unnecessary.


bo0tzz (approved):

C-Otto:
Thanks, I'm happy now :)
https://github.com/immich-app/immich/issues/12056

PR #12086, opened 2024-08-27
docs: sql query for duplicate files

mmomjian:


mertalev:
On a large instance, this worked but was relatively slow. This ends up being quite a bit faster:
```sql
SELECT T1.checksum, array_agg(T2.id) ids
FROM
  assets T1
    INNER JOIN assets T2
      ON T1.checksum = T2.checksum AND T1.id != T2.id AND T2."deletedAt" IS NULL
WHERE T1."deletedAt" IS NULL
GROUP BY T1.checksum;
```

jrasm91 (approved):

mmomjian:
> On a large instance, this worked but was relatively slow. This ends up being quite a bit faster:
> 
> ```sql
> SELECT T1.checksum, array_agg(T2.id) ids
> FROM
>   assets T1
>     INNER JOIN assets T2
>       ON T1.checksum = T2.checksum AND T1.id != T2.id AND T2."deletedAt" IS NULL
> WHERE T1."deletedAt" IS NULL
> GROUP BY T1.checksum;
> ```

Thanks! updated

mertalev (approved):

aviv926:
Out of curiosity, will this only work for external libraries? Because if I understand correctly immich by default rejects files with the same hash value (duplicate files).

mmomjian:
> Out of curiosity, will this only work for external libraries? Because if I understand correctly immich by default rejects files with the same hash value (duplicate files).

This will only work for *non*-external libraries. External libraries checksum based on the file path. Duplicate files are not rejected across different users for the internal library.

aviv926:
> > Out of curiosity, will this only work for external libraries? Because if I understand correctly immich by default rejects files with the same hash value (duplicate files).
> 
> This will only work for *non*-external libraries. External libraries checksum based on the file path. Duplicate files are not rejected across different users for the internal library.

Ah ok thanks for the clarification 💯

PR #12107, opened 2024-08-29
chore(server): log path when generating external thumbnail

etnoy:
This improves logging clarity when generating thumbs for external assets.

Question: We could do this for non-external assets as well, I think it might be even clearer than having cryptic asset ids. Agree/disagree?

### DIFF
--- a/server/src/services/media.service.ts
+++ b/server/src/services/media.service.ts
@@ -255,9 +255,13 @@ export class MediaService {
         throw new UnsupportedMediaTypeException(`Unsupported asset type for thumbnail generation: ${asset.type}`);
       }
     }
+
+    const assetInfo = asset.isExternal ? asset.originalPath : asset.id;
+
     this.logger.log(
-      `Successfully generated ${format.toUpperCase()} ${asset.type.toLowerCase()} ${type} for asset ${asset.id}`,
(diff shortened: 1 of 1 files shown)


bo0tzz:
IMO logging the asset ID is better - it may be more cryptic on first read, but it's a lot easier to go from the asset ID to viewing the full asset details (including path) since you can just put it into a URL, which you can't do with the path.

server/src/services/media.service.ts
```diff
@@ -255,9 +255,13 @@ export class MediaService {
         throw new UnsupportedMediaTypeException(`Unsupported asset type for thumbnail generation: ${asset.type}`);
       }
     }
+
+    const assetInfo = asset.isExternal ? asset.originalPath : asset.id;
+
     this.logger.log(
-      `Successfully generated ${format.toUpperCase()} ${asset.type.toLowerCase()} ${type} for asset ${asset.id}`,
+      `Successfully generated ${format.toUpperCase()} ${asset.type.toLowerCase()} ${type} in $for asset ${assetInfo}`,
```
danieldietzler, line 262: There's a `$` too much

danieldietzler (approved):
I agree with bo0tzz.

jrasm91 (approved):

PR #12111, opened 2024-08-29
fix(server): skip smtp validation if unchanged

michelheusschen:
Skip smtp validation if there are no changes to the smtp config. `isEqual(oldConfig, newConfig)` fails because it compares a plain object with a DTO instance and will always return `false`


danieldietzler (approved):

jrasm91 (commented):
Maybe we should convert it to a plain object when we emit it instead. This would prevent similar errors in other services in the future

michelheusschen:
> Maybe we should convert it to a plain object when we emit it instead. This would prevent similar errors in other services in the future

I’ve implemented your suggestion. To avoid reintroducing the issue, I’ve also kept an updated version of the equality check

jrasm91 (approved):

PR #12112, opened 2024-08-29
chore(e2e): use testcontainers

etnoy:
Use testcontainers instead of the homegrown docker compose startup.

This has many advantages, but especially when running local development and port 2283 is already taken

TODO:
* Fix cli tests
* playwright tests are not converted yet


jrasm91 (commented):
I don't think we want this. For one, it is possible to run a lot of e2e tests against the dev setup which I do regularly.

PR #12162, opened 2024-08-30
docs: update google oauth examples

malavolti:
I set up successfully the Google Authentication for my Immich App with Docker and I want to fix the example on the documentation to help other people to configure it.


jrasm91:
The .well-known part is optional, and it works without it. If you want to change demo to example, that's fine though.

malavolti:
> The .well-known part is optional, and it works without it. If you want to change demo to example, that's fine though.

Done

bo0tzz (approved):

PR #12180, opened 2024-08-31
fix(web): German translation for explorer

niklasfi:
Hi, the German translation for "explore" is off. "Entdeccker" as it is now is a misspelling of the German word "Entdecker" which means Explorer, but in an adventurer kind of way. In German we use the word "Explorer" (same as in English) to refer to the file app on Windows, and I think that would be the Best choice here.

Translation in context:
image

I am a native German speaker, so I'm relatively confident that what I am saying up there is correct.

### DIFF
--- a/web/src/lib/i18n/de.json
+++ b/web/src/lib/i18n/de.json
@@ -716,7 +716,7 @@
   "expired": "Verfallen",
   "expires_date": "Läuft am {date} ab",
   "explore": "Erkunden",
-  "explorer": "Entdeccker",
+  "explorer": "Explorer",
   "export": "Exportieren",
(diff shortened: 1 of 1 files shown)


danieldietzler (approved):
That works for me, thanks! For future reference you can also update translations on weblate :)

github-cli:
while Explorer is understood by most as that is how it is done in Windows itself as well
"Dateimanager" would probably the actual translation (that was used in Windows 3.11 (for those old enough to remember that, lol), to modernize it they renamed it to Windows Explorer even in the german translation I believe with Windows 95)

Datei-Explorer sounds a bit redundant, maybe it can be Immich Explorer instead 😄

PR #12220, opened 2024-09-01
refactor(server): library syncing

etnoy:
Merge all scan options into one
image

Library assets that are offline are moved to trash, with a little warning message attached to each asset
image

If the asset file is restored, the asset will be restored on next scan. If not, it'll be removed by the trash after the usual 30 days.

We remove isOffline and add a trashReason field to the asset entity


jrasm91:
I think it is time to consolidate the library scanning jobs. I don't think we need 4 separate jobs in a context menu like this.

jrasm91:
Can we have a single scan button/menu item that's opens a modal to submit a new scan job. It can use a drop-down to pick between a few options:
- new/added
- updated
- removed
- all

I think we can have a single library scanning job that takes in a body of options, that can use the selected type. Something like `{ type: 'new' }

etnoy:
> Can we have a single scan button/menu item that's opens a modal to submit a new scan job. It can use a drop-down to pick between a few options:
> 
> * new/added
> * updated
> * removed
> * all
> 
> I think we can have a single library scanning job that takes in a body of options, that can use the selected type. Something like `{ type: 'new' }`

The pr description is already outdated, I've cleaned it up somewhat. But I agree 100% and will try to incorporate this as well

mertalev:
I'd rather keep `isOffline` and use it for all assets instead of just library assets. It can be useful for the repair page, which is basically inaccessible for a lot of instances right now due to the amount of IO it does on each load.

etnoy:
> I'd rather keep `isOffline` and use it for all assets instead of just library assets. It can be useful for the repair page, which is basically inaccessible for a lot of instances right now due to the amount of IO it does on each load.
> 
> Edit: also, isn't it a little dangerous if someone happens to make a mistake when changing their mounts or something goes offline and a scan wipes out all their asset info?

Regarding the danger part, I've been thinking about this for a while. With isOffline, we aren't really helping the user at the moment. Only when viewing individual assets could you potentially spot that things are offline due to the red box on the right hand side, and this is only on web, not mobile. If someone (including myself) removes assets, they'll just run scan immediately followed by remove offline without any hesitation between the two steps.

This change ensures that the removal of deleted assets only happens manually, it's never automatically scheduled in any way. I'll add a simple check to each import path to make sure it's still there. This way we can exclude the simple cases of a network share being unavailable and make sure it doesn't cause the whole library to be deleted.

bo0tzz:
cc @zackpollard I think you were working on the repair page stuff right?

zackpollard:
> cc @zackpollard I think you were working on the repair page stuff right?

Yea, I put it on hold when I saw this rework was happening, I will continue after this gets merged :smile:

etnoy:
> > cc @zackpollard I think you were working on the repair page stuff right?
> 
> Yea, I put it on hold when I saw this rework was happening, I will continue after this gets merged 😄

You guys are talking about repair and I have no idea why. Are we using isOffline outside of external libraries? There's nothing in the code to suggest that

zackpollard:
> > > cc @zackpollard I think you were working on the repair page stuff right?
> > 
> > 
> > Yea, I put it on hold when I saw this rework was happening, I will continue after this gets merged 😄
> 
> You guys are talking about repair and I have no idea why. Are we using isOffline outside of external libraries? There's nothing in the code to suggest that

The plan was to change the repair page to run in the background, and given we already have a lot of logic for scanning "external" libraries, it'd make sense to extend this to all libraries and use that logic for the repair page. Otherwise we are maintaining two implementations of scanning for missing files.

server/src/dtos/library.dto.ts
```diff
@@ -95,6 +95,9 @@ export class ScanLibraryDto {
 
   @ValidateBoolean({ optional: true })
   refreshAllFiles?: boolean;
+
+  @ValidateBoolean({ optional: true })
+  removeDeleted?: boolean;
```
jrasm91, line 100: All of these options seems to be mutually exclusive, especially given how they are used in the web. Can we migrate this to an enum instead?

```ts
export enum LibraryAssetScan {
  ADDED='added', // files on disk without assets are added
  REMOVED='removed', // assets without files on disk are removed
  UPDATED='updated', // assets with files that have updated modification times are refreshed
}

Technically refreshing all assets isn't a scan, and I think it could be a separate endpoint. The implementation just needs to queue all assets for the library. You could argue that `libraryId` is an optional property to the "startJob" + metadata extraction job.

If we want to leave it in as another enum under `REFRESH='refresh'` that would be fine with me. 
```

etnoy:
This PR has taken another turn, now it moves offline files to trash instead of deleting them. Four scan options have been merged into one

server/src/interfaces/asset.interface.ts
```diff
@@ -142,64 +140,59 @@ export interface AssetUpdateDuplicateOptions {
   duplicateIds: string[];
 }
 
-export type AssetPathEntity = Pick<AssetEntity, 'id' | 'originalPath' | 'isOffline'>;
+export type AssetPathEntity = Pick<AssetEntity, 'id' | 'originalPath'>;
 
 export const IAssetRepository = 'IAssetRepository';
 
 export interface IAssetRepository {
-  getAssetsByOriginalPath(userId: string, partialPath: string): Promise<AssetEntity[]>;
```
danieldietzler, line 150: Did you sort these alphabetically? 😅
It's a mess in the diff lol

server/src/interfaces/search.interface.ts
```diff
@@ -57,12 +58,12 @@ export interface SearchStatusOptions {
   isEncoded?: boolean;
   isFavorite?: boolean;
   isMotion?: boolean;
-  isOffline?: boolean;
   isVisible?: boolean;
   isNotInAlbum?: boolean;
   type?: AssetType;
   withArchived?: boolean;
   withDeleted?: boolean;
+  trashReason?: AssetTrashReason | null;
```
danieldietzler, line 66: Why is it nullable here? Where's the difference between undefined and null for `trashReason` here?

server/src/repositories/asset.repository.ts
```diff
@@ -303,6 +294,7 @@ export class AssetRepository implements IAssetRepository {
   @Chunked()
   async restoreAll(ids: string[]): Promise<void> {
     await this.repository.restore({ id: In(ids) });
+    await this.updateAll(ids, { trashReason: null });
```
danieldietzler, line 297: Why not just work with `undefined` in this case as well? IMO `this.updateAll(ids, {})` is a bit simpler

server/src/services/asset.service.ts
```diff
@@ -286,6 +287,11 @@ export class AssetService {
         })),
       );
     } else {
+      if (!dto.trashReason) {
+        dto.trashReason = AssetTrashReason.DELETED;
```
danieldietzler, line 291: I'd prefer just destructuring `trashReason` in line 278 as well.

server/src/services/library.service.spec.ts
```diff
@@ -296,33 +223,48 @@ describe(LibraryService.name, () => {
...
-    it('should skip missing assets', async () => {
-      const mockAssetJob: ILibraryOfflineJob = {
-        id: assetStub.external.id,
-        importPaths: ['/'],
-        exclusionPatterns: [],
-      };
+  describe('handleQueueRemoveDeleted', () => {
+    it('should queue online check of existing assets', async () => {
+      libraryMock.get.mockResolvedValue(libraryStub.externalLibrary1);
+      storageMock.walk.mockImplementation(async function* generator() {});
```
danieldietzler, line 229: Can't you just `.mockImplementation(() => {})`?

server/src/services/library.service.spec.ts
```diff
@@ -585,12 +529,11 @@ describe(LibraryService.name, () => {
       expect(assetMock.create.mock.calls).toEqual([]);
     });
 
-    it('should not import an asset when mtime matches db asset', async () => {
+    it('should not refresh a file whose mtime matches existing asset', async () => {
```
danieldietzler, line 532:
```suggestion
    it('should not refresh a file which mtime matches existing asset', async () => {
```

server/src/services/library.service.spec.ts
```diff
@@ -607,12 +550,11 @@ describe(LibraryService.name, () => {
       expect(jobMock.queueAll).not.toHaveBeenCalled();
     });
 
-    it('should import an asset when mtime differs from db asset', async () => {
+    it('should refresh a file whose mtime differs from existing asset', async () => {
```
danieldietzler, line 553:
```suggestion
    it('should refresh a file which mtime differs from existing asset', async () => {
```

server/src/services/library.service.ts
```diff
@@ -366,258 +358,208 @@ export class LibraryService {
...
-      // TODO: We can likely remove this check in the second half of 2024 when all assets have likely been re-imported by all users
-      this.logger.debug(
-        `Asset is missing file extension, re-importing: ${assetPath}. Current incorrect filename: ${existingAssetEntity.originalFileName}.`,
-      );
-      doRefresh = true;
-    } else if (!job.force && stats && !existingAssetEntity.isOffline) {
-      // Asset exists on disk and in db and mtime has not changed. Also, we are not forcing refresn. Therefore, do nothing
-      this.logger.debug(`Asset already exists in database and on disk, will not import: ${assetPath}`);
-    }
+    if (asset) {
```
danieldietzler, line 377: I still find this quite hard to read/complex overall (lots of nesting/conditions).

server/src/services/library.service.ts
```diff
@@ -366,258 +358,208 @@ export class LibraryService {
...
-
-    if (!doImport && !doRefresh) {
-      // If we don't import, exit here
-      return JobStatus.SKIPPED;
-    }
+      if (asset.trashReason == AssetTrashReason.DELETED) {
+        // Asset is trashed by user, don't re-import. This is to prevent re-importing assets that are manually trashed by the user
+        this.logger.debug(`Asset is previously trashed by user, won't refresh: ${assetPath}`);
+        return JobStatus.SKIPPED;
+      } else if (asset.trashReason == AssetTrashReason.OFFLINE) {
```
danieldietzler, line 384: You don't need the else here.

server/src/services/library.service.ts
```diff
@@ -366,258 +358,208 @@ export class LibraryService {
...
+        this.logger.debug(`Asset is previously trashed as offline, restoring from trash: ${assetPath}`);
+        await this.assetRepository.restoreAll([asset.id]);
+        return JobStatus.SUCCESS;
 
-    let assetType: AssetType;
+        // TODO: can we send an event just like the asset restore in the trash service?
+      } else {
```
danieldietzler, line 390: You don't need the else here.

server/src/services/library.service.ts
```diff
@@ -366,258 +358,208 @@ export class LibraryService {
...
+        if (mtime.toISOString() === asset.fileModifiedAt.toISOString()) {
+          // Asset exists on disk and in db and mtime has not changed, do nothing
+          this.logger.debug(`Asset already exists in database and on disk, will not import: ${assetPath}`);
+          return JobStatus.SKIPPED;
+        } else {
```
danieldietzler, line 395: You don't need this else.

server/src/services/trash.service.ts
```diff
@@ -21,7 +22,14 @@ export class TrashService {
   async restoreAssets(auth: AuthDto, dto: BulkIdsDto): Promise<void> {
     const { ids } = dto;
     await requireAccess(this.access, { auth, permission: Permission.ASSET_DELETE, ids });
-    await this.restoreAndSend(auth, ids);
+
+    const assets = await this.assetRepository.getByIds(ids);
+    const deletedAssets = assets.filter((a) => a.trashReason === AssetTrashReason.DELETED);
```
danieldietzler, line 27:
```suggestion
    const deletedAssets = assets.filter(({ trashReason }) => trashReason === AssetTrashReason.DELETED);
```

server/src/services/trash.service.ts
```diff
@@ -32,7 +40,7 @@ export class TrashService {
     );
 
     for await (const assets of assetPagination) {
-      const ids = assets.map((a) => a.id);
+      const ids = assets.filter((a) => a.trashReason === AssetTrashReason.DELETED).map((a) => a.id);
```
danieldietzler, line 43:
```suggestion
      const ids = assets.filter(({ trashReason }) => trashReason === AssetTrashReason.DELETED).map((a) => a.id);
```

web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
```diff
@@ -136,7 +137,7 @@
         {#if showDownloadButton}
           <DownloadAction {asset} menuItem />
         {/if}
-        {#if asset.isTrashed}
+        {#if asset.trashReason == TrashReason.Deleted}
```
danieldietzler, line 140:
```suggestion
        {#if asset.trashReason === TrashReason.Deleted}
```

web/src/lib/components/asset-viewer/detail-panel.svelte
```diff
@@ -141,13 +143,20 @@
...
-        <div class="rounded-t bg-red-500 px-4 py-2 font-bold text-white">{$t('asset_offline')}</div>
+        <div class="rounded-t bg-red-500 px-4 py-2 font-bold text-white">
+          {$t('asset_offline')}
+        </div>
         <div class="rounded-b border border-t-0 border-red-400 bg-red-100 px-4 py-3 text-red-700">
           <p>
-            {$t('asset_offline_description')}
+            {#if $user?.isAdmin}
+              <p>{$t('admin.asset_offline_description')}</p>
+              <p class="text-sm">{$t('path')}: {asset.originalPath}</p>
```
danieldietzler, line 156: I think with regards to RTL languages this entire string needs to be localized (so passing the path as a variable), as the structure `Path: /path/to/asset` may not always work

danieldietzler (commented):
Regarding the translation files: I would just update the English file. Then, weblate will delete the keys in all the other files and add new ones, yes, but that may actually be what we want as translations may be inaccurate now.

server/src/services/library.service.ts
etnoy: It's in there for the log statement

server/src/services/trash.service.ts
etnoy: Thanks. Do you know of a way to do the filter already in the db call? I couldn't find a good way

server/src/services/library.service.ts
danieldietzler: You always return before, that's why you don't need the else

server/src/services/trash.service.ts
danieldietzler: cc @mertalev

server/src/controllers/library.controller.ts
```diff
@@ -67,14 +66,7 @@ export class LibraryController {
...
-  scanLibrary(@Param() { id }: UUIDParamDto, @Body() dto: ScanLibraryDto) {
-    return this.service.queueScan(id, dto);
-  }
-
-  @Post(':id/removeOffline')
-  @HttpCode(HttpStatus.NO_CONTENT)
-  @Authenticated({ admin: true })
-  removeOfflineFiles(@Param() { id }: UUIDParamDto) {
-    return this.service.queueRemoveOffline(id);
+  scanLibrary(@Param() { id }: UUIDParamDto) {
```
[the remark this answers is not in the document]
etnoy: It's an admin-only route so it should be fine :)

server/src/interfaces/search.interface.ts
etnoy: Fixed!

server/src/interfaces/asset.interface.ts
etnoy: Maybe I should move the sorting of repos to another PR. I can't stand them unsorted :)

server/src/services/asset.service.ts
etnoy: Done

server/src/services/library.service.spec.ts
etnoy: No, it didn't work out like that

server/src/services/library.service.spec.ts
etnoy: Nope, my version is correct :)

server/src/services/library.service.ts
etnoy: Got it!

server/src/services/library.service.ts
etnoy: Done

server/src/repositories/asset.repository.ts
danieldietzler: Nvm, has been discussed on Discord.

server/src/controllers/library.controller.ts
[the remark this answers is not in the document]
bo0tzz: We probably still want to have permission checks, especially if we're moving towards applying libraries more broadly, right?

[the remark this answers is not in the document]
etnoy: Good point, I'm adding the LIBRARY_UPDATE permission to the endpoint

web/src/lib/components/asset-viewer/detail-panel.svelte
etnoy: Fixed now

server/src/services/trash.service.ts
zackpollard: Should just be able to add `trashReason: AssetTrashReason.DELETED,` below line 38, right?

server/src/services/library.service.ts
```diff
@@ -366,258 +359,212 @@ export class LibraryService {
...
 
-  async handleOfflineCheck(job: ILibraryOfflineJob): Promise<JobStatus> {
+  async handleAssetOfflineCheck(job: ILibraryOfflineJob): Promise<JobStatus> {
     const asset = await this.assetRepository.getById(job.id);
 
-    if (!asset) {
-      // Asset is no longer in the database, skip
+    if (!asset || asset.trashReason) {
+      // Skip if asset is missing or already trashed
+      // We don't want to trash an asset that has already been trashed because it can otherwise re-appear on the timeline if an asset is re-imported
```
zackpollard, line 524: Not sure how I feel about this specifically, it leaves a bit of an open edge-case that won't be picked up until the next scan. Not sure how we'd resolve that really though while still using the trashReason stuff.

zackpollard (dismissed):
Some notes, but overall this looks like a great change and vastly simplifies the external libraries scanning code. Once comments are resolved I am happy for this to be merged in :smile:

server/src/entities/asset.entity.ts
```diff
@@ -170,6 +168,13 @@ export class AssetEntity {
   @OneToOne(() => AssetJobStatusEntity, (jobStatus) => jobStatus.asset, { nullable: true })
   jobStatus?: AssetJobStatusEntity;
 
+  @Column({
+    type: 'enum',
+    enum: AssetTrashReason,
+    nullable: true,
+  })
+  trashReason!: AssetTrashReason | null;
```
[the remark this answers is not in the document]
mertalev: I think it makes sense for it to be nullable, personally. This is something that's only really known when it's being trashed, so it should be set at that time.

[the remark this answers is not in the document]
zackpollard: Right, so it should be set to deleted. Everything that exists right now is deleted. What's the reason for not setting it?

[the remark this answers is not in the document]
etnoy: You bring up a good point @zackpollard. Any existing assets that are in the trash must be migrated to have trashReason set to deleted.

I also agree with @mertalev that trashReason should be nullable since the field will be null until it's trashed.

Therefore, I need to write a migration to set the field for all deleted assets.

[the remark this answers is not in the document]
mertalev: Are you talking about the migration? The value it should be set to depends on whether `deletedAt` or `isOffline` is true, so there should be a query that sets it with `case when "deletedAt" is not null then 'deleted' when "isOffline" then 'offline' end`. In the case where `isOffline` is true and `deletedAt` is null, it should also set `deletedAt`.

etnoy:
> Some notes, but overall this looks like a great change and vastly simplifies the external libraries scanning code. Once comments are resolved I am happy for this to be merged in 😄

Thanks! Happy that you like the concepts in this PR. I'll do some modifications based on your suggestions.

server/src/entities/asset.entity.ts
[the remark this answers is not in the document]
zackpollard: Honestly completely overlooked the case of it not being trashed 🤣 But yes we can do that for the existing trashed assets was more my point :P

server/src/services/trash.service.ts
etnoy: Done!

mobile/lib/extensions/collection_extensions.dart
```diff
@@ -72,13 +72,14 @@ extension AssetListExtension on Iterable<Asset> {
   }
 
   /// Filters out offline assets and returns those that are still accessible by the Immich server
+  /// TODO: isOffline is removed from Immich, so this method is not useful anymore
```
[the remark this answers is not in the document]
etnoy: Yes, but I have no idea if I can remove the nonOfflineOnly function

server/src/services/library.service.ts
```diff
@@ -366,258 +359,212 @@ export class LibraryService {
...
-        `Asset is missing file extension, re-importing: ${assetPath}. Current incorrect filename: ${existingAssetEntity.originalFileName}.`,
-      );
-      doRefresh = true;
-    } else if (!job.force && stats && !existingAssetEntity.isOffline) {
-      // Asset exists on disk and in db and mtime has not changed. Also, we are not forcing refresn. Therefore, do nothing
-      this.logger.debug(`Asset already exists in database and on disk, will not import: ${assetPath}`);
-    }
+  private async refreshExistingAsset(asset: AssetEntity) {
+    if (asset.trashReason == AssetTrashReason.DELETED) {
+      // Asset is trashed by user, don't re-import. This is to prevent re-importing assets that are manually trashed by the user
```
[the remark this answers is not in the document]
etnoy: You're right, the log messages I added make the comments redundant

server/src/services/library.service.ts
etnoy: The whole reason for the trashReason complexity is to handle this edge case. What edge case is left open?

server/src/services/library.service.ts
```diff
@@ -366,258 +359,212 @@ export class LibraryService {
...
+      this.logger.debug(`Asset is previously trashed as offline, restoring from trash: ${asset.originalPath}`);
+      await this.assetRepository.restoreAll([asset.id]);
+      return JobStatus.SUCCESS;
 
-    if (stats && existingAssetEntity?.isOffline) {
-      // File was previously offline but is now online
-      this.logger.debug(`Marking previously-offline asset as online: ${assetPath}`);
-      await this.assetRepository.update({ id: existingAssetEntity.id, isOffline: false });
-      doRefresh = true;
+      // TODO: can we send an event just like the asset restore in the trash service?
```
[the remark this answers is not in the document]

server/src/entities/asset.entity.ts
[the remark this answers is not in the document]
jrasm91: The question seems to be "do we want `null` to imply `deleted`". I am fine making it non-nullable, always requiring a value, and setting it to `deleted` when a request comes in through the api.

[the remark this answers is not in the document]
mertalev: With kysely, null fields can be omitted from DB responses entirely, so it's generally better to set something null than to have a default that has the same meaning.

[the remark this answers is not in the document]
jrasm91: Very free queries return deleted data in the first place, so I don't think that's a very strong argument.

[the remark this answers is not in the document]
mertalev: But if `trashReason` were always set to a value, it'd always be part of the response even if the asset isn't trashed. I guess you could omit it from the select instead, but that's not as clean.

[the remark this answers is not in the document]
jrasm91: Oh right. I think I meant to recommend it always have a value _when an asset has been deleted_. So null when not deleted, deleted when trashed and offline when it goes offline. I guess that means nullable, but not "null implies deleted".

server/src/services/library.service.ts
zackpollard: If an asset is trashed and goes offline what happens? We don't change the reason right? So if the user then untrashes the asset it could be in their timeline until the next scan but be broken. An edge case though and not something we can easily mitigate.

jrasm91: Hah, I guess we could make another enum "trashed _and_ offline"

server/src/entities/asset.entity.ts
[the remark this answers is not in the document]
mertalev: Ah, that makes sense. I also agree with that.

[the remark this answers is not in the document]
mertalev: Maybe there should be a check constraint on this like `("deletedAt" is null) = ("trashReason" is null)`, so we have an invariant that either both are null or neither are.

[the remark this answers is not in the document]
etnoy: Just to confirm: I currently implemented this by having trashReason be null for all non-trashed item. When an item is trashed by the api, it's set to deleted, and when the file is offline it's set to offline. This is the way we want, right?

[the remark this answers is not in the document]
zackpollard: Yes that sounds right to me 🙂

server/src/interfaces/job.interface.ts
```diff
@@ -76,12 +76,12 @@ export enum JobName {
...
   // library management
-  LIBRARY_SCAN = 'library-refresh',
-  LIBRARY_SCAN_ASSET = 'library-refresh-asset',
-  LIBRARY_REMOVE_OFFLINE = 'library-remove-offline',
-  LIBRARY_CHECK_OFFLINE = 'library-check-offline',
+  LIBRARY_QUEUE_SYNC_FILES = 'library-scan-new',
+  LIBRARY_QUEUE_SYNC_ASSETS = 'library-queue-remove-deleted',
+  LIBRARY_SYNC_FILE = 'library-refresh-asset',
+  LIBRARY_SYNC_ASSET = 'library-remove-deleted',
   LIBRARY_DELETE = 'library-delete',
```
danieldietzler, line 83: Nit: can we use the enum names as the values instead of using new words? To me this is somewhat confusing.

server/src/services/library.service.ts
```diff
@@ -250,15 +243,14 @@ export class LibraryService {
     return mapLibrary(library);
   }
 
-  private async scanAssets(libraryId: string, assetPaths: string[], ownerId: string, force = false) {
+  private async scanAssets(libraryId: string, assetPaths: string[], ownerId: string) {
```
jrasm91, line 246: Maybe we should rename this to syncFiles? Also maybe `syncFiles(assetPaths: string, { libraryId, ownerId }: { libraryId: string, ownerId: string })` would look better

server/src/services/library.service.ts
```diff
@@ -630,55 +545,66 @@ export class LibraryService {
...
-    if (validImportPaths.length === 0) {
+    if (validImportPaths) {
+      const assetsOnDisk = this.storageRepository.walk({
+        pathsToCrawl: validImportPaths,
+        includeHidden: false,
+        exclusionPatterns: library.exclusionPatterns,
+        take: JOBS_LIBRARY_PAGINATION_SIZE,
+      });
+
+      let crawledAssets = 0;
```
jrasm91, line 556: Maybe just `count`.

jrasm91 (approved):
Thanks for working on this @etnoy and @zackpollard!

server/src/services/library.service.ts
```diff
@@ -366,258 +358,181 @@ export class LibraryService {
...
     }
 
-    let doImport = false;
-    let doRefresh = false;
+    this.logger.log(`Importing new library asset: ${assetPath}`);
 
-    if (job.force) {
-      doRefresh = true;
+    const library = await this.repository.get(job.id, true);
+    if (library?.deletedAt) {
```
danieldietzler, line 385: Shouldn't we check if the library actually exists? I guess it's already covered wherever we queue the sync, but it still feels odd not sanity-checking that here

server/src/services/library.service.ts
```diff
@@ -630,55 +545,66 @@ export class LibraryService {
...
+      let crawledAssets = 0;
+
+      for await (const assetBatch of assetsOnDisk) {
+        crawledAssets += assetBatch.length;
+        this.logger.debug(`Discovered ${crawledAssets} asset(s) on disk for library ${library.id}...`);
+        await this.scanAssets(job.id, assetBatch, library.ownerId);
+        this.logger.verbose(`Queued scan of ${assetBatch.length} crawled asset(s) in library ${library.id}...`);
+      }
+
+      if (crawledAssets) {
```
danieldietzler, line 565: I'd prefer being explicit here, so `if (count > 0)`

danieldietzler (approved):

PR #12222, opened 2024-09-01
feat(web): load original panorama image when zoomed in to 75% or above

TapuCosmo:
## Description
This PR allows the photo sphere viewer to load the original resolution image when the user zooms in to 75% or above. The currently existing photo sphere viewer uses the preview resolution image at all zoom levels, resulting in a loss of quality that is especially noticeable when zoomed in.

### DIFF
--- a/web/src/lib/components/asset-viewer/panorama-viewer.svelte
+++ b/web/src/lib/components/asset-viewer/panorama-viewer.svelte
@@ -1,11 +1,12 @@
 <script lang="ts">
   import { getAssetOriginalUrl, getKey } from '$lib/utils';
+  import { isWebCompatibleImage } from '$lib/utils/asset-utils';
   import { AssetMediaSize, AssetTypeEnum, viewAsset, type AssetResponseDto } from '@immich/sdk';
   import type { AdapterConstructor, PluginConstructor } from '@photo-sphere-viewer/core';
   import { fade } from 'svelte/transition';
   import LoadingSpinner from '../shared-components/loading-spinner.svelte';
   import { t } from 'svelte-i18n';
-  export let asset: Pick<AssetResponseDto, 'id' | 'type'>;
+  export let asset: AssetResponseDto;
 
   const photoSphereConfigs =
     asset.type === AssetTypeEnum.Video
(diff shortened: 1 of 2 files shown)


web/src/lib/components/asset-viewer/photo-sphere-viewer-adapter.svelte
```diff
@@ -28,6 +30,18 @@
       maxFov: 180,
       fisheye: true,
     });
+
+    if (originalImageUrl) {
+      const zoomHandler = ({ zoomLevel }: events.ZoomUpdatedEvent) => {
+        // zoomLevel range: [0, 100]
+        if (Math.round(zoomLevel) >= 75) {
```
michelheusschen, line 37: It might be worth removing this check and loading the original on any zoom event, considering that loading the original can take some time

web/src/lib/components/asset-viewer/photo-sphere-viewer-adapter.svelte
```diff
@@ -28,6 +30,18 @@
...
+          // Replace the preview with the original
+          viewer.setPanorama(originalImageUrl, { showLoader: false, speed: 150 }).catch(() => {});
```
michelheusschen, line 39: If the original image is unsupported a `The panorama cannot be loaded` overlay will show up. I think we should do one or both of the following:
- Only load the image when `isWebCompatibleImage(asset)` is true
- Fall back to the previous image when loading fails

web/src/lib/components/asset-viewer/panorama-viewer.svelte
```diff
@@ -27,14 +27,23 @@
...
     return url;
   };
+
+  const originalImageUrl = asset.type === AssetTypeEnum.Image ? getAssetOriginalUrl(asset.id) : null;
 </script>
 
 <div transition:fade={{ duration: 150 }} class="flex h-full select-none place-content-center place-items-center">
   <!-- the photo sphere viewer is quite large, so lazy load it in parallel with loading the data -->
-  {#await Promise.all([loadAssetData(), import('./photo-sphere-viewer-adapter.svelte'), ...photoSphereConfigs])}
+  {#await Promise.all( [loadAssetData(), originalImageUrl, import('./photo-sphere-viewer-adapter.svelte'), ...photoSphereConfigs], )}
```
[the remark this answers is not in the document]
TapuCosmo: Fixed

web/src/lib/components/asset-viewer/photo-sphere-viewer-adapter.svelte
TapuCosmo: Added `isWebCompatibleImage` check to set the `originalImageUrl`, and changed the catch for `setPanorama` to set it back to the preview image.

web/src/lib/components/asset-viewer/photo-sphere-viewer-adapter.svelte
TapuCosmo: I put in this check mainly to avoid fetching relatively large files (~57 MB original vs 330 kB preview in my case) if the user does not zoom in enough to where the details in the higher resolution original image are significant. The threshold is kind of arbitrary right now though, and user experience will probably vary based on configured preview resolutions.

michelheusschen (approved):

PR #12226, opened 2024-09-02
feat(server): add OAuth environment variables

bdsoha:
🔐 Add the ability to configure OAuth using environment variables.

Related #11045


bo0tzz:
We already have the config file for settings like this (https://immich.app/docs/install/config-file), personally I'm against introducing environment variables for them as well.

bdsoha:
The config file is excellent, but *secrets* should not be passed as plaintext.
IMO, there shouldn't be a difference between the `DB_*`, `REDIS_*`, or even the other `IMMICH_*` already available.

For users who feel comfortable storing secrets in config files, they can continue to use the current implementation.
More secure deployments can opt-in to using `env`s.

bo0tzz:
You can use environment variable placeholders in the config file, then run `envsubst` on it just before passing it to Immich. That's plenty secure.

> there shouldn't be a difference between the DB_*, REDIS_*, or even the other IMMICH_* already available.

There is, because those settings are about infrastructure that Immich needs to even start up, and don't exist in the config file.

bo0tzz:
I just did some config templating from an external secret in k8s pretty recently, yes it's more complex than pulling from an env var but it works just fine. 

https://github.com/immich-app/devtools/blob/4c8d603add876bd7e1fe54c91a5eb35732f1d558/kubernetes/apps/tools/containerssh/app/deployment.yaml#L19-L35

bdsoha:
Can I conclude from this thread, it is not in Immich's roadmap to support _env-based_ configurations?

jrasm91:
At this point in time we are not planning to support changing individual system config via env variable.

PR #12230, opened 2024-09-02
docs: fixing example docker compose

Biepa:
Change needed so the following statement included in the docs a bit below makes sense: NOTE: We have to use the `/mnt/media/christmas-trip` path and not the `/mnt/nas/christmas-trip` path since all paths have to be what the Docker containers see.


mmomjian:
Can we also fix /home/user/old-pics to /mnt/media/old-pics? Same issue. Ty

Biepa:
@mmomjian Done.

mmomjian (approved):

alextran1502 (approved):

PR #12254, opened 2024-09-03
fix(server): remove thumbnailAt in asset_job_status for missing thumbnails

mPyKen:
Similar to https://github.com/immich-app/immich/pull/12225, I realized some assets had `thumbnailAt` set, but had no thumbnail file in `asset_files`. In https://github.com/immich-app/immich/pull/11908/files#diff-99da4a5a2a767c94716e167a5d49f089778c22365cf403e92808711beb6125b9, `thumbnailAt` was set to `NOW()` if `thumbnailPath IS NOT NULL`, whereas `asset_file` was only created when `"thumbnailPath" IS NOT NULL AND "thumbnailPath" != ''` in https://github.com/immich-app/immich/pull/11861/files#diff-09097609a560037cd9fc82b6d5de63f184101d01c8d14a53182de0d5e649d4e0. This PR fixes actually missing thumbnails to be recognized as missing, again.

```
UPDATE "asset_job_status" j
SET "thumbnailAt" = NULL
WHERE j."thumbnailAt" IS NOT NULL
AND NOT EXISTS (
	SELECT 1 FROM public.asset_files f
	WHERE j."assetId" = f."assetId"
	AND f."type" = 'thumbnail'
	AND f."path" IS NOT NULL
)
```

### DIFF
--- a/server/src/migrations/1725327902980-RemoveThumbailAtForMissingThumbnails.ts
+++ b/server/src/migrations/1725327902980-RemoveThumbailAtForMissingThumbnails.ts
@@ -1,0 +1,12 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class RemoveThumbailAtForMissingThumbnails1725327902980 implements MigrationInterface {
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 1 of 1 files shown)


mertalev:
Ugh, I hate empty strings... The same thing applies to previews, right? It'd be best to handle them too.

mPyKen:
Interestingly, I didn't find such cases for previews. I guess we managed not to create empty strings for previews so far. I can still add that migration if need be

mertalev:
Weirdly enough, the old preview column apparently defaulted to null while the thumbnail column defaulted to an empty string. I guess there's no need to handle it then.

mertalev (approved):

PR #12255, opened 2024-09-03
feat(web): filter shared links for albums

TapuCosmo:
## Description
Adds a path to display only the shared links corresponding to an album. Changes the View links button from albums to go to this path instead of the generic sharedlinks path that displays shared links for all assets and albums.

I don't know where to put a button to view links for individual shares, so this feature is only fully implemented for albums right now. Additionally, I'm not sure what the best way to handle individual shares that include multiple assets is. Should it only display shared links that contain all the selected assets or that contain any of the selected assets?


bo0tzz:
This implements #9989, right?

zackpollard:
This was implemented in #15948 and you can find it on your instance @ https://my.immich.app/shared-links?filter=album

PR #12260, opened 2024-09-03
ci: more path filtering, path filtering happens in pre-job so all jobs can be required

zackpollard:
This PR moves our path filtering from the github workflow path filtering to a pre-job. This allows us to have more flexibility on filtering specific jobs with path filtering, as well as meaning skipped jobs will show as a success, allowing us to require all jobs on all PRs.

Additional path filters added in this PR:
- build-mobile now path filters on mobile/**
- docker builds were split into machine-learning and server, now have path filtering for those builds (steps were duplicated as no easy way to de-duplicate that right now, worth it for the saved runner concurrency imo)
- static-analysis now path filters on mobile/**
- tests now path filter on a range of different paths depending on the job

This should massively reduce our job concurrency and for small PRs just touching specific sub-sections of the codebase, should massively reduce the time waiting for jobs to complete :smile_cat:

### DIFF
--- a/.github/workflows/build-mobile.yml
+++ b/.github/workflows/build-mobile.yml
@@ -16,10 +16,28 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  pre-job:
+    runs-on: ubuntu-latest
+    outputs:
+      should_run: ${{ steps.found_paths.outputs.mobile == 'true' || steps.should_force.outputs.should_force == 'true' }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - id: found_paths
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            mobile:
+              - 'mobile/**'
+      - name: Check if we should force jobs to run
+        id: should_force
+        run: echo "should_force=${{ github.event_name == 'workflow_call' || github.event_name == 'workflow_dispatch' }}" >> "$GITHUB_OUTPUT"
+
   build-sign-android:
     name: Build and sign Android
+    needs: pre-job
     # Skip when PR from a fork
-    if: ${{ !github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]' }}
+    if: ${{ !github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]' && needs.pre-job.outputs.should_run == 'true' }}
     runs-on: macos-14
 
     steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/build-mobile.yml
```diff
@@ -16,10 +16,28 @@ concurrency:
...
+      - name: Check if we should force jobs to run
+        id: should_force
+        run: echo "should_force=${{ github.event_name == 'workflow_call' || github.event_name == 'workflow_dispatch' }}" >> "$GITHUB_OUTPUT"
+
   build-sign-android:
     name: Build and sign Android
+    needs: pre-job
     # Skip when PR from a fork
-    if: ${{ !github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]' }}
+    if: ${{ !github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]' && needs.pre-job.outputs.should_skip == 'false' }}
```
bo0tzz, line 40: The output is called `should_run` but this uses `should_skip`.

.github/workflows/docker.yml
```diff
@@ -17,47 +17,149 @@ permissions:
...
+    env:
+      image: immich-server
+      context: .
+      file: server/Dockerfile
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platforms: linux/amd64,linux/arm64
+            device: cpu
```
bo0tzz, line 162: Afaict we don't need the matrix here anymore. I believe device is not used, and platforms can just be included straight up in the build-push-action call. There's also still some stale uses of matrix.suffix in the image tags.

.github/workflows/static_analysis.yml
```diff
@@ -10,8 +10,27 @@ concurrency:
...
+            mobile:
+              - 'mobile/**'
+      - name: Check if we should force jobs to run
+        id: should_force
+        run: echo "should_force=${{ github.event_name == 'release' }}" >> "$GITHUB_OUTPUT"
+
   mobile-dart-analyze:
     name: Run Dart Code Analysis
+    needs: pre-job
+    if: ${{ needs.pre-job.outputs.should_skip == 'false' }}
```
bo0tzz, line 33: Uses `should_skip` but output is `should_run`

bo0tzz (changes requested):
Generally looks good, just a few details.

bo0tzz (approved):

PR #12270, opened 2024-09-03
feat(web): Use mouse scroll to resize crop in editor

Webrow:
First PR for this project, I don't know if you want all code commented because this is not available in the releases yet.
Reasoning, while editing the images I'd like to be quick in sorting through them, rotating, and making images ready for sharing on other platforms.


zackpollard:
Hey, we decided we won't accept PRs to the editor until it's actually available in web. It still requires some fundamental changes. Once those are done, please feel free to make a new PR with these changes again :slightly_smiling_face:

PR #12288, opened 2024-09-03
feat(server): support lightroom tags

jrasm91:
Add support for reading tags from `HierarchicalSubject` 
- Treat `|` as hierarchy symbol
- Replace `/` characters with `|` (after parsing hierachy)

Closes #12196


alextran1502 (approved):

danieldietzler (approved):
LGTM

PR #12319, opened 2024-09-04
fix(text): Typo fix in translation

JavierMonton:
A simple typo fix for the spanish translation. It should be `original` rather than `oeiginal`.

imagen


jrasm91:
can you just make this change in weblate?

https://hosted.weblate.org/translate/immich/immich/es/

PR #12330, opened 2024-09-04
feat(web): add skip link to sidebar

ben-basten:
## Description

Show a skip link in the sidebar for the folders and tags pages.

In my production immich instance, I have ~5000 tags, which is difficult to navigate past with a keyboard. Offering a skip link gives users a quick way to navigate to the asset grid or breadcrumbs.

## How Has This Been Tested?

- [x] Verify that the top "Skip to content" skip link still works
- [x] Use the new skip link in both the "Folders" and "Tags" pages to move focus to the header
- [x] Verify that the admin and main page sidebars do not have skip links

## Screenshots

<details><summary>Expand for screenshot</summary>
<p>

sidebar-skip-link

</p>
</details>


alextran1502 (approved):

michelheusschen (approved):
This works for now, but the folder/tags tree should have accessibility and keyboard features and then this skip link wouldn't be needed

ben-basten:
> This works for now, but the folder/tags tree should have accessibility and keyboard features and then this skip link wouldn't be needed

I agree, some keyboard work in particular would be awesome. Perhaps a top level accordion in the tree would serve a similar purpose as adding a skip link.

PR #12335, opened 2024-09-05
feat(server): refresh face detection

mertalev:
## Description

This PR allows re-running face detection without clearing face data. This can be done on one or more assets with the "Refresh faces" option, or for all assets in the job panel with a new Refresh button.

It compares existing and new bounding boxes for an asset with the IoU metric. If a new bounding box is above the IoU threshold, it's considered a duplicate of an existing face and skipped. If an existing face doesn't appear in the new set of bounding boxes, it's removed.

This IoU-based matching has an additional benefit in allowing faces from non-machine learning sources to be associated with an embedding. This means that detected faces can match against people listed in face metadata instead of creating duplicate people.

To make the three options of Reset / Refresh / Missing work with the current API, a distinction is made between `force` being `undefined` and `false`. The former maps to Refresh, and the latter maps to Missing.

face-detection-job-panel

refresh-faces

Testing:
- Increasing the detection threshold and re-running removes existing faces below the new threshold
- Lowering the threshold and re-running adds new faces
- Refreshing with the same face detection settings doesn't change anything
- "Missing" and "All" (now called "Reset") behave the same
- Refreshing faces for a single asset works as expected


pyorot:
seems like time to put all the Refresh… options in a sub-menu

radh21301:
Hey. Would this also work when we change the facial recognition model? Like from buffalo_l to antelopev2? I suppose not?

mertalev:
No, there are a few more things that would need to happen to let you switch without resetting everything. But it's a big step toward that.

sammyke007:
Will this be in the next update?

jrasm91 (approved):
Looks good. Can you add a TODO to not use undefined in the API (for force)? Imo we should migrate to a more explicit API in the future instead of using undefined for a third state.

mertalev:
Do you know how to make the PR label validation check happy?

PR #12339, opened 2024-09-05
feat(web): add download shortcut on the timeline & asset viewer

### LINKED ISSUE #11375: [Feature] Download from the timeline with a keyboard shortcut (SHIFT+D)
### The bug

When you are on the homepage and selecting multiple images the shortcut to download (SHIFT + D) does not work. I can confirm that the delete shortcut (DEL) does work as expected.

Ubuntu Server 22.04 LTS

### Version of Immich Server

v1.110.0

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

```bash
1. Select 1 or multiple items from the homepage
2. Press SHIFT + D on your keyboard
3. Nothing happens
4. When pressing DEL on your keyboard the items do get deleted as expected
```

### Relevant log output

_No response_

### Additional information

_No response_

imondrag:
We currently tell users that the download shortcut is "shift+d" but it's not actually implemented!

This PR will address #11375

### DIFF
--- a/web/src/lib/components/photos-page/actions/download-action.svelte
+++ b/web/src/lib/components/photos-page/actions/download-action.svelte
@@ -2,6 +2,7 @@
   import CircleIconButton from '$lib/components/elements/buttons/circle-icon-button.svelte';
   import { downloadArchive, downloadFile } from '$lib/utils/asset-utils';
   import MenuOption from '../../shared-components/context-menu/menu-option.svelte';
+  import { shortcut } from '$lib/actions/shortcut';
   import { getAssetControlContext } from '../asset-select-control-bar.svelte';
   import { mdiCloudDownloadOutline, mdiFileDownloadOutline, mdiFolderDownloadOutline } from '@mdi/js';
   import { t } from 'svelte-i18n';
@@ -25,6 +26,8 @@
 
   $: menuItemIcon = getAssets().size === 1 ? mdiFileDownloadOutline : mdiFolderDownloadOutline;
 </script>
+
+<svelte:window use:shortcut={{ shortcut: { key: 'd', shift: true }, onShortcut: handleDownloadFiles }} />
 
 {#if menuItem}
   <MenuOption text={$t('download')} icon={menuItemIcon} onClick={handleDownloadFiles} />


michelheusschen (commented):
It's better to add the shortcut directly in the download action to avoid duplicating logic and to make sure it also works for shared links

https://github.com/immich-app/immich/blob/main/web/src/lib/components/photos-page/actions/download-action.svelte

web/src/lib/components/photos-page/asset-grid.svelte
```diff
@@ -459,6 +459,18 @@
...
 
+  const downloadSelectedAssets = async () => {
+    if ($selectedAssets.size === 1) {
+      const [asset] = $selectedAssets;
+      assetInteractionStore.clearMultiselect();
+      await downloadFile(asset);
+      return;
+    }
+
+    assetInteractionStore.clearMultiselect();
```
jrasm91, line 470: We should probably clear the selection _after_ the download happens. Also, can you wrap this in a try/catch that uses `handleError` to send a notification to the user if something goes wrong.

michelheusschen: The `downloadFile` promise only resolves once the file is fully downloaded. Clearing the selection afterwards could be problematic, because the user can change the selection in the meantime. This function already does some error handling and will show a notification if the download fails, so I'm not sure if the try/catch is even needed.

jrasm91: I see.

imondrag:
> It's better to add the shortcut directly in the download action to avoid duplicating logic and to make sure it also works for shared links
> 
> https://github.com/immich-app/immich/blob/main/web/src/lib/components/photos-page/actions/download-action.svelte

Great idea! It turns out the asset viewer already had the shortcut implemented. It was simply missing from the photos page

https://github.com/immich-app/immich/blob/main/web/src/lib/components/asset-viewer/actions/download-action.svelte

michelheusschen (approved):

PR #12366, opened 2024-09-05
feat: optimize copy image to clipboard

martabal:
Remove the `copy-image-clipboard` dependency. When copying the asset to the clipboard, use the asset from the photoviewer instead of doing a new network request.

### DIFF
--- a/web/src/lib/i18n/en.json
+++ b/web/src/lib/i18n/en.json
@@ -685,7 +685,8 @@
     "unable_to_update_settings": "Unable to update settings",
     "unable_to_update_timeline_display_status": "Unable to update timeline display status",
     "unable_to_update_user": "Unable to update user",
-    "unable_to_upload_file": "Unable to upload file"
+    "unable_to_upload_file": "Unable to upload file",
+    "unable_to_write_to_clipboard": "Unable to write to clipboard"
   },
   "exif": "Exif",
   "exit_slideshow": "Exit Slideshow",
--- a/web/src/lib/i18n/en.json
+++ b/web/src/lib/i18n/en.json
@@ -685,8 +685,7 @@
     "unable_to_update_settings": "Unable to update settings",
     "unable_to_update_timeline_display_status": "Unable to update timeline display status",
     "unable_to_update_user": "Unable to update user",
-    "unable_to_upload_file": "Unable to upload file",
-    "unable_to_write_to_clipboard": "Unable to write to clipboard"
+    "unable_to_upload_file": "Unable to upload file"
   },
   "exif": "Exif",
   "exit_slideshow": "Exit Slideshow",
--- a/web/src/lib/utils/asset-utils.ts
+++ b/web/src/lib/utils/asset-utils.ts
@@ -556,7 +556,7 @@ const HtmlImageElementToBlob = async (imageElement: HTMLImageElement) => {
   throw new Error('Canvas context is null');
 };
 
-const ImageSourceToBlob = async (imageSource: string) => {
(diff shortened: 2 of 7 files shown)


web/src/lib/i18n/en.json
```diff
@@ -685,7 +685,8 @@
     "unable_to_update_settings": "Unable to update settings",
     "unable_to_update_timeline_display_status": "Unable to update timeline display status",
     "unable_to_update_user": "Unable to update user",
-    "unable_to_upload_file": "Unable to upload file"
+    "unable_to_upload_file": "Unable to upload file",
+    "unable_to_write_to_clipboard": "Unable to write to clipboard"
```
jrasm91, line 689: Let's use the existing `copy_error` which is used the other times clipboard stuff fails.

web/src/lib/utils/asset-utils.ts
```diff
@@ -527,3 +527,49 @@ export const archiveAssets = async (assets: AssetResponseDto[], archive: boolean
...
+          throw new Error('Canvas conversion to Blob failed');
+        }
+      });
+    });
+  }
+
+  throw new Error('Canvas context is null');
+};
+
+const ImageSourceToBlob = async (imageSource: string) => {
```
jrasm91, line 559:
```suggestion
const urlToBlob = async (imageSource: string) => {
```

web/src/lib/utils/asset-utils.ts
```diff
@@ -527,3 +527,49 @@ export const archiveAssets = async (assets: AssetResponseDto[], archive: boolean
 export const delay = async (ms: number) => {
   return new Promise((resolve) => setTimeout(resolve, ms));
 };
+
+export const canCopyImageToClipboard = (): boolean => {
+  return !!(navigator.clipboard && window.ClipboardItem);
+};
+
+const HtmlImageElementToBlob = async (imageElement: HTMLImageElement) => {
```
jrasm91, line 535:
```suggestion
const imgToBlob = async (imageElement: HTMLImageElement) => {
```

jrasm91 (commented):
Nice!

jrasm91 (approved):

jrasm91:
I tested this an confirmed copying images still works as expected.

web/src/lib/utils/asset-utils.ts
```diff
@@ -527,3 +527,41 @@ export const archiveAssets = async (assets: AssetResponseDto[], archive: boolean
...
+export const canCopyImageToClipboard = (): boolean => {
+  return !!(navigator.clipboard && window.ClipboardItem);
+};
+
+const imgToBlob = async (imageElement: HTMLImageElement) => {
+  const canvas = document.createElement('canvas');
+  const context = canvas.getContext('2d');
+
+  canvas.width = imageElement.width;
+  canvas.height = imageElement.height;
```
michelheusschen, line 540: This doesn't seem right. It just takes the dimensions of the image element, but we should be using the dimensions of the original image.

martabal: This how it currently works. It always copy the image shown in the asset viewer. Do you think we should always copy the original asset (even if the asset is incompatible with the browser) ?

michelheusschen: This isn't about thumbnail vs original, but about the image dimensions of the copied image. The dimensions of the asset shown in the photo viewer are used for the canvas, so if you resize your browser the copied image will have different dimensions.

martabal: Oh sorry for the confusion, it should be fixed

PR #12401, opened 2024-09-06
feat(server): start up folder checks

### LINKED ISSUE #5804: Startup checks

jrasm91:
At some pre-checks before starting up the server:
- Fail to start the server if `.immich` files don't exist in each folder (encoded-video, thumbs, library, upload, profile)
- Fail to start the server if `.immich` cannot be written in each folder (same as above)
- Track in the database if `.immich` has been written yet, if it hasn't write them on start-up

Tested the server failed to start in these situations:
- After successful start up, remove volume mounts, and restart
- Run container as 1000 instead of root
- Run volume as read only

This should help avoid some of these common situations:
- Container starts up but the library is on a network mount that is unavailable
- Container starts up but is being run as a user that doesn't have permission to the files
- User moves folders on the host system and incorrectly adjusts the volume mount, resulting in the folder being in the wrong location internally

Closes #5804

### DIFF
--- a/server/src/entities/system-metadata.entity.ts
+++ b/server/src/entities/system-metadata.entity.ts
@@ -12,12 +12,14 @@ export class SystemMetadataEntity<T extends keyof SystemMetadata = SystemMetadat
 }
 
 export type VersionCheckMetadata = { checkedAt: string; releaseVersion: string };
+export type SystemFlags = { mountFiles: boolean };
 
 export interface SystemMetadata extends Record<SystemMetadataKey, Record<string, any>> {
-  [SystemMetadataKey.REVERSE_GEOCODING_STATE]: { lastUpdate?: string; lastImportFileName?: string };
-  [SystemMetadataKey.FACIAL_RECOGNITION_STATE]: { lastRun?: string };
   [SystemMetadataKey.ADMIN_ONBOARDING]: { isOnboarded: boolean };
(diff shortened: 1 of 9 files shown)


danieldietzler (approved):

NicholasFlamy:
> * Track in the database if `.immich` has been written yet, if it hasn't write them on start-up

So if for some reason someone changed their `/upload` mount (maybe from the main HDD to their cache SSD) and didn't copy the folder and deleted the old one, I would assume immich wouldn't start because the new `/upload` folder is empty. What would be the solution for this user? Create a file and name it `.immich`?

jrasm91:
This is specifically why the feature was added. If you didn't copy a folder it would imply you have stuff in the database referring to files that no longer exist on disk and that leads to all types of issues.

For the case that you chose not to copy a folder because it was empty, than yeah simply creating `.immich` inside the folder would fix the issue.

NicholasFlamy:
Alright, good to know. That makes sense.

PR #12413, opened 2024-09-06
feat(web): Add Partner Sharing Avatars to Timeline and Info Cards

x24git:
This Feature adds user avatars to the timeline for both partner sharing, and shared albums. 

### Timeline
The avatar uses the user's avatar image (if one is set) or the server's information is used to make an avatar using the user's assigned color and first letter of their name.
(In the screenshot below, the partner's icon is an Orange 'T'). Implements a simple store for user data to efficiently pull user data for the timelines.
image
### New Settings
In order to control this feature there is a new setting added in the sharing config (changed from partner sharing to generic sharing).
image

### Asset View
Additionally, there is improvements to the asset view to make it clear the source of the image (either shared, when in an album, or from partner sharing in the main timeline)
image

Is the companion feature for the same capability set on the mobile app #12089


x24git:
Implements Feature Request https://github.com/immich-app/immich/discussions/7608

mertalev:
You might want to rebase - the diff is looking pretty gnarly

x24git:
Cleaned up commit history and rebased on main. Thanks @mertalev

x24git:
Resolved all merge conflicts and test cases. Ready to merge

x24git:
Resolved all merge conflicts and test cases. Ready to merge

web/src/lib/components/assets/thumbnail/thumbnail.svelte
danieldietzler, line 1: What I'd do here is:
- get rid of the reactive `shareUser` variable
- have `getShareUser` return the result of the getUser... call
- use `{#await getShareUser() then shareUser}` inside the `if` in the HTML
That should make it easier to follow and cleaner imo

web/src/lib/i18n/en.json
```diff
@@ -1155,6 +1155,8 @@
   "show_slideshow_transition": "Show slideshow transition",
   "show_supporter_badge": "Supporter badge",
   "show_supporter_badge_description": "Show a supporter badge",
+  "show_user_thumbnails": "Show user thumbnails",
+  "show_user_thumbnails_description": "Show user avatars on timelinle for shared albums and partners",
```
danieldietzler, line 1159:
```suggestion
  "show_user_thumbnails_description": "Show user avatars on timeline for shared albums and partners",
```

web/src/lib/stores/users.store.ts
danieldietzler, line 1: I'm not sure if we need a complete store just for this. I'll let that to the web experts though

web/src/lib/components/assets/thumbnail/thumbnail.svelte
x24git: Nice insight. Much cleaner now!

web/src/lib/stores/users.store.ts
x24git: If there is another store that is worth integrating into, I don't think that would be an issue. We do need a store though, otherwise there would be API requests for User info for each thumbnail in the asset view.

aviv926:
Maybe it should be placed under Partner Sharing?
image

x24git:
> Maybe it should be placed under Partner Sharing? image

Agreed, updated the PR description with the new placement

aviv926:
> > Maybe it should be placed under Partner Sharing?
> 
> Agreed, updated the PR description with the new placement. Swapped out the Partner Sharing title to a generic Sharing configuration, and then moved partners into an accordion box. Allow for future sharing specific settings to be added in the future.
> image
> 

Looks good I like that, but it's not my job to decide so let's see what the opinion of the core development team is.

alextran1502:
Closing due to stale. We can look into this at a later time

PR #12418, opened 2024-09-07
fix(server): deprioritize METADATA_EXTRACTION to let LINK_LIVE_PHOTOS complete first

### LINKED ISSUE #12306: LINK_LIVE_PHOTOS jobs lost when clearing queue after METADATA_EXTRACTION job completes
### The bug

The `METADATA_EXTRACTION` job [sets metadataExtractedAt](https://github.com/immich-app/immich/blob/c3a8ddaaf2dfe1c968c10e202bebecb346b8ff86/server/src/services/metadata.service.ts#L251C5-L254C8). If one clears the `Extract Metadata` queue, the subsequent `LINK_LIVE_PHOTOS` job will not execute as they are appended to the queue [after the `METADATA_EXTRACTION` job](https://github.com/immich-app/immich/blob/c3a8ddaaf2dfe1c968c10e202bebecb346b8ff86/server/src/services/job.service.ts#L243C9-L243C93). A restart of `Extract Metadata` job on 'missing' assets does not pick up the linkage jobs but completely ignores them.

Background: I have been wondering why the `Extract Metadata` queue does not decrease in size and have been stopping, clearing, and restarting the queue multiple times as it looked like a bug. What actually happens is that for each processed `METADATA_EXTRACTION` job, a `LINK_LIVE_PHOTOS` job is appended to the *same* queue. Thus, on restart, the queue size decreased as `metadataExtractedAt` was set, but now the `LINK_LIVE_PHOTOS` jobs have not been run for a large part of my assets.

Suggested solutions:
a) Set metadataExtractedAt only after `LINK_LIVE_PHOTOS` has completed and prioritize the LINK_LIVE_PHOTOS job so that the queue size actually decreases instead of waiting for all the `METADATA_EXTRACTION` jobs to complete first.
b) Merge `LINK_LIVE_PHOTOS` job into `METADATA_EXTRACTION` job
c) Create a separate queue to be able to run the `LINK_LIVE_PHOTOS` jobs by themselves

ubuntu

### Version of Immich Server

v1.113.1

none

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
none
```

### Your .env content

```Shell
none
```

### Reproduction steps

Start Metadata Extraction on multiple missing assets, let it extract from 1 asset, cancel job, clear queue. Start Metadata Extraction again. LINK_LIVE_PHOTOS has not been run on the first extracted asset.

### Relevant log output

_No response_

### Additional information

_No response_

mPyKen:
Least intrusive change to partially fix https://github.com/immich-app/immich/issues/12306.

When stopping and clearing the METADATA_EXTRACTION queue midway, METADATA_EXTRACTION queue must be forcefully filled and run again to LINK_LIVE_PHOTOS as LINK_LIVE_PHOTOS jobs are processed after all the METADATA_EXTRACTION jobs. Hence, LINK_LIVE_PHOTOS jobs should be prioritized over METADATA_EXTRACTION.

This also fixes UX such that the queue size actually decreases on the job status page evenly over time.

### DIFF
--- a/server/src/repositories/job.repository.ts
+++ b/server/src/repositories/job.repository.ts
@@ -253,6 +253,9 @@ export class JobRepository implements IJobRepository {
       case JobName.STORAGE_TEMPLATE_MIGRATION_SINGLE: {
         return { jobId: item.data.id };
       }
+      case JobName.METADATA_EXTRACTION: {
+        return { priority: 1 };
+      }
       case JobName.GENERATE_PERSON_THUMBNAIL: {
         return { priority: 1 };
       }


mertalev:
Hi! Thank you for the PR, but we can't merge this as it is. Metadata extraction is required for most other jobs, so deprioritizing it is not a good idea.

The situation you're describing can be solved in other ways instead, such as by tracking live photo linking separately from metadata extraction in the job status table.

jrasm91:
Closing per merts comment.

mertalev:
The issue is that the jobs in the other queues can't run until metadata extraction is finished, so lowering its priority has a cascade effect on the job system.

mPyKen:
For `JobName.QUEUE_METADATA_EXTRACTION` Q, `JobName.LINK_LIVE_PHOTOS` L and `JobName.METADATA_EXTRACTION` M, we currently complete jobs of METADATA_EXTRACTION queue in this order:
`QMMMMMLLLLL`
whereas this PR changes this order to
`QMLMLMLMLML`
The first jobs in other queues will still start after the first M job. As LINK_LIVE_PHOTOS is just a couple of database queries, I thought it wouldn't affect the job system that much. Is that not the case?

mertalev:
I've seen cases where metadata extraction was the bottleneck and the other queues were waiting on it. I just think a different approach would accomplish the same result without that tradeoff.

PR #12419, opened 2024-09-07
feat(server): Shared albums content in search results

lukasdotcom:
In the map you can currently show assets that are in a shared album. This adds the same ability to smartsearch/metadatasearch, showing assets shared with you in a shared album, by checking the Shared Albums option in a search. Implemented for mobile and web.


server/src/utils/database.ts
```diff
@@ -68,7 +68,22 @@ export function searchAssetBuilder(
   builder.andWhere(_.omitBy(id, _.isUndefined));
 
   if (options.userIds) {
-    builder.andWhere(`${builder.alias}.ownerId IN (:...userIds)`, { userIds: options.userIds });
+    if (options.withSharedAlbums && !options.isNotInAlbum) {
+      builder
+        .leftJoin(`${builder.alias}.albums`, 'albums')
+        .leftJoin(`albums.albumUsers`, 'albumUsers')
```
mertalev, line 74: These are one-to-many relationships, so you can't do these joins without duplicating rows and messing with pagination. There are other ways to write it, like using `where exists()` or adding a CTE that you do an inner join on.

Lukasdotcom: thanks

jrasm91:
Thanks for working on this. Unfortunately, we're doing a feature freeze for all sharing related features until the development team has some time to discuss. For more details, see #12614.

Lukasdotcom:
@jrasm91 would that mean that #12317 should also be closed or is that one still okay.

jrasm91:
TBD. It is definitely a bug so we might merge it in anyways.

PR #12434, opened 2024-09-07
fix: update profile picture in the navigation-bar

martabal:
When setting a new profile picture, change automatically the profile picture from the navigation-bar without reloading the entire page.


web/src/lib/components/shared-components/user-avatar.svelte
```diff
@@ -13,6 +13,7 @@
     email: string;
     profileImagePath: string;
     avatarColor: UserAvatarColor;
+    updatedAt: string;
```
jrasm91, line 16: No I mean like send it from the server in the api response.

martabal: Sure. But can this information be used to know when a user update his private information (like email, password) ?

jrasm91: You think exposing updatedAt would be bad because it leaks that other information? 

If so, let's add a new column like profileChangedAt and set it when the profile picture changes.

martabal: > If so, let's add a new column like profileChangedAt and set it when the profile picture changes.

Yes, IMO that's the best solution. I'll add it next week

PR #12438, opened 2024-09-07
feat(web): move search options into a modal

ben-basten:
## Description

Moving the search options into a modal, rather than displaying them below the search bar. Benefits of this change include:

- Guaranteed space in all screen sizes for the filters
- Focus trapping, so users can never accidentally tab off the end of the filters
- Automatically return focus back to the button in the search bar when the modal is closed

## Screenshots

<details><summary>Wide screen width</summary>
<p>

filter-modal-desktop

</p>
</details> 

<details><summary>Mobile screen width</summary>
<p>

filter-modal-mobile

</p>
</details>


jrasm91 (approved):

alextran1502:
Thank you for the PR. One change I think would make the model look better is incorporating the scroll bar inside the modal and appearing below the curved corner. Please see the example below

image

PR #12489, opened 2024-09-09
feat(web): show timezone information in detail view

C-Otto:
before:
Thu, 23:45 GMT-03:00

after:
Thu, 23:45 Chile Summer Time (-03:00)

The timezone name is localized and can be quite long (e.g. "Chilenische Sommerzeit", "Nordamerikanische Ostküsten-Sommerzeit"). This might cause line breaks.

The offset is not included if the timezone name includes "UTC":

* Thu, 23:45 Chile Summer Time (-03:00)
* Thu, 23:45 UTC
* Thu, 23:45 UTC+2

image
image

### DIFF
--- a/web/src/lib/components/asset-viewer/detail-panel.svelte
+++ b/web/src/lib/components/asset-viewer/detail-panel.svelte
@@ -296,10 +296,13 @@
                     weekday: 'short',
                     hour: 'numeric',
                     minute: '2-digit',
-                    timeZoneName: 'longOffset',
                   },
                   { locale: $locale },
                 )}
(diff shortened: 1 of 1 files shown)


bo0tzz (approved):

jrasm91 (approved):

C-Otto:
I'm converting this to a draft. It works, but as discussed in Discord, the time zone name information cannot easily be changed. It is derived from GPS coordinates and, if some time zone is shown, there's the chance that it is not correct. I think we should only show correct or at least fixable information. As the time zone name is not important for the actual display (orderinging of assets), we could also ignore it and continue relying on (and showing) the time zone offset instead.

PR #12514, opened 2024-09-09
feat(web): manually link live photos

jrasm91:
Add an option to manually link live photos when two assets are selected (and one is a photo and the other a video) by adding the option "Link motion video" to the menu.

Screenshot from 2024-09-09 12-23-28

### DIFF
--- a/web/src/lib/components/photos-page/actions/link-live-photo-action.svelte
+++ b/web/src/lib/components/photos-page/actions/link-live-photo-action.svelte
@@ -1,0 +1,44 @@
+<script lang="ts">
+  import CircleIconButton from '$lib/components/elements/buttons/circle-icon-button.svelte';
+  import type { OnLink } from '$lib/utils/actions';
+  import { AssetTypeEnum, updateAsset } from '@immich/sdk';
+  import { mdiMotionPlayOutline, mdiTimerSand } from '@mdi/js';
+  import { t } from 'svelte-i18n';
+  import MenuOption from '../../shared-components/context-menu/menu-option.svelte';
+  import { getAssetControlContext } from '../asset-select-control-bar.svelte';
+
+  export let onLink: OnLink;
+  export let menuItem = false;
+
+  let loading = false;
+
+  const text = $t('link_motion_video');
+  const icon = mdiMotionPlayOutline;
+
+  const { clearSelect, getOwnedAssets } = getAssetControlContext();
+
+  const handleLink = async () => {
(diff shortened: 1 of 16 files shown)


web/src/lib/components/photos-page/actions/link-live-photo-action.svelte
```diff
@@ -0,0 +1,44 @@
...
+    clearSelect();
+    loading = false;
+  };
+</script>
+
+{#if menuItem}
+  <MenuOption {text} {icon} onClick={handleLink} />
+{/if}
+
+{#if !menuItem}
```
danieldietzler, line 38:
```suggestion
{:else}
```

danieldietzler (approved):

web/src/lib/components/photos-page/actions/link-live-photo-action.svelte
jrasm91: nty

pyccl:
Is this connecting when storing files or two files? Are there still two files when downloading? Is the newly downloaded file also a static image (separated from the video)?

jrasm91:
Live photos have 2 flavors android and iOS. Android is a single file with the video embedded in the jpeg. iOS is two separate assets that are linked together. This PR is adding the option to manually link 2 separate assets together to form an iOS version of a live photo. When that version is downloaded you get 2 separate assets.

waclaw66:
How to unlink when a mistake is made?
Video icon overlaps another icon, panorama in my case...

obrazek

jrasm91:
Lol I guess I should have figured this would be inevitable. Why in the world did you link a panorama to a video?

waclaw66:
> Lol I guess I should have figured this would be inevitable. Why in the world did you link a panorama to a video?

On purpose, testing edge cases ;)

NightMean:
Thanks for this! :) If I understand correctly, this links the file only for immich but does not use the Content Identifier metadata that Apple uses on iOS. With the Content Identifier you can import the photo back to Apple Photos and retain the Live Photo as well. There is a separate project which can does this but only on a MacOS. 
https://github.com/RhetTbull/makelive

Also, it might be worth taking a look at https://github.com/PetrVys/MotionPhoto2/ which can create single HEIC live photo file that is compatible in Google Photos/iOS and of course Immich.

koenhendriks:
Is there any automated way to do this? I've just imported ~800GB from a Google Photo's export and all the iOS live photos have become 2 separate files.

I really don't want to do this by hand 😔

Issue #9392, opened 2024-05-11
dev: linting newline after HTML elements


mmomjian:
### The bug

When running `npm run format:fix`, the following error is not caught.

```
<details>
<summary>Check if checksums are enabled</summary>
` ` `bash
```

Should be: 
```
<details>
<summary>Check if checksums are enabled</summary>

` ` `bash
```

Without a new line after the HTML blocks, the code block is not be rendered properly. Can we add a linting check to make sure that there is a new line after each HTML element?

jrasm91:
I tested this and it rendered correctly for me. If you have a specific example that doesn't work let me know and I can take another look.

mmomjian:
Still happens for me with the following snippet from the FAQ: 

```
### The immich_model-cache volume takes up a lot of space, what could>

If you installed several models and chose not to use some of them, it>

<details>
<summary>Steps</summary>
` ` `bash
```

I’m not sure to see if I’m using an old version of the NPM formatter?

mmomjian:
Or did you mean it renders the HTML ok without the new line?

jrasm91:
Can you attach of screenshot of the before/after with the difference highlighted because I am not sure what I am looking for.

mmomjian:
Looks like it works now #12547 , ty

mmomjian:
https://github.com/mdx-js/mdx/issues/2263

danieldietzler:
@mmomjian why did you open this again?

jrasm91:
Well, I'm just going to close this as wont-fix 😛

PR #12551, opened 2024-09-10
docs: details for windows users how to change docker volume

mmomjian:


mmomjian:
image
image
image

@jrasm91 I think this is the issue from #9392

jrasm91 (approved):

PR #12574, opened 2024-09-11
feat(web): unlink live photos

jrasm91:
Follow-up to #12514 by also allowing to _unlink_ a motion video. Hidden assets don't have thumbnails so one is queued when it becomes visible again. UX should improve once the timeline auto-refresh is re-enabled.

A few notes:
- Technically two live photos share the same motion video (editor photo + original photo). Unlinking always unhides the motion asset.
- There is no way to know if the selected asset is an android or iOS motion video so the option always shows up, even though the operation will error for android.

[Screencast from 09-11-2024 10:15:57 AM.webm](https://github.com/user-attachments/assets/07ec144a-af03-4827-8e6b-f2b4f30ec1f4)

image


web/src/lib/components/photos-page/actions/link-live-photo-action.svelte
```diff
@@ -1,44 +1,75 @@
...
   let loading = false;
 
-  const text = $t('link_motion_video');
-  const icon = mdiMotionPlayOutline;
+  $: text = unlink ? $t('unlink_motion_video') : $t('link_motion_video');
+  $: icon = unlink ? mdiLinkOff : mdiMotionPlayOutline;
 
   const { clearSelect, getOwnedAssets } = getAssetControlContext();
 
+  const onClick = () => (unlink ? handleUnlink() : handleLink());
```
danieldietzler, line 23:
```suggestion
  const onClick = unlink ? handleUnlink : handleLink;
```

danieldietzler (approved):

waclaw66:
There is a display bug after motion video unlink...
|before link|linked|unlinked|
|-|-|-|
|obrazek|obrazek|obrazek|

Unlinked video is not shown, there is just some month artefact. Refresh of a page fixes that.

danieldietzler:
> There is a display bug after motion video unlink...
> before link 	linked 	unlinked
> obrazek 	obrazek 	obrazek
> 
> Unlinked video is not shown, there is just some month artefact. Refresh of a page fixes that.

That bug is unrelated to this feature. We currently don't update the timeline on websocket events anymore (due to performance issues), so any change that impacts assets in the timeline requires a reload

PR #12593, opened 2024-09-12
docs: scaling immich guide

bo0tzz:
This isn't completely a guide because (as addressed) the details will be specific to each deployment, but it should at least cover most questions.


danieldietzler (approved):

alextran1502 (approved):
LGTM, thanks

PR #12643, opened 2024-09-13
docs: add hint for healthcheck.start_interval

kastl-ars:
Add a hint to comment out the `healthcheck.start_interval` setting when running with (rootless) Podman and Docker compose.

(https://github.com/immich-app/immich/discussions/12638)

### DIFF
--- a/docs/docs/install/docker-compose.mdx
+++ b/docs/docs/install/docker-compose.mdx
@@ -80,6 +80,10 @@ The Compose file './docker-compose.yml' is invalid because:
 See the previous paragraph about installing from the official docker repository.
 :::
 
+:::info Healthcheck start_interval
+If you get an error `can't set healthcheck.start_interval as feature require Docker Engine v25 or later`, it helps to comment out the line for `start_interval` in the PostgreSQL section of the `docker-compose.yml` file.
+:::
+
 :::tip
 For more information on how to use the application, please refer to the [Post Installation](/docs/install/post-install.mdx) guide.
 :::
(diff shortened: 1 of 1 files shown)


mmomjian:
So I think this is actually due to the old version of docker compose, not podman, can we make the tip a bit more generic to the error being an old version?

kastl-ars:
> So I think this is actually due to the old version of docker compose, not podman, can we make the tip a bit more generic to the error being an old version?

Sure. I only encountered it with Podman, I did not check with docker.

In my case, I am using docker-compose-2.29.2-2.1.x86_64, which is the latest version.

kastl-ars:
> So I think this is actually due to the old version of docker compose, not podman, can we make the tip a bit more generic to the error being an old version?

Done.

mmomjian (approved):

docs/docs/install/docker-compose.mdx
```diff
@@ -80,6 +80,10 @@ The Compose file './docker-compose.yml' is invalid because:
 See the previous paragraph about installing from the official docker repository.
 :::
 
+:::info Healthcheck start_interval
+If you get an error `can't set healthcheck.start_interval as feature require Docker Engine v25 or later`, it helps to comment out the line for `start_interval` in the PostgreSQL section of the `docker-compose.yml` file.
```
mertalev, line 84:
```suggestion
:::info Health check start interval
If you get an error `can't set healthcheck.start_interval as feature require Docker Engine v25 or later`, it helps to comment out the line for `start_interval` in the `database` section of the `docker-compose.yml` file.
```

danieldietzler (approved):
With Mert's suggestion this looks good to me

jrasm91 (approved):

Issue #10415, opened 2024-06-17
Deleting all photos for a date in memory lane causes page to freeze


ben-basten:
### The bug

Deleting all photos for a date in memory lane causes the page to freeze, with an error in the browser console. This bug only occurs when all photos for a date are deleted. A page refresh is required to resume using immich.

### The OS that Immich Server is running on

Demo instance

### Version of Immich Server

v1.106.4

### Version of Immich Mobile App

n/a

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
N/A - demo instance
```

### Your .env content

```Shell
N/A - demo instance
```

### Reproduction steps

```bash
1. Open a memory lane memory
2. Scroll down to see the asset grid with all photos for that date
3. Select all photos in the asset grid, using the checkboxes
4. Delete the photos using the ellipses menu in the top right of the screen
```

### Relevant log output

```shell
Uncaught (in promise) TypeError: m[18] is undefined
    Immutable 43
11.BOQw4eMJ.js:1:9630
```

jrasm91:
Fixed in #12649

PR #12649, opened 2024-09-13
fix(web): memory viewer

### LINKED ISSUE #8897: Navigating using arrow keys in a gallery under memories switches view back to the timeline
### The bug

I'm unable to switch between pictures using arrow keys in the photo galleries that are displayed when you scroll down under the memories.

Raspberry OS Bookworm

### Version of Immich Server

v1.101.0

irrelevant

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
I don't think this is relevant.
```

### Your .env content

```Shell
I don't think this is relevant.
```

### Reproduction steps

```bash
0. You have to have multiple memories displayed on the top of the timeline
1. Open any memory
2. Scroll down to see the photos of the memory (can be only single one too!)
3. Open a random photo
4. Press left or right arrow
5. Suddenly, you're back on the main timeline
```

### Relevant log output

_No response_

### Additional information

No logs in the console.
When you are in a gallery of the first memory, the left arrow works okay; for the last memory, the right arrow works okay.
I suppose it probably is a conflict between binding keys for the memory view and this gallery view.

jrasm91:
Fix a few bugs with the memory viewer component.

- Arrow keys work on the asset viewer (fixes #8897)
- You can delete the last asset in a memory without a null pointer exception
- Archiving photos from a memory remove them from the memory
- Change the url to only store/track the current assetId

### DIFF
--- a/web/src/lib/components/memory-page/memory-viewer.svelte
+++ b/web/src/lib/components/memory-page/memory-viewer.svelte
@@ -1,8 +1,9 @@
 <script lang="ts">
-  import { goto } from '$app/navigation';
+  import { afterNavigate, goto } from '$app/navigation';
   import { page } from '$app/stores';
+  import { intersectionObserver } from '$lib/actions/intersection-observer';
+  import { resizeObserver } from '$lib/actions/resize-observer';
   import { shortcuts } from '$lib/actions/shortcut';
-
   import CircleIconButton from '$lib/components/elements/buttons/circle-icon-button.svelte';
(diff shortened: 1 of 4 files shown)


danieldietzler (approved):
LGTM

alextran1502 (approved):
Thanks for fixing the navigation in asset viewer

ben-basten:
Does this fix #10415 too?

PR #12712, opened 2024-09-15
feat(server): parse offset from "Image_UTC_Data" (Samsung)

C-Otto:
A Samsung phone might provide the local time (e.g. 09:00) without any timezone or offset information. If the file also includes the non-standard trailer tag "TimeStamp" in "Image_UTC_Data", we can use the unix timestamp contained within to deduce the offset.

As an example, if the local date/time is "2024-09-15T09:00" and the unix timestamp is 1726408800 (which is 2024-09-15T16:00 UTC), we know that the offset is -07:00.

Also see
https://github.com/exiftool/exiftool/blob/0f63a780906abcccba796761fc2e66a0737e2f16/lib/Image/ExifTool/Samsung.pm#L996-L1001

### DIFF
--- a/server/src/services/metadata.service.spec.ts
+++ b/server/src/services/metadata.service.spec.ts
@@ -1,4 +1,5 @@
 import { BinaryField, ExifDateTime } from 'exiftool-vendored';
+import { DateTime } from 'luxon';
 import { randomBytes } from 'node:crypto';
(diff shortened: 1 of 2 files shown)


jrasm91:
Closing in favor of upstream fix.

C-Otto:
Fixed in #13080

PR #12724, opened 2024-09-16
feat: use template migration to migrate external library files

rhyst:
POC of using template migration to migrate an external library into the main immich upload library.

I tested this locally with ~50 files and it seemed to work but will wait for confirmation that this is useful and/or doing things in an acceptable way before continuing with it.

From this discussion: https://github.com/immich-app/immich/discussions/12562

### DIFF
--- a/open-api/immich-openapi-specs.json
+++ b/open-api/immich-openapi-specs.json
@@ -2814,6 +2814,41 @@
         ]
       }
     },
+    "/libraries/{id}/migrate": {
+      "post": {
+        "operationId": "migrateLibrary",
+        "parameters": [
+          {
+            "name": "id",
+            "required": true,
+            "in": "path",
+            "schema": {
+              "format": "uuid",
+              "type": "string"
+            }
+          }
+        ],
+        "responses": {
+          "204": {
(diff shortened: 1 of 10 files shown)


zackpollard:
Hey, thanks for the PR. Unfortunately external libraries are currently under a feature freeze. This is due to us making vast changes to how they work with the rest of the codebase, trying to align them with how the default upload library works to reduce extra complexity across the codebase. I think this is something we may want in the future, once that work is complete, however right now we wouldn't be looking to merge something like this into Immich. Hopefully this doesn't put you off contributing to Immich in future, if you want feedback on implementing things before writing up a PR, you can join our [discord](https://discord.immich.app) or create a github discussion :slightly_smiling_face:

etnoy:
And adding to Zack's discussion, we might be able to revisit the PR and these changes after the feature freeze is over.

Again, thanks for your PR.

rhyst:
No problem, I did not invest much time yet (thanks to the immich team for making the codebase easy to work with!). I will try to remember to revisit this if I see the relevant updates in the changelogs.

PR #12741, opened 2024-09-17
feat(server): sort assets randomly from the API 'api/search/metadata' endpoint by including 'order': 'rand' in the API call.

jschwalbe:
> [!WARNING]
> `GET /api/assets/random` is now deprecated in favor of `POST /api/search/random`

Goal is to simplify getting random photos for a kiosk. The new API call works as such:

```
curl -L "http://localhost:2283/api/search/random" \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json' \
  -H "x-api-key: $API_KEY"  \
  -d  '{ "size": 100 }'
```

### DIFF
--- a/server/src/interfaces/search.interface.ts
+++ b/server/src/interfaces/search.interface.ts
@@ -116,6 +116,7 @@ export interface SearchPeopleOptions {
 
 export interface SearchOrderOptions {
   orderDirection?: 'ASC' | 'DESC';
+  random?: boolean;
 }
 
 export interface SearchPaginationOptions {
--- a/server/src/repositories/search.repository.ts
+++ b/server/src/repositories/search.repository.ts
@@ -73,8 +73,13 @@ export class SearchRepository implements ISearchRepository {
   async searchMetadata(pagination: SearchPaginationOptions, options: AssetSearchOptions): Paginated<AssetEntity> {
     let builder = this.assetRepository.createQueryBuilder('asset');
     builder = searchAssetBuilder(builder, options);
-
     builder.orderBy('asset.fileCreatedAt', options.orderDirection ?? 'DESC');
+
+    if (options.random) {
+      // TODO replace with complicated SQL magic after kysely migration
+      builder.addSelect('RANDOM() as r').orderBy('r');
+    }
+
     return paginatedBuilder<AssetEntity>(builder, {
       mode: PaginationMode.SKIP_TAKE,
       skip: (pagination.page - 1) * pagination.size,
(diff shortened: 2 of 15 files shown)


server/src/repositories/asset.repository.ts
```diff
@@ -219,7 +219,11 @@ export class AssetRepository implements IAssetRepository {
   getAll(pagination: PaginationOptions, options: AssetSearchOptions = {}): Paginated<AssetEntity> {
     let builder = this.repository.createQueryBuilder('asset').leftJoinAndSelect('asset.files', 'files');
     builder = searchAssetBuilder(builder, options);
-    builder.orderBy('asset.createdAt', options.orderDirection ?? 'ASC');
+    if (options.orderDirection === 'RAND') {
+      builder.addOrderBy('RANDOM()');
```
jrasm91, line 223: I think this causes a full table scan unless I'm mistaken.

mertalev: It does, yes. The "official" fast way to do it is to use `TABLESAMPLE`, but it doesn't handle filters well since they apply *after* it fetches a row (or rows).

I think you have to get crafty to make it fast and work well with filters, like taking advantage of the id or checksum column's inherent randomization and comparing a random uuid or hash against the table.

server/src/interfaces/search.interface.ts
```diff
@@ -114,7 +114,7 @@ export interface SearchPeopleOptions {
 }
 
 export interface SearchOrderOptions {
-  orderDirection?: 'ASC' | 'DESC';
+  orderDirection?: 'ASC' | 'DESC' | 'RAND';
```
mertalev, line 117: I think an API along the lines of `order: { direction: 'ASC' | 'DESC', field: 'fileCreatedAt' | 'random' } would be nicer and more extensible with other sort options in the future.

jschwalbe: I didn't want to make sweeping changes as this would seem to change quite a few things. Any suggestions about how to go forward with this? Thanks

mertalev: You can change references and types for `orderDirection: <dir>` to `order: { direction: <dir> }` style and have the `field` property be specific to `searchMetadata`.

danieldietzler (approved):

jrasm91 (approved):

server/src/repositories/search.repository.ts
```diff
@@ -73,10 +73,17 @@ export class SearchRepository implements ISearchRepository {
   async searchMetadata(pagination: SearchPaginationOptions, options: AssetSearchOptions): Paginated<AssetEntity> {
     let builder = this.assetRepository.createQueryBuilder('asset');
     builder = searchAssetBuilder(builder, options);
-
+    let mode = PaginationMode.SKIP_TAKE;
     builder.orderBy('asset.fileCreatedAt', options.orderDirection ?? 'DESC');
+
+    if (options.random) {
+      mode = PaginationMode.LIMIT_OFFSET;
```
mertalev, line 80: I don't think limit/offset is safe here because there are possible joins with one-to-many or many-to-many relations.

mertalev: A thought: you could make a separate `/search/random` endpoint that takes the same options as the smart search endpoint but without the query field.

jrasm91: I see. The other pagination throws an error I did not know how to fix. If you have time can you take a look?

jrasm91: Nevermind, I think I figured it out.

danieldietzler (approved):
LGTM

mertalev (approved):

PR #12742, opened 2024-09-17
fix(web): modal scrolling with dropdown menus

ben-basten:
## Description

Fix the scrolling container in the modals so content can never appear below the bottom buttons. 

This was done by pulling the bottom buttons outside of the scrolling container.

**Also in this change:**

- Fix the "Edit date and time" + "Tag assets" modals being obscured in Safari. Applied the same fix as #9756.
- Expand the link text in the "Tag assets" modal to include more context than just "here".

**For a follow-up effort**

Using fixed positioning for the combobox dropdown menu, so it will appear above all other content on the page and be able to expand outside of the modal for easier use. I'm thinking this can be built similar to the context menu.

<details><summary>v1.113.0 modal screenshots for reference</summary>
<p>

Dropdown menus were never previously able to expand outside of a modal. In some cases it would appear above other content, and in other cases it would be hidden behind.

Screenshot 2024-09-16 at 07 17 14

Screenshot 2024-09-16 at 07 19 12

</p>
</details> 

## How Has This Been Tested?

- [x] Open all modals in the asset detail sidebar in Safari on Mac and iPad to verify that they're fully visible.
- [x] Try dropdown menus in the "Filter options", "Edit date and time", and "Tag assets" modals to make sure scrolling is behaving as expected.

## Screenshots

| Before | After |
| ------ | ----- |
| bug | tag-assets-modal |

<details><summary>Filter options modal</summary>
<p>

filter-options-modal

</p>
</details>


jrasm91:
The drop-down in the after screenshot looks pretty unable. It only shows a single item, is that correct?

ben-basten:
> The drop-down in the after screenshot looks pretty unusable. It only shows a single item, is that correct?

Yeah, I was trying to separate the bugfix from adding the new positioning feature. I can close this and bundle them both together.

PR #12755, opened 2024-09-17
refactor(server): view repository

jrasm91:
- Move into new repository
- Stop using `getBuilder`


danieldietzler (approved):

mertalev (approved):
Nit: the "view" naming is too generic. It doesn't really describe what the service/repository is about. `FolderRepository` or `FolderViewRepository` would be more descriptive.

Issue #11833, opened 2024-08-15
Navigation resets due to slow loading asset


C-Otto:
### The bug

When I quickly move from asset A to asset B and then quickly back to asset A (by pressing "right" immediately followed by "left"), asset A is shown and then replaced by asset B (after a short delay). I assume that this is related to the delay of loading the original file for asset B.

Instead of showing asset B, the related request should be aborted or silently ignored once completed.

I think this also happens when navigating back or pressing escape and is quite confusing.

### The OS that Immich Server is running on

Debian Stable

### Version of Immich Server

v1.112.1

### Version of Immich Mobile App

none

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Reproduction steps

1. Navigate to album
2. Press "right" immediately followed by "left"

danieldietzler:
Is this with the recent improvements to the asset viewer still an issue?

C-Otto:
Yes, I don't see any change.

PR #12765, opened 2024-09-18
fix(web): image preload and cancelation

midzelis:
This is the beginning of a multi-part set of PRs, each of which will stand on its own, mostly because it will be easier to review, and understand smaller changes, and its generally a best practice.  

In this PR, the base functionality for image preload and cancelation is fixed, previously broken in #9684. 

Additionally, video loads can also be canceled while they are still being buffered by pausing the video, and clearing the src attribute. 

Also, the memory-lane component has been retrofitted to use image-thumbnail instead of bare images. In general, all thumbnails should be loaded via the image-thumbnail (or thumbnail) components. 

The main part of this PR is the introduction of an image-manager component - it orchestrates the loading/cancelation of images. 

A future PR will ensure that all images are loaded via the image-manager component.


C-Otto:
Just a hunch, this might be related to #11833

zackpollard:
Closing for now, will come back to it later with a new PR

PR #12779, opened 2024-09-18
fix(web): only show the timezone when it is known

jrasm91:
Instead of guessing a timezone and showing an obviously wrong date. Show the local date without a time zone. (notice the day could actually change before, depending on what client timezone you were in)

| Before | After |
| - | - |
| image | image |


danieldietzler (approved):

zackpollard (approved):

sammyke007:
It broke correct time settings for me. exiftool shows the correct time, Immich is configured with the correct TZ env. variable, but it still shows the time 2h off since v1.116.0

jrasm91:
> It broke correct time settings for me. exiftool shows the correct time, Immich is configured with the correct TZ env. variable, but it still shows the time 2h off since v1.116.0

Please post appropriate details here: 

https://github.com/immich-app/immich/discussions/12650

PR #12791, opened 2024-09-19
refactor(mobile): album api repository for album service

fyfrey:
- This PR adds a new repository to perform action with the Album API.
- The `AlbumApiRepository` is used in the AlbumService that no longer depends on OpenAPI! :-)
- replaces usage of `AlbumResponseDto` throughout AlbumService and SyncService with `Album` (needed a few nasty tricks to get there, sadly)
- finally add some unit tests for AlbumService methods (now that API access can me easily mocked)
- new linting checking openapi usage (limit to all files still using it as of this PR)

### DIFF
--- a/mobile/lib/interfaces/album_api.interface.dart
+++ b/mobile/lib/interfaces/album_api.interface.dart
@@ -1,0 +1,40 @@
+import 'package:immich_mobile/entities/album.entity.dart';
+
+abstract interface class IAlbumApiRepository {
+  Future<Album> get(String id);
+
+  Future<List<Album>> getAll({bool? shared});
+
+  Future<Album> create(
+    String name, {
+    required Iterable<String> assetIds,
+    Iterable<String> sharedUserIds = const [],
+  });
+
+  Future<({List<String> added, List<String> duplicates})> addAssets(
+    String albumId,
+    Iterable<String> assetIds,
+  );
+
+  Future<({List<String> removed, List<String> failed})> removeAssets(
+    String albumId,
+    Iterable<String> assetIds,
+  );
+
+  Future<Album> addUsers(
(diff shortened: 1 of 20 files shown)


mobile/lib/interfaces/album_api.interface.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:immich_mobile/entities/album.entity.dart';
+
+abstract interface class IAlbumApiRepository {
```
jrasm91, line 3: Ideally, this eventually is just `IAlbumRepository`, right?

mobile/lib/interfaces/album_api.interface.dart
```diff
@@ -0,0 +1,40 @@
...
+  Future<Album> get(String id);
```
jrasm91, line 4: Do you mind trying to keep the order consistent in all the repos? I usually prefer something like: CRUD, then relations, then misc stuff. Here it would be
- getAll
- create
- update
- delete
- addAssets
- removeAssets
- addUsers
- removeUser

jrasm91 (approved):

mobile/lib/interfaces/album_api.interface.dart
fyfrey: Not sure about that.

We now have
* 1 repository to access Albums in the local DB (AlbumRepo),
* 1 repository to access albums on the device via platform calls (AlbumMediaRepo)
* 1 repository to access albums on the server via open API (AlbumApiRepo)

jrasm91: Right. The API one might go away depending on how the sync process works long term. Either way this send fine for now.

Issue #10208, opened 2024-06-12
The photo is displayed in the wrong direction


PolarisGitYHNL:
### The bug

The photo is displayed in the wrong direction.
The photo source file is a vertical photo, and it displays normally in both the Synology album and MT Photos (another Docker based album), as shown in the following figure:
<img width="1916" alt="image" src="https://github.com/immich-app/immich/assets/165972224/9f412f9e-2ef6-4d01-8676-bae686837fcf">
<img width="1919" alt="image" src="https://github.com/immich-app/immich/assets/165972224/362457e6-4d49-4a64-8278-d75073f5d774">
In IMMICH, the shape of the thumbnail is normal and displayed as a vertical version, as shown in the following figure:
<img width="1919" alt="image" src="https://github.com/immich-app/immich/assets/165972224/b57854a5-3908-4cd7-a210-05497c2333d8">
However, the images and photos in the thumbnail are folded down to display horizontally, as shown in the following figure:
<img width="1919" alt="image" src="https://github.com/immich-app/immich/assets/165972224/313256d7-f51c-4b42-95c1-33e44f780d20">
The same applies to displaying errors in mobile apps:
![image](https://github.com/immich-app/immich/assets/165972224/346ec6a5-b9cd-4848-a6be-68b8d200512b)

### The OS that Immich Server is running on

Sinology DSM docker

### Version of Immich Server

v1.106.2

### Version of Immich Mobile App

v1.106.1

### Platform with the issue

- [X] Server
- [X] Web
- [X] Mobile

### Your docker-compose.yml content

```YAML
Perhaps the problem is unrelated to this
```

### Your .env content

```Shell
Perhaps the problem is unrelated to this
```

### Reproduction steps

```bash
None
```

bo0tzz:
Does the EXIF of the files have any rotation metadata?

Snowknight26:
libraw probably needs an update.

Potential workaround: https://github.com/immich-app/immich/issues/9845#issuecomment-2145840833

PolarisGitYHNL:
> Does the EXIF of the files have any rotation metadata?
YES
One line in EXIF:
Orientation Rotate 270 CW

PolarisGitYHNL:
sorry
my mistake(I have reopened it)

jrasm91:
Can you refresh metadata and then refresh thumbnails for this picture and let me know if it is still orientated incorrectly?

[...]

PR #12792, opened 2024-09-19
refactor: orientation

### LINKED ISSUE #3022: Orientation enum
Based on discussion in #3003.

  * Since `exiftool-vendored` disables the print conversion for the Orientation tag, I believe that guarantees we can get only `1-8` values for the orientation of photos.
  * `ffprobe` returns other values for the orientation (`-90`, `90`, etc.)

We can change the type of the orientation field in `ExifEntity` and `ExifResponseDto` to enum to provide a better experience for all the clients.

jrasm91:
Fixes #3022

Use an enum for exif orientation, including sending it back in the asset response dtos.

### DIFF
--- a/server/src/entities/exif.entity.ts
+++ b/server/src/entities/exif.entity.ts
@@ -1,4 +1,5 @@
 import { AssetEntity } from 'src/entities/asset.entity';
+import { Orientation } from 'src/enum';
 import { Index, JoinColumn, OneToOne, PrimaryColumn } from 'typeorm';
 import { Column } from 'typeorm/decorator/columns/Column.js';
 import { Entity } from 'typeorm/decorator/entity/Entity.js';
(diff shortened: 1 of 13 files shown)


danieldietzler (approved):

danieldietzler:
Might also fix #10208?

server/src/entities/exif.entity.ts
```diff
@@ -25,8 +26,8 @@ export class ExifEntity {
   @Column({ type: 'bigint', nullable: true })
   fileSizeInByte!: number | null;
 
-  @Column({ type: 'varchar', nullable: true })
-  orientation!: string | null;
+  @Column({ type: 'enum', enum: Orientation, nullable: true })
```
mertalev, line 29: For this case, the keys are basically already enum values. It'd be nice to make this a `"char"` (not `char`) and use a check constraint that it's between 1-8. It's 1 byte instead of 4 and doesn't have the string vs number ambiguity that an enum would have.

PR #12796, opened 2024-09-19
chore(web): remove stray dateTimeOriginal reference

jrasm91:
Basically:
- `asset.fileCreatedAt` should only ever be used for sorting within a bucket
- `dateTimeOriginal` should only be used for displaying in the detail view when there is a timezone, and we can probably deduce this from `localDateTime` itself in the future
- `localDateTime` should be used for everything else


bo0tzz (approved):
We should probably document these distinctions somewhere

PR #12812, opened 2024-09-20
fix: rework file handling so we always explicitly create, overwrite or both

zackpollard:
Due to https://github.com/nodejs/node/issues/41093 we are unable to open hidden files with the 'w' flag on windows and some samba shares. This got me thinking, and realistically we should never actually need to open files with the w flag anyway. In all contexts within immich we know if we are opening a file for creating or overwriting, so we can explicitly use those actions in those contexts. This means we now have more protections when creating file, as if somehow a file exists on the filesystem, but we didn't expect it to, the code will error rather than overwrite it.


danieldietzler (approved):

jrasm91 (approved):
While you are here, do you want to change the messages to use UPLOAD_LOCATION instead of MEDIA_LOCATION?

alextran1502 (approved):

zackpollard:
> While you are here, do you want to change the messages to use UPLOAD_LOCATION instead of MEDIA_LOCATION?

Done 😄

PR #12813, opened 2024-09-20
docs: add section for Traefik Reverse Proxy

KuramaSyu:
This adds a section for an example Traefik Proxy config. This is needed because the readingTimeouts leed to video-uploads failing after 60 seconds uploading. See [here](https://github.com/immich-app/immich/discussions/12564) for more information. In the example the issue is "resolved" by increasing the default timeout to 600 seconds instead of 60 seconds.

### DIFF
--- a/docs/docs/administration/reverse-proxy.md
+++ b/docs/docs/administration/reverse-proxy.md
@@ -80,9 +80,8 @@ entryPoints:
     # this section needs to be added
     transport:
       respondingTimeouts:
-        readTimeout: 600s
-        writeTimeout: 600s
-        idleTimeout: 600s
+        readTimeout: 0
+        idleTimeout: 0
 ```
 
 The second part is in the `docker-compose.yml` file where immich is in. Add the Traefik specific labels like in the example.
--- a/docs/docs/administration/reverse-proxy.md
+++ b/docs/docs/administration/reverse-proxy.md
@@ -64,3 +64,73 @@ Below is an example config for Apache2 site configuration.
    ProxyPreserveHost On
 </VirtualHost>
 ```
+
+### Traefik Proxy example config
+
+The example below is for Traefik version 3.
+
+The most important is to increase the `respondingTimeouts` of the entrypoint used by immich. In this example of entrypoint `websecure` for port `443`. Per default it's set to 60s which leeds to videos stop uploading after 1 minute (Error Code 499). In the example it's set to 10 minutes. Hence video-uploads will fail after 10 minutes instead of one.
+
+`traefik.yaml`
+```yaml
+[...]
+entryPoints:
+  websecure:
+    address: :443
+    AsDefault: true
+    http:
+      tls:
+        certresolver: cloudflare-resolver
+        domains:
+          - main: "*.domain.com"
+            sans: 
+              - "domain.com"
+    # this section needs to be added
+    transport:
+      respondingTimeouts:
+        readTimeout: 600s
+        writeTimeout: 600s
+        idleTimeout: 600s
+[...]
+```
+
+The second part is in the `docker-compose.yml` file where immich is in. Add your external network, where traefik is in, inside the `networks` section and add all the traefik labels like in the example.
+
+`docker-compose.yml`
+```yaml
+[...]
+services:
+  immich-server:
+    container_name: immich_server
+    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
+    volumes:
+      - ${UPLOAD_LOCATION}:/usr/src/app/upload
+      - /etc/localtime:/etc/localtime:ro
+    env_file:
+      - .env
+    ports:
+      - 2283:3001
+    networks:
+      # this is the network where traefik is in
+      - proxy
+      # this is the network where all other immich containers are in
+      - default
+    depends_on:
+      - redis
+      - database
+    restart: always
+    healthcheck:
+      disable: false
+    labels:
+      traefik.enable: true
+      # increase readingTimeouts for the entrypoint used here
+      traefik.http.routers.immich.entrypoints: websecure  
+      traefik.http.routers.immich.rule: Host(`immich.your-domain.com`)
+      traefik.http.services.immich.loadbalancer.server.port: 3001
+[...]
+
+# add the external network where traefik is in
+networks:
+  proxy:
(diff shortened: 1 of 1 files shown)


bo0tzz (commented):
Please try to keep the config as minimal as possible, just like in the other examples.

Does this enable websockets as well?

docs/docs/administration/reverse-proxy.md
```diff
@@ -64,3 +64,73 @@ Below is an example config for Apache2 site configuration.
...
+  websecure:
+    address: :443
+    AsDefault: true
+    http:
+      tls:
+        certresolver: cloudflare-resolver
+        domains:
+          - main: "*.domain.com"
+            sans: 
+              - "domain.com"
```
bo0tzz, line 87: Is this section required? The other examples just show a plain http setup.

docs/docs/administration/reverse-proxy.md
```diff
@@ -64,3 +64,73 @@ Below is an example config for Apache2 site configuration.
...
+      - database
+    restart: always
+    healthcheck:
+      disable: false
+    labels:
+      traefik.enable: true
+      # increase readingTimeouts for the entrypoint used here
+      traefik.http.routers.immich.entrypoints: websecure  
+      traefik.http.routers.immich.rule: Host(`immich.your-domain.com`)
+      traefik.http.services.immich.loadbalancer.server.port: 3001
```
bo0tzz, line 129:
```suggestion
  immich-server:
    ...
    labels:
      traefik.enable: true
      # increase readingTimeouts for the entrypoint used here
      traefik.http.routers.immich.entrypoints: websecure  
      traefik.http.routers.immich.rule: Host(`immich.your-domain.com`)
      traefik.http.services.immich.loadbalancer.server.port: 3001
```
No need to reproduce the whole block here

zp33dy: I think at least the networks section where the network for traefik is added is needed. Otherwise it wont work.

zp33dy:
> Does this enable websockets as well?
As far as I know it does, if the websocket also uses port 443 what it should do.

docs/docs/administration/reverse-proxy.md
zp33dy: No, for http it's not required. I'll remove it

docs/docs/administration/reverse-proxy.md
mmomjian: Thanks for the PR. in general we don’t teach people how to set up the proxy themself, just the Immich specific stuff. The network setup would vary greatly between different setups so thats why it may not be best to include. Also there is no requirement to use separate networks, they can all be on one.

mmomjian: Maybe you can add a comment that says something like `# make sure that Traefik is able to communicate with the network of the Immich server`

docs/docs/administration/reverse-proxy.md
```diff
@@ -64,3 +64,52 @@ Below is an example config for Apache2 site configuration.
...
+[...]
+entryPoints:
+  websecure:
+    address: :443
+    # this section needs to be added
+    transport:
+      respondingTimeouts:
+        readTimeout: 600s
+        writeTimeout: 600s
+        idleTimeout: 600s
```
dvbthien, line 85:
```suggestion
    transport:
      respondingTimeouts:
        readTimeout: 0
        idleTimeout: 0
```

docs/docs/administration/reverse-proxy.md
zp33dy: I updated it by replaceing the networking stuff with a last information sentence. Is it fine now?

docs/docs/administration/reverse-proxy.md
dvbthien: I think respondingTimeouts need to change as above. Value 0 is no timeout exists

zp33dy: I commited it. But isn't it a bit of bad practice to never terminate a connection?

docs/docs/administration/reverse-proxy.md
```diff
@@ -64,3 +64,40 @@ Below is an example config for Apache2 site configuration.
...
+`traefik.yaml`
+```yaml
+[...]
+entryPoints:
+  websecure:
+    address: :443
+    # this section needs to be added
+    transport:
+      respondingTimeouts:
+        readTimeout: 0
```
mmomjian, line 83: Looks like set to 0 here, which is no timeout, not 10 minutes

Can we also add writeTimeout?

docs/docs/administration/reverse-proxy.md
mmomjian: I see this now. I think we should go back to 600s like the other proxy examples. I disagree with removing writeTimeout

docs/docs/administration/reverse-proxy.md
mmomjian: Hey, thanks for your comment. I don't agree with using a timeout of 0, all our other examples use 10 minutes which has worked well.

mmomjian: Hey, thanks for your comment. I don't agree with using a timeout of 0, all our other examples use 10 minutes which has worked well. Regardless of what we pick, it's never good to depend on defaults - we should specify all 3 values as the default can change in a later release.

zp33dy: > Would prefer 10 minutes like other examples

I agree. Same for idle timeout. I think it's bad practise to not terminate connections at any point. I'll change it back.

dvbthien: Thanks, I don't know of other examples using 10 minutes. To maintain consistency with other documents, may use 10 minutes. In my configuration, I use a value of 0 so i just suggest it

mmomjian (approved):

danieldietzler (approved):

danieldietzler:
@zp33dy could you fix the formatting issues?

zp33dy:
> @zp33dy could you fix the formatting issues?

Can I see somewhere what the issues are?

danieldietzler:
> > @zp33dy could you fix the formatting issues?
> 
> Can I see somewhere what the issues are?

Run `npm run format`

Issue #12839, opened 2024-09-22
Only 1 tag displayed in web display


djmccullers:
### The bug

On Immich v1.115.0, when using the "Tag" feature, 

![Immich_1](https://github.com/user-attachments/assets/89abaa5e-e62c-4a80-9555-d046956b0cc2)

only the top most tag is ever displayed.

![Immich_2](https://github.com/user-attachments/assets/7c04d383-90da-4000-ab28-dff3abe0c008)

### The OS that Immich Server is running on

Linux Mint 22.0

### Version of Immich Server

v1.115.0

### Version of Immich Mobile App

N/A

### Platform with the issue

- [X] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
name: immich
#      image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
#      image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-latest}
#      image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-next}
services:
  immich-server:
    container_name: immich_server
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-next}
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - /etc/localtime:/etc/localtime:ro
    env_file:
      - .env
    ports:
      - 2283:3001
    depends_on:
      - redis
      - database
    restart: always

  immich-machine-learning:
    container_name: immich_machine_learning
    image: ghcr.io/immich-app/immich-machine-learning:${IMMICH_VERSION:-next}
    volumes:
      - model-cache:/cache
    env_file:
      - .env
    restart: always

  redis:
    container_name: immich_redis
    image: docker.io/redis:6.2-alpine@sha256:d6c2911ac51b289db208767581a5d154544f2b2fe4914ea5056443f62dc6e900
    healthcheck:
      test: redis-cli ping || exit 1
    restart: always

  database:
    container_name: immich_postgres
    image: docker.io/tensorchord/pgvecto-rs:pg14-v0.2.0@sha256:90724186f0a3517cf6914295b5ab410db9ce23190a2d9d0b9dd6463e3fa298f0
    environment:
      POSTGRES_PASSWORD: ${DB_PASSWORD}
      POSTGRES_USER: ${DB_USERNAME}
      POSTGRES_DB: ${DB_DATABASE_NAME}
[...]

PR #12848, opened 2024-09-22
feat(web): fixed combobox positioning

ben-basten:
## Description

Dropdown options can appear above all other page content and expand outside of modals, making them easier to use. When there is limited space to show the dropdown menu below the text input, it will move to top positioning. This is similar to the solution proposed in #11810. 

The dropdown menu is automatically closed if the input box is hidden by another element.

**Also in this change:**

- Reworking the scroll container in full screen modals, to prevent an issue where the bottom buttons scroll with modal content incorrectly.
- Fix the "Edit date and time" + "Tag assets" modals being obscured in Safari. Applied the same fix as #9756.
- Expand the link text in the "Tag assets" modal to include more context than just "here".

## How Has This Been Tested?

**Testing matrix**

Above/below refer to the position of the dropdown menu, relative to the input box.

Due to poor results with the dropdown menu appearing "above" on mobile devices, it will only appear below when a mobile virtual keyboard or zoom is applied.

| OS/Browser           | Virtual keyboard open                                                                            | Pinch zoom in                                                     | Default zoom support |
| -------------------- | ------------------------------------------------------------------------------------------------ | ----------------------------------------------------------------- | -------------------- |
| iOS/Safari           | Below: ✅ Above: ❌ Slight zoom when an input field is clicked breaks positioning above          | Below: ✅ Above: ❌                                               | Below/Above: ✅      |
| iPadOS/Safari        | ✅                                                                                               | Below: ✅ Above: ❌                                               | Below/Above: ✅      |
| Android/Chrome       | 🟡 Works if dropdown is triggered in a position that is above where the virtual keyboard appears | Below/Above: ❌ Loses position tracking                           | Below/Above: ✅      |
| Android/Firefox      | ✅                                                                                               | Below/Above: ❌ Loses position tracking                           | Below/Above: ✅      |
| MacOS/Safari         | N/A                                                                                              | 🟡 Untested - do not have a trackpad to test pinch zoom in Safari | Below/Above: ✅      |
| MacOS/Chrome+Firefox | N/A                                                                                              | N/A                                                               | Below/Above: ✅      |

**Combobox locations**

- Search options: city, country, state, camera make, camera model
- App settings: language
- App settings: time locale
- Change date modal
- Tag asset modal
- Admin job status: create job modal

**Testing considerations**

- Dropdown with no scrolling options, above and below
- Dropdown with scrolling options, above and below
- Dropdown moves as the page scrolls, repositions to top if not enough room
- Dropdown moves as the page vertically resizes, repositions to top if not enough room

## Screenshots

https://github.com/user-attachments/assets/10561de9-90d8-40fe-8315-c850fab8fa03

<p>

create-job
edit-date-and-time
tag-assets

</p>
</details>


danieldietzler:
Fixes #12839

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -53,15 +53,45 @@
...
   }
 
+  $: position = calculatePosition(bounds);
+
+  onMount(() => {
+    observer.observe(input);
+    const scrollableAncestor = input.closest('.overflow-y-auto, .overflow-y-scroll');
+    scrollableAncestor?.addEventListener('scroll', onPositionChange);
+    window.visualViewport?.addEventListener('resize', onPositionChange);
+    window.visualViewport?.addEventListener('scroll', onPositionChange);
```
danieldietzler, line 85: Can't you apply the event listeners "svelte-style"? So on the respective component (or `svelte:window`) with `on:<event>` (e.g. `on:scroll`)?

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -116,8 +147,66 @@
     searchQuery = '';
     onSelect(selectedOption);
   };
+
+  const calculatePosition = (boundary: DOMRect | undefined) => {
+    dropdownDirection = getComboboxDirection(boundary);
+
+    if (!boundary) {
+      return undefined;
```
danieldietzler, line 155:
```suggestion
      return;
```

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -116,8 +147,66 @@
...
+    }
+
+    const vv = window.visualViewport;
```
danieldietzler, line 158: I get that `visualViewport` is a bit lengthy, but I'm not a fan of acronyms 😅

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -116,8 +147,66 @@
...
+      top: `${boundary.bottom + offsetTop}px`,
+      left: `${left}px`,
+      width: `${boundary.width}px`,
+      maxHeight: maxHeight(availableHeight - dropdownOffset),
+    };
+  };
+
+  const maxHeight = (size: number) => {
+    return `min(${size}px,18rem)`;
+  };
```
danieldietzler, line 183:
```suggestion
  const maxHeight = (size: number) => `min(${size}px,18rem)`;
```

web/src/lib/components/shared-components/combobox.svelte
```diff
@@ -116,8 +147,66 @@
...
+      return 'bottom';
+    }
+
+    const viewportHeight = window.visualViewport?.height || 0;
+    const heightBelow = viewportHeight - boundary.bottom;
+    const heightAbove = boundary.top;
+
+    const isViewportScaled = viewportHeight && Math.floor(viewportHeight) !== Math.floor(window.innerHeight);
+
+    return heightBelow <= 225 && heightAbove > heightBelow && !isViewportScaled ? 'top' : 'bottom';
```
danieldietzler, line 203: Where does this 225 come from?

web/src/lib/components/shared-components/combobox.svelte
ben-basten: Agreed, good call.

web/src/lib/components/shared-components/combobox.svelte
ben-basten: I don't think I can.

- `scrollableAncestor` - this contains the nearest ancestor component with a scrollbar and attaches a listener. This element is not inside of the combobox.
- `window.visualViewport` - the scroll and resize listeners specifically need to be attached to the `visualViewport`, to detect screen changes like a virtual keyboard appearing, or the browser app navigation appearing/disappearing as you scroll on a mobile phone. I don't think this can be done with `svelte:window`, but I'm open to ideas?

web/src/lib/components/shared-components/combobox.svelte
ben-basten: I pulled this number into a variable in the code to better explain what it does.

This is the hardcoded breakpoint to decide when the dropdown should be displayed above the input box. If the available height for the dropdown is less than 225px and there's more space above than below, the dropdown will be displayed above.

225 felt to me like the right balance between not switching to the top too early, and still being able to see enough of the options when the dropdown is positioned below for it to still be useful.

ben-basten: https://github.com/user-attachments/assets/01325906-1dbe-4c7b-930c-64a1addb8ffa

Here's a quick demo showing what that breakpoint looks like. 225 is the moment when the dropdown switches from below to above. 

Note that the same breakpoint applies to short dropdown menus that aren't scrolling at all - it doesn't depend on the height of the dropdown, but rather the height available.

web/src/lib/components/shared-components/combobox.svelte
danieldietzler: Ah looks like you're right, sorry.

web/src/lib/components/shared-components/combobox.svelte
danieldietzler: Awesome, thanks for clarifying it here and in the code! :))

danieldietzler (approved):
The code looks good to me!

jrasm91 (approved):

PR #12869, opened 2024-09-23
fix: web e2e

jrasm91:
Looks like web utils was resetting the database _before_ microservices was up and running. The system metadata > system flags can't be cleared before microservices does the folder mount checks.

### DIFF
--- a/e2e/src/utils.ts
+++ b/e2e/src/utils.ts
@@ -156,8 +156,7 @@ export const utils = {
 
       for (const table of tables) {
         if (table === 'system_metadata') {
-          // prevent reverse geocoder from being re-initialized
-          sql.push(`DELETE FROM "system_metadata" where "key" != 'reverse-geocoding-state';`);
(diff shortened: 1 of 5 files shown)


e2e/src/utils.ts
```diff
@@ -157,7 +157,7 @@ export const utils = {
       for (const table of tables) {
         if (table === 'system_metadata') {
           // prevent reverse geocoder from being re-initialized
-          sql.push(`DELETE FROM "system_metadata" where "key" != 'reverse-geocoding-state';`);
+          sql.push(`DELETE FROM "system_metadata" where "key" NOT IN ('reverse-geocoding-state', 'system-flags');`);
```
bo0tzz, line 160: nits: comment is probably redundant + will not clearning the system flags potentially cause issues elsewhere/later?

bo0tzz (approved):

danieldietzler (approved):

e2e/src/utils.ts
jrasm91: It might yeah. It's the easiest option right now though. In the future we might be able to wait for microservices to fully start up with a custom implementation.

PR #12879, opened 2024-09-23
refactor(mobile): more repositories

fyfrey:
- ExifInfoRepository
- ActivityApiRepository
- initial AssetApiRepository


mobile/lib/repositories/activity_api.repository.dart
```diff
@@ -0,0 +1,67 @@
...
+    return _toActivity(response);
+  }
+
+  @override
+  Future<void> delete(String id) {
+    return checkNull(_api.deleteActivity(id));
+  }
+
+  @override
+  Future<int> countComments(String albumId, {String? assetId}) async {
```
jrasm91, line 51: What about just stats or getStatistics

jrasm91 (approved):

jrasm91:
I think you also need to rebase and then regenerate open API

PR #12891, opened 2024-09-24
refactor(mobile): use repositories in a number of services

fyfrey:
- UserService
- PartnerService
- HashService
- MemoryService
- PersonService
- SearchService
- StackService


mobile/lib/interfaces/user_api.interface.dart
```diff
@@ -0,0 +1,20 @@
...
+abstract interface class IUserApiRepository {
+  Future<List<User>> getAll();
+  Future<({String profileImagePath})> createProfileImage({
+    required String name,
+    required Uint8List data,
+  });
+  Future<List<User>> getPartners(Direction direction);
+  Future<void> removePartner(String id);
+  Future<User> addPartner(String id);
+  Future<User> updatePartner(String id, {required bool inTimeline});
```
jrasm91, line 14: Might be worth putting this in a partner API repo (get, create, update, delete)

mobile/lib/repositories/asset_api.repository.dart
```diff
@@ -22,4 +26,26 @@ class AssetApiRepository extends BaseApiRepository
     );
     return Asset.remote(response);
   }
+
+  @override
+  Future<List<Asset>> search({List<String> personIds = const []}) async {
+    final List<Asset> result = [];
+    bool hasNext = true;
+    int currentPage = 1;
+    while (hasNext) {
```
jrasm91, line 35: Maybe we should add a to-do to make the mobile implementation use pagination, and not fetch all the results all the time.

mobile/lib/services/user.service.dart
```diff
@@ -71,13 +46,18 @@ class UserService {
...
   Future<List<User>?> getUsersFromServer() async {
-    final List<User>? users = await _getAllUsers();
-    final List<User>? sharedBy =
-        await _partnerService.getPartners(PartnerDirection.by);
-    final List<User>? sharedWith =
-        await _partnerService.getPartners(PartnerDirection.with_);
+    List<User>? users;
+    try {
+      users = await _userApiRepository.getAll();
+    } catch (exception) {
```
jrasm91, line 52: Can we at least log these errors?

jrasm91 (approved):

PR #12927, opened 2024-09-25
fix(server): http error parsing on endpoints without a default response

### LINKED ISSUE #10166: [Enhancement] - Better smtp error
### The bug

By default, when `Ignore certificate errors` is disabled and that you forget to enable it (in my case, I know that my certificate works but is invalid...), the default notification is only a `Error 400` that is thrown : see the screen below : 

Capture d’écran 2024-06-11 à 22 33 58

Maybe could it be better to throw a more precise error in this case, like `SMTP certificate error !` ??

Thanks for your hard work !

Linux 5.10.0-30-amd64 | Debian 5.10.218-1

### Version of Immich Server

v1.106.2

not used yet 

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
not needed ig :)
```

### Your .env content

```Shell
not needed neither :) (sorry I'm too lazy, and since it's only some front thing...)
```

### Reproduction steps

```bash
1. go in the settings
2. have an invalid certificate
3. forgot do disable verification
4. boom ! the notif 
...
```

### Relevant log output

_No response_

### Additional information

_No response_

jrasm91:
Fixes #10166

You can see in the open api spec, that this endpoint did not have a default resopnse, so open-api assumed it was a string and did not parse the response. This is why `handleError` was unable to extract the message correctly.

This PR fixes that parsing, but also adds a default response for this endpoint (returns `messageId`).

| Before | After |
| - | - |
| image | image |

### DIFF
--- a/open-api/immich-openapi-specs.json
+++ b/open-api/immich-openapi-specs.json
@@ -3491,6 +3491,13 @@
         },
         "responses": {
           "200": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/TestEmailResponseDto"
+                }
+              }
+            },
             "description": ""
           }
         },
@@ -12346,6 +12353,17 @@
             "type": "boolean"
           }
         },
+        "type": "object"
+      },
+      "TestEmailResponseDto": {
+        "properties": {
+          "messageId": {
+            "type": "string"
+          }
+        },
+        "required": [
+          "messageId"
+        ],
         "type": "object"
       },
       "TimeBucketResponseDto": {
--- a/open-api/typescript-sdk/src/fetch-client.ts
+++ b/open-api/typescript-sdk/src/fetch-client.ts
@@ -656,6 +656,9 @@ export type SystemConfigSmtpDto = {
     replyTo: string;
     transport: SystemConfigSmtpTransportDto;
 };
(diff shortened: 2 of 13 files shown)


open-api/immich-openapi-specs.json
```diff
@@ -3491,6 +3491,13 @@
         },
         "responses": {
           "200": {
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/TestEmailResponseDto"
+                }
+              }
+            },
```
jrasm91, line 3500: Previously no default response type.

open-api/typescript-sdk/src/fetch-client.ts
```diff
@@ -2220,7 +2223,10 @@ export function addMemoryAssets({ id, bulkIdsDto }: {
 export function sendTestEmail({ systemConfigSmtpDto }: {
     systemConfigSmtpDto: SystemConfigSmtpDto;
 }, opts?: Oazapfts.RequestOpts) {
-    return oazapfts.ok(oazapfts.fetchText("/notifications/test-email", oazapfts.json({
```
jrasm91, line 2223: Changed from fetchText to fetchJson

zackpollard (approved):

PR #12934, opened 2024-09-25
user removal from option menu on the top in shared album

Pranav-8bit:


bo0tzz:
Can you add a PR description with what these changes do?

PR #13006, opened 2024-09-27
feat(server): separate quality for thumbnail and preview images

mertalev:
## Description

It can be desirable to set the target quality differently between thumbnails and previews. Some admins might want to increase preview quality, but without making thumbnails unnecessarily larger. Others who mainly view the original image may wish to lower preview quality to save space.

This PR changes the config to have separate quality entries for thumbnails and previews. It also restructures the image settings to reflect the common settings between thumbnails and previews.

This is not a *breaking* change per se, but admins who set custom image settings in their config file will use the defaults until their config file is corrected. Admins who use the database via the admin UI are migrated over to the new config structure.

The image settings in the admin UI are now grouped by preview and thumbnail. I opted to set `isOpen` true for these accordions so they're expanded when first viewing without extra clicks.

Fixes #12886

## How Has This Been Tested?

Tested by setting an extremely low quality for one output and an extremely high quality for another, verifying that refreshing thumbnails for an asset finished successfully, and that the two images are of very different quality.

image-settings


jrasm91 (approved):
Nice!

PR #13012, opened 2024-09-28
feat(server): generate all thumbnails for an asset in one job

mertalev:
### Description

This PR changes image thumbnail generation to first decode and resize the input image to a preview-sized raw buffer, then reuse that buffer as the input for the preview, thumbnail and thumbhash. This has several advantages:
  * As reading from disk and decoding the image are the slowest steps of thumbnail generation, this makes generating thumbnails *significantly* faster
  * It is more cache-efficient due to handling everything to do with a file at once instead of spreading this out into jobs that run far apart from one another
  * It is easier to understand thumbnail generation progress when there is a 1:1 relationship between jobs and assets. It's currently a common source of confusion that the number of jobs keeps growing upward, and it's difficult to understand how much progress has been made
  * Other steps during thumbnail generation only need to run once, such as extracting embedded images or running FFprobe

A concern I had was around RAM usage. Remarkably, this approach uses *less* RAM than the current one despite the use of an in-memory buffer. Peak memory usage dropped from 2.47GiB to 1.22GiB during testing, and the average is also lower compared to preview generation.

Video thumbnails are not optimized in this PR to keep it smaller and more focused. A later PR will use a similar approach through FFmpeg. In the meantime, I refactored the code such that it will be easy to make this change. I intentionally made it generate video thumbnails sequentially instead of using `Promise.all` to avoid excessive resource usage.

### Testing

I ran thumbnail generation on all assets and confirmed that jobs were processed successfully. Setting different image settings and enabling the "extract embedded preview" setting all functioned as expected. I additionally ran an image-only comparison on all assets:

thumbnail-generation-memory-usage

In order from left to right:
  - Thumbnail generation start on main
  - Thumbnail generation previews completed on main (starting thumbnails)
  - Thumbnail generation finished on main
  - Thumbnail generation started in PR
  - Thumbnail generation finished in PR

On main, generating the previews took 43 minutes and completing all jobs took 76 minutes , while this PR completed all thumbnail generation tasks in 44 minutes. For this particular library with default settings, this is a ~73% speedup for images. The gap will widen with more RAW or high megapixel images and when the "extract embedded preview" setting is enabled.


server/src/repositories/media.repository.ts
```diff
@@ -57,19 +59,12 @@ export class MediaRepository implements IMediaRepository {
...
-  async generateThumbnail(input: string | Buffer, output: string, options: ThumbnailOptions): Promise<void> {
-    // some invalid images can still be processed by sharp, but we want to fail on them by default to avoid crashes
-    const pipeline = sharp(input, { failOn: options.processInvalidImages ? 'none' : 'error', limitInputPixels: false })
-      .pipelineColorspace(options.colorspace === Colorspace.SRGB ? 'srgb' : 'rgb16')
-      .rotate();
-
-    if (options.crop) {
-      pipeline.extract(options.crop);
-    }
+  async decodeImage(input: string, options: DecodeToBufferOptions) {
```
jrasm91, line 62: Unnecessary async, right?

server/src/services/media.service.spec.ts
```diff
@@ -438,233 +482,189 @@ describe(MediaService.name, () => {
...
+        assetMock.getById.mockResolvedValue(assetStub.image);
+        const thumbhashBuffer = Buffer.from('a thumbhash', 'utf8');
+        mediaMock.generateThumbhash.mockResolvedValue(thumbhashBuffer);
 
-    it('should skip invisible assets', async () => {
-      assetMock.getByIds.mockResolvedValue([assetStub.livePhotoMotionAsset]);
+        const previewFormat = type === AssetFileType.PREVIEW ? format : ImageFormat.JPEG;
+        const thumbnailFormat = type === AssetFileType.THUMBNAIL ? format : ImageFormat.WEBP;
+        const previewPath = `upload/thumbs/user-id/as/se/asset-id-preview.${previewFormat}`;
+        const thumbnailPath = `upload/thumbs/user-id/as/se/asset-id-thumbnail.${thumbnailFormat}`;
```
jrasm91, line 497: This test has little value imo. Looking at it, it is not clear what it is doing. It takes reading and understanding the _test_ code and checking the different branching going on here. Tests should not have branching and should be very simple. I would prefer 4 individual, trivial tests that are easy to read.

server/src/services/media.service.ts
```diff
@@ -181,141 +174,127 @@ export class MediaService {
...
-      this.logger.debug(`Deleting old thumbnail for asset ${asset.id}`);
-      await this.storageRepository.unlink(thumbnailFile.path);
+    if (asset.thumbhash != generated.thumbhash) {
+      await this.assetRepository.update({ id: asset.id, thumbhash: generated.thumbhash });
     }
 
-    await this.assetRepository.upsertFile({ assetId: asset.id, type: AssetFileType.THUMBNAIL, path: thumbnailPath });
-    await this.assetRepository.update({ id: asset.id, updatedAt: new Date() });
-    await this.assetRepository.upsertJobStatus({ assetId: asset.id, thumbnailAt: new Date() });
+    await this.assetRepository.upsertJobStatus({ assetId: asset.id, previewAt: new Date(), thumbnailAt: new Date() });
```
jrasm91, line 232: In the future should we combine these into one column?

jrasm91 (approved):
Looks great!

server/src/services/media.service.spec.ts
mertalev: That's fair! The reason I did it like this is mainly so it can handle more formats or image types being added in the future. I can get rid of the test branching by splitting it into two jobs for previews and thumbnails.

server/src/services/media.service.ts
mertalev: I think so. There's not much point to having both when they get set at the same time.

server/src/services/media.service.spec.ts
jrasm91: The goal of the unit test should be to cover the branches in the code. And there should ideally be a test for each branch. We don't have to use every format if it doesn't run different code paths.

martabal (commented):
Hey, that's great. Thanks !

aviv926:
It's great. I wonder how many assets you ran the test on?

mertalev:
> It's great. I wonder how many assets you ran the test on?

I ran it on 3,500 images.

PR #13016, opened 2024-09-28
chore(docs): Update nginx example config

PetrSvirak:
This tweak in the nginx example should make all requests made by immich web app [go through the SSL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Upgrade-Insecure-Requests). 

I stumbled upon the lack of the header when setting up nginx to redirect wan traffic to an immich instance and noticed a lot of tracking still wanting to use HTTP instead of HTTPS protocol.

> I was unsure how to file an improvement for docs properly, so I created this PR instead. Please, if this change is too niche (not universal enough in your eyes) or requires an unnecessary amount of effort on your side to merge, feel free to close the PR.

### DIFF
--- a/docs/docs/administration/reverse-proxy.md
+++ b/docs/docs/administration/reverse-proxy.md
@@ -18,10 +18,11 @@ server {
     client_max_body_size 50000M;
 
     # Set headers
-    proxy_set_header Host              $http_host;
-    proxy_set_header X-Real-IP         $remote_addr;
-    proxy_set_header X-Forwarded-For   $proxy_add_x_forwarded_for;
-    proxy_set_header X-Forwarded-Proto $scheme;
+    proxy_set_header Host                      $http_host;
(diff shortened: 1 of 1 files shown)


bo0tzz:
This is generally a sensible thing to do, but I'm not sure how I feel about it here since we deliberately keep these configs as minimal as possible and it doesn't have any HTTPS config for that reason.

PetrSvirak:
I will only add that I did not know how or whether to update Caddy, Apache, and Traefik in a similar fashion - that might be a thing missing in this PR even if changing nginx would look like a good idea.

The reason I created this PR anyway was that the TLS was basically the only reason I set up a reverse proxy for my instance. But I might be side-tracking a bit: So let's close this PR?

jrasm91:
While useful, it is a bit out of scope for immich itself so I will go ahead and close this.

PR #13040, opened 2024-09-30
fix(server): `/search/random` failing with certain options

### LINKED ISSUE #13021: API: POST /search/random returns internal server error
### The bug

https://demo.immich.app/doc#/Search/searchRandom

calling the `POST /search/random` endpoint with parameters like `withExif` or `personIds` returns an internal server error

Unraid

### Version of Immich Server

v1.116

-

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
not needed
```

### Your .env content

```Shell
not needed
```

### Reproduction steps

1. Go to https://demo.immich.app/doc#/Search/searchRandom (make sure you are logged in)
2. use request body:
```
{
  "withExif": true,
  "page": 1,
  "size": 100
}
```
3. Execute

### Relevant log output

```shell
Immich Logs:
ERROR [Api:ErrorInterceptor~2tquej6l] Database error: QueryFailedError: for SELECT DISTINCT, ORDER BY expressions must appear in select list


Server Response:
Error: response status is 500

{
  "message": "Failed to search random",
  "error": "Internal Server Error",
  "statusCode": 500,
  "correlationId": "p93x6n1q"
}
```

### Additional information

```
curl -X 'POST' \
  'https://demo.immich.app/api/search/random' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "withExif": true,
  "page": 1,
  "size": 100
}'
```

mertalev:
## Description

The current query only works when no relations are requested. When they are, TypeORM does an additional query with a `distinct` clause that conflicts with the sort order. This PR makes a few changes to address this:

* Changes the query to not rely on ordering and instead use a where clause leveraging the random nature of UUIDs. This has the additional advantage of making the query scale much better to larger libraries
* Does two simultaneous queries with opposite constraints: this is to ensure that the requested number of results is provided if enough matching rows exist in the DB. Filtering only in one direction of the random UUID risks excluding too many rows. This can be done more efficiently using `UNION ALL`, but the fact that TypeORM does two queries makes it infeasible right now
* Fixes the `withPeople` option that's currently broken
* Removes pagination from the `/search/random` endpoint since it makes things more complicated and there isn't a compelling use-case for it here

Fixes #13021

## How Has This Been Tested?

Tested that the endpoint works with a combination of `withExif`, `withStacked` and `withPeople` and returns different rows in each call.

### DIFF
--- a/mobile/openapi/lib/api/search_api.dart
+++ b/mobile/openapi/lib/api/search_api.dart
@@ -383,7 +383,7 @@ class SearchApi {
   /// Parameters:
   ///
   /// * [RandomSearchDto] randomSearchDto (required):
-  Future<SearchResponseDto?> searchRandom(RandomSearchDto randomSearchDto,) async {
+  Future<List<AssetResponseDto>?> searchRandom(RandomSearchDto randomSearchDto,) async {
     final response = await searchRandomWithHttpInfo(randomSearchDto,);
     if (response.statusCode >= HttpStatus.badRequest) {
       throw ApiException(response.statusCode, await _decodeBodyBytes(response));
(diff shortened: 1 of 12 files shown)


jrasm91 (approved):
Looks much better, thanks!

PR #13041, opened 2024-09-30
docs: update config file

mertalev:
## Description

The config file in the docs is extremely outdated. We should either auto-generate this or add a check for it in PRs.


alextran1502 (approved):

aviv926:
I know we use the Config file at https://demo.immich.app/
I'm not sure where it's loaded from but it seems to be updated manually (?)
Maybe it can be imported for documentation if it is updated frequently.

bo0tzz:
The demo config file is at https://github.com/immich-app/demo/blob/main/config.yaml

PR #13068, opened 2024-09-30
feat: preload fonts

martabal:
This PR:
- preload fonts for better loading speed. The `hook.server.ts` is only used to replace the `%app.font%` and `%app.monofont%` variables with the actual location of the `ttf` files.
- remove the `WEB_ROOT` constant which is no longer used.

| Before | After |
| :---: | :---: |
| image | image |

`index.html`: 

| Before | After |
| :---: | :---: |
| image | image|


jrasm91 (approved):
Neat! Are the first set of screenshots backwards?

martabal:
> Neat! Are the first set of screenshots backwards?

Nope this is to show that ttf files are fetched right after receiving `index.html`. Before, they were fetched long after (we don't see when they are fetched in the 'before' screenshot lol)

PR #13086, opened 2024-10-01
fix: deletedAt not set for offline assets during 1.116.0 migration

zackpollard:
Since 1.116.0 we now assume deletedAt is set for offline assets, which makes them show in trash and get hidden in other screens. This PR will add a migration to set deletedAt for all offline files at the time it runs.


bo0tzz (approved):
Presumably this overrides deletedAt for assets that went offline between 1.116.0 and this migration, is that an issue?

zackpollard:
> Presumably this overrides deletedAt for assets that went offline between 1.116.0 and this migration, is that an issue?

Fixed this with a change in the query.

PR #13091, opened 2024-10-01
fix(docs): Windows backup script

aviv926:
The method shown in the Immich documents for backing up on Windows does not work well and it does not properly back up city names in languages ​​other than English.

Cities in Hebrew and Arabic languages ​​would appear as ┘ו╪»┘ך┘ז╪⌐ ╪¿┘ז┘ך ┘ך╪º╪│and caused problems after restoring from a backup.

A solution to this problem is to backup directly from the Docker container and then copy the file to Windows.

The script will create a folder named `Immich DB backups` on the desktop where a file named `immich_dump.sql.gz` will be stored.

It's important for me to say, I'm not a big expert in PS that's why I used artificial intelligence to write this script, it works well and does the job properly.

I will understand if you don't want to include it in the documentation, however if we don't include it we will have to remove the option of backing up the database in Windows and instead put a link to the gits I created or link it to the Community Guides, or find a way to back up the database properly , as you see fit :)

https://github.com/user-attachments/assets/0532a157-f274-4f75-afa6-6c4519557cf5

### DIFF
--- a/docs/docs/administration/backup-and-restore.md
+++ b/docs/docs/administration/backup-and-restore.md
@@ -65,13 +65,13 @@ The location and name of the original backup should be:
 
 1. docker compose down -v # CAUTION! Deletes all Immich data to start from scratch
 2. rm -rf DB_DATA_LOCATION # CAUTION! Deletes all Immich data to start from scratch
-4. docker compose pull # Update to latest version of Immich (if desired)
-5. docker compose create # Create Docker containers for Immich apps without running them
-6. docker start immich_postgres # Start Postgres server
-7. sleep 10 # Wait for Postgres server to start up
-8. [Download](https://github.com/immich-app/immich/tree/main/docs/docs/administration/files/immich-windows-restore-script.ps1) the restore script
-9. Right click and click on Edit
-10. Run the script from the windows PowerShell ISE
+3. docker compose pull # Update to latest version of Immich (if desired)
+4. docker compose create # Create Docker containers for Immich apps without running them
+5. docker start immich_postgres # Start Postgres server
+6. sleep 10 # Wait for Postgres server to start up
+7. [Download](https://github.com/immich-app/immich/tree/main/docs/docs/administration/files/immich-windows-restore-script.ps1) the restore script
+8. Right click and click on Edit
+9. Run the script from the windows PowerShell ISE
 
 </TabItem>
 </Tabs>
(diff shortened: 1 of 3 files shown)


bo0tzz:
I don't really understand the issue

> A solution to this problem is to backup directly from the Docker container and then copy the file to Windows.

As far as I can tell the existing setup already runs the pg dump inside of the container.

aviv926:
> I don't really understand the issue
> 
> > A solution to this problem is to backup directly from the Docker container and then copy the file to Windows.
> 
> As far as I can tell the existing setup already runs the pg dump inside of the container.

From an experiment I did when I ran the script:
docker exec -t immich_postgres pg_dumpall --clean --if-exists --username=postgres | Set-Content -Encoding utf8 "C:\path\to\backup\dump.sql"

This returned a file with the names of the cities in Hebrew and Arabic corrupted, in addition even if a user uses an external directory with paths that contain Hebrew or Arabic characters it corrupts them.

This is how it looks:
image


I tried to figure out if the problem is related to Set-Content -Encoding utf8 but I couldn't get it to work even when I changed it to Windows-1256 for Arabic or Windows-1255 for Hebrew.



Only when I ran inside the docker container the backup command
pg_dumpall --clean --if-exists --username=postgres | gzip > "/path/to/backup/dump.sql.gz"

And I checked the backup content, it displayed the Hebrew and Arabic letters correctly.

This is how it should look:
image

zackpollard:
Hey, we have chatted about this and since we now have backups integrated into the server itself, we don't think this is needed. It would be better to update the restore section to refer to those integrated backups. If you want to add a guide for windows manual backups, lets make that a community guide that we can link out to and we can convert the manual backup section here to point to the integrated automatic backups.

aviv926:
> Hey, we have chatted about this and since we now have backups integrated into the server itself, we don't think this is needed. It would be better to update the restore section to refer to those integrated backups. If you want to add a guide for windows manual backups, lets make that a community guide that we can link out to and we can convert the manual backup section here to point to the integrated automatic backups.

Hi Zack how are you? Thanks, yes this PR is no longer needed for the reasons you mentioned and I think it's good that it is because now it is a feature added to the code so it will get updates accordingly and avoid future problems. I believe I will update the documentation to explain about the new feature the team added (thanks for that! As always 0:) By the way, you can check with the team about https://github.com/immich-app/immich/pull/14911

PR #13092, opened 2024-10-01
feat(server): better mount checks

jrasm91:
- Ignore errors when `.immich` already exists, but mount checks aren't enabled
- Include a link in the error message to the docs
- Add a docs page about system integrity
- Add an env option to skip mount checks

### DIFF
--- a/docs/docs/administration/system-integrity.md
+++ b/docs/docs/administration/system-integrity.md
@@ -1,0 +1,47 @@
+# System Integrity
+
+## Folder checks
+
+:::info
+The folders considered for these checks include: `upload/`, `library/`, `thumbs/`, `encoded-video/`, `profile/`
+:::
+
+When Immich starts, it performs a series of checks in order to validate that it can read and write files to the volume mounts used by the storage system. If it cannot perform all the required operations, it will fail to start. The checks include:
+
+- Creating an initial hidden file (`.immich`) in each folder
+- Reading a hidden file (`.immich`) in each folder
+- Overwriting a hidden file (`.immich`) in each folder
+
+The checks are designed to catch the following situations:
+
+- Incorrect permissions (cannot read/write files)
+- Missing volume mount (`.immich` files should exist, but are missing)
+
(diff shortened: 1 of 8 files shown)


docs/docs/administration/system-integrity.md
```diff
@@ -0,0 +1,47 @@
...
+- File does not exist - user manually deleted it and should be manually re-created (`touch .immich`)
+- File does not exist - user restored from a backup, but did not restore each folder (user should restore all folders or manually create `.immich` in any missing folders)
+
+### Ignoring the checks
+
+The checks are designed to catch common problems that we have seen users have in the past, but if you want to disable them you can set the following environment variable:
+
+```
+IMMICH_IGNORE_MOUNT_CHECK_ERRORS=true
+```
```
bo0tzz, line 47: Do we want to add a disclaimer here along the lines of "if you've used this env var then unfortunately we can't help you with mountpoint issues"?

server/src/repositories/config.repository.ts
```diff
@@ -10,6 +10,9 @@ export class ConfigRepository implements IConfigRepository {
         skipMigrations: process.env.DB_SKIP_MIGRATIONS === 'true',
         vectorExtension: getVectorExtension(),
       },
+      storage: {
+        ignoreMountCheckErrors: (process.env.IMMICH_IGNORE_MOUNT_CHECK_ERRORS ?? 'true') === 'true',
```
bo0tzz, line 14: Doesn't this default to disabling the checks?

bo0tzz (approved):

mmomjian:
Can we add an entry to the env vars page for this?

server/src/repositories/config.repository.ts
danieldietzler: Yeah lol

jrasm91:
> Can we add an entry to the env vars page for this?

Sure.

danieldietzler (approved):

PR #13100, opened 2024-10-01
refactor(server): config file env

jrasm91:
Proxy requests to access `IMMICH_CONFIG_FILE` through `IConfigRepository.getEnv()` and simplify associated tests.


danieldietzler (approved):
Awesome!

PR #13102, opened 2024-10-01
add env_file to all docker services for timezone consistency

cwallace:
This PR adds env_file to all Docker containers to ensure timezone settings are consistent throughout the stack. By centralizing the environment configurations, we avoid any potential issues caused by mismatched timezones between services.

### DIFF
--- a/docker/docker-compose.prod.yml
+++ b/docker/docker-compose.prod.yml
@@ -48,6 +48,8 @@ services:
   redis:
     container_name: immich_redis
     image: redis:6.2-alpine@sha256:2d1463258f2764328496376f5d965f20c6a67f66ea2b06dc42af351f75248792
+    env_file:
+      - .env
     healthcheck:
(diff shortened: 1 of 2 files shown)


bo0tzz:
The redis container definitely isn't involved in any (user-facing) timezone stuff.

I'm not all too stoked about just blanket passing all the env vars to containers that don't need them. If the Postgres container can meaningfully make use of the TZ env var then we should pass it explicitly instead.

jrasm91:
Feel free to add environment entries to specific containers but we're not going to add the env to any other container.

Phlogi:
@bo0tzz and @jrasm91 The env_file entry is missing on latest release on the database service: https://github.com/immich-app/immich/blob/v1.118.0/docker/docker-compose.yml

Why? Please re-apply this change.

bo0tzz:
What do you mean re-apply? The PR was closed and we didn't add this change.

PR #13127, opened 2024-10-02
chore: add more languages

danieldietzler:


pyorot:
Serbian was already in there… where did you get your languages from? 🧐

danieldietzler:
> Serbian was already in there… where did you get your languages from? 🧐

There were only Serbian (Cyrillic) and Serbian (Latin). Are those not different?

pyorot:
> > Serbian was already in there… where did you get your languages from? 🧐
> 
> There were only Serbian (Cyrillic) and Serbian (Latin). Are those not different? To answer your question, we get requests to add new languages and then we do :)

I was wondering who would’ve requested Serbian then :P

> Serbian is practically the only European standard language whose speakers are fully functionally digraphic, using both Cyrillic and Latin alphabets.

Those are the language’s two writing systems, basically

danieldietzler:
Ah I see. Are you on our discord (https://discord.immich.app) ? That'd be easier to discuss this with others who speak the language as well

pyorot:
> Ah I see. Are you on our discord (https://discord.immich.app) ? That'd be easier to discuss this with others who speak the language as well

Yep, I’ll give you a bell

zackpollard (approved):

PR #13128, opened 2024-10-02
feat: add QR code for shared link

jnav:
Simple patch to add QR code to the dialog where link for album sharing is generated

2024-10-02-immich_QR


web/src/lib/utils/qrcode.js
danieldietzler, line 1: Please properly import dependencies and don't just copy (minified) code over.

danieldietzler (changes requested):

web/src/lib/utils/qrcode.js
jnav: @danieldietzler  thank you for the "kick" - i will try to figure it out how it should be done properly

jnav:
this time i am using svelte-qrcode to generate the QR code and its being added as module, not as library. Hope that's better now

web/src/lib/utils/qrcode.js
danieldietzler: The file is still in the commit. Please remove it as well :)

web/src/lib/components/shared-components/create-share-link-modal/create-shared-link-modal.svelte
```diff
@@ -94,6 +97,8 @@
       });
       sharedLink = makeSharedLinkUrl($serverConfig.externalDomain, data.key);
       onCreated();
+      showQRCode = true;
+
```
danieldietzler, line 101:
```suggestion
```

web/src/lib/components/shared-components/create-share-link-modal/create-shared-link-modal.svelte
```diff
@@ -23,6 +25,7 @@
   export let onCreated: () => void = () => {};
 
   let sharedLink: string | null = null;
+  let showQRCode: boolean | null = false;
```
danieldietzler, line 28: When should `showQRCode` be null?

web/src/lib/utils/qrcode.js
jnav: uff, still learning, sorry - should be fixed with latest commit

web/src/lib/components/shared-components/create-share-link-modal/create-shared-link-modal.svelte
jnav: you are right, boolean is never null. Ups, will fix it

jnav: @danieldietzler thank you for your patience - i am new to this svelte/js world

danieldietzler: Happy to help :)

danieldietzler (approved):
Looks good to me. I'm not sure if we want a fixed width for the QR code though

danieldietzler:
Also, please fix the typescript issues.

alextran1502:
Hello, thank you for the PR. What is the use case of this PR? How are you planning to share the QR code?

jnav:
hi @alextran1502 , my use case is as follows - i shoot people on the street and when i am done i create album and share that code over phone to the person who will later have photos there for download. I dont share any phone numbers or contact details to send link later. I understand that this is a change in UI and not everyone would like to have it always on - if you want i will try to add switch to User settings (feature toggle) and by default QR code would be disabled. Second option would be to click on "generate QR" button once link is created. This way it would be less "pushy" feature

CrushedAsian255:
Any progress on this PR?

jrasm91:
I definitely like the idea of generating a QR code for stuff like this, but I think the user experience could be better. I am wondering if after generating the link we have a button that when you click it, it opens a full screen modal with the QR code in it. It could have some buttons in the footer to download it, etc. 

Is something you want to take a look at? Otherwise I can push some changes to this branch.

jnav:
added button to generate QR code, hope its less intrusive solution. I wanted to scroll to the bottom of QR code after its being generated, but i failed %-|

thank you for your comments/reviews

2024-10-25-QR-button

zackpollard:
Hey, I think this is getting closer, however I think I would still like to see a small change here. This QR code feature is good, but I think the button is still far to prominent for the amount of people that will likely use it. I'd personally like to see this as a small QR code symbol within or next to the link box itself. Similar to the copy button. This seems much more reusable in other areas if we have links, and also much smaller and out of the way.

zackpollard:
Superseded by #16543

PR #13159, opened 2024-10-03
chore: some more unit tests :)

danieldietzler:
Screenshot 2024-10-03 at 19 15 02
Screenshot 2024-10-03 at 19 15 42


jrasm91 (approved):
tyty

PR #13160, opened 2024-10-03
refactor(server): worker env

jrasm91:
`start.sh immich` and `start.sh microservices` have been deprecated in favor of using a single container and/or worker include/exclude environment variables.This PR removes the previous fallback, which mapped `immich` and `microservices` to their respective environment variable equivalents.


zackpollard (approved):
LGTM

PR #13185, opened 2024-10-04
feat: align compose ports

jrasm91:
Change the internal port from 3001 to 2283, to align with the external port.

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -45,7 +45,6 @@ services:
         soft: 1048576
         hard: 1048576
     ports:
-      - 3001:3001
       - 9230:9230
       - 9231:9231
     depends_on:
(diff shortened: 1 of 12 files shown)


Miladiir:
👎

danieldietzler (approved):

HybridSarcasm:
Why was this important to change? I don’t see any supporting issue stating why this was necessary. As it is, this seems like a “we wanted to” change. If so, seems like a poor reason for such a breaking change.

bo0tzz:
Immich is still unstable beta software, and we're taking the liberty to change things to be better and cleaner even if it's not strictly necessary.

jrasm91:
> to align with the external port.

Which has implications with troubleshooting, support, install instructions, demos, etc.

PR #13207, opened 2024-10-05
chore: auth unit tests

danieldietzler:
Screenshot 2024-10-05 at 18 01 24

The remaining lines are untestable currently because we cannot mock out the OIDC handling, since that isn't abstracted away in a repo.

### DIFF
--- a/server/src/services/auth.service.ts
+++ b/server/src/services/auth.service.ts
@@ -192,10 +192,6 @@ export class AuthService extends BaseService {
 
   async authorize(dto: OAuthConfigDto): Promise<OAuthAuthorizeResponseDto> {
     const config = await this.getConfig({ withCache: false });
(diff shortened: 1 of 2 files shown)


server/src/services/auth.service.ts
```diff
@@ -192,10 +192,6 @@ export class AuthService extends BaseService {
 
   async authorize(dto: OAuthConfigDto): Promise<OAuthAuthorizeResponseDto> {
     const config = await this.getConfig({ withCache: false });
-    if (!config.oauth.enabled) {
-      throw new BadRequestException('OAuth is not enabled');
-    }
```
danieldietzler, line 197: This already happens in `this.getOAuthClient`

jrasm91 (approved):

PR #13212, opened 2024-10-05
feat(docs): update roadmap

aviv926:


docs/src/pages/roadmap.tsx
```diff
@@ -78,6 +80,7 @@ import React from 'react';
 import { Item, Timeline } from '../components/timeline';
 
 const releases = {
+  'v1.114.0': new Date(2024, 8, 6),
```
danieldietzler, line 83:
```suggestion
  'v1.114.0': new Date(2024, 8, 5),
```

docs/src/pages/roadmap.tsx
```diff
@@ -231,18 +234,31 @@ const roadmap: Item[] = [
...
     icon: mdiTagMultiple,
     iconColor: 'orange',
     title: 'Tags',
     description: 'Tag your photos and videos',
     release: 'v1.113.0',
   }),
+  withRelease({
+    icon: mdiFolderSync,
+    iconColor: 'green',
+    title: 'Ablume sync (mobile)',
```
danieldietzler, line 253:
```suggestion
    title: 'Album sync (mobile)',
```

docs/src/pages/roadmap.tsx
```diff
@@ -231,18 +234,31 @@ const roadmap: Item[] = [
...
+    description: 'Sync or mirror an album from your phone to the Immich server',
+    release: 'v1.113.0',
+  }),
   withRelease({
     icon: mdiFolderMultiple,
     iconColor: 'brown',
     title: 'Folders',
-    description: 'View your photos and videos in folders',
+    description: 'View your external photos and videos in folders',
```
danieldietzler, line 261: That does not only apply to external libraries.

docs/src/pages/roadmap.tsx
danieldietzler: Nvm 😅

docs/src/pages/roadmap.tsx
aviv926: Maybe it's worth mentioning in some way that these are external folders as well?
When I first read it it looked like it was an addition of something "simple".

I know a lot of people have been waiting for this feature, maybe for those following the roadmap it can be confusing.

docs/src/pages/roadmap.tsx
aviv926: image
lol I'm a bit confused

docs/src/pages/roadmap.tsx
danieldietzler: > Browse your photos and videos in their folder structure

?

danieldietzler: Not a huge fan tbh, idk....

aviv926: > > Browse your photos and videos in their folder structure
> 
> ?

In fact I think it explains this feature well.

docs/src/pages/roadmap.tsx
```diff
@@ -231,18 +234,31 @@ const roadmap: Item[] = [
 ];
 
 const milestones: Item[] = [
+  withRelease({
+    icon: mdiFaceRecognition,
+    title: 'Metadata Face Import',
+    description: 'Read face metadata in Digikam format during metadata extraction',
```
jrasm91, line 240: Maybe just say during import

jrasm91 (approved):

docs/src/pages/roadmap.tsx
aviv926: Since this is currently [exclusive](https://github.com/immich-app/immich/releases#:~:text=Only%20face%20metadata%20in%20Digikam%20format%20is%20supported) to face metadata from Digikam, I thought it would be necessary to mention it in the roadmap for those who follow there.
what do you think?

jrasm91: I think it is fine to add here like you have it, but maybe we just say "during import" instead of "during metadata extraction" which is a bit technical.

aviv926: :)

danieldietzler (approved):

PR #13237, opened 2024-10-07
feat(ml): round-robin device assignment

mertalev:
## Description

This PR enables multi-GPU support by adding a new `MACHINE_LEARNING_DEVICE_IDS` environmental variable that accepts a comma-separated list of device IDs. This list is used to assign a single device ID to each gunicorn worker. The approach here takes advantage of assymetric inheritance of environmental variables from parent to child when forking, setting a different value for each worker.

Fixes #13140

## How Has This Been Tested?

Confirmed the device ID distribution when workers = devices, workers > devices and workers < devices all work in round-robin priority. When load-tested with Locust, the device ID used when loading models reflects the assigned ID.

In draft, pending tests.

### DIFF
--- a/machine-learning/app/config.py
+++ b/machine-learning/app/config.py
@@ -39,6 +39,10 @@ class Settings(BaseSettings):
         case_sensitive = False
         env_nested_delimiter = "__"
 
+    @property
+    def device_id(self) -> int:
+        return os.environ.get("MACHINE_LEARNING_DEVICE_ID", "0")
+
 
 class LogSettings(BaseSettings):
     immich_log_level: str = "info"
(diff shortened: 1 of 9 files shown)


machine-learning/app/config.py
```diff
@@ -39,6 +39,10 @@ class Config:
         case_sensitive = False
         env_nested_delimiter = "__"
 
+    @property
+    def device_id(self) -> int:
+        return os.environ.get("MACHINE_LEARNING_DEVICE_ID", "0")
```
mertalev, line 44: The reason this is a property is because the environmental variables are normally evaluated before the pre-fork hook runs. Using a property makes it evaluate the post-fork value instead.

jrasm91 (approved):
Nice!

bo0tzz (commented):
Neat. Should this be added to the docs as well?

mertalev:
> Neat. Should this be added to the docs as well?

Yup, definitely.

PR #13244, opened 2024-10-07
docs: Add DB name to backup-and-restore

Mraedis:
Add database name to the restore command and document it in the notes
Minor alignment changes

### DIFF
--- a/docs/docs/administration/backup-and-restore.md
+++ b/docs/docs/administration/backup-and-restore.md
@@ -34,14 +34,15 @@ docker exec -t immich_postgres pg_dumpall --clean --if-exists --username=postgre
 docker compose down -v  # CAUTION! Deletes all Immich data to start from scratch
 ## Uncomment the next line and replace DB_DATA_LOCATION with your Postgres path to permanently reset the Postgres database
 # rm -rf DB_DATA_LOCATION # CAUTION! Deletes all Immich data to start from scratch
-docker compose pull     # Update to latest version of Immich (if desired)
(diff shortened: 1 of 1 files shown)


bo0tzz (approved):

danieldietzler (approved):

mmomjian:
I don’t think this is correct/needed. When restoring a pg_dumpall, the Immich database may not yet exist. It’s better practice to connect to the default Postgres database since that will always exist.

Mraedis:
So what happens when you restore a backup that had a different DB name from default then? Is the DB created when running the command? (in which case the flag is unneeded in both cases)

mmomjian:
> So what happens when you restore a backup that had a different DB name from default then? Is the DB created when running the command? (in which case the flag is unneeded in both cases)

Yes, when restoring pg_dumpall it will create all the databases that were present in the old cluster.

Mraedis:
I'll just re-commit with the indent changes and a note on the username.🙂

docs/docs/administration/backup-and-restore.md
```diff
@@ -103,6 +105,7 @@ services:
 Then you can restore with the same command but pointed at the latest dump.
 
 ```bash title='Automated Restore'
+# Be sure to check the username and database name if you changed them from defaults
```
mmomjian, line 108:
```suggestion
# Be sure to check the username if you changed it from default
```

mmomjian (approved):

Issue #13053, opened 2024-09-30
GPS geo location wrong: GPSLongitudeRef and GPSLatitudeRef not taken into account


skycryer:
### The bug

So I have it looks like around 1k images that are shown on the wrong place on the map. I checked my Apple Photos and Export data. All have the correct gps data and are shown correct on Apple Photos. But on Immich they are completely off, not just a few miles but a few thousand.

Here is the metadata of the image

```
<x:xmpmeta xmlns:x="adobe:ns:meta/" x:xmptk="XMP Core 6.0.0">
   <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#">
      <rdf:Description rdf:about=""
            xmlns:exif="http://ns.adobe.com/exif/1.0/"
            xmlns:photoshop="http://ns.adobe.com/photoshop/1.0/"
            xmlns:dc="http://purl.org/dc/elements/1.1/">
         <exif:GPSLongitude>14.160669499999999</exif:GPSLongitude>
         <exif:GPSLongitudeRef>W</exif:GPSLongitudeRef>
         <exif:GPSHPositioningError>1</exif:GPSHPositioningError>
         <exif:GPSLatitudeRef>N</exif:GPSLatitudeRef>
         <exif:GPSLatitude>28.184478333333335</exif:GPSLatitude>
         <exif:GPSTimeStamp>2024-09-25T09:16:21Z</exif:GPSTimeStamp>
         <photoshop:DateCreated>2019-05-19T12:40:50+01:00</photoshop:DateCreated>
         <dc:subject>
            <rdf:Seq>
               <rdf:li>Elitsa</rdf:li>
            </rdf:Seq>
         </dc:subject>
      </rdf:Description>
   </rdf:RDF>
</x:xmpmeta>
```

The majority of my images seem to be on the correct location on the map, but this 1k are not. I have no idea why.

### The OS that Immich Server is running on

Ubuntu 22.x

### Version of Immich Server

v116.2

### Version of Immich Mobile App

latest

### Platform with the issue

- [X] Server
- [X] Web
- [X] Mobile

### Your docker-compose.yml content

```YAML
https://github.com/alexbelgium/hassio-addons/tree/master/immich
```

### Your .env content

```Shell
https://github.com/alexbelgium/hassio-addons/tree/master/immich
```

### Reproduction steps

Add the gps data from the xml file, see above, and you see the image will be located in Lybia, not on Fuerteventrura,

[...]

PR #13246, opened 2024-10-07
Add support for GPSLatitudeRef and GPSLongitudeRef in sidecar files.

aehiggi:
Changes to metadata service and repository to handle GPSLat/LongRef tags in sidecar files.

Please see comments in #13053.


zackpollard:
Still a draft and not the direction we want to go, so I am going to close this. Looking at the linked issue, someone mentions exiftool may be adding something that helps with this case anyway.

PR #13256, opened 2024-10-07
fix(mobile): do not removed not backup asset when selecting the correspond options

### LINKED ISSUE #11428: bug(mobile): delete from device doesn't work properly
### The bug

There is fortunately a display bug only, that deletes all selected assets including non backed ones when "Delete Backed Up Only" is selected. See video.
When Immich is restarted all allegedly deleted non backed up assets show up again.

|deleting|after restart|
|-|-|
|<video src="https://github.com/user-attachments/assets/ea639bc9-78b8-4385-8cab-c5bc8079d2e3">|Screenshot_20240729_100700|

What's wrong?
- Immich temprarily removes non backed up assets from view
- shows wrong number of deleted assets in the toats

Fedora 40

### Version of Immich Server

v.1.110.0

v.1.110.0

### Platform with the issue

- [ ] Server
- [ ] Web
- [X] Mobile

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

```bash
- select mixture of backed and non backed up assets
- Delete from device - select Delete Backed Up Only
- all selected assets including non backed up are allegedly deleted
```

### Relevant log output

_No response_

### Additional information

_No response_

yashrajjain726:
Fixed #11428


mobile/lib/widgets/asset_grid/multiselect_grid.dart
```diff
@@ -204,11 +204,26 @@ class MultiselectGrid extends HookConsumerWidget {
...
+            msg: 'No backed-up assets selected for deletion.',
+            gravity: ToastGravity.BOTTOM,
+          );
+          return;
+        }
+
+        // Proceed with deletion of filtered assets
         final isDeleted = await ref
             .read(assetProvider.notifier)
             .deleteLocalOnlyAssets(localIds, onlyBackedUp: onlyBackedUp);
```
shenlong-tanwen, line 225: The `onlyBackup` parameter is already handled in `deleteLocalOnlyAssets` method. The fix to this issue should ideally be made inside the method.

shenlong-tanwen (changes requested):

yashrajjain726:
@shenlong-tanwen updated the implementation inside the deleteLocalOnlyAssets func:

yashrajjain726:
@shenlong-tanwen  Any update on review ?

mobile/lib/providers/asset.provider.dart
```diff
@@ -78,27 +80,48 @@ class AssetNotifier extends StateNotifier<bool> {
...
+          msg: 'No backed-up assets selected for deletion.',
+          gravity: ToastGravity.BOTTOM,
+        );
+        return false;
+      }
+
+      // Continue with deletion of the filtered assets
       final assets = onlyBackedUp
-          ? deleteAssets.where((e) => e.storage == AssetState.merged)
-          : deleteAssets;
```
shenlong-tanwen, line 89: This section of code already filters the total list passed based on the backed up status of the asset. i.e, `e.storage == AssetState.merged` essentially filters only those that are already backed up. You added the same filtering more explicitly before this thus making this iteration useless

mobile/lib/providers/asset.provider.dart
```diff
@@ -78,27 +80,48 @@ class AssetNotifier extends StateNotifier<bool> {
...
       final localDeleted = await _deleteLocalAssets(assets);
       if (localDeleted.isNotEmpty) {
-        final localOnlyIds = deleteAssets
+        final localOnlyIds = assetsToDelete
             .where((e) => e.storage == AssetState.local)
             .map((e) => e.id)
             .toList();
         // Update merged assets to remote only
-        final mergedAssets =
-            deleteAssets.where((e) => e.storage == AssetState.merged).map((e) {
```
shenlong-tanwen, line 98: This portion of the code is actually where the logical error is. So if we do not need the toast message if there are not assets to delete, the entirety of the change would only be switching the list from `deleteAssets` to `assets` in the current implementation

Basically, the issue is that, we filter the list of assets and get the final list to delete but when making the changes to the local DB, we filter the assets again from the original list passed and not the filtered list

mobile/lib/providers/asset.provider.dart
```diff
@@ -78,27 +80,48 @@ class AssetNotifier extends StateNotifier<bool> {
...
+          .where((e) => e.isLocal && (!onlyBackedUp || e.isRemote))
+          .toList();
+
+      if (assetsToDelete.isEmpty) {
+        // No assets to delete, exit early
+        ImmichToast.show(
+          context: context,
+          msg: 'No backed-up assets selected for deletion.',
+          gravity: ToastGravity.BOTTOM,
+        );
```
shenlong-tanwen, line 102: Toast handling can be made in the widget and not the provider. In general, we shouldn't be passing `BuildContext` inside business logic unless we have no other option.

mobile/lib/widgets/asset_grid/multiselect_grid.dart
```diff
@@ -204,10 +204,11 @@ class MultiselectGrid extends HookConsumerWidget {
       processing.value = true;
       try {
         final localIds = selection.value.where((a) => a.isLocal).toList();
```
shenlong-tanwen, line 206: The best way to handle this is to use the `onlyBackedUp` flag to filter assets directly here instead of having the filtering logic in the provider. Having the logic here will also allow you to show the toast when there are no assets to delete.

shenlong-tanwen (changes requested):
You could do one of the following:

i) You could use your previous implementation of having the filtering logic in the widget and remove all the filtering logic in the provider as it is redundant and has errors
ii) Fix the logical issue in the provider and remove your filtering logic and toast handling in the widget as the assets are already filtered in the provider

Also fix any issues from the dart analyzer as well

Apologies for not being more clearer with the previous comment. Thanks for working on this.

yashrajjain726:
Thanks @shenlong-tanwen , I have updated the logic as per requirement. And, also removed UI logic from the provider code, as you mentioned and have implemented it from code perpective

mobile/lib/widgets/asset_grid/multiselect_grid.dart
```diff
@@ -203,19 +203,33 @@ class MultiselectGrid extends HookConsumerWidget {
...
-                .tr(args: ["${localIds.length}"]),
+                .tr(args: ["$deletedCount"]),
             gravity: ToastGravity.BOTTOM,
           );
+
+          // Reset the selection
           selectionEnabledHook.value = false;
+
+          // Trigger a refresh to update the UI after deletion
+          await ref.read(assetProvider.notifier).getAllAsset();
```
shenlong-tanwen, line 232: Why is this needed?

shenlong-tanwen (commented):
Kindly fix the dart analysis issues. i.e, there are unused imports in the provider file

shenlong-tanwen:
@yashrajjain726 The changes are good.Can you respond to the only remaining comment on why the refresh is needed?

yashrajjain726:
I have replied to it long time ago, please check.

yashrajjain726:
Hi @shenlong-tanwen , Can you please check again, I have updated as required.

shenlong-tanwen (approved):
Looks great! Thank you

yashrajjain726:
@alextran1502 Any update on the review ?

PR #13259, opened 2024-10-07
fix(server): Allow commas and braces in import paths

### LINKED ISSUE #10699: Having a comma in a path added to external library breaks scanning
### The bug

When having multiple paths that have commas in them added to an external library breaks the library and it can't access any assets.

For example, I have an External library named test1:

- I have added a path /projects/2024/06/cars/export
- This library works

The second example I have an External library named test2:

- I have added a path /projects/2024/06/cars/export, 28.6.2024
- This library works

Third example I have an External library named test3:

- I have added multiple paths:
  - /projects/2024/06/red-cars/export, 28.6.2024
  - /projects/2024/06/white-cars/export, 28.6.2024
- This library won't find any assets because I have more than one path with a comma. I would have to guess that somewhere in the backend something gets split by commas.

For the record, I have been using commas in folders for a long time and it is an OS issue.

Docker 20.10.24, Docker Compose v2.21.0, Unraid 6.12.3 (kernal  6.1.38-Unraid )

### Version of Immich Server

v1.106.2

N/A

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
I don't have one
```

### Your .env content

```Shell
I don't have one
```

### Reproduction steps

```bash
1. Add a new external library
2. Somewhere on the filesystem create two or more folders containing commas (for example test,1 and test,2)
3. Add *each* folder separately to the *same* external library 
4. Watch how the external library won't detect any assets
```

### Relevant log output

```shell
[Nest] 200  - 06/29/2024, 8:49:14 AM   DEBUG [Microservices:LibraryService] Found 0 asset(s) when crawling import paths /projects/2024/06/28.6.2024, Export
```

### Additional information

_No response_

etnoy:
Fixes #10699

### DIFF
--- a/server/src/repositories/storage.repository.ts
+++ b/server/src/repositories/storage.repository.ts
@@ -156,7 +156,9 @@ export class StorageRepository implements IStorageRepository {
       return Promise.resolve([]);
     }
 
-    return glob(this.asGlob(pathsToCrawl), {
+    const globbedPaths = pathsToCrawl.map((path) => this.asGlob(path));
+
+    return glob(globbedPaths, {
       absolute: true,
       caseSensitiveMatch: false,
       onlyFiles: true,
@@ -172,7 +174,9 @@ export class StorageRepository implements IStorageRepository {
       return emptyGenerator();
     }
 
-    const stream = globStream(this.asGlob(pathsToCrawl), {
+    const globbedPaths = pathsToCrawl.map((path) => this.asGlob(path));
+
+    const stream = globStream(globbedPaths, {
       absolute: true,
       caseSensitiveMatch: false,
       onlyFiles: true,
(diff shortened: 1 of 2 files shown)


server/src/repositories/storage.repository.ts
```diff
@@ -172,10 +174,15 @@ export class StorageRepository implements IStorageRepository {
...
-    const stream = globStream(this.asGlob(pathsToCrawl), {
+    const globbedPaths = pathsToCrawl.map((path) => this.asGlob(path));
+
+    this.logger.debug(globbedPaths);
+
+    const stream = globStream(globbedPaths, {
       absolute: true,
       caseSensitiveMatch: false,
       onlyFiles: true,
+      unique: true,
```
mertalev, line 185: Don't add this. It makes the glob slower and use more memory.

etnoy: Ok, will remove!

jrasm91 (approved):
LGTM. This problem also exists for the CLI.

etnoy:
> LGTM. This problem also exists for the CLI.

Good catch! It's on my todo

PR #13283, opened 2024-10-08
fix(server): parse ExifDuration

### LINKED ISSUE #13095: Video duration from XMP persisted as "[object object]"
### The bug

I was getting an error on the web when a single video came into view. The js error (below) was erroring because the assets.Duration field was set to [object object] in the database. 
If I remove the xmpDM:duration element from the sidecar file and refresh the metadata the correct duration of 00:22:11.360 is saved to the table and the error in the UI goes away. I suspect this is because the duration is being read from the file and the sidecar file and it's creating an array?

The fix for me was to remove the element from the sidecar file but I wanted to raise it because I don't know how many other files will have this issue.

`Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'split')`
from the below function
`function bl(t) {
    const e = t.split(":");
    e[2] = e[2].split(".").slice(0, 2).join(".");
    const [l,a,n] = e.map(Number);
    return We.fromObject({
        hours: l,
        minutes: a,
        seconds: n
    }).as("seconds")
}`

Ubuntu 22.04

### Version of Immich Server

v1.116.2

v1.116.1 build.177

### Platform with the issue

- [X] Server
- [X] Web
- [ ] Mobile

```YAML
immich:
    container_name: immich
    #command: [ "start.sh", "immich" ]
    volumes:
      - ./immich/library:/usr/src/app/upload/library
      - ./immich/cache/upload:/usr/src/app/upload/upload
      - ./immich/cache/thumbs:/usr/src/app/upload/thumbs
      - ./immich/cache/profile:/usr/src/app/upload/profile
      - ./immich/cache/encoded-video:/usr/src/app/upload/encoded-video
      - ${HOMEDATA}/Pictures:/photos:ro
    env_file:
      - .env
    ports:
      - 2283:3001
    environment:
      DB_USERNAME: ${ADMIN_USER}
      DB_PASSWORD: ${ADMIN_PASSWORD}
      DB_DATABASE_NAME: immich
      DB_HOSTNAME: immich_postgres
      REDIS_HOSTNAME: immich_redis
      TZ: ${TIMEZONE}
    depends_on:
      - immich_redis
      - immich_postgres
    restart: always
    labels:
      autoheal-app: true
  
  immich-machine-learning: #needs to be hyphenated otherwise the url doesn't work
    volumes:
      - ./immich/library:/usr/src/app/upload/library
      - ./immich/cache/upload:/usr/src/app/upload/upload
      - ./immich/cache/thumbs:/usr/src/app/upload/thumbs
      - ./immich/cache/profile:/usr/src/app/upload/profile
      - ./immich/cache/encoded-video:/usr/src/app/upload/encoded-video
      - ./immich-machine-learning/model-cache:/cache
    restart: always
    environment:
      TZ: ${TIMEZONE}
    labels:
      autoheal-app: true
  
  immich_redis:
    image: docker.io/redis:6.2-alpine@sha256:d6c2911ac51b289db208767581a5d154544f2b2fe4914ea5056443f62dc6e900
    container_name: immich_redis
    restart: always
    environment:
      TZ: ${TIMEZONE}
    healthcheck:
      test: ["CMD", "redis-cli", "ping"]
      interval: 10s
      timeout: 5s
      retries: 5
    volumes:
      - ./immich-redis:/data
    labels:
      autoheal-app: true
  
  immich_postgres:
    environment:
      POSTGRES_PASSWORD: ${ADMIN_PASSWORD}
      POSTGRES_USER: ${ADMIN_USER}
      POSTGRES_DB: immich
    ports:
      - "5432:5432"
    volumes:
      - ./immich-postgres:/var/lib/postgresql/data
    restart: always
    labels:
      autoheal-app: true
```

### Your .env content

```Shell
IMMICH_VERSION=v1.116.2
ADMIN_USER=***
ADMIN_PASSWORD=*********
USER_ID=1000
GROUP_ID=1000
TIMEZONE="Europe/London"
HOMEDATA="/mnt/home"
```

### Reproduction steps

1. Add `<xmpDM:duration xmpDM:scale="1/90000" xmpDM:value="119827200"/>` to the sidecar file.
2. Refresh metadata for that particular file.
3. See error in UI when scrolling past the video.
4. I tested with another video and got the same results.

### Relevant log output

_No response_

### Additional information

_No response_

C-Otto:
fixes #13095

### DIFF
--- a/server/src/services/metadata.service.ts
+++ b/server/src/services/metadata.service.ts
@@ -26,7 +26,7 @@ import {
   QueueName,
 } from 'src/interfaces/job.interface';
 import { ReverseGeocodeResult } from 'src/interfaces/map.interface';
-import { ImmichTags } from 'src/interfaces/metadata.interface';
+import { ExifDuration, ImmichTags } from 'src/interfaces/metadata.interface';
 import { BaseService } from 'src/services/base.service';
 import { isFaceImportEnabled } from 'src/utils/misc';
 import { usePagination } from 'src/utils/pagination';
@@ -243,7 +243,7 @@ export class MetadataService extends BaseService {
 
     await this.assetRepository.update({
       id: asset.id,
-      duration: exifTags.Duration?.toString() ?? null,
+      duration: this.durationToString(exifTags.Duration),
       localDateTime,
       fileCreatedAt: exifData.dateTimeOriginal ?? undefined,
     });
@@ -332,6 +332,18 @@ export class MetadataService extends BaseService {
     }
 
     return JobStatus.SUCCESS;
+  }
+
+  private durationToString(duration?: string | number | ExifDuration) {
+    if (!duration) {
+      return null;
+    }
+    if (!duration.hasOwnProperty('Value')) {
+      return duration.toString();
+    }
+    const exifDuration = duration as ExifDuration;
+    const seconds = exifDuration.Value * (exifDuration.Scale ?? 1);
+    return Duration.fromObject({ seconds }).toFormat('hh:mm:ss.SSS');
   }
 
   private async getExifTags(asset: AssetEntity): Promise<ImmichTags> {
--- a/server/src/services/metadata.service.spec.ts
+++ b/server/src/services/metadata.service.spec.ts
@@ -1128,6 +1128,25 @@ describe(MetadataService.name, () => {
         }),
       );
     });
+
+    it('should handle video durations parsed as ExifDuration with value and scale', async () => {
+      assetMock.getByIds.mockResolvedValue([assetStub.video]);
+      mediaMock.probe.mockResolvedValue(probeStub.videoStreamVertical2160p);
+      metadataMock.readTags.mockResolvedValue({
+        Duration: {
+          Value: 119_827_200,
+          Scale: 0.000_011_111_111_111_111_1, // 1/90000
+        },
+      });
+
+      await sut.handleMetadataExtraction({ id: assetStub.video.id });
+      expect(assetMock.update).toHaveBeenCalledWith(
+        expect.objectContaining({
+          // value * scale (in seconds)
+          duration: '00:22:11.413',
+        }),
+      );
+    });
   });
 
   describe('handleQueueSidecar', () => {


server/src/services/metadata.service.ts
```diff
@@ -334,6 +334,19 @@ export class MetadataService extends BaseService {
...
+    if (!duration) {
+      return null;
+    }
+
+    if (duration.hasOwnProperty('Value')) {
+      const exifDuration = duration as ExifDuration;
+      const seconds = exifDuration.Value * (exifDuration.Scale ?? 1);
+      return Duration.fromObject({ seconds }).toFormat('hh:mm:ss.SSS');
+    }
+    return duration?.toString();
```
danieldietzler, line 347: I'd probably flip the condition to keep the actual block within the if small

danieldietzler (approved):

C-Otto:
This seems to be a regression from #11610 and #4541. @jrasm91 and @michelheusschen could you please have a look?

C-Otto:
I tried, but I don't know how to create a random video (and the test won't work for images).

server/src/services/metadata.service.ts
```diff
@@ -334,6 +334,18 @@ export class MetadataService extends BaseService {
     return JobStatus.SUCCESS;
   }
 
+  private durationToString(duration?: string | number | ExifDuration) {
+    if (!duration) {
+      return null;
+    }
+    if (!duration.hasOwnProperty('Value')) {
+      return duration?.toString();
```
michelheusschen, line 342: Have you confirmed that returning `undefined` instead of `null` still works as expected? Maybe if there was a duration previously, but none after rerunning metadata extraction, it might not get updated correctly now.

C-Otto: No, good catch. I'll fix this. I'm not used to TypeScript (and I don't think I will ever get used to having null, 0, false, false-y, undefined to pick from).

C-Otto: Ah, actually duration is non-null-non-undefined at this point, so it's not an issue. I can remove the "?".

C-Otto:
Please also consider the logic itself, not just the code change. I believe that number inputs (seconds?) and strings that don't match the '00:00:00.000' format would also cause issues. As you, @michelheusschen and @jrasm91, have dealt with this logic before, could you please check that this is something we should do?

jrasm91:
It turns out that we should just ignore `Duration` if we find it in the sidecar file.

PR #13288, opened 2024-10-08
fix(server): People count after updating recognized faces count

### LINKED ISSUE #13245: People count doesn't respect minimum recognized faces
### The bug

Faces are shown according "Minimum recognized faces" settings, but People count is not affected, it still shows the same count.
If minimum is set to 1, faces and count is shown as 161,
if minimum is set to 2, 138 faces is shown, count is still 161.

obrazek
obrazek

Fedora 40

### Version of Immich Server

v1.117.0

not relevant

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

-

### Relevant log output

_No response_

### Additional information

_No response_

jedi04:
This PR fixes #13245

It updates the query used to fetch the total count of people keeping in consideration the recognized faces setting.

### DIFF
--- a/server/src/interfaces/person.interface.ts
+++ b/server/src/interfaces/person.interface.ts
@@ -78,7 +78,7 @@ export interface IPersonRepository {
   getRandomFace(personId: string): Promise<AssetFaceEntity | null>;
   getStatistics(personId: string): Promise<PersonStatistics>;
   reassignFace(assetFaceId: string, newPersonId: string): Promise<number>;
-  getNumberOfPeople(userId: string): Promise<PeopleStatistics>;
+  getNumberOfPeople(userId: string, options: PersonSearchOptions): Promise<PeopleStatistics>;
   reassignFaces(data: UpdateFacesData): Promise<number>;
   unassignFaces(options: UnassignFacesOptions): Promise<void>;
   update(person: Partial<PersonEntity>): Promise<PersonEntity>;
--- a/server/src/queries/person.repository.sql
+++ b/server/src/queries/person.repository.sql
@@ -250,22 +250,30 @@ WHERE
 
 -- PersonRepository.getNumberOfPeople
 SELECT
-  COUNT(DISTINCT ("person"."id")) AS "total",
-  COUNT(DISTINCT ("person"."id")) FILTER (
+  COUNT(person.id) AS "total",
+  COUNT(person.id) FILTER (
     WHERE
-      "person"."isHidden" = true
+      person."isHidden" = true
   ) AS "hidden"
 FROM
-  "person" "person"
-  LEFT JOIN "asset_faces" "face" ON "face"."personId" = "person"."id"
-  INNER JOIN "assets" "asset" ON "asset"."id" = "face"."assetId"
-  AND ("asset"."deletedAt" IS NULL)
-WHERE
-  "person"."ownerId" = $1
-  AND "asset"."isArchived" = false
-  AND "person"."thumbnailPath" != ''
-HAVING
-  COUNT("face"."assetId") != 0
+  (
+    SELECT
+      "person"."id" AS "id",
+      "person"."isHidden" AS "isHidden"
+    FROM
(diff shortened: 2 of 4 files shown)


jedi04:
> Thank you for the PR. Can you please run `make sql` from the root of the project?

Sure. I have run the command from root of project.

Thank you

hitech95:
@jedi04 This open another series of issues.
@alextran1502 Imho this is a workaround on a deeper issue. See #13551

zackpollard:
Hey @jedi04, I appreciate this has been open for quite a while. Just trying to go through some of our older PRs and cleanup. Is this ready to merge? I did a quick look through and I didn't really understand why you changed the SQL to have a subquery. From a brief look I feel like you could've just changed the HAVING statement to handle the minimum faces setting. Is there any other advantage to changing the query in this way?

zackpollard:
> @jedi04 This open another series of issues.

Could you expand on what other issues this opens in your eyes?

hitech95:
> > @jedi04 This open another series of issues.
> 
> Could you expand on what other issues this opens in your eyes?

See #13551

zackpollard:
> > > @jedi04 This open another series of issues.
> > 
> > 
> > Could you expand on what other issues this opens in your eyes?
> 
> See #13551

I saw that issue but it isn't really clear to me why this fix is directly a problem. It sounds like there are more issues in this area, but this PR feels like a bug regardless, it doesn't need to fix all of the problems. I think it's fine to look to merge this change but also look at a broader solution to the problems you mention separately.

hitech95:
> > > > @jedi04 This open another series of issues.
> > > 
> > > 
> > > Could you expand on what other issues this opens in your eyes?
> > 
> > 
> > See #13551
> 
> I saw that issue but it isn't really clear to me why this fix is directly a problem. It sounds like there are more issues in this area, but this PR feels like a bug regardless, it doesn't need to fix all of the problems. I think it's fine to look to merge this change but also look at a broader solution to the problems you mention separately.

This only fix the returned counte of the people, by limiting the real count to those that only have a at least N faces.
it does not fix teh underlaying issue that the person is created but the faces are not attached to it.

If I have a person Y with 3 faces the person is created. if I delete one of the pictures the person Y still exist. but the counter will return a lower count (as it dont meet the threshold). I could have already named the person.
The result would be still out of sink. As I can still navigate to the person and fetch the related pictures from the timeline.

IMHO: limiting the count is not the correct way to fix this, the correct way is to ensure that the DB is consistent with the data.

zackpollard:
> > > > > @jedi04 This open another series of issues.
> > > > 
> > > > 
> > > > Could you expand on what other issues this opens in your eyes?
> > > 
> > > 
> > > See #13551
> > 
> > 
> > I saw that issue but it isn't really clear to me why this fix is directly a problem. It sounds like there are more issues in this area, but this PR feels like a bug regardless, it doesn't need to fix all of the problems. I think it's fine to look to merge this change but also look at a broader solution to the problems you mention separately.
> 
> This only fix the returned counte of the people, by limiting the real count to those that only have a at least N faces. it does not fix teh underlaying issue that the person is created but the faces are not attached to it.
> 
> If I have a person Y with 3 faces the person is created. if I delete one of the pictures the person Y still exist. but the counter will return a lower count (as it dont meet the threshold). I could have already named the person. The result would be still out of sink. As I can still navigate to the person and fetch the related pictures from the timeline.
> 
> IMHO: limiting the count is not the correct way to fix this, the correct way is to ensure that the DB is consistent with the data.

I now understand the issue. In that case I agree, and it's probably better to fix that underlying issue then. In which case I think I will close this issue as it is a bandaid to the wider issue.

mlinko6:
Hi,

I have the same issue, but I'm totally lost in finding a solution... What should I do in this case? My minimal face recognition is set to 3... How can I solve the issue?

PR #13294, opened 2024-10-08
fix(server): duplicate faces, face insert query failing

mertalev:
## Description

The face detection logic can add a duplicate face when there is a machine learning-sourced match. The query to refresh faces is also buggy because TypeORM doesn't parameterize the second insert cte correctly... unless you call `getQuery` first.

## How Has This Been Tested?

Tested that refreshing and resetting with different thresholds work without any errors or duplicate faces.


danieldietzler (approved):
I find this hilarious lol

zackpollard (approved):
Kysely when?

PR #13320, opened 2024-10-09
Fix/mobile zh tw translation

ShinjuGoenji:
add translation of traditional chinese in mobile app (zh-TW)


ShinjuGoenji:
changelog: add translation of traditional chinese in mobile app (zh-TW)

PR #13342, opened 2024-10-10
fix(web): rtl text on map is hidden

zackpollard:


web/package.json
```diff
@@ -67,7 +67,7 @@
   "dependencies": {
     "@formatjs/icu-messageformat-parser": "^2.7.8",
     "@immich/sdk": "file:../open-api/typescript-sdk",
-    "@mapbox/mapbox-gl-rtl-text": "^0.3.0",
+    "@mapbox/mapbox-gl-rtl-text": "0.2.3",
```
danieldietzler, line 70: So this was a regression in the library?

danieldietzler (approved):

web/package.json
zackpollard: The library failed to load because they moved to compiling it as WASM. I tried fixing it when I initially added this and couldn't get it to work then either.

jrasm91: You need to separate this dependency out in renovate, etc. to prevent it from automatically being included in future updates, right?

PR #13345, opened 2024-10-10
chore: database service unit tests

danieldietzler:
Screenshot 2024-10-10 at 14 56 25
Screenshot 2024-10-10 at 14 57 26

### DIFF
--- a/server/vitest.config.mjs
+++ b/server/vitest.config.mjs
@@ -17,10 +17,10 @@ export default defineConfig({
         'src/services/index.ts',
       ],
       thresholds: {
-        lines: 80,
-        statements: 80,
-        branches: 85,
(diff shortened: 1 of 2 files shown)


server/vitest.config.mjs
```diff
@@ -17,10 +17,10 @@ export default defineConfig({
...
       ],
       thresholds: {
-        lines: 80,
-        statements: 80,
-        branches: 85,
-        functions: 80,
+        lines: 85,
+        statements: 85,
+        branches: 90,
+        functions: 85,
```
danieldietzler, line 23: Also bumped the thresholds :)

jrasm91 (approved):

PR #13359, opened 2024-10-10
Add step logging to test files for better structure

artshllk:
## Description

In this PR, I have added step-based logging across all test files to improve readability, clarity, and test structure. This will help us better understand each step during test execution and debugging, ensuring that each phase of the test is explicitly logged.

## Proposed Improvement

I also suggest we consider refactoring our approach to selectors:

Move all selectors to a class-based structure: By placing all selectors in a dedicated class, we can centralize their management, making it easier to update or modify them when needed.
Keep logic inside the test body: This would allow us to keep the test logic focused on behavior, improving maintainability and readability.
If the team agrees, I’m happy to create a new PR to implement this approach.

### DIFF
--- a/e2e/src/step.ts
+++ b/e2e/src/step.ts
@@ -1,0 +1,8 @@
+/**
+ * Tiny function to mark the action performed by the test in console.
+ *
+ * @param text name of the step performed by the test
+ */
+export function step(text: string) {
+  console.log(`🚩 STEP: ${text}`);
+}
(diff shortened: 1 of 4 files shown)


artshllk:
To further enhance our code quality and maintainability, we might consider adding some additional ESLint rules in the future. Here are a few suggestions:

`eqeqeq`: Enforce strict equality checks.
`no-console`: Disallow console statements for cleaner output.
`@typescript-eslint/no-unused-vars`: Help catch unused variables more effectively.
`@typescript-eslint/no-non-null-assertion`: Reduce potential null/undefined errors.
`prefer-const`: Encourage the use of const for variables that are never reassigned.

If anyone thinks this would be beneficial, just let me know, and I can create a PR for these updates!

mertalev:
Hi! Thanks for the PR, but I don't think the changes here add much value. It increases noise for successful tests and failed tests already show the line where they failed.

PR #13368, opened 2024-10-11
feat: Add support for preloading textual clip model without visual model

1-tempest:
Preloading the textual clip model is required for quick smart searches. 

I'd like to suggest having an ability to load the textual model without the dependency of the visual model to preserve system ram.


bo0tzz:
How does this PR relate to #12729?

PR #13377, opened 2024-10-11
fix(server): wrong image dimensions for RAW files (RAF, CR2) (also fixes face preview)

### LINKED ISSUE #13049: RAW files (RAF, CR2) have wrong face preview due to wrong dimensions
### The bug

This bug differs from https://github.com/immich-app/immich/issues/12440 in that the preview box correctly surrounds the person's face, but the preview image that shows in the People tab (or in file details) is not zoomed/cropped to the correct location in a photo.

This issue is fixed when manually selecting "Select Featured Photo." This cases the profile picture/preview for that particular person to be cropped correctly.

SCR-20240930-dbig-2
*(note - the image shown vs. source for "featured thumbnail" are slightly different photos (taken a few seconds apart), but the effect happens with that photo as well and I didn't want to re-blur for anonymity)*

image

*The overall \"People\" tab. Every one of these is a validly recognized person, the thumbnail ("Featured Photo") is just misaligned.*

Windows 10 Server

### Version of Immich Server

v1.116.2

n/a

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
#
#
#
#

name: immich

services:
  immich-server:
    container_name: immich_server
    # extends:
    volumes:
      - "D:/nas/Media:/home/user/photos1:ro"
    env_file:
      - .env
    ports:
      - 2283:3001
    depends_on:
      - redis
      - database
    restart: always
    healthcheck:
      disable: false

  immich-machine-learning:
    #   file: hwaccel.ml.yml
    volumes:
      - model-cache:/cache
    env_file:
      - .env
    restart: always
    healthcheck:
      disable: false
      
  redis:
    container_name: immich_redis
    image: docker.io/redis:6.2-alpine@sha256:2d1463258f2764328496376f5d965f20c6a67f66ea2b06dc42af351f75248792
    healthcheck:
    restart: always

  database:
    environment:
      POSTGRES_USER: ${DB_USERNAME}
    volumes:
      # Do not edit the next line. If you want to change the database storage location on your system, edit the value of DB_DATA_LOCATION in the .env file
    healthcheck:
      test: pg_isready --dbname='${DB_DATABASE_NAME}' --username='${DB_USERNAME}' || exit 1; Chksum="$$(psql --dbname='${DB_DATABASE_NAME}' --username='${DB_USERNAME}' --tuples-only --no-align --command='SELECT COALESCE(SUM(checksum_failures), 0) FROM pg_stat_database')"; echo "checksum failure count is $$Chksum"; [ "$$Chksum" = '0' ] || exit 1
      interval: 5m
      start_interval: 30s
      start_period: 5m
    command: ["postgres", "-c", "shared_preload_libraries=vectors.so", "-c", 'search_path="$$user", public, vectors', "-c", "logging_collector=on", "-c", "max_wal_size=2GB", "-c", "shared_buffers=512MB", "-c", "wal_compression=on"]
    restart: always

volumes:
  model-cache:
```

### Your .env content

```Shell

UPLOAD_LOCATION="D:/immich-library"
DB_DATA_LOCATION=./postgres

# TZ=America/[]

IMMICH_VERSION=release

# Please use only the characters `A-Za-z0-9`, without special characters or spaces
DB_PASSWORD=[redacted]

DB_USERNAME=postgres
DB_DATABASE_NAME=immich
```

### Reproduction steps

The issue:
1. Open Photos tab
2. Witness faces not being centered

The manual fix:
1. Click on a person in the People tab (or from elsewhere)
2. Top right menu->Select Featured Photo
3. (Choose a photo)
4. See that Featured Photo now appears correctly system-wide

### Relevant log output

_No response_

### Additional information

_No response_

C-Otto:
For some RAW formats/cameras, the ImageWidth/ImageHeight information represents the dimensions of the preview image, i.e. much smaller than the actual image. This also causes issues with (face/people) thumbnail generation. The information in ImageSize seems to be correct, so we use this instead.

fixes #13049


server/src/services/metadata.service.ts
```diff
@@ -334,6 +336,29 @@ export class MetadataService extends BaseService {
...
+      // width x height (e.g. 100x200)
+      const split = imageSize.split('x');
+      width = Number.parseInt(split[0]);
+      height = Number.parseInt(split[1]);
+    }
+    if (!width || !height) {
+      width = exifTags.ImageWidth ?? Number.NaN;
+      height = exifTags.ImageHeight ?? Number.NaN;
+    }
+    return { width, height };
```
mertalev, line 359:
```suggestion
    let [width, height] = exifTags.ImageSize?.split('x') || [];
    if (!width || !height) {
      [width, height] = [exifTags.ImageWidth, exifTags.ImageHeight];
    }
    return { width: Number.parseInt(width), height: Number.parseInt(height) };
```

server/src/services/metadata.service.ts
```diff
@@ -192,6 +192,8 @@ export class MetadataService extends BaseService {
     const { dateTimeOriginal, localDateTime, timeZone, modifyDate } = this.getDates(asset, exifTags);
     const { latitude, longitude, country, state, city } = await this.getGeo(exifTags, reverseGeocoding);
 
+    const imageDimensions = this.getImageDimensions(exifTags);
```
mertalev, line 195:
```suggestion
    const { width, height } = this.getImageDimensions(exifTags);
```

server/src/services/metadata.service.ts
```diff
@@ -209,8 +211,8 @@ export class MetadataService extends BaseService {
 
       // image/file
       fileSizeInByte: stats.size,
-      exifImageHeight: validate(exifTags.ImageHeight),
-      exifImageWidth: validate(exifTags.ImageWidth),
+      exifImageHeight: validate(imageDimensions.height),
+      exifImageWidth: validate(imageDimensions.width),
```
mertalev, line 215:
```suggestion
      exifImageHeight: validate(height),
      exifImageWidth: validate(width),
```

mertalev (approved):
Looks good! I just made a few suggestions to make it more idiomatic.

C-Otto:
I'll have a look at your suggestions, thanks. While other weird RAW formats keep popping up, I'm pausing work on this (see discussion in #13049).

server/src/services/metadata.service.ts
C-Otto: That won't work due to different types. I'd rather have long code that's readable than short code that's a mystery. This isn't perl, after all.

danieldietzler: I wouldn't consider Mert's suggestion mystery. I think it's a pretty elegant and idiomatic solution that's very well readable. If you need to juggle around with types however I can see that this becomes less pretty.

C-Otto: Well, I do. It doesn't work and uses typecasts (from number to string?). To me, that's quite a mystery.

mertalev: The cast is to cast `string | undefined` to string. The typing for `parseInt` only accepts string, but an undefined input results in NaN. You can avoid the type cast by changing `as string` to `|| ''`.

mertalev: Revised suggestion that handles typing more elegantly:
```suggestion
    let [width, height] = exifTags.ImageSize?.split('x').map((dim) => Number.parseInt(dim)) || [];
    if (!width || !height) {
      [width, height] = [exifTags.ImageWidth, exifTags.ImageHeight];
    }
    return { width, height };
```

C-Otto: That still doesn't work.

C-Otto: Maybe it's complex enough to prefer longer code that actually works (without employing elegant/mysterious code)? Just a suggestion.

mertalev: Can you clarify what doesn't work? I ran both functions in a REPL and they behave the same (except the suggested code uses `undefined` instead of `NaN`). The only change I needed to make for TS is adding a type hint for `dim` (quite normal).

Regardless, this is just a suggestion: you can keep the current code since this is what you prefer. I just think the suggestion is cleaner and more readable for something that (imo) isn't particularly complex.

C-Otto: It doesn't compile due to type issues. I realize that this is a suggestion, but I prefer suggestions that actually work. I see that your code is shorter and possibly also cleaner, but that won't help if it doesn't work. If you're unsure of how to provide working suggestions, I suggest in return to try them out yourself before.

server/src/services/metadata.service.ts
```diff
@@ -334,6 +336,29 @@ export class MetadataService extends BaseService {
...
+      const split = imageSize.split('x');
+      width = Number.parseInt(split[0]);
+      height = Number.parseInt(split[1]);
+    }
+    if (!width || !height) {
+      width = exifTags.ImageWidth ?? Number.NaN;
+      height = exifTags.ImageHeight ?? Number.NaN;
+    }
+    return { width, height };
+  }
```
mertalev, line 360:
```suggestion
  private getImageDimensions(exifTags: ImmichTags): { width?: number; height?: number } {
    /*
     * The "true" values for width and height are a bit hidden, depending on the camera model and file format.
     * For RAW images in the CR2 or RAF format, the "ImageSize" value seems to be correct,
     * but ImageWidth and ImageHeight are not correct (they contain the dimensions of the preview image).
     */

    let [width, height] = exifTags.ImageSize?.split('x').map((dim) => Number.parseInt(dim) || undefined) || [];
    if (!width || !height) {
      [width, height] = [exifTags.ImageWidth, exifTags.ImageHeight];
    }
    return { width, height };
  }
```

C-Otto:
@mertalev :rocket:?

mertalev (approved):

PR #13380, opened 2024-10-11
chore: Add viteconfig-timestamp to .gitignore

etnoy:
Ignore the spurious vite configs that sometimes pop up after testing
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ mobile/openapi/.openapi-generator/FILES
 open-api/typescript-sdk/build
 mobile/android/fastlane/report.xml
 mobile/ios/fastlane/report.xml
+
+vite.config.js.timestamp-*


jrasm91 (approved):

.gitignore
danieldietzler, line 1: The files in your screenshot are all in `web/`, so you should also edit the gitignore file over there. In fact, I don't even think this works.

danieldietzler (commented):
I have never run into those lol

.gitignore
jrasm91: Probably should just do **/...

danieldietzler: Or that, yea

etnoy: This works, all dangling config files were gone from the git list after this change. .gitignore does not need a leading **/

These files are sometimes left over after a failed e2e test. They're supposed to be cleaned up but sometimes this doesn't happen and I get annoyed. Hence this PR.

I still think this should be top-level

danieldietzler: > .gitignore does not need a leading **/

Oh ok, TIL!

danieldietzler (approved):

PR #13382, opened 2024-10-11
chore: media service unit tests

danieldietzler:
Screenshot 2024-10-11 at 22 21 08

### DIFF
--- a/server/src/services/media.service.spec.ts
+++ b/server/src/services/media.service.spec.ts
@@ -1,7 +1,10 @@
 import { Stats } from 'node:fs';
+import { SystemConfig } from 'src/config';
+import { AssetEntity } from 'src/entities/asset.entity';
 import { ExifEntity } from 'src/entities/exif.entity';
 import {
   AssetFileType,
+  AssetPathType,
   AssetType,
   AudioCodec,
   Colorspace,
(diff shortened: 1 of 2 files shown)


etnoy (approved):

server/src/services/media.service.spec.ts
```diff
@@ -797,6 +873,21 @@ describe(MediaService.name, () => {
       );
     });
 
+    it('should transcode when max bitrate is not a number', async () => {
+      mediaMock.probe.mockResolvedValue(probeStub.videoStream40Mbps);
+      systemMock.get.mockResolvedValue({ ffmpeg: { transcode: TranscodePolicy.BITRATE, maxBitrate: 'foo' } });
```
mertalev, line 878: This should probably be a config validation error.

mertalev (approved):

jrasm91 (approved):

server/src/services/media.service.spec.ts
danieldietzler: Agreed, but then https://github.com/immich-app/immich/blob/main/server/src/services/media.service.ts#L465-L467 wouldn't be necessary correct?

mertalev: Yeah, that would be redundant at that point

PR #13388, opened 2024-10-12
docs: Add more documentation for video transcoding settings.

kevincox:
This adds documentation on many of the video transcoding settings `ffmpeg.*`. I focused the documentation on values that aren't just passthough to ffmpeg settings but that are custom to Immich.

### DIFF
--- a/docs/docs/administration/system-settings.md
+++ b/docs/docs/administration/system-settings.md
@@ -205,4 +205,68 @@ When this option is enabled the `immich-server` will periodically make requests 
 
 ## Video Transcoding Settings
 
-The system administrator can define parameters according to which video files will be converted to different formats (depending on the settings). The settings can be changed in depth, to learn more about the terminology used here, refer to FFmpeg documentation for [H.264](https://trac.ffmpeg.org/wiki/Encode/H.264) codec, [HEVC](https://trac.ffmpeg.org/wiki/Encode/H.265) codec and [VP9](https://trac.ffmpeg.org/wiki/Encode/VP9) codec.
+The system administrator can configure which video files will be converted to different formats. The settings can be changed in depth, to learn more about the terminology used here, refer to FFmpeg documentation for [H.264](https://trac.ffmpeg.org/wiki/Encode/H.264) codec, [HEVC](https://trac.ffmpeg.org/wiki/Encode/H.265) codec and [VP9](https://trac.ffmpeg.org/wiki/Encode/VP9) codec.
+
+Which streams of a video file will be transcoded is determined by the [Transcode Policy](#ffmpeg.transcode). Streams that are transcoded use the following settings (config file name in brackets). Streams that are not transcoded are untouched and preserve their original settings.
+
+### Accepted containers (`ffmpeg.acceptedContainers`) {#ffmpeg.acceptedContainers}
+
+If the video asset's container format is not in this list, it will be remuxed to MP4 even if no streams need to be transcoded.
+
+The default set of accepted container formats is `mov`, `ogg` and `webm`.
+
+### Preset (`ffmpeg.preset`) {#ffmpeg.preset}
+
+The amount of "compute effort" to put into transcoding. These use [the preset names from h264](https://trac.ffmpeg.org/wiki/Encode/H.264#Preset) and will be converted to appropriate values for encoders that configure effort in different ways.
+
+The default value is `ultrafast`.
+
+### Audio codec (`ffmpeg.targetAudioCodec`) {#ffmpeg.targetAudioCodec}
+
+Which audio codec to use when the audio stream is being transcoded. Can be one of `mp3`, `aac`, `libopus`.
+
+The default value is `aac`.
+
+### Video Codec (`ffmpeg.targetVideoCodec`) {#ffmpeg.targetVideoCodec}
+
+Which video codec to use when the video stream is being transcoded. Can be one of `h264`, `hevc`, `vp9` or `av1`.
+
+The default value is `h264`.
+
+### Target resolution (`ffmpeg.targetResolution`) {#ffmpeg.targetResolution}
+
+When transcoding a video stream, downscale the largest dimension to this value while preserving aspect ratio. Videos are never upscaled.
+
+The default value is `720`.
+
+### Transcode policy (`ffmpeg.transcode`) {#ffmpeg.transcode}
+
+The transcoding policy configures which streams of a video asset will be transcoded. The transcoding decision is made independently for video streams and audio streams. This means that if a video stream needs to be transcoded, but an audio stream does not, then the video stream will be transcoded while the audio stream will be copied. If the transcoding policy does not require any stream to be transcoded and does not require the video to be remuxed, then no separate video file will be created.
+
+The default policy is `required`.
+
+#### All videos (`all`) {#ffmpeg.transcode-all}
+
+Videos are always transcoded. This ensures consistency during video playback.
+
+#### Don't transcode any videos (`disabled`) {#ffmpeg.transcode-disabled}
+
+Videos are never transcoded. This saves space and resources on the server, but may prevent playback on devices that don't support the source format (especially web browsers) or result in high bandwidth usage when playing high-bitrate files.
+
+#### Only videos not in an accepted format (`required`) {#ffmpeg.transcode-required}
+
+Video streams are transcoded when any of the following conditions are met:
+
+- The video is HDR.
+- The video is not in the yuv420p pixel format.
+- The video codec is not in `acceptedVideoCodecs`.
+
+Audio is transcoded if the audio codec is not in `acceptedAudioCodecs`.
+
+#### Videos higher than max bitrate or not in an accepted format (`bitrate`) {#ffmpeg.transcode-bitrate}
+
+In addition to the conditions in `required`, video streams are also transcoded if their bitrate is over `maxBitrate`.
+
+#### Videos higher than target resolution or not in an accepted format (`optimal`) {#ffmpeg.transcode-optimal}
+
+In addition to the conditions in `required`, video streams are also transcoded if the horizontal **and** vertical dimensions are higher than [`targetResolution`](#ffmpeg.targetResolution).


server/src/enum.ts
```diff
@@ -231,10 +231,30 @@ export enum UserPathType {
 export type PathType = AssetPathType | PersonPathType | UserPathType;
 
 export enum TranscodePolicy {
+  // Videos are always transcoded.
   ALL = 'all',
+  
+  // Videos are transcoded if they don't mee the format requirements or are too high resolution.
```
crash123, line 237: meet

server/src/enum.ts
```diff
@@ -231,10 +231,30 @@ export enum UserPathType {
...
+  //
+  // Currently this means that the video is transcoded if any of the following conditions are met.
+  // - Both video dimensions are greater than `targetResolution`.
+  // - Any of the conditions for `REQUIRED`.
   OPTIMAL = 'optimal',
+  
+  // Videos are transcoded if they don't mee the format requirements or the video bitrate is over `maxBitrate`
```
crash123, line 244: meet

server/src/enum.ts
```diff
@@ -231,10 +231,30 @@ export enum UserPathType {
...
+  //
+  // Note: This is a superset of `REQUIRED`.
   BITRATE = 'bitrate',
+  
+  // Videos are only transcoded if they don't meed the format requirements.
```
crash123, line 249: meet

server/src/enum.ts
```diff
@@ -231,10 +231,30 @@ export enum UserPathType {
...
+  //
+  // Currently this means that the video is transcoded if any of the following conditions are met.
+  // - The video codec is not in `acceptedVideoCodecs`
+  // - The video is HDR.
+  // - The audia codec is not in `acceptedAudioCodecs`
```
crash123, line 254: audio

mmomjian:
I feel like this should be written up in the Immich docs, not as comments in a file.

danieldietzler (dismissed):
I agree, this shouldn't be a comment in the code as people who read the code typically know how to figure it out themselves anyways. If anything, this should help people figuring out the correct setting in the GUI

mertalev:
It can go in the transcode policy description in a more condensed form.

kevincox:
Can you clarify where you would like to see this?

kevincox:
> the use of "optimal" as the resolution-based policy has always bothered me. Calling it "resolution" would be easier to understand.

A nice improvement would be to separate our resolution and bitrate into flags rather than presets. Right now you can pick either but not both, example (I would like to transcode if either resolution or bitrate is over the preference). Or maybe it would be better to specify this similar to video formats where there is allowed and the target when transcoding specified separately. So you would have transcode policy Always, Required, Never then have targetBitrate, maxBitrate and targetResolution, maxResolution. But that is a bigger change than just documenting the current state.

mertalev:
In the video transcoding settings in the admin UI.

kevincox:
Hmm, the problem with putting it there is that if you are setting the values from the config file you can't actually see the descriptions of individual settings.

[Screencast from 2024-10-12 13-03-32.webm](https://github.com/user-attachments/assets/7c87563c-f68d-47bd-b4d7-bca54762925e)

It also doesn't list the settings "keyword". Maybe we need some sort of documentation that can appear both in the UI and in some place for people using the config file?

mertalev:
I agree setting them as toggles would be nice to mix and match. It'd implicitly document how it works as well, such as having toggles for HDR etc.

mertalev:
> Hmm, the problem with putting it there is that if you are setting the values from the config file you can't actually see the descriptions of individual settings.
> 
> 
> 
> [Screencast from 2024-10-12 13-03-32.webm](https://github.com/user-attachments/assets/7c87563c-f68d-47bd-b4d7-bca54762925e)
> 
> 
> 
> It also doesn't list the settings "keyword". Maybe we need some sort of documentation that can appear both in the UI and in some place for people using the config file?
> 
> 

Hmm, I guess putting in the docs and linking to it in the admin UI would be best. There'd be more room to explain config file usage that way. It's okay to put it in the section you mentioned in the OP for the scope of this PR, but in the future it'd be nicer to have a separate transcoding page instead.

docs/docs/administration/system-settings.md
kevincox, line 1: Ok, I've moved the docs to that page. I also expanded a bit to cover most of the Immich-specific settings in there. For now I have left out things that are basically ffmpeg flags.

I used the config file attribute as the heading name, but I'm not sure if this is the best option as it won't make as much sense for people configuring things in the UI. I could use the english names in the headers and mention the config properties in the body instead if that is preferable. I don't know if that would be confusing to people using Immich in other localizations but I AFAICT the docs aren't localized so I don't know if there is a better option.

docs/docs/administration/system-settings.md
```diff
@@ -206,3 +206,60 @@ When this option is enabled the `immich-server` will periodically make requests
...
+
+Audio is transcoded if the audio codec is not in `acceptedAudioCodecs`.
+
+#### `bitrate`
+
+In addition to the conditions in `required`, video is also transcoded if their bitrate is over `maxBitrate`.
+
+#### `optimal`
+
+In addition to the conditions in `required`, video is also transcoded if the horizontal **and** vertical resolutions are higher than `targetResolution`.
```
mertalev, line 265:
```suggestion
In addition to the conditions in `required`, video is also transcoded if the horizontal **and** vertical dimensions are higher than `targetResolution`.
```

docs/docs/administration/system-settings.md
mertalev: You can use the English title for them and put the JSON path in parentheses, clarifying at the start that this is what the parentheses mean.

kevincox: Done.

alextran1502:
Thank you for the PR. However, the subtitles in the settings are self-explanatory, and breaking them down in the documentation will cause a mismatch when we change them, creating a maintenance burden. I will close the PR for the reason above.

kevincox:
The settings UI is one thing but there is currently zero documentation for the config file. I had to read the source to figure it out which I assume is not considered acceptable UX.

alextran1502:
@kevincox ah that is fair, sorry I misunderstood the purpose of the PR

docs/docs/administration/system-settings.md
```diff
@@ -206,3 +206,60 @@ When this option is enabled the `immich-server` will periodically make requests
...
+
+#### Don't transcode any videos (`disabled`) {#ffmpeg.transcode-disabled}
+
+Videos are never transcoded. This saves space and resources on the server, but may prevent playback on devices that don't support the source format (especially web browsers) or result in high bandwidth usage when playing high-bitrate files.
+
+#### Only videos not in an accepted format (`required`) {#ffmpeg.transcode-required}
+
+Video is transcoded when any of the following conditions are met:
+
+- The video is HDR.
```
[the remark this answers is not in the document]
kevincox: How should I handle this? That change isn't live yet. I can prepare a PR to merge once that goes in or we can add a note to that PR if it doesn't land before this one.

docs/docs/administration/system-settings.md
```diff
@@ -206,3 +206,60 @@ When this option is enabled the `immich-server` will periodically make requests
...
+
+#### All videos (`all`) {#ffmpeg.transcode-all}
+
+Videos are always transcoded. This ensures consistency during video playback.
+
+#### Don't transcode any videos (`disabled`) {#ffmpeg.transcode-disabled}
+
+Videos are never transcoded. This saves space and resources on the server, but may prevent playback on devices that don't support the source format (especially web browsers) or result in high bandwidth usage when playing high-bitrate files.
+
+#### Only videos not in an accepted format (`required`) {#ffmpeg.transcode-required}
```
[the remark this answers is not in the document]
kevincox: I've added that property to the docs in a separate section.

docs/docs/administration/system-settings.md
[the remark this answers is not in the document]
mertalev: It's live now :P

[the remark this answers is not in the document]
kevincox: Added the condition.

docs/docs/administration/system-settings.md
```diff
@@ -206,3 +206,67 @@ When this option is enabled the `immich-server` will periodically make requests
...
+
+Video streams are transcoded when any of the following conditions are met:
+
+- The video is HDR.
+- The video is not in the yuv420p colour space.
```
mertalev, line 261:
```suggestion
- The video is not in the yuv420p pixel format.
```

docs/docs/administration/system-settings.md
```diff
@@ -206,3 +206,67 @@ When this option is enabled the `immich-server` will periodically make requests
 ## Video Transcoding Settings
 
 The system administrator can define parameters according to which video files will be converted to different formats (depending on the settings). The settings can be changed in depth, to learn more about the terminology used here, refer to FFmpeg documentation for [H.264](https://trac.ffmpeg.org/wiki/Encode/H.264) codec, [HEVC](https://trac.ffmpeg.org/wiki/Encode/H.265) codec and [VP9](https://trac.ffmpeg.org/wiki/Encode/VP9) codec.
+
+Which streams of a video file will be transcoded is determined by the [Transcode Policy](#ffmpeg.transcode). Streams that are transcoded use the following settings (config file name in brackets). Streams that are not transcoded are untouched and preserve their original settings.
+
+### Accepted containers (`acceptedContainers`) {#ffmpeg.acceptedContainers}
+
+If the video asset's container format is not in this list it will be remuxed to MP4 even if no streams need to be transcoded.
```
mertalev, line 214:
```suggestion
If the video asset's container format is not in this list, it will be remuxed to MP4 even if no streams need to be transcoded.
```

docs/docs/administration/system-settings.md
```diff
@@ -206,3 +206,67 @@ When this option is enabled the `immich-server` will periodically make requests
...
+
+### Target resolution (`ffmpeg.targetResolution`) {#ffmpeg.targetResolution}
+
+When transcoding a video stream, downscale the largest dimension to this value while preserving aspect ratio. Videos are never upscaled.
+
+The default value is `720`.
+
+### Transcode policy (`ffmpeg.transcode`) {#ffmpeg.transcode}
+
+The transcoding policy configures which streams of a video asset will be transcoded. The transcoding decision is made independently for video streams and audio streams. This means that if a video stream needs to be transcoded, but an audio stream does not, then the video stream will be transcoded while the audio stream will be copied. If the transcoding policy does not require any stream to be transcoded, then no separate video file will be created.
```
mertalev, line 244:
```suggestion
The transcoding policy configures which streams of a video asset will be transcoded. The transcoding decision is made independently for video streams and audio streams. This means that if a video stream needs to be transcoded, but an audio stream does not, then the video stream will be transcoded while the audio stream will be copied. If the transcoding policy does not require any stream to be transcoded and does not require the video to be remuxed, then no separate video file will be created.
```
It is actually possible for a video to not be transcoded and simply remuxed.

mertalev (commented):
A few minor nits, otherwise looks good!

mertalev (approved):

PR #13389, opened 2024-10-12
chore(server): explicitly use svt-av1 encoder

kevincox:
SVT-AV1 is much, much faster than the reference libaom encoder. In most real-world cases libaom is too slow to be practical, often struggling to achieve 0.1fps for 1080p content.

I suspect this is what was intended in f1ca1794a182d71e45340a262a450e694f4f1600 when AV1 support was originally added as it sets the `-svtav1-params` flag which is ignored unless using the SVT-AV1 encoder. Furthermore there is lots of discussion about SVT-AV1 in the PR thread.

### DIFF
--- a/server/src/services/media.service.spec.ts
+++ b/server/src/services/media.service.spec.ts
@@ -1322,7 +1322,7 @@ describe(MediaService.name, () => {
         expect.objectContaining({
           inputOptions: expect.any(Array),
           outputOptions: expect.arrayContaining([
-            '-c:v av1',
+            '-c:v libsvtav1',
             '-movflags faststart',
             '-fps_mode passthrough',
             '-map 0:0',
--- a/server/src/utils/media.ts
+++ b/server/src/utils/media.ts
@@ -492,6 +492,10 @@ export class VP9Config extends BaseConfig {
 }
 
 export class AV1Config extends BaseConfig {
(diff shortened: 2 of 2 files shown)


mertalev:
The FFmpeg build we use actually doesn't have libaom, so `av1` uses SVT-AV1. It's good to be more explicit, though, especially for non-Docker installations.

mertalev (approved):

danieldietzler (approved):

mertalev:
Running `npm run format:fix` should fix the failing check.

martabal:
@mertalev do we need to add the svt-av1 package to the base image, or is it included by default with the jellyfin-ffmpeg package ?

mertalev:
> @mertalev do we need to add the svt-av1 package to the base image, or is it included by default with the jellyfin-ffmpeg package ?

It comes with jellyfin-ffmpeg, so no change is needed.

martabal:
> It comes with jellyfin-ffmpeg, so no change is needed.

Nice

martabal (approved):

n00mkrad:
Late to the party, but I agree with what's been said. I did not specify the encoder originally because aom is not in our ffmpeg build, but yes, explicit>implicit, so this is a good change.

Issue #13401, opened 2024-10-13
Immich web cannot use browser language for international display. Immich web 不能使用浏览器正在使用的中文展示


fa1seut0pia:
### The bug

我在浏览器控制台输入 `navigator.languages`,打印出 `['zh-CN']`
My browser `navigator.languages` console log

![image](https://github.com/user-attachments/assets/1577ee56-4767-4b0f-9018-94eaf6a8d704)

`zh-CN` 与 [web/src/lib/constants.ts#L323](https://github.com/immich-app/immich/blob/main/web/src/lib/constants.ts#L323) 中的 `code: 'zh-Hans'` 不匹配
This does not match the `zh-CN` in [web/src/lib/constants.ts#L323](https://github.com/immich-app/immich/blob/main/web/src/lib/constants.ts#L323) `zh-Hans`

应该要把 [web/src/lib/constants.ts#L323](https://github.com/immich-app/immich/blob/main/web/src/lib/constants.ts#L323) 中的`zh-Hans`修改为`zh-CN`?
Should I change the 'zh-Hans' in [web/src/lib/constants.ts#L323](https://github.com/immich-app/immich/blob/main/web/src/lib/constants.ts#L323) to `zh-CN`?

### The OS that Immich Server is running on

Debian

### Version of Immich Server

v1.117.0

### Version of Immich Mobile App

v1.117.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
https://github.com/immich-app/immich/releases/latest/download/docker-compose.yml

与Github仓库一致
Consistent with GitHub repository
```

### Your .env content

```Shell
https://github.com/immich-app/immich/releases/latest/download/example.env

与Github仓库一致
Consistent with GitHub repository
```

### Reproduction steps

1. Chrome 浏览器中(其他浏览器类似)，打开 chrome://settings/languages 添加并置顶中文
In Chrome browser (similar to other browsers), open chrome://settings/languages to add and set the top Chinese
Like this:

![image](https://github.com/user-attachments/assets/f342c04d-fd0b-4088-9da8-b38c656c3856)

3. 打开 Open [https://demo.immich.app](https://demo.immich.app/auth/login)

![image](https://github.com/user-attachments/assets/a6c4f837-ff5f-4373-a7aa-eec3035739fe)

### Relevant log output

```shell
-
```

### Additional information

[...]

PR #13412, opened 2024-10-13
fix/(web) Set the alias of zh-CN to zh-Hans

fa1seut0pia:
fix browser does not display local settings Chinese. #13401

### DIFF
--- a/web/src/lib/constants.ts
+++ b/web/src/lib/constants.ts
@@ -327,6 +327,12 @@ export const langs = [
   { name: 'Development (keys only)', code: 'dev', loader: () => Promise.resolve({}) },
 ];
 
+// browser is zh-CN, but weblate use zh-Hans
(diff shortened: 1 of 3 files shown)


danieldietzler (commented):
This PR does not make sense to me. We already distinguish `code` and `weblateCode` in line 317 and 323 of the `constants.ts` file respectively. So clearly, weblate uses `zh_SIMPLIFIED` and not `zh-Hans`, and the browser uses `zh-Hans`, right?

fa1seut0pia:
No，browser uses `zh-CN`

image

danieldietzler:
I meant "browser" as in, the code immich uses in web.

danieldietzler:
Try setting `code: 'zh-CN` in line 323 and undo all your other changes.

PR #13414, opened 2024-10-13
fix(web): change the code zh-Hans to zh-CN

### LINKED ISSUE #13401: Immich web cannot use browser language for international display. Immich web 不能使用浏览器正在使用的中文展示

fa1seut0pia:
Closes #13401 
Supersedes #13412 

By the way, `zh-Hant` has the same problem, so I changed it as well.

### DIFF
--- a/web/src/lib/constants.ts
+++ b/web/src/lib/constants.ts
@@ -314,13 +314,13 @@ export const langs = [
   { name: 'Vietnamese', code: 'vi', loader: () => import('$lib/i18n/vi.json') },
   {
     name: 'Chinese (Traditional)',
-    code: 'zh-Hant',
+    code: 'zh-TW',
     weblateCode: 'zh_Hant',
     loader: () => import('$lib/i18n/zh_Hant.json'),
   },
   {
     name: 'Chinese (Simplified)',
-    code: 'zh-Hans',
+    code: 'zh-CN',
     weblateCode: 'zh_SIMPLIFIED',
     loader: () => import('$lib/i18n/zh_SIMPLIFIED.json'),
   },


danieldietzler (approved):

bo0tzz:
Closes #13401

michelheusschen:
`zh-Hans` and `zh-Hant` look like the correct languages to me. The suggested changes also won't work for `zh-HK`, `zh-MO` and `zh-SG`. Maybe a mapping of countries to languages would work for all cases?

fa1seut0pia:
> `zh-Hans` and `zh-Hant` look like the correct languages to me. The suggested changes also won't work for `zh-HK`, `zh-MO` and `zh-SG`. Maybe a mapping of countries to languages would work for all cases?

Haha, exactly! But to make zh-HK, zh-MO, and zh-SG work properly, we should add configurations to `web/src/lib/constants.ts`, right? My initial alias solution used a similar "mapping." Of course, if `svelte-i18n` can directly support this kind of scenario, then we should just use `svelte-i18n` (I haven’t looked into it closely yet). What do you think?

PR #13434, opened 2024-10-14
feat: search album by name

martabal:
This new `searchAlbum` endpoint behaves like the `searchPerson` endpoint. It returns a list of albums based on the beginning of the album name you are searching for. Currently, album search based on name is done on the client side, which is not ideal when you have hundreds or thousands of albums.


server/src/repositories/album.repository.ts
```diff
@@ -302,4 +302,47 @@ export class AlbumRepository implements IAlbumRepository {
 
     return result.affected;
   }
+
+  @GenerateSql({ params: [DummyValue.UUID, DummyValue.STRING, undefined] })
+  getByName(userId: string, albumName: string, shared?: boolean): Promise<AlbumEntity[]> {
```
danieldietzler, line 307: This should be paginated, no? Only returning the first 1k results is optimistic 😅

martabal: It would be nice but I don't think that really matters: the goal of this endpoint is not to get all albums.

jrasm91: IMO we should try to have all of our endpoints behave in a similar manner. Search person vs search album vs search something else should all have the same method signatures.

danieldietzler: I strongly agree and I there are people out there with >10k albums, for whom a search result may very well include >1k albums. I think it's bad just not supporting that.

martabal: > I think it's bad just not supporting that.

I agree. But that's for a future PR :stuck_out_tongue_closed_eyes:

martabal: Generally speaking it would be nice to paginate ALL endpoints that return a list of items

danieldietzler: > I agree. But that's for a future PR 😝

Why? :P

martabal: > Why? :P

Because I agree with @jrasm91 too. Do we want the same method signatures or not ?

danieldietzler: Hm that's fair.

danieldietzler (approved):
LGTM, it'd be awesome if you could add (at least) unit tests though. :)

server/src/repositories/album.repository.ts
martabal: Well... I added pagination for both endpoints.

martabal:
@danieldietzler So the API for `searchPerson` has changed but only the web uses that endpoint. Should we add a breaking change label ?

danieldietzler:
> @danieldietzler So the API for `searchPerson` has changed but only the web uses that endpoint. Should we add a breaking change label ?

Yup, we probably should. It's great that we don't have mobile conflicts though

PR #13461, opened 2024-10-15
fix(server): ffmpeg matrices

lyynd:
This time, I'm pretty sure that the values are according to the docs. Didn't test it, but will try at some point. Probably won't be able to test with OpenCL or CUDA.

### DIFF
--- a/server/src/utils/media.ts
+++ b/server/src/utils/media.ts
@@ -629,11 +629,11 @@ export class NvencHwDecodeConfig extends NvencSwDecodeConfig {
     const colors = this.getColors();
     const tonemapOptions = [
       'desat=0',
-      `matrix=bt${colors.matrix}`,
-      `primaries=bt${colors.primaries}`,
+      `matrix=${colors.matrix}`,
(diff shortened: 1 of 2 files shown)


lyynd:
Also https://github.com/immich-app/immich/pull/13409 for context

server/src/utils/media.ts
```diff
@@ -629,11 +629,11 @@ export class NvencHwDecodeConfig extends NvencSwDecodeConfig {
     const colors = this.getColors();
     const tonemapOptions = [
       'desat=0',
-      `matrix=${colors.matrix}`,
-      `primaries=${colors.primaries}`,
+      `matrix=bt${colors.matrix}`,
```
mertalev, line 632: It'd be cleaner to override `getColors` in `BaseHWConfig` instead.

mertalev (approved):

akostadinov:
idk if this affects performance or it is the ffmpeg 7 update but I see AV1 encoding slightly slower on my server with 1.118.2 compared to 1.117

just FYI

PR #13462, opened 2024-10-15
feat: show warning when running main branch build

bo0tzz:
image
image

### DIFF
--- a/web/src/lib/components/shared-components/side-bar/server-status.svelte
+++ b/web/src/lib/components/shared-components/side-bar/server-status.svelte
@@ -15,10 +15,13 @@
 
   let isOpen = false;
 
-  $: version = $serverVersion ? `v${$serverVersion.major}.${$serverVersion.minor}.${$serverVersion.patch}` : null;
-
   let info: ServerAboutResponseDto;
   let versions: ServerVersionHistoryResponseDto[] = [];
+
+  $: isMain = info?.sourceRef === 'main';
+  $: version = isMain
+    ? 'main ⚠️'
+    : $serverVersion && `v${$serverVersion.major}.${$serverVersion.minor}.${$serverVersion.patch}`;
 
   onMount(async () => {
     await requestServerInfo();
--- a/web/src/lib/components/shared-components/side-bar/server-status.svelte
+++ b/web/src/lib/components/shared-components/side-bar/server-status.svelte
@@ -10,18 +10,18 @@
     type ServerAboutResponseDto,
     type ServerVersionHistoryResponseDto,
   } from '@immich/sdk';
+  import Icon from '$lib/components/elements/icon.svelte';
+  import { mdiAlert } from '@mdi/js';
 
   const { serverVersion, connected } = websocketStore;
 
   let isOpen = false;
 
+  $: isMain = info?.sourceRef === 'main' && info.repository === 'immich-app/immich';
+  $: version = $serverVersion ? `v${$serverVersion.major}.${$serverVersion.minor}.${$serverVersion.patch}` : null;
+
   let info: ServerAboutResponseDto;
   let versions: ServerVersionHistoryResponseDto[] = [];
-
-  $: isMain = info?.sourceRef === 'main';
-  $: version = isMain
-    ? 'main ⚠️'
-    : $serverVersion && `v${$serverVersion.major}.${$serverVersion.minor}.${$serverVersion.patch}`;
 
   onMount(async () => {
     await requestServerInfo();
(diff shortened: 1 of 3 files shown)


web/src/lib/components/shared-components/side-bar/server-status.svelte
```diff
@@ -15,11 +15,14 @@
...
-  $: version = $serverVersion ? `v${$serverVersion.major}.${$serverVersion.minor}.${$serverVersion.patch}` : null;
-
   let info: ServerAboutResponseDto;
   let versions: ServerVersionHistoryResponseDto[] = [];
 
+  $: isMain = info?.sourceRef === 'main';
+  $: version = isMain
+    ? 'main ⚠️'
+    : $serverVersion && `v${$serverVersion.major}.${$serverVersion.minor}.${$serverVersion.patch}`;
+
```
bo0tzz, line 25: Feels kind of messy, suggestions welcome

web/src/lib/i18n/en.json
```diff
@@ -816,6 +816,7 @@
   "look": "Look",
   "loop_videos": "Loop videos",
   "loop_videos_description": "Enable to automatically loop a video in the detail viewer.",
+  "main_branch_warning": "⚠\uFE0F You're running a build from the main branch. We strongly recommend using a release version!",
```
bo0tzz, line 819: I think the tests don't like this emoji? Not sure what to do about that.

web/src/lib/components/shared-components/side-bar/server-status.svelte
danieldietzler: I think you want to keep the existing ternary so that it'll still be null. Not prettier though 😅

web/src/lib/i18n/en.json
danieldietzler: I'm assuming you also tried pasting the emote there?

bo0tzz: I pasted it in WebStorm which caused that weird \uFE0F. Pasting it in nvim didn't actually show the emoji in the editor, but did fix the issue lol.

danieldietzler: LMFAO noice

web/src/lib/i18n/en.json
```diff
@@ -816,6 +816,7 @@
   "look": "Look",
   "loop_videos": "Loop videos",
   "loop_videos_description": "Enable to automatically loop a video in the detail viewer.",
+  "main_branch_warning": "⚠️You're running a build from the main branch. We strongly recommend using a release version!",
```
jrasm91, line 819: we should probably use an icon in the code and avoid embedding an emoji in the translated text imo.

danieldietzler: I thought so too but it'd break in RTL languages I think

jrasm91: That can be fixed other ways though. There is something very brittle about emojis in text. I prefer icons over emojis in general, and this way if we wanted to change it we only do it in one place.

bo0tzz: Do we have a way to handle RTL for that yet? Or is it something I can leave for later? I think I prefer the emoji over an icon, but I do agree that having it in the i18n strings isn't ideal.

bo0tzz: Otoh, leaving the emoji in the text means that it too could be localized, if ⚠️ isn't as meaningful as a warning icon in another language.

web/src/lib/components/shared-components/side-bar/server-status.svelte
```diff
@@ -15,11 +15,14 @@
 
   let isOpen = false;
 
-  $: version = $serverVersion ? `v${$serverVersion.major}.${$serverVersion.minor}.${$serverVersion.patch}` : null;
-
   let info: ServerAboutResponseDto;
   let versions: ServerVersionHistoryResponseDto[] = [];
 
+  $: isMain = info?.sourceRef === 'main';
```
jrasm91, line 21: This is just build metadata and it could be from an entirely different source repository. I think we should have a repositoryUrl AND sourceRef check for the warning.

danieldietzler (approved):

jrasm91 (approved):
Looks great! Did we want the conditions to be the same in both places? Right now it is source ref in one and repo and source ref in the other.

bo0tzz:
Woops, missed that one, ty.

jrasm91 (approved):

jrasm91 (approved):

PR #13506, opened 2024-10-16
refactor(server): external domain fallback

jrasm91:
Move another `process.env` reference to `configRepository.getEnv()`


danieldietzler (approved):
Clean, I like it!

PR #13507, opened 2024-10-16
refactor(server): move host env to config repo

jrasm91:
Move another process.env reference to configRepository.


bo0tzz (commented):
Should this be called IMMICH_HOST?

jrasm91:
> Should this be called IMMICH_HOST?

Indeed. It even says that in the docs. I wonder what happened there.

danieldietzler (approved):

bo0tzz:
Changing the env var name is technically a breaking change, should we surface that?

bo0tzz (approved):

PR #13517, opened 2024-10-16
fix(server): Allow passwordless users when oauth enabled

### LINKED ISSUE #13274: Cannot create passwordless user, even though passwords are disabled
### The bug

Unable to create accounts without providing a password, even though I have password logins disabled and I'm using OAuth2 (working)

Debian

### Version of Immich Server

v1.117.0

N/A

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

## Steps

1. Go to `Administration`
2. `Settings`
3. `Auth settings`
4. `Oauth` -> enabled and working
5. `Password Login` -> disabled
6. Go to `Users`
7. `Create user`
8. Provide `Email`, disable `Require user to change password on first login`
9. Click `Create`

## Expected

User gets created and can auth with OAuth2, assuming user has account with matching email with the OAuth2 provider.

## Actual

Error `Please fill in this field` appears at the password field.

### Relevant log output

```shell
N/A
```

### Additional information

_No response_

jedi04:
This PR fixes : #13274 

It allows creation of passwordless users when Oauth is enabled in config.

### DIFF
--- a/server/src/dtos/user.dto.ts
+++ b/server/src/dtos/user.dto.ts
@@ -62,7 +62,6 @@ export class UserAdminCreateDto {
   @Transform(toEmail)
   email!: string;
 
-  @IsNotEmpty()
   @IsString()
   password!: string;
 
--- a/server/src/services/user-admin.service.ts
+++ b/server/src/services/user-admin.service.ts
@@ -26,6 +26,10 @@ export class UserAdminService extends BaseService {
 
   async create(dto: UserAdminCreateDto): Promise<UserAdminResponseDto> {
     const { notify, ...rest } = dto;
+    const config = await this.getConfig({ withCache: false });
+    if (!config.oauth.enabled && !rest.password) {
+      throw new BadRequestException('password is required');
+    }
     const user = await createUser({ userRepo: this.userRepository, cryptoRepo: this.cryptoRepository }, rest);
 
     await this.eventRepository.emit('user.signup', {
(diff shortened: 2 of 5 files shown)


jrasm91 (commented):
Looks good! You can use the $featureFlag.oauthEnabled store instead I think

jedi04:
> Looks good! You can use the $featureFlag.oauthEnabled store instead I think

@jrasm91 Thanks for the suggestion. I have updated it to use featureFlags.

danieldietzler (approved):

server/src/dtos/user.dto.ts
```diff
@@ -62,7 +62,6 @@ export class UserAdminCreateDto {
   @Transform(toEmail)
   email!: string;
 
-  @IsNotEmpty()
```
jrasm91, line 65: I think this just allows empty strings, but it should probably have Optional({ nullable: true }) and still keep the is not empty. We should also test that you cannot login with an empty string of password login is enabled still

jedi04: We have set password to be string in the user entity. So would it be okay to insert null values?
https://github.com/immich-app/immich/blob/3f663106e8093824617128d7558423284c4741cd/server/src/entities/user.entity.ts#L33

And during login we have added check for not allowing empty password
https://github.com/immich-app/immich/blob/3f663106e8093824617128d7558423284c4741cd/server/src/services/auth.service.ts#L329

jrasm91: Oh, I see that we don't allow null in the database. We can change that later I think. We'll allow empty now, and move to `null` in the future.

PR #13536, opened 2024-10-17
feat(web): Change relink person icon from minus to pencil

CrushedAsian255:
The re-link person icon is currently a minus symbol. This can be confusing as it looks like a "remove person" button. Changing it to a pencil makes it clear it is an editing operation, not a removing operation.

### DIFF
--- a/web/src/lib/components/faces-page/person-side-panel.svelte
+++ b/web/src/lib/components/faces-page/person-side-panel.svelte
@@ -15,9 +15,8 @@
     type AssetFaceResponseDto,
     type PersonResponseDto,
   } from '@immich/sdk';
-  import { mdiAccountOff } from '@mdi/js';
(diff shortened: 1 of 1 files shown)


CrushedAsian255:
> Label error. Requires exactly 1 of: changelog:.*. Found:

what does this mean?

jrasm91:
Can you add screen shots of the page before and after this change?

hitech95:
Before:
immagine
After
immagine

CrushedAsian255:
> Before: immagine After immagine

Not sure why the icon is slightly smaller in the second image, that's a bug.

jrasm91 (approved):

NicholasFlamy:
> Before: immagine After immagine

James from LTT!?

PR #13552, opened 2024-10-17
refactor(server): auth enums

jrasm91:


server/src/enum.ts
```diff
@@ -3,6 +3,28 @@ export enum AuthType {
   OAUTH = 'oauth',
 }
 
+export enum ImmichCookie {
```
danieldietzler, line 6: I was hoping js supports emotes in variable names but unfortunately that's not the case :( (`Immich🍪`)

danieldietzler (approved):

server/src/enum.ts
jrasm91: ........

PR #13560, opened 2024-10-17
refactor(server): telemetry ports env

jrasm91:


server/src/repositories/config.repository.spec.ts
```diff
@@ -178,4 +181,41 @@ describe('getEnv', () => {
...
+  describe('network', () => {
+    it('should return default network options', () => {
+      const { network } = getEnv();
+      expect(network).toEqual({
+        trustedProxies: [],
+      });
+    });
+
+    it('should parse trusted proxies', () => {
+      process.env.IMMICH_TRUSTED_PROXIES = '10.1.0.0,10.2.0.0';
```
danieldietzler, line 194: Mind making one of those a larger subnet (e.g. /24) for validation or whatever later? :)

danieldietzler (approved):

PR #13564, opened 2024-10-17
refactor(server): telemetry env

jrasm91:


danieldietzler (approved):
Awesome, I'm digging those refactorings!

server/src/repositories/config.repository.ts
```diff
@@ -30,6 +30,8 @@ const asSet = (value: string | undefined, defaults: ImmichWorker[]) => {
   return new Set(values.length === 0 ? defaults : (values as ImmichWorker[]));
 };
 
+const parseBoolean = (value: string | undefined, defaultValue: boolean) => (value ? value === 'true' : defaultValue);
+
 const getEnv = (): EnvData => {
```
mertalev, line 35: We should really cache this at this point.

mertalev (approved):

PR #13567, opened 2024-10-17
feat(ml): better multilingual search with nllb models

mertalev:
### Description

The NLLB models have the highest quality results for multilingual text in our model catalog. However, these models expect the input to explicitly indicate the language of the text. Since we don't provide this, the results are worse than other multilingual models.

This PR makes the machine learning service accept a `language` option that it maps to the corresponding FLORES200 token that NLLB expects. The server is updated to accept this parameter and forward it to the machine learning service, while web and mobile are updated to provide the language based on current user settings. Search still works without it, and the option only has an impact when using an NLLB model.

### How Has This Been Tested?

I tested on web and can confirm the results for Turkish are significantly better. Changing the language in the user settings has a marked effect on the ranking. I haven't tested mobile so I'm not sure if all the language codes there line up with the map in the machine learning service.


machine-learning/app/models/constants.py
```diff
@@ -66,6 +66,65 @@
 SUPPORTED_PROVIDERS = ["CUDAExecutionProvider", "OpenVINOExecutionProvider", "CPUExecutionProvider"]
 
 
+WEBLATE_TO_FLORES200 = {
```
jrasm91, line 69: I wonder where this mapping should live. To some extent the server knows which model it is using when it sends the request so you could argue this belongs there. What do you think?

jrasm91 (approved):
Awesome

machine-learning/app/models/constants.py
mertalev: I'd say it's better for it to live here - it makes things easier if you query ML directly or use the models in another context like in notebooks.

web/src/routes/(user)/search/[[photos=photos]]/[[assetId=id]]/+page.svelte
```diff
@@ -144,6 +144,7 @@
       page: nextPage,
       withExif: true,
       isVisible: true,
+      language: $lang,
```
Comment (martabal) [inline web/src/routes/(user)/search/[[photos=photos]]/[[assetId=id]]/+page.svelte:147]: Should we add a new setting? I like to have my UI in English, but I prefer to use my native language for searching.

Comment (mertalev) [reply web/src/routes/(user)/search/[[photos=photos]]/[[assetId=id]]/+page.svelte:147]: Sure, we could have a Search Language user preference separate from the UI language.

Comment (bo0tzz) [reply web/src/routes/(user)/search/[[photos=photos]]/[[assetId=id]]/+page.svelte:147]: Would it make sense to also have this be something the admin can override?

Comment (mertalev) [reply web/src/routes/(user)/search/[[photos=photos]]/[[assetId=id]]/+page.svelte:147]: Do you mean making it use a certain search language by default for all users? That could be nice, possibly doing the same for the UI language as well. But I think we should avoid making too many changes at once in this PR, so the admin stuff can be done separately.

Comment (bo0tzz) [reply web/src/routes/(user)/search/[[photos=photos]]/[[assetId=id]]/+page.svelte:147]: A few related FRs came up that I consolidated into #13651.

mertalev:
Sure! I got caught up trying to make it work in mobile, but the language codes for localization were different IIRC. The web portion should be good though after rebasing.

mertalev:
Tested on web and mobile with a variety of languages and confirmed the language is passed to the ML service, which uses it to pass the right language token to the NLLB model.

bo0tzz:
Do the model benchmarks account for this token being passed?

mertalev:
The results in the docs are when the language token is passed correctly. The models might not even make the list when no token is passed because the input is interpreted as English.

PR #13569, opened 2024-10-18
docs: improve install/requirements and install/docker-compose

pyorot:
Giving back changes that mostly would've helped me starting out with Immich:
1. requirements/software: distinguish Docker Engine and Desktop; clarify that Compose plugin is included with recommended installation guides;
2. put the hw-accel config file downloads in an optional note;
3. add a hook after .env file setup suggesting docker-compose.yml edits for optional features;
4. adding a sentence that "docker compose up -d" takes care of automatic restarts;
5. linking docker version error note back to "do the recommended install thoroly" with explanation why;
6. added info on purging obsolete containers;
7. moved extra install/docker-compose recipe details from quick-start to install/docker-compose.

The idea is to explain some of the basics of Docker (1, 4, 6) and Linux (5) as we go, and streamline out-of-order steps (2, 3, 7). Once I've looked over the compiled docs, I'll set this as ready for review and you can tell me which edits you fancy :).


docs/docs/install/docker-compose.mdx
```diff
@@ -93,20 +97,36 @@ For more information on how to use the application, please refer to the [Post In
...
 
-```bash title="Upgrade Immich"
+```bash title="Upgrade and restart Immich"
 docker compose pull && docker compose up -d
 ```
 
+To clean up disk space, the old version's obsolete containers can be deleted with the following command, which deletes all stopped containers. As such, make sure to start Immich (with the previous command) before running it.
+
+```bash title="Delete all stopped containers"
+docker system prune
```
mmomjian, line 127:
```suggestion
docker image prune
```

docs/docs/install/docker-compose.mdx
```diff
@@ -93,20 +97,36 @@ For more information on how to use the application, please refer to the [Post In
...
 
-When a new version of Immich is [released][releases], the application can be upgraded with the following commands, run in the directory with the `docker-compose.yml` file:
+When a new version of Immich is [released][releases], the application can be upgraded and restarted with the following commands, run in the directory with the `docker-compose.yml` file:
 
-```bash title="Upgrade Immich"
+```bash title="Upgrade and restart Immich"
 docker compose pull && docker compose up -d
 ```
 
+To clean up disk space, the old version's obsolete containers can be deleted with the following command, which deletes all stopped containers. As such, make sure to start Immich (with the previous command) before running it.
```
mmomjian, line 124:
```suggestion
To clean up disk space, the old version's obsolete containers can be deleted with the following command, which deletes all unused container images. As such, make sure to start Immich (with the previous command) before running it.
```

docs/docs/install/requirements.md
```diff
@@ -8,18 +8,20 @@ Hardware and software requirements for Immich:
...
+Immich requires [**Docker**](https://docs.docker.com/get-started/get-docker/) with the **Docker Compose plugin**:
+
+- **Docker Engine**: This CLI variant is suitable for Linux servers or Windows via WSL2.
+- **Docker Desktop**: This GUI variant is suitable for Linux (desktop), Windows or macOS.
+
+The Compose plugin will be installed by both Docker Engine and Desktop by following the linked installation guides; it can also be [separately installed](https://docs.docker.com/compose/install/) (pick the plugin, not the standalone).
 
 :::note
-Immich requires the command `docker compose` - the similarly named `docker-compose` is [deprecated](https://docs.docker.com/compose/migrate/) and is no longer compatible with Immich.
+Immich requires the command `docker compose` (the plugin); the similarly named `docker-compose` (the standalone) is [deprecated](https://docs.docker.com/compose/migrate/) and is no longer compatible with Immich.
```
mmomjian, line 19: I think this plugin vs standalone talk is more confusing, not less

docs/docs/install/requirements.md
```diff
@@ -8,18 +8,20 @@ Hardware and software requirements for Immich:
...
 :::
 
 ## Hardware
 
-- **OS**: Recommended Linux operating system (Ubuntu, Debian, etc).
-  - Windows is supported with [Docker Desktop on Windows](https://docs.docker.com/desktop/install/windows-install/) or [WSL 2](https://docs.docker.com/desktop/wsl/).
-  - macOS is supported with [Docker Desktop on Mac](https://docs.docker.com/desktop/install/mac-install/).
+- **OS**: Linux is recommended, and Windows and macOS are supported: any version or distribution that Docker can be installed on (see above).
```
mmomjian, line 24: Not fond of this change, would prefer it how we had it.

docs/docs/install/docker-compose.mdx
```diff
@@ -54,22 +52,28 @@ Optionally, you can enable hardware acceleration for machine learning and transc
...
+## Step 3 - Start the containers
 
-From the directory you created in Step 1, (which should now contain your customized `docker-compose.yml` and `.env` files) run `docker compose up -d`.
+From the directory you created in Step 1 (which should now contain your customized `docker-compose.yml` and `.env` files), run `docker compose up -d`.
 
 ```bash title="Start the containers using docker compose command"
 docker compose up -d
 ```
 
+The `restart: always` entries in `docker-compose.yml` for each container mean they will always restart after system reboot or program crashes.
```
mmomjian, line 73: This isn't needed here (or really at all IMO, this is `man docker` stuff)

docs/docs/install/docker-compose.mdx
```diff
@@ -54,22 +52,28 @@ Optionally, you can enable hardware acceleration for machine learning and transc
...
 
-- Populate custom database information if necessary.
 - Populate `UPLOAD_LOCATION` with your preferred location for storing backup assets.
 - Consider changing `DB_PASSWORD` to a custom value. Postgres is not publically exposed, so this password is only used for local authentication.
   To avoid issues with Docker parsing this value, it is best to use only the characters `A-Za-z0-9`.
 - Set your timezone by uncommenting the `TZ=` line.
+- Populate custom database information if necessary.
+
+:::info Optional Features
+You can make the required edits to `docker-compose.yml` to add external libraries or enable hardware acceleration now by following [their guides](#setting-up-optional-features) up to and including those edits.
```
mmomjian, line 62:
```suggestion
You can edit `docker-compose.yml` to add external libraries or enable hardware acceleration now by following [their guides](#setting-up-optional-features).
```

mmomjian (commented):
thanks, I have some thoughts

pyorot:
Thanks for the review; I'll check it out tomo.

docs/docs/install/requirements.md
pyorot: For immich’s purposes strictly, it is; however, the [Docker Compose installation](https://docs.docker.com/compose/install/) docs that this page links specifically give the choice between installing Docker Desktop, the Compose plugin, or the Compose standalone.

docs/docs/install/docker-compose.mdx
pyorot: Wasn’t aware of images vs containers but looks good, will merge.

docs/docs/install/docker-compose.mdx
pyorot: The purpose of this line is to inform those of us who don’t use or care about Docker that the “compose up” command starts immich as a persistent service, which IMO is a worry that would cross the minds of users who aren’t culturally immersed in Linux to the point of intuitively expecting the behaviour. 

The point isn’t to explain “restart: always” but rather to say “this command starts the container and ensures it automatically restarts”. I’ll reword it to make that clearer.

docs/docs/install/docker-compose.mdx
pyorot: I think this is a fair simplification; it’s just annoying that the optional guides don’t split apart the yaml file edits from the settings changes made to the running program, so users who follow this tip would have to flip-flop between the two pages.

docs/docs/install/docker-compose.mdx
pyorot: I believe this command can be run with the containers stopped so I deleted that second sentence as well.

pyorot:
image

Another thing that crossed my mind as a potential change to add is that the order of these features is a bit weird; I'd swap External Libraries and Hardware Transcoding.

docs/docs/install/requirements.md
bo0tzz: It gives the choice but also pretty clearly notes that the standalone is deprecated, so I don't think we need to repeat that here.

docs/docs/install/docker-compose.mdx
bo0tzz: A topic we find ourselves constantly trying to balance is how much basic information to include in our docs. We're documenting Immich, and so I don't think we should expect to be teaching people sysadmin basics. If someone is wondering about details of the restart behaviour, they can find answers to that in plenty of places already.

pyorot: I changed it earlier to something closer to what I intended:

> This starts immich and ensures it automatically restarts after system reboots or crashes (thanks to the `restart: always` entries in `docker-compose.yml`).

The bracketed bit is more for curiosity and can be deleted cos it’s somewhat evident from the non-bracketed bit, if one also reads the yaml file.

But yeah I wouldn’t consider this teaching as much as explaining what that command actually does. On further thought, I recall seeing people forget the d flag, which is kind of a similar mistake – we’re trying to say the command starts the program as a service, so not blocking the console, but as a background thing that auto restarts. Basically that’s what I wanted.

docs/docs/install/requirements.md
pyorot: That’s fair yeah; I just figured people might want to join the dots on which installation option corresponds to which command, as an explanation for why the old command doesn’t work. I did delete the reference to “v2” over “v1” since I could hardly find it on Docker’s website. But yeah, omitting those brackets is fine.

docs/docs/install/docker-compose.mdx
pyorot: New sentence: I am feeling:

> This starts immich as a background service, ensuring it restarts after system reboots or crashes.

Or with btw teaching:

> This starts immich as a background service (per the -d flag), ensuring it restarts after system reboots or crashes (per the `restart` fields in `docker-compose.yml`).

pyorot:
Updated all my suggestions and am happy to stop working on this and allow whichever parts to be merged.

> Another thing that crossed my mind as a potential change to add is that the order of these features is a bit weird; I'd swap External Libraries and Hardware Transcoding.

Turns out there’re no sidebar front-matter parameters in the whole features folder so I’ll let a different patch sort that out. The capitalisation of the Supported Formats page’s headings is also inconsistent with the other pages in that folder.

docs/docs/install/docker-compose.mdx
pyorot: actually i tried running `docker image prune` with the containers stopped and i feel like `docker compose up -d` then redownloaded some stuff? unsure, needs checking

docs/docs/install/requirements.md
pyorot: > That’s fair yeah; I just figured people might want to join the dots on which installation option corresponds to which command, as an explanation for why the old command doesn’t work. I did delete the reference to “v2” over “v1” since I could hardly find it on Docker’s website. But yeah, omitting those brackets is fine.

done

> I'd suggest replacing `compatible` with `supported`. New Synology nas systems use docker-compose and it seems to be running fine for me.

and done

docs/docs/install/requirements.md
pyorot: reverted cos cba

alextran1502 (approved):

pyorot:
thank 🙇

PR #13588, opened 2024-10-18
refactor(server): telemetry

jrasm91:


danieldietzler (commented):
Looks really good so far, I quite like it!

danieldietzler (approved):

server/src/repositories/telemetry.repository.ts
```diff
@@ -33,20 +48,120 @@ class MetricGroupRepository implements IMetricGroupRepository {
...
+        prefix: 'immich',
+        description: `The elapsed time in ${unit} for the ${startCase(className)} to ${propertyName.toLowerCase()}`,
+        unit,
+        valueType: contextBase.ValueType.DOUBLE,
+      });
+
+      descriptor.value = function (...args: any[]) {
+        const start = performance.now();
+        // TODO: wrap with try/catch
+        const result = method.apply(this, args);
```
mertalev, line 152: Why would you wrap it?

server/src/repositories/telemetry.repository.ts
```diff
@@ -33,20 +48,120 @@ class MetricGroupRepository implements IMetricGroupRepository {
...
 
     this.api = new MetricGroupRepository(metricService).configure({ enabled: apiMetrics });
     this.host = new MetricGroupRepository(metricService).configure({ enabled: hostMetrics });
     this.jobs = new MetricGroupRepository(metricService).configure({ enabled: jobMetrics });
     this.repo = new MetricGroupRepository(metricService).configure({ enabled: repoMetrics });
   }
+
+  setup({ repositories }: { repositories: ClassConstructor<unknown>[] }) {
+    const { telemetry } = this.configRepository.getEnv();
+    if (!telemetry.enabled || !telemetry.repoMetrics) {
```
mertalev, line 112: This is a slight difference in behavior, right? `IMMICH_METRICS` takes priority over `IMMICH_IO_METRICS` with this change. You mentioned changing this to include/exclude so it doesn't matter much, but if you don't think you'll get to that before the next release then you'd want to update the docs on the new behavior.

mertalev (approved):

server/src/repositories/telemetry.repository.ts
jrasm91: Oh you are right. For some reason I thought we did something on error that was missing.

server/src/repositories/telemetry.repository.ts
jrasm91: Enabled gets set to true here if any are enabled., which is how you had it before. I agree it is confusing though.

mertalev: Oh true, I forgot about that too lol. That's fine then

Issue #11887, opened 2024-08-18
Changing date from web UI doesn't respect timezone's DST


MMauro94:
### The bug

When selecting a a timezone in the date/time change dialog, you can also select a IANA timezone identifier, such as `Europe/Rome`. These timezones support changes in offset (for example, due to Daylight Savings Time).

However, it seems that timezones in this dialog are fixed at a predetermined offset (the one for the current date?), while they should take the correct offset based on the selected date/time.

### The OS that Immich Server is running on

Debian 12

### Version of Immich Server

v1.112.1

### Version of Immich Mobile App

v1.112.1

### Platform with the issue

- [X] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
version: "3.8"

services:
  immich-server:
    container_name: immich-server
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - /etc/localtime:/etc/localtime:ro
    env_file:
      - .env
    depends_on:
      - redis
      - database
    restart: always

  immich-machine-learning:
    container_name: immich-machine-learning
    image: ghcr.io/immich-app/immich-machine-learning:${IMMICH_VERSION:-release}
    volumes:
      - model-cache:/cache
    env_file:
      - .env
    restart: always


  redis:
    container_name: immich-redis
    image: redis:6.2-alpine
    restart: always


  database:
    container_name: immich-postgres
    image: tensorchord/pgvecto-rs:pg14-v0.2.0
    env_file:
      - .env
    environment:
      POSTGRES_PASSWORD: ${DB_PASSWORD}
      POSTGRES_USER: ${DB_USERNAME}
      POSTGRES_DB: ${DB_DATABASE_NAME}
    volumes:
      - pgdata:/var/lib/postgresql/data
    restart: always


volumes:
  pgdata:
    name: immich-db
  model-cache:
    name: immich-model-cache

networks:
  default:
    external:
      name: web_services
```

### Your .env content

```Shell
DB_HOSTNAME=immich-postgres
DB_USERNAME=<REDACTED>
DB_PASSWORD=<REDACTED>
DB_DATABASE_NAME=immich
REDIS_HOSTNAME=immich-redis
UPLOAD_LOCATION=<REDACTED>/immich
TYPESENSE_API_KEY=<REDACTED>
PUBLIC_LOGIN_PAGE_MESSAGE=

[...]

PR #13599, opened 2024-10-19
chore(web): add tests for ChangeDate component

darrelhong:
Intended to look into issue #11887 but I believe it has already been resolved by #12315.

There was a comment to add some tests to verify the functionality and I think unit tests fit this use case very well so that's what is added in this PR.

This should give more confidence in making future changes to the ChangeDate component.


web/src/lib/components/shared-components/change-data.spec.ts
danieldietzler, line 1: The file should be called `change-date.spec.ts`

danieldietzler (approved):

web/src/lib/components/shared-components/change-data.spec.ts
darrelhong: ooh great catch, how did i name it that? 😂

jrasm91 (approved):

PR #13604, opened 2024-10-19
chore: move weblate translations to root directory

ben-basten:
## Description

Move the Weblate translations into the root directory of the repository. This starts setting the groundwork for mobile and web to share the same translations.

Moving the translation JSON files is in its own commit, in case any changes are needed.

## How Has This Been Tested?

- All PR checks are passing
- Check that immich-server runs as expected when using the `immich-server:pr-13604` image
- Dev server for web works, using `make dev` and `make dev-update`


ben-basten:
A couple of open questions:

- Are JSON files with ICU message formatting the desired direction for mobile?
- Should country codes be added to the translation files?

jrasm91:
> A couple of open questions:
> 
> * Are JSON files with ICU message formatting the desired direction for mobile?
> * Should country codes be added to the translation files?

For 1, I believe so. For 2, I have no idea, but maybe eventually. Just moving them to the top level is a good start.

ben-basten:
> For 1, I believe so. For 2, I have no idea, but maybe eventually. Just moving them to the top level is a good start.

Ok, just figured I'd check. From what I can tell, it doesn't look like the current mobile [easy_localization](https://pub.dev/packages/easy_localization) library supports ICU formatting.

jrasm91 (approved):

Issue #13638, opened 2024-10-21
Repair page is missing from admin sidebar


polgarc:
### The bug

The `Repair` page is missing from the admin sidebar:

<details>
<summary>screenshot 1</summary>

![image](https://github.com/user-attachments/assets/ee7dc2e0-2206-43d1-8285-be6a16290426)

</details>

Note if I type the proper URL, I can access the page:

<details>
<summary>screenshot 1</summary>

![image](https://github.com/user-attachments/assets/9cc65e93-12c1-4f6a-8e5d-1cf45bc36bc6)

</details>

Or am I missing something and the `Repair` page was moved to somewhere else? The [docs](https://v1.118.2.archive.immich.app/docs/administration/repair-page) says it should be under the admin settings.

### The OS that Immich Server is running on

Debian 12

### Version of Immich Server

v1.118.2

### Version of Immich Mobile App

v1.118.0

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
name: immich

services:
  immich-server:
    container_name: immich_server
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    # extends:
      # file: hwaccel.transcoding.yml
      # service: cpu # set to one of [nvenc, quicksync, rkmpp, vaapi, vaapi-wsl] for accelerated transcoding
    volumes:
      - data:/usr/src/app/upload
      - /etc/localtime:/etc/localtime:ro
      - /mnt/photo:/mnt/photo
    env_file:
      - .env
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.immich.entrypoints=websecure"
      - "traefik.http.routers.immich.rule=Host(`immich.mydomain.cc`)"
      - "traefik.http.routers.immich.tls.certresolver=myresolver"
      - "traefik.http.routers.immich.tls=true"
    depends_on:
      - redis
      - database
    restart: always
    networks:
      - immich
      - traefik_net

  immich-machine-learning:
    container_name: immich_machine_learning
    # For hardware acceleration, add one of -[armnn, cuda, openvino] to the image tag.
    # Example tag: ${IMMICH_VERSION:-release}-cuda
    image: ghcr.io/immich-app/immich-machine-learning:${IMMICH_VERSION:-release}
    extends: # uncomment this section for hardware acceleration - see https://immich.app/docs/features/ml-hardware-acceleration
      file: hwaccel.ml.yml
[...]

PR #13640, opened 2024-10-21
docs: add warning to repair page

polgarc:
Add a warning to the repair page docs, as the link was removed from the sidebar in #12294.

This'll hopefully stop people from opening issues when they don't find the page, see #13638.


mmomjian:
The repair page still exists and will return at some point. We can add a note at the top saying “this feature is currently disabled and will be reworked in the near future” but I don’t see a point to deleting all this work.

mmomjian (approved):

jrasm91 (approved):

PR #13704, opened 2024-10-23
feat(web): Implement keep this delete others for asset stacks

bdavis2-PCTY:
Adds a new button when viewing a stacked asset to "Keep this, delete others."

This will result in deleting all the assets that are in the stack, except for the one that is currently being viewed. 

image


web/src/lib/components/asset-viewer/actions/keep-this-delete-others.svelte
```diff
@@ -0,0 +1,21 @@
...
+
+  const handleKeepThisDeleteOthers = async () => {
+    const keptAsset = await keepThisDeleteOthers(asset.id, stack.id);
+    if (keptAsset) {
+      onAction({ type: AssetAction.UNSTACK, assets: [keptAsset] });
+    }
+  };
+</script>
+
+<MenuOption icon={mdiPinOutline} onClick={handleKeepThisDeleteOthers} text={'Keep this, delete others'} />
```
danieldietzler, line 21: This text should be internationalized

web/src/lib/utils/asset-utils.ts
```diff
@@ -438,6 +439,32 @@ export const deleteStack = async (stackIds: string[]) => {
   }
 };
 
+export const keepThisDeleteOthers = async (keepId: string, stackId: string) => {
+  const $t = get(t);
+
+  try {
+    const stack = await getStack({ id: stackId });
+    const assetToKeep = stack.assets.find(x => x.id === keepId);
+    const assetsToDeleteIds = stack.assets.filter(x => x.id !== keepId).map(x => x.id);
```
danieldietzler, line 448: Please don't use random, single character variable names

web/src/lib/components/asset-viewer/actions/keep-this-delete-others.svelte
bdavis2-PCTY: Oops, thought I did this before making the PR.

Fix has been pushed.

web/src/lib/utils/asset-utils.ts
bdavis2-PCTY: Code has been updated with more meaningful variable names here.

web/src/lib/components/asset-viewer/actions/keep-this-delete-others.svelte
```diff
@@ -0,0 +1,22 @@
...
+  import { mdiPinOutline } from '@mdi/js';
+  import type { OnAction } from './action';
+  import { t } from 'svelte-i18n';
+
+  export let stack: StackResponseDto;
+  export let asset: AssetResponseDto;
+  export let onAction: OnAction;
+
+  const handleKeepThisDeleteOthers = async () => {
+    const keptAsset = await keepThisDeleteOthers(asset.id, stack.id);
```
mertalev, line 15: You already have the asset and stack here, no? Just change the `keepThisDeleteOthers` to accept those instead of getting the IDs and querying for them again.

web/src/lib/utils/asset-utils.ts
```diff
@@ -438,6 +439,32 @@ export const deleteStack = async (stackIds: string[]) => {
   }
 };
 
+export const keepThisDeleteOthers = async (keepId: string, stackId: string) => {
+  const $t = get(t);
+
+  try {
+    const stack = await getStack({ id: stackId });
+    const assetToKeep = stack.assets.find((asset) => asset.id === keepId);
```
mertalev, line 447: You can remove these after changing the function signature.

web/src/lib/components/asset-viewer/actions/keep-this-delete-others.svelte
bdavis2-PCTY: Function signature has been updated to accept the Asset & Stack objects instead of ids. Originally implemented this way as passing ids to the functions seemed to be the most common way elsewhere, but I completely agree that passing in the objects themselves is much more efficient.

web/src/lib/utils/asset-utils.ts
bdavis2-PCTY: Commit has been pushed to address this.

web/src/lib/utils/asset-utils.ts
```diff
@@ -438,6 +440,26 @@ export const deleteStack = async (stackIds: string[]) => {
   }
 };
 
+export const keepThisDeleteOthers = async (keepAsset: AssetResponseDto, stack: StackResponseDto) => {
+  const $t = get(t);
+
+  try {
+    const assetsToDeleteIds = stack.assets.filter((asset) => asset.id !== keepAsset.id).map((asset) => asset.id);
+    await deleteStacks({ bulkIdsDto: { ids: [stack.id] } });
+    await deleteAssets({ assetBulkDeleteDto: { ids: assetsToDeleteIds } });
```
mertalev, line 449: `deleteAssets` will permanently delete these assets if the trash feature is disabled, so you'll want to display a deletion warning and get user confirmation. There should be other examples of this in the code.

bdavis2-PCTY: Good to know.  I'll look into doing this!

bdavis2-PCTY: Confirmation dialog has been implemented when clicking "Keep this, delete others."
image

i18n/en.json
```diff
@@ -462,6 +462,7 @@
   "confirm": "Confirm",
   "confirm_admin_password": "Confirm Admin Password",
   "confirm_delete_shared_link": "Are you sure you want to delete this shared link?",
+  "confirm_keep_this_delete_others": "Are you sure you want to keep this asset and delete the others?",
```
mertalev, line 465:
```suggestion
  "confirm_keep_this_delete_others": "All other assets in the stack will be deleted except for this asset. Are you sure you want to continue?",
```

web/src/lib/utils/asset-utils.ts
mertalev, line 449: It's better to switch the order of these two requests. If the delete fails, it will avoid the stack becoming unstacked.

bdavis2-PCTY: Sorry for the delay. Good suggestion. Updated in next commit.

bdavis2-PCTY:
As you all can see, I totally messed up the git commit log by accident. 

I've made a new branch to clear all that here: https://github.com/immich-app/immich/pull/14217/files

bo0tzz:
For future reference, you don't need to create new PRs to clean up a mess, just fix it locally and then force-push to your PR branch.

PR #13705, opened 2024-10-23
refactor(server): telemetry env variables

danieldietzler:
> [!CAUTION]
> The following env variables have been removed
> - `IMMICH_METRICS`
> - `IMMICH_API_METRICS`
> - `IMMICH_IO_METRICS`
> - `IMMICH_JOB_METRICS`
> - `IMMICH_REPO_METRICS`
> Use `IMMICH_TELEMETRY_INCLUDE`/`IMMICH_TELEMETRY_EXCLUDE`.
>
> ### Examples:
>
> ```
> IMMICH_TELEMETRY_INCLUDE=all
> ```
>
> ```
> IMMICH_TELEMETRY_INCLUDE=all
> IMMICH_TELEMETRY_EXCLUDE=repo
> ```
>
> ```
> IMMICH_TELEMETRY_INCLUDE=api,repo
> ```

### DIFF
--- a/docs/docs/install/environment-variables.md
+++ b/docs/docs/install/environment-variables.md
@@ -183,15 +183,10 @@ Other machine learning parameters can be tuned from the admin UI.
 
 ## Prometheus
 
-| Variable                       | Description                                                                                   | Default | Containers | Workers            |
-| :----------------------------- | :-------------------------------------------------------------------------------------------- | :-----: | :--------- | :----------------- |
-| `IMMICH_METRICS`<sup>\*1</sup> | Toggle all metrics (one of [`true`, `false`])                                                 |         | server     | api, microservices |
-| `IMMICH_API_METRICS`           | Toggle metrics for endpoints and response times (one of [`true`, `false`])                    |         | server     | api, microservices |
-| `IMMICH_HOST_METRICS`          | Toggle metrics for CPU and memory utilization for host and process (one of [`true`, `false`]) |         | server     | api, microservices |
-| `IMMICH_IO_METRICS`            | Toggle metrics for database queries, image processing, etc. (one of [`true`, `false`])        |         | server     | api, microservices |
-| `IMMICH_JOB_METRICS`           | Toggle metrics for jobs and queues (one of [`true`, `false`])                                 |         | server     | api, microservices |
-
-\*1: Overridden for a metric group when its corresponding environmental variable is set.
+| Variable                   | Description                                                                                                           | Default | Containers | Workers            |
+| :------------------------- | :-------------------------------------------------------------------------------------------------------------------- | :-----: | :--------- | :----------------- |
+| `IMMICH_TELEMETRY_INCLUDE` | Collect these telemetries. List of `host`, `api`, `io`, `repo`, `job`. Note: You can also specify `all` to enable all |         | server     | api, microservices |
+| `IMMICH_TELEMETRY_EXCLUDE` | Do not collect these telemetries. List of `host`, `api`, `io`, `repo`, `job`                                          |         | server     | api, microservices |
 
 ## Docker Secrets
 
--- a/server/src/enum.ts
+++ b/server/src/enum.ts
@@ -363,3 +363,11 @@ export enum ImmichWorker {
   API = 'api',
   MICROSERVICES = 'microservices',
 }
+
+export enum ImmichTelemetry {
+  HOST = 'host',
+  API = 'api',
+  IO = 'io',
+  REPO = 'repo',
+  JOB = 'job',
+}
--- a/server/src/repositories/config.repository.ts
+++ b/server/src/repositories/config.repository.ts
@@ -2,7 +2,7 @@ import { Injectable } from '@nestjs/common';
 import { join } from 'node:path';
 import { citiesFile, excludePaths } from 'src/constants';
 import { Telemetry } from 'src/decorators';
-import { ImmichEnvironment, ImmichWorker, LogLevel } from 'src/enum';
+import { ImmichEnvironment, ImmichTelemetry, ImmichWorker, LogLevel } from 'src/enum';
 import { EnvData, IConfigRepository } from 'src/interfaces/config.interface';
 import { DatabaseExtension } from 'src/interfaces/database.interface';
 import { QueueName } from 'src/interfaces/job.interface';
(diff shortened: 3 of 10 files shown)


docs/docs/install/environment-variables.md
```diff
@@ -183,13 +183,10 @@ Other machine learning parameters can be tuned from the admin UI.
...
-| `IMMICH_HOST_METRICS`          | Toggle metrics for CPU and memory utilization for host and process (one of [`true`, `false`]) |         | server     | api, microservices |
-| `IMMICH_IO_METRICS`            | Toggle metrics for database queries, image processing, etc. (one of [`true`, `false`])        |         | server     | api, microservices |
-| `IMMICH_JOB_METRICS`           | Toggle metrics for jobs and queues (one of [`true`, `false`])                                 |         | server     | api, microservices |
+| Variable                   | Description                                                                  | Default | Containers | Workers            |
+| :------------------------- | :--------------------------------------------------------------------------- | :-----: | :--------- | :----------------- |
+| `IMMICH_TELEMETRY_INCLUDE` | Collect these telemetries. List of `all`, `host`, `api`, `io`, `repo`, `job` |         | server     | api, microservices |
+| `IMMICH_TELEMETRY_EXCLUDE` | Do not collect these telemetries. List of `host`, `api`, `io`, `repo`, `job` |         | server     | api, microservices |
```
mertalev, line 189: You should elaborate on this to explain the behavior. I think the Prometheus doc also references the old envs and should be updated.

docs/docs/install/environment-variables.md
```diff
@@ -183,13 +183,10 @@ Other machine learning parameters can be tuned from the admin UI.
...
 
 \*1: Overridden for a metric group when its corresponding environmental variable is set.
```
mertalev, line 191: Outdated footnote.

docs/docs/install/environment-variables.md
danieldietzler: Yeah, I didn't want this column to become too long. Maybe I'll just explain it in the prom docs?

mertalev: Sounds good! You should change the wording for `all` here though: it's an alternative to a list, not a normal list option.

danieldietzler: Good point!

jrasm91: Yeah I would have list of be everything minus `all` and then either add an asterisk or inline put something like `NOTE: use all to include everything` or something like that.

server/src/enum.ts
```diff
@@ -363,3 +363,11 @@ export enum ImmichWorker {
   API = 'api',
   MICROSERVICES = 'microservices',
 }
+
+export enum ImmichTelemetries {
```
jrasm91, line 367: I would probably make this singular, none of the others are plural (`ImmichEnvironments`, `RouteKeys`, `MetadataKeys`, etc.)

server/src/repositories/config.repository.ts
```diff
@@ -69,12 +68,18 @@ const getEnv = (): EnvData => {
...
-  const hostMetrics = parseBoolean(process.env.IMMICH_HOST_METRICS, globalEnabled);
-  const apiMetrics = parseBoolean(process.env.IMMICH_API_METRICS, globalEnabled);
-  const repoMetrics = parseBoolean(process.env.IMMICH_IO_METRICS, globalEnabled);
-  const jobMetrics = parseBoolean(process.env.IMMICH_JOB_METRICS, globalEnabled);
-  const telemetryEnabled = globalEnabled || hostMetrics || apiMetrics || repoMetrics || jobMetrics;
+  const telemetryInclude =
+    process.env.IMMICH_TELEMETRY_INCLUDE === 'all'
+      ? Object.values(ImmichTelemetries).join(',')
+      : process.env.IMMICH_METRICS_INCLUDE;
+  const includedTelemetries = asSet(telemetryInclude, [] as ImmichTelemetries[]);
```
jrasm91, line 75: I think we should let `asSet` convert the string input to a set. If the set has a single item and it is `all`, then replace it. This will also prevent you from needing to do this weird string manipulation right here.

server/src/repositories/config.repository.ts
```diff
@@ -168,11 +173,11 @@ const getEnv = (): EnvData => {
...
-      enabled: telemetryEnabled,
-      hostMetrics,
-      apiMetrics,
-      repoMetrics,
-      jobMetrics,
+      enabled: telemetries.size > 0,
+      hostMetrics: telemetries.has(ImmichTelemetries.HOST),
+      apiMetrics: telemetries.has(ImmichTelemetries.API),
+      repoMetrics: telemetries.has(ImmichTelemetries.REPO),
+      jobMetrics: telemetries.has(ImmichTelemetries.JOB),
```
jrasm91, line 180: for workers, we just passed it in as `{ workers }`. Maybe we should do the same for telemetries as well. The services that use this can change their usage from `if(telemetry.hostMetrics)` to `if(telementry.has(ImmichTelemetry.HOST))` and `if(telemetry.size > 0)`, etc. I think that would be nice and consistent with workers.

jrasm91 (commented):
Looks great! Can you add some more tests similar to workers, to make sure we're covering all the conditions?

server/src/repositories/config.repository.ts
danieldietzler: Yeah that's a good point

server/src/repositories/config.repository.ts
```diff
@@ -69,12 +68,21 @@ const getEnv = (): EnvData => {
...
+  const isAll = (telemetries: Set<ImmichTelemetry | 'all'>): telemetries is Set<'all'> =>
+    telemetries.has('all') && telemetries.size === 1;
+
+  let includedTelemetries = asSet<ImmichTelemetry | 'all'>(process.env.IMMICH_TELEMETRY_INCLUDE, []);
+  if (isAll(includedTelemetries)) {
+    includedTelemetries = new Set(Object.values(ImmichTelemetry)) as Set<ImmichTelemetry>;
+  }
+
+  const excludedTelemetries = asSet<ImmichTelemetry>(process.env.IMMICH_TELEMETRY_EXCLUDE, []);
+  const telemetries = setDifference(includedTelemetries as Set<ImmichTelemetry>, excludedTelemetries);
```
mertalev, line 80:
```suggestion
  const includedTelemetries =
    process.env.IMMICH_TELEMETRY_INCLUDE === 'all'
      ? asSet<ImmichTelemetry>(process.env.IMMICH_TELEMETRY_INCLUDE, [])
      : new Set(Object.values(ImmichTelemetry));

  const excludedTelemetries = asSet<ImmichTelemetry>(process.env.IMMICH_TELEMETRY_EXCLUDE, []);
  const telemetries = setDifference(includedTelemetries, excludedTelemetries);
```

jrasm91: this is what I was thinking too @danieldietzler

server/src/repositories/telemetry.repository.ts
```diff
@@ -99,17 +99,18 @@ export class TelemetryRepository implements ITelemetryRepository {
...
+    this.host = new MetricGroupRepository(metricService).configure({ enabled: metrics.has(ImmichTelemetry.HOST) });
+    this.jobs = new MetricGroupRepository(metricService).configure({ enabled: metrics.has(ImmichTelemetry.JOB) });
+    this.repo = new MetricGroupRepository(metricService).configure({ enabled: metrics.has(ImmichTelemetry.REPO) });
   }
 
   setup({ repositories }: { repositories: ClassConstructor<unknown>[] }) {
     const { telemetry } = this.configRepository.getEnv();
-    if (!telemetry.enabled || !telemetry.repoMetrics) {
+    const { metrics } = telemetry;
+    if (metrics.size === 0 || !metrics.has(ImmichTelemetry.REPO)) {
```
jrasm91, line 113: I believe we only need the second condition here now.

jrasm91 (approved):

server/src/repositories/config.repository.ts
danieldietzler: Ooooh that makes so much more sense lol

server/src/repositories/telemetry.repository.ts
danieldietzler: Yeah that's true

jrasm91 (approved):

PR #13707, opened 2024-10-23
chore(docs): add clarifying note for custom file location migration

jrstrunk:
Just wanted to add a clarifying note on what to do if an immich instance is running already. I had an instance running and because this guide said nothing about copying files but did specify to "restart" the containers (implying that they were already running at one point), I assumed Immich must have some sort of migration job to run after file paths are updated. I realized this was not the case when my containers failed to start after updating the paths. So I just wanted to add a little line to save anyone else that bit of confusion :) 

Thanks for the great app!

### DIFF
--- a/docs/docs/guides/custom-locations.md
+++ b/docs/docs/guides/custom-locations.md
@@ -33,6 +33,8 @@ services:
       - /etc/localtime:/etc/localtime:ro
 ```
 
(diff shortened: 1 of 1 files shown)


jrstrunk:
Hmm actually my immich instance seems a little strange after manually moving the files, maybe we shouldn't officially suggest this. I am rebuilding my thumbnails now and hope that will fix it, but this PR probably shouldn't be merged.

PR #13714, opened 2024-10-24
fix: script in app.html

martabal:
The only place where we have js is in `app.html` and that's causing multiple issues:
- Linters don't work here  
- ~Shared variables like `colorTheme` have to be kept in sync across multiple files.~

That PR fixes all these issues it by moving that script to a complete new ts file `fouc.ts`. That script is also simplified. 
During the first page load/the build, the import statement in `fouc.ts` is directly replaced by the entire content of the imported file to not have to deal with imports. Then the ts is transpiled to js and that js replaces the %app.fouc% placeholder in `app.html`.

The script in `app.html` is now:
image


web/src/hooks.server.ts
```diff
@@ -1,12 +1,22 @@
...
 import type { Handle } from '@sveltejs/kit';
 
 // only used during the build to replace the variables from app.html
 export const handle = (async ({ event, resolve }) => {
   return resolve(event, {
     transformPageChunk: ({ html }) => {
-      return html.replace('%app.font%', overpass).replace('%app.monofont%', overpassMono);
+      const themePrepared = theme.replaceAll(/^export\s+/gm, '');
+      const foucPrepared = fouc.replaceAll(/^import.*$/gm, themePrepared);
+      const scriptFouc = `<script>${transpileFile(foucPrepared)}</script>`;
```
jrasm91, line 14: Why do you have to strip out imports/exports? Why can this not be loaded as a module?

web/src/lib/utils.ts
```diff
@@ -25,6 +25,7 @@ import { mdiCogRefreshOutline, mdiDatabaseRefreshOutline, mdiHeadSyncOutline, md
 import { sortBy } from 'lodash-es';
 import { init, register, t } from 'svelte-i18n';
 import { derived, get } from 'svelte/store';
+import { ModuleKind, transpileModule } from 'typescript';
```
jrasm91, line 28: This file is used in other places so we don't want to be importing from typescript here, because that makes it a runtime dependency.

web/src/lib/utils/fouc.ts
jrasm91, line 1: I'd prefer something like `web/src/lib/app.ts`

web/src/hooks.server.ts
martabal: All variables and methods should be placed within the script block. Correct me if I'm wrong, but if you include imports here, the browser will need to fetch those modules before executing the script, which will defeat the purpose.

web/src/lib/utils/app-utils.ts
jrasm91, line 1: Why would you not put this inside hooks.server.ts directly? It's not used for anything else. It won't ever be used for anything else in the future either.

web/src/hooks.server.ts
jrasm91: The solution isn't to manually resolve the dependencies by hand though. We should probably have the script be dependency-less and throw an error if there are any imports being used.

web/src/hooks.server.ts
```diff
@@ -1,9 +1,16 @@
 import overpass from '$lib/assets/fonts/overpass/Overpass.ttf?url';
 import overpassMono from '$lib/assets/fonts/overpass/OverpassMono.ttf?url';
-import { transpileFile } from '$lib/utils/app-utils';
 import fouc from '$lib/utils/app?raw';
 import theme from '$lib/utils/theme?raw';
 import type { Handle } from '@sveltejs/kit';
+import { ModuleKind, transpileModule } from 'typescript';
+
+export const transpileFile = (content: string) => {
```
jrasm91, line 8: No need for this to be exported.

web/src/hooks.server.ts
martabal: So no shared modules? It's quiet nice to have `theme.ts` so we are sure that if one day we change one of the values here, there will be a lint/build failure

jrasm91: I think there is value in getting linting, type-checking, etc. for this file. But you are getting that at the expense of manually managing module imports here, which I think isn't ideal. What if in the future `theme.ts` imports something from another file? Unless I'm mistaken that would just break this. What ends up happening is that you basically have to keep the dependency tree for this file in sync with this piece of code here. That's the job of a bundler, which you are in part replicating here. The two options I'd like to see instead would be either:
1. This script has _no dependencies_, but other files can import from it. TLDR - move all the stuff that needs to be inlined, into this file and have everything else export it.
2. Define a custom entrypoint for vite which pulls everything in automatically into a single file and then inline that.

web/src/lib/utils/app.ts
```diff
@@ -0,0 +1,18 @@
+// should be the same values as the one in perferences.store.ts
+interface ThemeSetting {
+  value: Theme;
+  system: boolean;
+}
+
+// should be the same values as the ones in constants.ts
+enum Theme {
```
jrasm91, line 8: Does it not work to export these?

martabal: Nope:
```bash
 9:07:52 PM [vite] Error when evaluating SSR module /src/lib/utils/app.ts:
 |- ReferenceError: localStorage is not defined
     at /usr/src/app/src/lib/utils/app.ts:14:21
```

jrasm91: Can we tell svelte that this isn't a SSR module?

PR #13734, opened 2024-10-25
fix: duplicated library scan jobs and api server library watch

zackpollard:
The library cron job wasn't created using the lock, so we would create a cron job for every server and microservices instance


danieldietzler (commented):
Unit tests? KEKWait

jrasm91 (approved):

Issue #13566, opened 2024-10-17
Wrong order of photos on "X years ago" feature


kikendo:
### The bug

Photos taken with different devices, set to different timezones, on the same day, have the wrong order on the "X years ago" feature, on both web and mobile.
They show up in chronological order in the timeline view, as they should

### The OS that Immich Server is running on

Debian 5.10

### Version of Immich Server

v1.118.2

### Version of Immich Mobile App

1.118.0 build 180

### Platform with the issue

- [ ] Server
- [X] Web
- [X] Mobile

### Your docker-compose.yml content

```YAML
.
```

### Your .env content

```Shell
.
```

### Reproduction steps

Have photos on the same date that have different timezones. I'm not sure it's the device creating them but the timezones.

kikendo:
Adding to the above I can confirm that the problem is that the order in which images are shown on that feature only follows the time and ignores the timezone, so you would see something like this:
1) Photo taken at 1:30 GMT+3
2) Photo taken at 1:31 GMT+3
3) Photo taken at 1:34 GMT-5
4) Photo taken at 1:38 GMT+3

etc.

PR #13741, opened 2024-10-26
fix(server): account for time zone information when deriving localDateTime

darrelhong:
Was looking into #13566 and initially thought the a solution would be to sort the images before returning. But noticed it uses localDateTime in the query and it does not account for the time zone. I think the fix is to update the date calculation logic in metadata service to offset with time zone information if any.

**BEFORE**
Screenshot 2024-10-26 at 9 36 04 AM

**AFTER**
Screenshot 2024-10-26 at 9 45 45 AM

Conveniently it looks like an adjacent test case covers this scenario so I have updated the expected return object for the failing test.

### DIFF
--- a/server/src/services/metadata.service.spec.ts
+++ b/server/src/services/metadata.service.spec.ts
@@ -835,7 +835,7 @@ describe(MetadataService.name, () => {
         id: assetStub.image.id,
         duration: null,
         fileCreatedAt: dateForTest,
-        localDateTime: dateForTest,
+        localDateTime: new Date('1970-01-01T23:00:00.000Z'),
       });
     });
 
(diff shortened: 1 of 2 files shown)


jrasm91:
This is working as designed although we need to document it better. `localDateTime` is only used to show should yyyy-mm-dd hh:mm:ss and never has a timezone offset because it represents the time the picture was taken, wherever it was taken. In your example the picture was taken at 7:59 so we want this to have a localDateTime of 7:59. We just always ignore the timezone part of it. Maybe in the future we will store it without a timezone but that also has its own set of complications. You can use asset.fileCreatedAt or exif.dateTimeOriginal to get a timezone adjusted date.

darrelhong:
gotcha, would a better fix be to update the `getByDayOfYear` query to sort by `entity. fileCreatedAt` instead?

jrasm91:
Yup, this should sort by fileCreatedAt

https://github.com/immich-app/immich/blob/main/server%2Fsrc%2Frepositories%2Fasset.repository.ts#L96

PR #13752, opened 2024-10-26
fix: remove duplicateIds on unique assets

### LINKED ISSUE #11637: Duplicate group with just one image after manually deleting the other(s)
### The bug

If I manually delete a (duplicate) image, the "review duplicates" utility shows the corresponding group with the single (!) leftover image. I expect that this is never shown, either silently ignored (when otherwise shown) or properly cleaned up (when deleting an image from the duplicate group).

Debian Stable

### Version of Immich Server

v1.111.0

None

### Platform with the issue

- [X] Server
- [X] Web
- [ ] Mobile

Default from v1.111.0, with minor changes (listen to 127.0.0.1 only, add two volumes for external libraries)

### Your .env content

Default, aside from paths and passwords

### Reproduction steps
In my case I'm using an external library. Deleting from within the web UI (which just deletes metadata, but not the file itself?) or from outside (and rescanning the library for missing files) doesn't seem to make a difference.

1. Add image to two separate locations in external library
2. Run duplicate scan
3. Manually delete one of the duplicates
4. Check "review duplicates view"

### Relevant log output

_No response_

### Additional information

_No response_

Pranay-Pandey:
fixes: #11637 

## Issue Overview
A bug was identified in the review duplicates utility where, after manually deleting a duplicate image, the corresponding group was still displayed with a single leftover image. 

## Changes 
We validate the duplicates in duplicates utility page by checking if each duplicate set has atleast 2 assets. If not we update the assets to remove duplicates ids from them and do not show them on the page.

### DIFF
--- a/server/src/services/duplicate.service.ts
+++ b/server/src/services/duplicate.service.ts
@@ -16,13 +16,15 @@ export class DuplicateService extends BaseService {
   async getDuplicates(auth: AuthDto): Promise<DuplicateResponseDto[]> {
     const res = await this.assetRepository.getDuplicates({ userIds: [auth.user.id] });
     const uniqueAssetIds: string[] = [];
-    const duplicates = mapDuplicateResponse(res.map((a) => mapAsset(a, { auth, withStack: true }))).filter((duplicate) => {
-      if (duplicate.assets.length === 1) {
-        uniqueAssetIds.push(duplicate.assets[0].id);
-        return false;
-      }
-      return true;
-    });
+    const duplicates = mapDuplicateResponse(res.map((a) => mapAsset(a, { auth, withStack: true }))).filter(
+      (duplicate) => {
+        if (duplicate.assets.length === 1) {
+          uniqueAssetIds.push(duplicate.assets[0].id);
+          return false;
+        }
+        return true;
+      },
+    );
     if (uniqueAssetIds.length > 0) {
       void this.assetRepository.updateAll(uniqueAssetIds, { duplicateId: null });
     }
--- a/server/src/services/duplicate.service.ts
+++ b/server/src/services/duplicate.service.ts
@@ -26,7 +26,11 @@ export class DuplicateService extends BaseService {
       },
     );
     if (uniqueAssetIds.length > 0) {
-      void this.assetRepository.updateAll(uniqueAssetIds, { duplicateId: null });
+      try {
+        await this.assetRepository.updateAll(uniqueAssetIds, { duplicateId: null });
+      } catch (error: any) {
+        this.logger.error(`Failed to remove duplicateId from assets: ${error.message}`);
+      }
     }
     return duplicates;
   }
--- a/server/src/services/duplicate.service.ts
+++ b/server/src/services/duplicate.service.ts
@@ -15,8 +15,18 @@ import { usePagination } from 'src/utils/pagination';
 export class DuplicateService extends BaseService {
   async getDuplicates(auth: AuthDto): Promise<DuplicateResponseDto[]> {
     const res = await this.assetRepository.getDuplicates({ userIds: [auth.user.id] });
-
-    return mapDuplicateResponse(res.map((a) => mapAsset(a, { auth, withStack: true })));
+    const uniqueAssetIds: string[] = [];
+    const duplicates = mapDuplicateResponse(res.map((a) => mapAsset(a, { auth, withStack: true }))).filter((duplicate) => {
+      if (duplicate.assets.length === 1) {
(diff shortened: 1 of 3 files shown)


web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.ts
```diff
@@ -1,13 +1,33 @@
...
+  });
+  rectifyDuplicate(uniqueAssetIds);
+  return validDuplicates;
+};
+
 export const load = (async ({ params }) => {
   await authenticate();
   const asset = await getAssetInfoFromParam(params);
-  const duplicates = await getAssetDuplicates();
+  const Allduplicates = await getAssetDuplicates();
```
Comment (mertalev) [inline web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.ts:29]: This should be camel-case. Also, while you're changing this, could you also put these promises (except `authenticate`) in a `Promise.all` call so they happen in parallel?

web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.ts
```diff
@@ -1,13 +1,33 @@
...
+  const duplicates = processDuplicates(Allduplicates);
```
Comment (mertalev) [inline web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.ts:30]: This is up to preference, but I think it'd be cleaner if this returned the valid and unique duplicates and you call `rectifyDuplicates` on the latter.

mertalev (commented):
Some code style comments, but looks good overall! I was thinking of adding a nightly check for this, but lazily checking when you fetch them is smarter.

web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.ts
Comment (Pranay-Pandey) [reply web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.ts:30]: Do we mean the getAssetDuplicates function to return these two lists?

Comment (mertalev) [reply web/src/routes/(user)/utilities/duplicates/[[photos=photos]]/[[assetId=id]]/+page.ts:30]: Yup!

bo0tzz:
Would this be better to do on the server side? It'd save a bunch of round trips etc.

mertalev:
Yeah, that makes sense to me.

server/src/services/duplicate.service.ts
```diff
@@ -15,8 +15,20 @@ import { usePagination } from 'src/utils/pagination';
...
+      (duplicate) => {
+        if (duplicate.assets.length === 1) {
+          uniqueAssetIds.push(duplicate.assets[0].id);
+          return false;
+        }
+        return true;
+      },
+    );
+    if (uniqueAssetIds.length > 0) {
+      void this.assetRepository.updateAll(uniqueAssetIds, { duplicateId: null });
```
danieldietzler, line 29: Why are you not awaiting this?

Pranay-Pandey: My reasoning is that the update operation is non-critical to the response, so it doesn’t need to hold up the function's main purpose, which is to return valid duplicates.

danieldietzler: Not awaiting the promise will make catching a potential error impossible though. So, if the update should fail for whatever reason, the process is f'ed

Pranay-Pandey: I see. 
Would a try catch block be the correct way of handling this?

danieldietzler: No. You need to await it. It's not like this is gonna take long either way though, so whatever

server/src/services/duplicate.service.ts
```diff
@@ -15,8 +15,24 @@ import { usePagination } from 'src/utils/pagination';
 export class DuplicateService extends BaseService {
   async getDuplicates(auth: AuthDto): Promise<DuplicateResponseDto[]> {
     const res = await this.assetRepository.getDuplicates({ userIds: [auth.user.id] });
-
-    return mapDuplicateResponse(res.map((a) => mapAsset(a, { auth, withStack: true })));
+    const uniqueAssetIds: string[] = [];
```
bo0tzz, line 18: Naming nitpick: These aren't necessarily _unique_, right? Not sure what a better name would be though.

bo0tzz (approved):
Nice, much better, thanks!

server/src/services/duplicate.service.ts
mertalev: `singletonDuplicateIds`?

mertalev (approved):

PR #13773, opened 2024-10-28
feat: built-in automatic database backups

zackpollard:
image


server/src/services/backup.service.ts
```diff
@@ -0,0 +1,89 @@
...
+  async onBootstrap() {
+    const config = await this.getConfig({ withCache: false });
+
+    const { database } = config.backups;
+
+    // This ensures that backups only occur in one microservice
+    this.backupLock = await this.databaseRepository.tryLock(DatabaseLock.BackupDatabase);
+
+    if (this.backupLock) {
+      this.jobRepository.addCronJob(
```
bo0tzz, line 25: What does this mean in a scenario where there's multiple microservices workers (eg a scaled setup) and the one that got this lock is removed? Is the cron job in memory meaning it will no longer run? And are there any other (cron)jobs that applies to?

server/src/services/backup.service.ts
```diff
@@ -0,0 +1,89 @@
...
+import { OnEvent } from 'src/decorators';
+import { DatabaseLock } from 'src/interfaces/database.interface';
+import { ArgOf } from 'src/interfaces/event.interface';
+import { JobName, JobStatus } from 'src/interfaces/job.interface';
+import { BaseService } from 'src/services/base.service';
+import { handlePromiseError } from 'src/utils/misc';
+import { validateCronExpression } from 'src/validation';
+
+@Injectable()
+export class BackupService extends BaseService {
```
bo0tzz, line 12: nit, should this be DatabaseBackupService?

web/src/lib/components/admin-page/settings/backup-settings/backup-settings.svelte
```diff
@@ -0,0 +1,84 @@
...
+</script>
+
+<div>
+  <div in:fade={{ duration: 500 }}>
+    <form autocomplete="off" on:submit|preventDefault>
+      <div class="ml-4 mt-4 flex flex-col gap-4">
+        <SettingSwitch
+          title={$t('admin.backup_database_enable_description')}
+          {disabled}
+          bind:checked={config.library.scan.enabled}
```
bo0tzz, line 36: Wrong bind?

bo0tzz (commented):
I realise it's still draft ;) Just a few stray thoughts

server/src/services/backup.service.ts
zackpollard: I planned for this to be where all backup logic lives going forwards.

server/src/services/backup.service.ts
zackpollard: This is a fair point, but an issue we already have with the setup. I am not sure how we would easily resolve this right now. I'm not sure if we explicitly support dynamic scaling right now. Maybe we should force cron creation to run on the api server, as they are less likely to be dynamically scaled I would expect.

server/src/config.ts
```diff
@@ -156,6 +163,13 @@ export interface SystemConfig {
 }
 
 export const defaults = Object.freeze<SystemConfig>({
+  backups: {
```
jrasm91, line 166: No other keys in this list are plurals.

server/src/main.ts
```diff
@@ -24,6 +28,10 @@ function bootstrapWorker(name: ImmichWorker) {
   worker.on('exit', (exitCode) => {
     if (exitCode !== 0) {
       console.error(`${name} worker exited with code ${exitCode}`);
+      if (apiProcess && name !== 'api') {
+        console.error('Killing api process');
+        apiProcess.kill('SIGTERM');
+      }
```
jrasm91, line 34: It would be nice if you threw this into it's own PR.

server/src/services/backup.service.ts
```diff
@@ -0,0 +1,178 @@
...
+          if (code !== 0) {
+            this.logger.error(`Gzip failed with code ${code}`);
+            reject(`Gzip failed with code ${code}`);
+            this.logger.error(gzipLogs);
+            return;
+          }
+          if (pgdump.exitCode !== 0) {
+            this.logger.error(`Gzip exited with code 0 but pgdump exited with ${pgdump.exitCode}`);
+            return;
+          }
```
jrasm91, line 157: this doesn't have a resolve/reject, does that imply an error event was thrown and you already returned something elsewhere?

zackpollard: Yea, this is just handling the case that pgdump errored but gzip exits successfully, we have to handle that case differently.

jrasm91: I mean where is the promise resolved or rejected in this scenario? In the `on('error')` event handler?

server/src/services/backup.service.ts
```diff
@@ -0,0 +1,179 @@
...
+      backup: { database: config },
+    } = await this.getConfig({ withCache: false });
+
+    const backupsFolder = StorageCore.getBaseFolder(StorageFolder.BACKUPS);
+    const files = await this.storageRepository.readdir(backupsFolder);
+    const failedBackups = files.filter((file) => file.match(/immich-db-backup-\d+\.sql\.gz\.tmp$/));
+    const backups = files
+      .filter((file) => {
+        return file.match(/immich-db-backup-\d+\.sql\.gz$/);
+      })
```
danieldietzler, line 67:
```suggestion
      .filter((file) => file.match(/immich-db-backup-\d+\.sql\.gz$/))
```

server/src/services/backup.service.ts
```diff
@@ -0,0 +1,179 @@
...
+          }
+          this.storageRepository
+            .rename(backupFilePath, backupFilePath.replace('.tmp', ''))
+            .then(() => {
+              resolve();
+            })
+            .catch((error) => {
+              this.logger.error('Backup failed with error', error);
+              reject(error);
+            });
```
danieldietzler, line 167: I'd prefer awaiting it and wrapping it in a try catch

server/src/services/storage.service.spec.ts
```diff
@@ -74,9 +79,41 @@ describe(StorageService.name, () => {
...
+          'encoded-video': true,
+          library: false,
+          profile: true,
+          thumbs: true,
+          upload: true,
+        },
+      });
+
+      await expect(sut.onBootstrap()).resolves.toBeUndefined();
+
```
danieldietzler, line 100: I am confused. Didn't you move that stuff out?

web/src/lib/components/admin-page/settings/backup-settings/backup-settings.svelte
```diff
@@ -0,0 +1,99 @@
...
+          title={$t('admin.backup_database_enable_description')}
+          {disabled}
+          bind:checked={config.backup.database.enabled}
+        />
+
+        <div class="flex flex-col my-2 dark:text-immich-dark-fg">
+          <label class="font-medium text-immich-primary dark:text-immich-dark-primary text-sm" for="expression-select">
+            {$t('admin.cron_expression_presets')}
+          </label>
+          <select
```
danieldietzler, line 43: There's also `<SettingSelect>`, not sure if you want to use that instead.

server/src/services/backup.service.ts
jrasm91: Ideally this lives outside of the promise which is only needed to interface with the callbacks, but those are done at this point in time

jrasm91 (commented):
Sweet. Can you add screenshots?

server/src/services/backup.service.ts
```diff
@@ -0,0 +1,179 @@
...
+          this.logger.error('Backup failed, could not spawn backup process');
+          reject('Backup failed, could not spawn backup process');
+          return;
+        }
+
+        if (!gzip.stderr || !gzip.stdout) {
+          this.logger.error('Backup failed, could not spawn gzip process');
+          reject('Backup failed, could not spawn gzip process');
+          return;
+        }
```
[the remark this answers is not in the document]
zackpollard: Yea, appears since we moved to the version that can't ever have null streams, we no longer need these cases 😄

server/src/services/backup.service.ts
zackpollard: Moved it just below, then we don't even need another try catch :)

docs/docs/administration/backup-and-restore.md
```diff
@@ -21,6 +21,17 @@ The recommended way to backup and restore the Immich database is to use the `pg_
 It is not recommended to directly backup the `DB_DATA_LOCATION` folder. Doing so while the database is running can lead to a corrupted backup that cannot be restored.
 :::
 
+### Automatic Database Backups
+
+Immich will automatically create database backups by default. The backups are stored in `UPLOAD_LOCATION/backups`.  
+You can adjust the schedule and amount of kept backups in the [admin settings](http://my.immich.app:2283/admin/system-settings?isOpen=backup).
```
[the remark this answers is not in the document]
zackpollard: Good spot, apparently Boet's nice bookmarklet doesn't remove the port :(

server/src/repositories/storage.repository.ts
```diff
@@ -2,9 +2,10 @@ import { Inject, Injectable } from '@nestjs/common';
 import archiver from 'archiver';
 import chokidar, { WatchOptions } from 'chokidar';
 import { escapePath, glob, globStream } from 'fast-glob';
-import { constants, createReadStream, existsSync, mkdirSync } from 'node:fs';
+import fsSync, { constants, createReadStream, existsSync, mkdirSync } from 'node:fs';
```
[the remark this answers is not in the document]
zackpollard: Yea, I think this import got messed up when it got refactored

web/src/lib/components/admin-page/settings/backup-settings/backup-settings.svelte
zackpollard: Nice, wasn't aware of that, will make a separate PR to update this on the library settings too :)

danieldietzler (approved):
Great job!

PR #13775, opened 2024-10-28
chore(docs): Encode db dump in UTF-8 without BOM for Windows

amorphobia:
Related to #11787 

On Windows, the encoding “utf8” actually means “UTF-8 with BOM”. What we need is “utf8NoBOM”. However it’s not available from the Windows 10 built-in PowerShell.

This change suggests `WriteAllLines` function from the pre-installed .net framework, which uses real UTF-8 encoding by default.

https://learn.microsoft.com/en-us/dotnet/api/system.io.file.writealllines?view=netframework-4.6

> The default behavior of the [WriteAllLines(String, IEnumerable<String>)](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.writealllines?view=netframework-4.6#system-io-file-writealllines(system-string-system-collections-generic-ienumerable((system-string)))) method is to write out data by using UTF-8 encoding without a byte order mark (BOM).

### DIFF
--- a/docs/docs/administration/backup-and-restore.md
+++ b/docs/docs/administration/backup-and-restore.md
@@ -49,7 +49,7 @@ docker compose up -d            # Start remainder of Immich apps
   <TabItem value="Windows system (PowerShell)" label="Windows system (PowerShell)">
 
 ```powershell title='Backup'
-docker exec -t immich_postgres pg_dumpall --clean --if-exists --username=postgres | Set-Content -Encoding utf8 "C:\path\to\backup\dump.sql"
+[System.IO.File]::WriteAllLines("C:\absolute\path\to\backup\dump.sql", (docker exec -t immich_postgres pg_dumpall --clean --if-exists --username=postgres))
 ```
 
 ```powershell title='Restore'


NicholasFlamy:
It works, but I don't understand the need for it. Please enlighten me.

amorphobia:
The previous PR #11787 said

> Powershell does not use UTF-8 encoding by default, which makes database dumps incompatible between Windows and Linux. This PR makes Windows save the dump in UTF-8 format.

And UTF-8 without BOM is the de facto encoding in Linux. Some software may not handle BOM well. Just in case BOM-less is really needed.

I personally do not have preference on BOM. It's up to the maintainers.

NicholasFlamy:
> UTF-8 without BOM is the de facto encoding in Linux. Some software may not handle BOM well. Just in case BOM-less is really needed.

I presume the Automatic backups as well as the manual Linux backups save to this format as well so I take it that having the manual Windows backup save in this format makes sense.

NicholasFlamy (approved):
Tested the command. It does work. LGTM

zackpollard (approved):

zackpollard:
LGTM, thanks for the change and the explanation!

PR #13785, opened 2024-10-28
feat(server): use tonemapx for software tone-mapping

### LINKED ISSUE #13709: Corrupt transcoded video with rkmpp decoder.
### The bug

X265 4K videos, transcoded down to 720p x265 using rkmmp hardware accelerated encoding and decoding are sometimes corrupted causing looped playback and strange colours and artifacts when played back.
Video seem fine with HW decoding disabled.
It does not happen every time and is inconsistent but happens often.

Armbian Bookworm 

### Version of Immich Server

V1.118.2

V1.118.0

### Platform with the issue

- [ ] Server
- [ ] Web
- [ ] Mobile

```YAML
Same as default
```

### Your .env content

```Shell
Same as default
```

### Reproduction steps

1.
2.
3.
...

### Relevant log output

_No response_

### Additional information

_No response_

mertalev:
### Description

This PR changes the current zscale-based tone-mapping filter chain to tonemapx, a custom filter that's significantly faster.

Testing on a 4K60 HDR video showed major improvements in overall transcoding speed, especially for higher resolutions. For a 4K target, the tonemapx-based command is 69% faster in real time.

| Filter | Target Resolution | Real Time (s) | User Time (s) |
| ----- | ------------------- | ---------- | ------------ | 
| zscale | 720p | 50.167 | 600.08 |
| tonemapx | 720p | **40.85** | **540.65** |
| zscale | 1080p | 72.79 | 1491.05 |
| tonemapx | 1080p | **50.06** | **655.72** |
| zscale | 2160p | 172.63 | 2244.80 |
| tonemapx | 2160p | **102.10** | **1393.57** |

I also set `peak=100` for all tonemap filters following the advice of Jellyfin devs, as videos often have inaccurate data for peak luminance.

Additionally, as tonemapx does not have an `npl` setting, this setting is removed from the system config.

Note that because `tonemapx` is a custom filter exclusive to `jellyfin-ffmpeg`, Immich distributions that use other FFmpeg builds will be unable to use software tone-mapping. This means that thumbnail generation and transcoding for HDR videos will fail with the default configuration. I've reached out to the maintainers for the Arch and Nix packages to smoothen this transition.

### Testing

I tested with different tone-mapping algorithms and tone-mapping disabled, as well as with hardware encoding (but not decoding) enabled on both HDR and SDR source videos. All were transcoded successfully. The same applies for thumbnail generation.

Fixes #13709


danieldietzler (approved):

jrasm91 (approved):
Looks good! Should we write a migration to clear out this value from the database for those who have changed the default to something else?

mertalev:
Sure! It'd avoid warning logs about it.

PR #13801, opened 2024-10-29
feat: mount checks on a folder level

zackpollard:


server/src/services/storage.service.ts
```diff
@@ -19,25 +20,36 @@ export class StorageService extends BaseService {
...
             this.logger.log(`Writing initial mount file for the ${folder} folder`);
             await this.createMountFile(folder);
           }
 
           await this.verifyReadAccess(folder);
           await this.verifyWriteAccess(folder);
+
+          if (!flags.mountChecks) {
+            flags.mountChecks = {};
+          }
```
danieldietzler: Why is this necessary? That should be defined either way in line 23-25, no?

danieldietzler (approved):

jrasm91 (approved):

PR #13802, opened 2024-10-29
fix: shutdown api process when another worker exits unexpectedly

zackpollard:


server/src/main.ts
```diff
@@ -10,12 +10,16 @@ if (immichApp) {
...
 
+let apiProcess: ChildProcess | undefined;
+
 function bootstrapWorker(name: ImmichWorker) {
   console.log(`Starting ${name} worker`);
 
   const execArgv = process.execArgv.map((arg) => (arg.startsWith('--inspect') ? '--inspect=0.0.0.0:9231' : arg));
   const worker =
-    name === 'api' ? fork(`./dist/workers/${name}.js`, [], { execArgv }) : new Worker(`./dist/workers/${name}.js`);
+    name === 'api'
```
jrasm91, line 20: Can you change this to use the enum?

jrasm91 (approved):

jrasm91 (approved):

PR #13817, opened 2024-10-29
refactor(server): env validation

jrasm91:
Replace joi with class-validator.

Technically this doesn't support local `.env` files _directly_. A `.env` file is still passed into the container through docker. If we need `.env` file support by the main node.js process we can add a call to `dotenv.config()` back in.


danieldietzler (approved):
Great job! 🎉

bo0tzz (approved):
Dig it

PR #13834, opened 2024-10-30
Simplify running as non-root user using the env file and tmpfs

eiqnepm:
Allows to easily change the user which all the containers use. I changed the model cache from a Docker managed volume to a tmpfs as it's the only way to have the owner and group apply automatically without the user having to create directory and apply the owner and group manually

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -29,6 +29,7 @@ services:
     restart: always
     healthcheck:
       disable: false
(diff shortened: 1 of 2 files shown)


bo0tzz:
Thanks for the PR! We want to keep the default compose file as straightforward as possible, but if the non-root setup can be clarified in the docs then a PR for that would be welcome. (Note there's a similar PR in #13257)

PR #13849, opened 2024-10-31
docs: improve custom-locations wording to be easier to read

slamp:
It's only grammatical change

### DIFF
--- a/docs/docs/guides/custom-locations.md
+++ b/docs/docs/guides/custom-locations.md
@@ -40,8 +40,8 @@ docker compose up -d
 ```
 
 :::note
-Due to the underlying properties of docker bind mounts, it is not recommended to mount the `upload/` and `library/` folders as separate bind mounts if they are on the same device.
-For this reason, we mount the hard drive (HDD) or the network storage (NAS) to `/usr/src/app/upload` and then mount the folders we want to access under that folder.
+Because of the underlying properties of docker bind mounts, it is not recommended to mount the `upload/` and `library/` folders as separate bind mounts if they are on the same device.
+For this reason, we mount the HDD or the network storage (NAS) to `/usr/src/app/upload` and then mount the folders we want to access under that folder.
 
 The `thumbs/` folder contains both the small thumbnails displayed in the timeline and the larger previews shown when clicking into an image. These cannot be separated.
 
--- a/docs/docs/guides/custom-locations.md
+++ b/docs/docs/guides/custom-locations.md
@@ -1,15 +1,15 @@
 # Files Custom Locations
 
-This guide explains storing generated and raw files with docker's volume mount in different locations.
+This guide explains how to store generated and raw files with docker's volume mount in different locations.
 
 :::caution Backup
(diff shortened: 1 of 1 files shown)


docs/docs/guides/custom-locations.md
```diff
@@ -40,13 +40,12 @@ docker compose up -d
 ```
 
 :::note
-Because of the underlying properties of docker bind mounts, it is not recommended to mount the `upload/` and `library/` folders as separate bind mounts if they are on the same device.
-For this reason, we mount the HDD or network storage to `/usr/src/app/upload` and then mount the folders we want quick access to below this folder.
+Due to the underlying properties of docker bind mounts, it is not recommended to mount the `upload/` and `library/` folders as separate bind mounts if they are on the same device.
+For this reason, we mount the hard drive (HDD) or the network storage (NAS) to `/usr/src/app/upload` and then mount the folders we want to access under that folder.
 
-The `thumbs/` folder contains both the small thumbnails shown in the timeline, and the larger previews shown when clicking into an image. These cannot be split up.
+The `thumbs/` folder contains both the small thumbnails displayed in the timeline and the larger previews shown when clicking into an image. The latter cannot be separated.
```
bo0tzz, line 46:
```suggestion
The `thumbs/` folder contains both the small thumbnails displayed in the timeline and the larger previews shown when clicking into an image. These cannot be separated.
```

docs/docs/guides/custom-locations.md
```diff
@@ -40,13 +40,12 @@ docker compose up -d
 ```
 
 :::note
-Because of the underlying properties of docker bind mounts, it is not recommended to mount the `upload/` and `library/` folders as separate bind mounts if they are on the same device.
-For this reason, we mount the HDD or network storage to `/usr/src/app/upload` and then mount the folders we want quick access to below this folder.
+Due to the underlying properties of docker bind mounts, it is not recommended to mount the `upload/` and `library/` folders as separate bind mounts if they are on the same device.
+For this reason, we mount the hard drive (HDD) or the network storage (NAS) to `/usr/src/app/upload` and then mount the folders we want to access under that folder.
```
bo0tzz, line 44: "the HDD or network storage" is fine IMO. No need to explain those concepts :P

bo0tzz (commented):
Thanks for the PR!

docs/docs/guides/custom-locations.md
slamp: Sure! I will amend the commit later this weekend

pyorot: Just came across this; “because of” is actually better there because it describes a verb not a noun. If you don’t like how it sounds then you can swap it for “owing to”, which is synonymous.

slamp:
I revert to 'because of', I removed hard drive but I kept (NAS) to ease the search.

docs/docs/guides/custom-locations.md
slamp: Modification done, see comment below

slamp:
Is there anything else I can do to help the PR progress?

alextran1502 (approved):

PR #13856, opened 2024-10-31
chore: move library settings cron expression to setting select element

zackpollard:


mertalev:
What's the motivation of this change? I think you should still be able to set the cron to something besides one of the presets.

zackpollard:
> What's the motivation of this change? I think you should still be able to set the cron to something besides one of the presets.
> 
> Oh, never mind. I thought this was removing the text field.

I was in the middle of replying :laughing:

mertalev (approved):

web/src/lib/components/admin-page/settings/library-settings/library-settings.svelte
```diff
@@ -60,25 +61,13 @@
...
-                bind:value={config.library.scan.cronExpression}
-              >
-                {#each cronExpressionOptions as { title, expression }}
-                  <option value={expression}>{title}</option>
-                {/each}
-              </select>
-            </div>
+            <SettingSelect
+              options={cronExpressionOptions}
+              disabled={disabled || !config.library.scan.cronExpression}
```
mertalev, line 66: Is it intentional that you changed this from `enabled` to `cronExpression`?

zackpollard: Oops, copy paste error :sweat_smile: Fixed

PR #13859, opened 2024-10-31
fix(web): time zone dependent test

jrasm91:
Fix tests which sometimes break, depending on what timezone you are in.


benmccann (approved):
thanks!!

danieldietzler (approved):

PR #13875, opened 2024-11-01
chore: prune Weblate translations

ben-basten:
## Description

Cleaning up Weblate keys that are causing Dart errors with the code generation process in the [slang Flutter library](https://pub.dev/packages/slang), which will potentially be used on mobile for translations.

* `force_re-scan_library_files` - the hyphen causes problems with the code generation
* `note_apply_storage_label_to_previously_uploaded assets` - the space causes problems with code generation
* `durations` - the object results in additional Dart classes being generated that are not needed and cause Dart inheritance errors, since the object is not present in the `en.json`

The cleanup needs to be done across all languages, because any translation JSON file that has a malformed key will cause Dart errors.

## How Has This Been Tested?

- Searched the code to verify that there are no uses of the keys on the web


i18n/ar.json
```diff
@@ -483,7 +483,7 @@
   "confirm_password": "تأكيد كلمة المرور",
   "contain": "محتواة",
   "context": "السياق",
-  "continue": "متابعة",
```
jrasm91, line 486: Why is this a problem? Seems like it should be a bug with the implementation not something we should have to change.

ben-basten: The `slang` tool generates Dart helper classes to provide fully typed translations. It looks like this:

https://github.com/immich-app/immich/blob/37a57d75fd84ec149ee39b64b1a0c6ae1b7aaea3/mobile/lib/i18n/strings_en.g.dart#L197

If I leave the key as "continue", I get the following error when building or running flutter:

```
Error (Xcode): lib/i18n/strings_en.g.dart:197:13: Error:
'continue' can't be used as an identifier because it's a
keyword.
```

The cause is that the generated code contains a getter using a [Dart reserved keyword](https://dart.dev/language/keywords) as a name. I'm not sure how the upstream library would fix this? Open to thoughts here.

jrasm91: Id assume the upstream library should not generate invalid code. It should either throw an error or use an alternative naming scheme. It's probably worth opening an issue with them about it. I guess it'd be easy enough to change it either way though.

ben-basten: We'll see what they say.

https://github.com/slang-i18n/slang/issues/257

ben-basten: @jrasm91 it sounds like this is the expected behavior for the upstream, and it would be a new feature to include a warning that a key is using a reserved keyword. Would you prefer leaving the "continue_text" change out of this PR, until we're more confident about the migration to Slang?

ben-basten: Looks like a configuration option to fix this has been added in the latest version of Slang. I reverted the change to `continue`.

jrasm91: Awesome!

Tienisto:
@ben-basten A few things to consider:

- Slang supports automatic [recasing](https://pub.dev/packages/slang#-recasing), so you could set `key_case: camel`, then there is no problem with hyphens or spaces - although especially `note_apply_storage_label_to_previously_uploaded assets` seems like a typo and not a feature
- Related to this PR but since I couldn't find the PR of your [mobile-i18n](https://github.com/immich-app/immich/tree/feat/mobile-i18n) branch: You should set `fallback_strategy: base_locale_empty_string` ([doc](https://pub.dev/packages/slang#-fallback)) so that slang fallbacks to base locale if the current locale has an empty string. Alternatively, Weblate supports the [cleanup blank string addon](https://docs.weblate.org/en/latest/admin/addons.html#addon-weblate-cleanup-blank)

ben-basten:
@Tienisto thanks for your time and feedback!

> * Slang supports automatic [recasing](https://pub.dev/packages/slang#-recasing), so you could set `key_case: camel`, then there is no problem with hyphens or spaces - although especially `note_apply_storage_label_to_previously_uploaded assets` seems like a typo and not a feature

Good to know! In this case, it's a typo that needs to be cleaned up.

> * Related to this PR but since I couldn't find the PR of your [mobile-i18n](https://github.com/immich-app/immich/tree/feat/mobile-i18n) branch: You should set `fallback_strategy: base_locale_empty_string` ([doc](https://pub.dev/packages/slang#-fallback)) so that slang fallbacks to base locale if the current locale has an empty string. Alternatively, Weblate supports the [cleanup blank string addon](https://docs.weblate.org/en/latest/admin/addons.html#addon-weblate-cleanup-blank)

Thanks for the pointer, I'll be looking into both of those things for our Immich translations.

> These tools help to migrate to slang seamlessly without affecting other parts of Immich.

One thing that I've been wondering about is Slang namespaces. It looks like the current naming of translation files in [the i18n/ directory](https://github.com/immich-app/immich/tree/main/i18n) won't work, without adding a prefix of `strings_` to all of the names. Is it possible to keep the current naming structure and disable namespaces?

Tienisto:
Unfortunately, the prefix `strings_` (or any other prefix) is currently needed so that Slang knows what is namespace and what is locale (even if there is only one namespace). Maybe this can be added in the future though.

ben-basten:
Closing for now, I've been short on time to implement translations on mobile.

PR #13931, opened 2024-11-05
chore: better observability for geodata import

zackpollard:


mertalev:
How many lines are there? This is going to be a lot of logs, right? It might be better log at larger intervals.

zackpollard:
> How many lines are there? This is going to be a lot of logs, right? It might be better log at larger intervals.

There's 200k entries, so 200 logs, I think this is fine as it only happens on update.

mertalev:
Yeah, but this is going to be in a lot of log reports and bury other things. Can you make it run every 10k? It's only 20 lines that way. Or maybe adjust the interval by log level if you think it's important to be able to log more frequently than that.

zackpollard:
> Yeah, but this is going to be in a lot of log reports and bury other things. Can you make it run every 10k? It's only 20 lines that way. Or maybe adjust the interval by log level if you think it's important to be able to log more frequently than that.

Sure, I don't really care either way. Just means another statement where we already had one.

mertalev (approved):

PR #13948, opened 2024-11-06
fix(server): thumbnail rotation when using embedded previews

### LINKED ISSUE #12720: Photo rotation of embeded previews wrong
### The bug

When I upload raw pictures into Immich and create thumbnails with the "Use embedded preview" feature enabled, Immich rotates vertical pictures 90 degrees. When I turn off the "Use embedded preview" feature, the pictures are rotated correctly, but the colors are displayed incorrectly.

Unraid 7.0.0-beta.2

### Version of Immich Server

v1.115.0

v1.115.0 build.159

### Platform with the issue

- [ ] Server
- [X] Web
- [X] Mobile

```YAML
NA
```

### Your .env content

```Shell
NA
```

### Reproduction steps

1.Turn on "Use embedded preview" for thumbnail and preview images
2.Upload vertical Sony compressed RAW files

### Relevant log output

_No response_

### Additional information

Embedded preview:
Immich Test picture embedded preview

Embedded preview off:
Immich Test picture no embedded preview

Raw File:
https://drive.google.com/file/d/1vZvrNVNe3mjakngBEM52I6JpdLPtZtX9/view?usp=drivesdk

zhaoterryy:
## Description

Fixes #12720

Missing orientation data from RAW extracted previews because embedded jpgs are missing these EXIF tags. Without this EXIF data `sharp` was producing landscape images regardless of rotation. This tackles that issue by copying over the `Orientation` EXIF tag from the original asset.


mertalev:
Thanks for the PR! Since we already run exiftool on the original before this, try using this metadata (accessible in an asset's `exifInfo`) instead of reading the original again. And rather than writing orientation to the embedded image, you can change the interface of `decodeImage` to optionally accept a rotation value and pass this to sharp's `.rotate` method.

zhaoterryy:
By transforming the orientation value into an angle for `sharp.rotate()` we would also have to handle mirroring ourselves. Is it worth to forgo sharp's auto orientation based off EXIF to avoid the `exiftool.write()` call? I'm not sure so let me know what's best.

mertalev:
I'm planning to change the extraction to use `extractBinaryTagToBuffer` to avoid the overhead of writing, reading and then deleting a temporary file. Since there's no `.write` API for buffers (exiftool runs in a separate process), it'd be best not to rely on it here.

You can reference sharp's [orientation mapping](https://github.com/lovell/sharp/blob/1533bf995acda779313fc178d2b9d46791349961/src/pipeline.cc#L1351).

server/src/interfaces/media.interface.ts
```diff
@@ -126,7 +127,7 @@ export interface ProbeOptions {
 
 export interface IMediaRepository {
   // image
-  extract(input: string, output: string): Promise<boolean>;
+  extract(asset: AssetEntity, outputPath: string): Promise<boolean>;
```
jrasm91, line 130: Can you change this to take a second argument? 

```ts
extract(output: string, options: { orientation?: number })
```

This interface should not be using AssetEntity.

zhaoterryy:
> I'm planning to change the extraction to use `extractBinaryTagToBuffer` to avoid the overhead of writing, reading and then deleting a temporary file. Since there's no `.write` API for buffers (exiftool runs in a separate process), it'd be best not to rely on it here.
> 
> You can reference sharp's [orientation mapping](https://github.com/lovell/sharp/blob/1533bf995acda779313fc178d2b9d46791349961/src/pipeline.cc#L1351).

Thanks for the quick response! That reference is useful. I'll refactor the PR taking your comments into account

server/src/repositories/media.repository.ts
```diff
@@ -92,6 +100,23 @@ export class MediaRepository implements IMediaRepository {
     return pipeline.resize(options.size, options.size, { fit: 'outside', withoutEnlargement: true });
   }
 
+  private getRotationAngleFromOrientation(orientation: ExifOrientation | undefined) {
+    if (!orientation) return {};
+
+    const valueMap: Record<ExifOrientation, { angle?: number, flip?: boolean, flop?: boolean }> = {
```
mertalev, line 106: You can make this a constant in `constants.ts` and add `as const`. That would also let you omit the type hint here since TS will know the exact object and its properties.

server/src/services/media.service.ts
```diff
@@ -214,7 +214,8 @@ export class MediaService extends BaseService {
       const colorspace = this.isSRGB(asset) ? Colorspace.SRGB : image.colorspace;
       const processInvalidImages = process.env.IMMICH_PROCESS_INVALID_IMAGES === 'true';
 
-      const decodeOptions = { colorspace, processInvalidImages, size: image.preview.size };
+      const orientation = Number(asset.exifInfo?.orientation);
```
mertalev, line 217: Prefer making it undefined instead of NaN when `asset.exifInfo?.orientation` is undefined.

mertalev (commented):
Looks good! Just minor comments.

mertalev (approved):
Great job!

zhaoterryy:
Thank you! Need another approval due to broken test

PR #13956, opened 2024-11-06
fix: docker link

danieldietzler:
The old link now points to the docker desktop versions, which isn't what people usually want


Mraedis (approved):
Makes sense to me

bo0tzz (approved):

PR #13985, opened 2024-11-07
refactor: remove smart info table

jrasm91:
Drop `smart_info` table which is no longer being populated by any jobs.


danieldietzler (approved):

mertalev (approved):

server/src/migrations/1730989238718-DropSmartInfoTable.ts
```diff
@@ -0,0 +1,11 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class DropSmartInfoTable1730989238718 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP TABLE smart_info`);
+  }
+
+  public async down(): Promise<void> {
+    // not implemented
```
wzru, line 9: Why not implemented here? If user want to downgrade, smart_info should be created again.

PR #13995, opened 2024-11-07
fix(server): attempt to delete failed backups immediately after failure

zackpollard:


danieldietzler (approved):

server/src/services/backup.service.spec.ts
```diff
@@ -207,5 +208,11 @@ describe(BackupService.name, () => {
       const result = await sut.handleBackupDatabase();
       expect(result).toBe(JobStatus.FAILED);
     });
+    it('should ignore unlink failing and still return failed job status', async () => {
+      processMock.spawn.mockReturnValueOnce(mockSpawn(1, '', 'error'));
+      storageMock.unlink.mockRejectedValue(new Error('error'));
+      const result = await sut.handleBackupDatabase();
+      expect(result).toBe(JobStatus.FAILED);
```
danieldietzler, line 215: You should probably expect `unlink` to be called, right?

danieldietzler (approved):

jrasm91 (approved):

PR #13996, opened 2024-11-07
chore: backups custom location and config file docs

zackpollard:


bo0tzz (approved):

NicholasFlamy:
Is it supposed to be `backup` or `backups`?

zackpollard:
> Is it supposed to be `backup` or `backups`?

https://github.com/immich-app/immich/blob/main/server/src/config.ts#L18

`backup` is correct

zackpollard:
Ah, but the folder on disk is backups, you're right. I will correct this tomorrow

PR #14001, opened 2024-11-07
chore(web): Update the new version announcement text

Snowknight26:
The new version announcement text has been updated to fix some errors in the existing language, making it clearer that a new version of Immich (instead of application) is available.

I imagine this is Alex's text so keeping this PR as draft for now to ensure that the change is ok to make.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -1283,7 +1283,7 @@
   "variables": "Variables",
   "version": "Version",
   "version_announcement_closing": "Your friend, Alex",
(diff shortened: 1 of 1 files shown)


danieldietzler (approved):

bo0tzz:
Maybe we should also update this to note reading the release notes carefully instead of just mentioning updating the compose/.env.

jrasm91 (approved):
You can change this on weblate though.

Snowknight26:
> You can change this on weblate though.

Ahh thanks for the heads-up. Last time I made commits changes to English still had to be done via code. I can update it that way if that's the preferred approach.

danieldietzler:
> > You can change this on weblate though.
> 
> Ahh thanks for the heads-up. Last time I made commits changes to English still had to be done via code. I can update it that way if that's the preferred approach.

Both is fine with weblate (assuming there aren't any conflicts, which there shouldn't in this case) :)

alextran1502 (approved):

PR #14036, opened 2024-11-09
feat(web): stable json settings export

mcarbonne:
When exporting settings to json (or copying to clipboard), json output isn't stable between immich versions and keys are sometimes re-ordered.

Proposal: sort (recursively) the json output (equivalent to `jq -S`).

### DIFF
--- a/web/src/routes/admin/system-settings/+page.svelte
+++ b/web/src/routes/admin/system-settings/+page.svelte
@@ -65,15 +65,16 @@
   type SettingsComponent = ComponentType<SvelteComponent<SettingsComponentProps>>;
 
   // https://stackoverflow.com/questions/16167581/sort-object-properties-and-json-stringify/43636793#43636793
-  const jsonReplacer = (key, value) =>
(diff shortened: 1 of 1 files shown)


mertalev (approved):

mertalev:
There are some failing checks for typing and linting. You can suppress the `unicorn/no-array-reduce` rule in this case.

mcarbonne:
> There are some failing checks for typing and linting. You can suppress the `unicorn/no-array-reduce` rule in this case.

For the moment I've only tested that my modifications do work. I haven't managed yet to run linters, formatters... (I need to set up a devcontainer).

mcarbonne:
@mertalev everything is fixed !
I'll open another PR for the devcontainers setup.

jrasm91:
Might be better to have this be implemented in the server endpoint instead of clientside.

mertalev:
I think it's better for it to be client-side. It doesn't matter for normal config requests, only for exporting or copy/paste.

PR #14083, opened 2024-11-11
feat(server): use pg_dumpall version that matches the database version

zackpollard:


jrasm91 (approved):

server/src/services/backup.service.ts
```diff
@@ -101,14 +101,23 @@ export class BackupService extends BaseService {
       `immich-db-backup-${Date.now()}.sql.gz.tmp`,
     );
 
+    const databaseVersion = await this.databaseRepository.getPostgresVersion();
+    const databaseMajorVersion = Number.parseInt(databaseVersion.split('.')[0], 10);
```
danieldietzler, line 105: It's semver, right? Should we use the semver package instead then?

danieldietzler (approved):

server/src/services/backup.service.ts
zackpollard: Moved to checking a range with semver

docker/docker-compose.dev.yml
bo0tzz, line 1: Are these changes intentional?

zackpollard: Just trying to fix these damn semver changes :laughing: Accidentally committed those, will cleanup in a minute

PR #14085, opened 2024-11-11
refactor(server): filter on-event handlers by worker

jrasm91:
Add the option to only register event handlers for specific workers


danieldietzler (approved):
Very nice!

PR #14086, opened 2024-11-11
refactor(mobile): video controls

mertalev:
### Description

* Breaks up video controls into smaller widgets and marks more things as `const`. Besides improving reusability, this means that less of the widget tree is rebuilt when the position changes and more work can be done during compilation.
* Makes duration text take up fixed amount of space. Different numbers can have different widths, which means that the slider's width slightly jittered when the position changed.
* Pauses video while seeking. This doesn't matter too much for the current player as seeking gets stuck, but with the native player it avoids you "fighting" against the video's playback while seeking.

### Testing

Tested that the video controls look and feel the same but without jitter, that the widest possible duration texts don't overflow the text box, that the slider never shifts around when seeking, and that the video pauses and plays when seeking.

### DIFF
--- a/mobile/lib/widgets/asset_viewer/video_position.dart
+++ b/mobile/lib/widgets/asset_viewer/video_position.dart
@@ -10,25 +10,6 @@ import 'package:immich_mobile/widgets/asset_viewer/formatted_duration.dart';
 import 'package:immich_mobile/widgets/asset_viewer/video_mute_button.dart';
 
 class VideoPosition extends HookConsumerWidget {
-  static void _onChangedDummy(_) {}
-  static const placeholder = Row(
-    children: [
-      FormattedDuration(Duration.zero),
-      Expanded(
-        child: Slider(
-          value: 0.0,
-          min: 0,
-          max: 100,
-          thumbColor: Colors.white,
-          activeColor: Colors.white,
-          inactiveColor: whiteOpacity75,
-          onChanged: _onChangedDummy,
-        ),
-      ),
-      FormattedDuration(Duration.zero),
-      VideoMuteButton(),
-    ],
-  );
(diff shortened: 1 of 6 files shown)


mobile/lib/widgets/asset_viewer/video_position.dart
```diff
@@ -0,0 +1,77 @@
...
+import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:immich_mobile/constants/immich_colors.dart';
+import 'package:immich_mobile/providers/asset_viewer/video_player_controls_provider.dart';
+import 'package:immich_mobile/providers/asset_viewer/video_player_value_provider.dart';
+import 'package:immich_mobile/widgets/asset_viewer/formatted_duration.dart';
+import 'package:immich_mobile/widgets/asset_viewer/video_mute_button.dart';
+
+class VideoPosition extends HookConsumerWidget {
+  static void _onChangedDummy(_) {}
+  static const placeholder = Row(
```
shenlong-tanwen, line 14: Can you update this to be a private widget instead?

shenlong-tanwen (commented):
Change looks great. I'll test it once and will approve it

jrasm91 (approved):

alextran1502:
Thank you for the work on this! Tested, and the functionalities are good! One minor UI bug is from Android for the video duration here

image

alextran1502 (approved):
Mert, feel free to merge after the Android duration UI bug is fixed!

mertalev:
> Thank you for the work on this! Tested, and the functionalities are good! One minor UI bug is from Android for the video duration here
> 
> image

I can add a few pixels to the text box width, but is there a more reliable way to know what the smallest non-overflowing size is for a display? 😅

PR #14103, opened 2024-11-12
feat: support geesefs by allowing to skip utimes

### LINKED ISSUE #4445: object storage
Object storage support has been widely requested (https://github.com/immich-app/immich/discussions/1683) and is something we're keen to support. The limitations imposed by object storage happen to be beneficial for data resilience and consistency, as it makes features like the storage template infeasible. Issues like orphaned assets (#2877) or asset availability (#4442)  would be resolved completely.

As discussed on the orphaned assets issue (#2877), I'd like to propose a new storage layout designed for object storage, with scalability and resilience as priorities.

Where:
- `<asset id>` is a unique ID for an asset, ideally a random UUID. [UUIDv7](https://www.ietf.org/archive/id/draft-peabody-dispatch-new-uuid-format-01.html#name-uuidv7-layout-and-bit-order) may serve to be beneficial due to its property of natural order. If not, UUIDv4 should be sufficient.
- `<original asset filename>` is the original filename of an asset, as it was uploaded.

```sh
.
└── <asset id>/
    ├── <original asset filename>
    ├── sidecar.xml
    └── thumbnails/
        ├── small.jpg
        └── large.webp
```

The above structure should serve to efficiently scale with resiliency and flexibility. The unique 'directory' for an asset can contain additional files like edits, colour profiles, thumbnails or anything else.

The original file and filename is preserved in case of an unlikely full database loss, where it should be possible to restore most information in such a scenario. This property is also good for humans, or if a full export is required. A directory of vague filenames without extensions would be quite unhelpful. I feel this strikes a good balance between legibility and a resiliency.

I have also considered content addressable storage (CAD), as it would save space in the event of duplicate uploads but consider it to be impractical due to complexity and the previous concern of legibility. I believe this should instead be deferred to the underlying storage provider, which can make much better decisions about how to store opaque binary blobs.

Part of this effort will require some changes to the storage interfaces (#1011) and the actual object storage implementation should use the [AWS S3 SDK](https://www.npmjs.com/package/@aws-sdk/client-s3) ([docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/)). Most, if not all object storage systems should have an S3-compatible API.

### LINKED ISSUE #14005: Upload fails when `utime` is not available
### The bug

It looks like Immich tries to set the `utime` of the uploaded files, but certain storage implementations such as [geesefs](https://github.com/yandex-cloud/geesefs) do not allow this.  
  
To properly fix the issue, I think it would be convenient to allow users to disable / mock `utime` when a specific environment variable is specified.

https://github.com/immich-app/immich/blob/75340985969b371d73279622343a84adf856a92b/server/src/repositories/storage.repository.ts#L62-L64

```
Error: EPERM: operation not permitted, utime 'upload/upload/e9209f4c-7ad4-437d-a
f25-1c059623c8e1/24/cd/24cd5cc9-a7db-4d4d-86d5-d7d4675f69df.jpeg'
    at async Object.utimes (node:internal/fs/promises:1128:10)
    at async AssetMediaService.create (/usr/src/app/dist/services/asset-media.se
rvice.js:273:9)
    at async AssetMediaService.uploadAsset (/usr/src/app/dist/services/asset-med
ia.service.js:96:27)
    at async AssetMediaController.uploadAsset (/usr/src/app/dist/controllers/ass
et-media.controller.js:40:29)
```

Alpine

### Version of Immich Server

v1.120.1

N/A

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

1. Use [geesefs](https://github.com/yandex-cloud/geesefs) (or any non-POSIX compliant FS) and mount an S3 bucket to the uploads directory
2. Try to upload a file

### Relevant log output

```shell
Error: EPERM: operation not permitted, utime 'upload/upload/e9209f4c-7ad4-437d-a
    at async Object.utimes (node:internal/fs/promises:1128:10)
    at async AssetMediaService.create (/usr/src/app/dist/services/asset-media.se
    at async AssetMediaService.uploadAsset (/usr/src/app/dist/services/asset-med
    at async AssetMediaController.uploadAsset (/usr/src/app/dist/controllers/ass
```

### Additional information

_No response_

denysvitali:
This fixes #14005 by adding an environment variable that allows to skip the `utimes` call on filesystems that do not support it.  
  
Thanks to this fix, it's possible to use an S3 storage in Immich via the [geesefs](https://github.com/yandex-cloud/geesefs) FUSE filesystem.  
  
This could indirectly fix #4445 too (or at least provide a workaround).

### DIFF
--- a/docs/docs/install/environment-variables.md
+++ b/docs/docs/install/environment-variables.md
@@ -43,6 +43,7 @@ These environment variables are used by the `docker-compose.yml` file and do **N
 | `IMMICH_PROCESS_INVALID_IMAGES`     | When `true`, generate thumbnails for invalid images                                       |                              | server                   | microservices      |
 | `IMMICH_TRUSTED_PROXIES`            | List of comma separated IPs set as trusted proxies                                        |                              | server                   | api                |
 | `IMMICH_IGNORE_MOUNT_CHECK_ERRORS`  | See [System Integrity](/docs/administration/system-integrity)                             |                              | server                   | api, microservices |
+| `IMMICH_SKIP_UTIMES`                | Does not call `utime` on files after processing. This can be useful for certain filesystems such as [geesefs](https://github.com/yandex-cloud/geesefs) that do not support it | `false` | server |  |
 
 \*1: `TZ` should be set to a `TZ identifier` from [this list][tz-list]. For example, `TZ="Etc/UTC"`.
 `TZ` is used by `exiftool` as a fallback in case the timezone cannot be determined from the image metadata. It is also used for logfile timestamps and cron job execution.
--- a/docs/docs/install/environment-variables.md
+++ b/docs/docs/install/environment-variables.md
@@ -43,7 +43,7 @@ These environment variables are used by the `docker-compose.yml` file and do **N
 | `IMMICH_PROCESS_INVALID_IMAGES`     | When `true`, generate thumbnails for invalid images                                       |                              | server                   | microservices      |
 | `IMMICH_TRUSTED_PROXIES`            | List of comma separated IPs set as trusted proxies                                        |                              | server                   | api                |
 | `IMMICH_IGNORE_MOUNT_CHECK_ERRORS`  | See [System Integrity](/docs/administration/system-integrity)                             |                              | server                   | api, microservices |
-| `IMMICH_SKIP_UTIMES`                | Does not call `utime` on files after processing. This can be useful for certain filesystems such as [geesefs](https://github.com/yandex-cloud/geesefs) that do not support it | `false` | server |  |
+| `IMMICH_SKIP_UTIMES`                | Does not call [`utimes`](https://linux.die.net/man/2/utimes) on files after processing. This can be useful for certain filesystems such as [geesefs](https://github.com/yandex-cloud/geesefs) that do not support it | `false` | server |  |
 
 \*1: `TZ` should be set to a `TZ identifier` from [this list][tz-list]. For example, `TZ="Etc/UTC"`.
 `TZ` is used by `exiftool` as a fallback in case the timezone cannot be determined from the image metadata. It is also used for logfile timestamps and cron job execution.
--- a/server/src/dtos/env.dto.ts
+++ b/server/src/dtos/env.dto.ts
@@ -66,6 +66,9 @@ export class EnvDto {
   @Optional()
   IMMICH_REPOSITORY_URL?: string;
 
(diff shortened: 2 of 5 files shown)


jrasm91 (commented):
I think I'd rather see utime errors be ignored

zackpollard:
Personally, and as was mentioned in the issue when it was opened, we don't want to support non-POSIX filesystems. This is a rabbit hole that we don't really want to get into, I can see it ending with us adding these kind of patches for a bunch of things filesystems should be providing.

denysvitali:
> Personally, and as was mentioned in the issue when it was opened, we don't want to support non-POSIX filesystems. This is a rabbit hole that we don't really want to get into, I can see it ending with us adding these kind of patches for a bunch of things filesystems should be providing.

Supporting _some_ FUSE filesystems (especially S3) might be a quick win. Personally I think this one line change won't hurt and might help more users in the same situation.

zackpollard:
> > Personally, and as was mentioned in the issue when it was opened, we don't want to support non-POSIX filesystems. This is a rabbit hole that we don't really want to get into, I can see it ending with us adding these kind of patches for a bunch of things filesystems should be providing.
> 
> Supporting _some_ FUSE filesystems (especially S3) might be a quick win. Personally I think this one line change won't hurt and might help more users in the same situation.

We have plenty of people using S3 without needing to make this concession, I am not familiar with those setups specifically, however I imagine their mount tool just ignores utime updates instead of throwing errors.

denysvitali:
FYI - I fixed this on my side by patching the `fs` library:

```js
    const fs = require('fs');
    const util = require('util');

    // Store the original method
    const originalUtimes = fs.utimes;
    const originalUtimesSync = fs.utimesSync;

    // Patch the async version
    fs.utimes = function(path, atime, mtime, callback) {
    // Just call callback successfully without doing the operation
    if (callback) process.nextTick(callback);
    return Promise.resolve();
    };

      // Patch the sync version
      fs.utimesSync = function(path, atime, mtime) {
      // Do nothing and return
      return;
    };

    // Patch the promises version if it exists
    if (fs.promises) {
      fs.promises.utimes = util.promisify(fs.utimes);
    }
```

Then `NODE_OPTIONS` can be set to `--require /patch/fs-patch.js`.

This way, we don't make modifications to immich's source code but still get geesefs support.

PR #14126, opened 2024-11-13
feat(mobile): new video slider ui

mertalev:
### Description

This PR introduces a cleaner look for the video controls:
* Expanded slider
* Soft gradient background
* Symmetric duration and position
* Removed mute button
* Thicker timestamp font with full opacity for readability

The core idea is to make the slider as wide as possible by moving other components above it. This makes the controls take more vertical space, but the mostly transparent background makes it look light and unobtrusive despite this.

## Testing

Tested on iOS with a physical 15 Pro on portrait and landscape.

#### Before

IMG_3336

#### After

IMG_3334

#### Before

IMG_3337

#### After

IMG_3333


alextran1502 (approved):
Nice work!

PR #14151, opened 2024-11-14
refactor: rename searchMetadata to searchAssets

jrasm91:
The operation name `searchMetadata` has been renamed to `searchAssets`. The URL, params, and all other parts of the endpoint signature remain unchanged.


alextran1502:
This will be a breaking change for projects that use openapi, right?

danieldietzler (approved):

danieldietzler:
> This will be a breaking change for projects that use openapi, right?

No, because the endpoint is the same. Names are just meaningless

alextran1502:
@danieldietzler 
image

OpenAPI generates the method name based on the controller's method name

Issue #11216, opened 2024-07-19
Serbian Cyrillic wrong font


medolino2009:
### The bug

Hi guys, i am translator of Serbian latin and Cyrillic language, Immich renders whole Cyrillic with "wrong" font...very hard to read when you have it in Serbian Cyrillic... 
I tried to change language to Russian and that language renders with readable font for example... I searched google fonts and "Alumni Sans" font renders nice readble Serbian Cyrillic. I hope you can change this. 

Mobile app don't support Serbian Cyrillic yet (i can choose Serbian Cyrillic but language changes to English), if this also could be fixed, would be splendid

btw. Thanx a lot for amazing software, i've already got license for it. Keep up great job.!!!

Thanx in advance.

Best regards

Miki

### The OS that Immich Server is running on

Unraid docker server

### Version of Immich Server

v.1.109.2

### Version of Immich Mobile App

v.1.109.0

### Platform with the issue

- [ ] Server
- [x] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
It is translation trouble
```

### Your .env content

```Shell
It is translation trouble
```

### Reproduction steps

```bash
1.Change to Serbian Cyrillic in web
2.Language changes to Serbian Cyrillic
3.Font is dificult to read
...
```

jrasm91:
Are you saying that the current font does not look/work very well in that language? Can you show screenshots of it and pictures of it with the new font?

medolino2009:
> Are you saying that the current font does not look/work very well in that language? Can you show screenshots of it and pictures of it with the new font?

Yes, that is exactly what i was saying

This is wrong font (dificult to read because of some letters):
![Immich cyrillic wrong font](https://github.com/user-attachments/assets/69f2b66a-87ff-431f-87f6-1d20ad4a778c)

This is font that works wel in Immich:
![Immich Alumni Sans font](https://github.com/user-attachments/assets/e382e1fd-a27d-411f-bb64-245bdd7714d3)

i think Alumni Sans Regular 400 looks best.

Thanx in advance.

Best regards

Miki

VelocityRa:
@medolino2009 Does Serbian Latin render incorrectly too?

meesfrensel:
Immich recently changed to a different font. Does that work better for cyrililc?

VelocityRa:
> Immich recently changed to a different font. Does that work better for cyrililc?

Yes! At least on my end on one of the devices where I had the issue (Samsung S23), Greek now works well!

[...]

PR #14158, opened 2024-11-15
feat(mobile): exclude locales from overpass font

johnstef99:
Fixes #10436 

As mentioned [here](https://github.com/immich-app/immich/issues/10436#issuecomment-2429775832) this PR adds a list of languages that should be excluded from using Overpass as a font. These languages will use the default Flutter font based on the platform.

As of right now the only known to me language having a problem is Greek, feel free to exclude more if needed.

### DIFF
--- a/mobile/lib/main.dart
+++ b/mobile/lib/main.dart
@@ -10,6 +10,7 @@ import 'package:flutter/services.dart';
 import 'package:flutter_displaymode/flutter_displaymode.dart';
 import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:immich_mobile/extensions/build_context_extensions.dart';
+import 'package:immich_mobile/providers/locale_provider.dart';
 import 'package:immich_mobile/utils/download.dart';
 import 'package:intl/date_symbol_data_local.dart';
 import 'package:timezone/data/latest.dart';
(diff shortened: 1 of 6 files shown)


mobile/lib/main.dart
```diff
@@ -212,20 +213,31 @@ class ImmichAppState extends ConsumerState<ImmichApp>
...
-        debugShowCheckedModeBanner: false,
-        themeMode: ref.watch(immichThemeModeProvider),
-        darkTheme: getThemeData(colorScheme: immichTheme.dark),
-        theme: getThemeData(colorScheme: immichTheme.light),
-        routeInformationParser: router.defaultRouteParser(),
-        routerDelegate: router.delegate(
-          navigatorObservers: () => [TabNavigationObserver(ref: ref)],
+    return ProviderScope(
+      overrides: [
+        localeProvider.overrideWithValue(context.locale),
```
alextran1502, line 218: What is the purpose of this `overrides` property?

johnstef99: It sets the `localeProvider` value for the rest subtree of widgets every time that `context.locale` gets updated. Do you have an alternative way of settings the initial value of `localeProvider` and then updating it?

VelocityRa:
> As of right now the only known to me language having a problem is Greek, feel free to exclude more if needed.

Serbian Cyrillic has the issue too: https://github.com/immich-app/immich/issues/11216

johnstef99:
> > As of right now the only known to me language having a problem is Greek, feel free to exclude more if needed.
> 
> Serbian Cyrillic has the issue too: #11216, could be added now. Might also want to tweak the filtering to take the country code into account too - Serbian Cyrillic is specifically `Locale('sr', 'Cyrl')`, there's Serbian Latin too which presumably renders fine (I asked on the issue to make sure).

Ok I will update it to filter with region/country subtag and exclude `Locale('sr', 'Cyrl')`

alextran1502:
The `localProvider` isn't used correctly here, as the value never updates. The theme is already initialized when the locale changes, so you don't need the provider.

You can achieve the same effect by removing the `localProvider` across the app then changing the `getThemeData` method signature to the following

image

johnstef99:
Have you tested that the localeProvider's value never changes? Because I can't reproduce it. When the EasyLocalization parent widget rebuilds the localeProvider's value gets updated. 

Let me explain why I added this provider:

1. In the `MapThemeOveride` widget we also need the locale to get the theme data. If `conext.locale` was used here, it would be difficult for the widget to be tested (and there are already written tests for this widget). The already written test will have to be refactored to have as parent widget the EasyLocalization.
2. I have opened [an other MR](https://github.com/immich-app/immich/pull/14089#issue-2650586726) that also needs the `localeProvider` for the reason that some UI elements are cached by providers. That results on not refreshing the UI after the user changing language.

alextran1502 (approved):

PR #14170, opened 2024-11-15
fix(mobile): unnecessary rebuilds from partner share notifier

mertalev:
### Description

I noticed this notifier was triggering a rebuild of the app bar because `[]` is not equal to `[]` (different objects). This PR changes this to use a deep comparison instead. Also makes the watching happen only after the initial query finishes to make sure there's no race condition.

### DIFF
--- a/mobile/lib/providers/partner.provider.dart
+++ b/mobile/lib/providers/partner.provider.dart
@@ -10,13 +10,13 @@ import 'package:isar/isar.dart';
 
 class PartnerSharedWithNotifier extends StateNotifier<List<User>> {
   PartnerSharedWithNotifier(Isar db, this._ps) : super([]) {
+    Function eq = const ListEquality<User>().equals;
     final query = db.users.filter().isPartnerSharedWithEqualTo(true).sortById();
     query.findAll().then((partners) {
-      if (partners.isNotEmpty) {
+      if (!eq(state, partners)) {
         state = partners;
       }
     }).then((_) {
-      Function eq = const ListEquality<User>().equals;
       query.watch().listen((partners) {
         if (!eq(state, partners)) {
           state = partners;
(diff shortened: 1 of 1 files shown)


jrasm91 (approved):

mobile/lib/providers/partner.provider.dart
```diff
@@ -9,9 +10,19 @@ import 'package:isar/isar.dart';
 
 class PartnerSharedWithNotifier extends StateNotifier<List<User>> {
   PartnerSharedWithNotifier(Isar db, this._ps) : super([]) {
-    final query = db.users.filter().isPartnerSharedWithEqualTo(true);
-    query.findAll().then((partners) => state = partners);
-    query.watch().listen((partners) => state = partners);
+    final query = db.users.filter().isPartnerSharedWithEqualTo(true).sortById();
+    query.findAll().then((partners) {
+      if (partners.isNotEmpty) {
```
fyfrey, line 15: I think this should use the same eq function as below.

mobile/lib/providers/partner.provider.dart
```diff
@@ -31,9 +42,19 @@ final partnerSharedWithProvider =
 
 class PartnerSharedByNotifier extends StateNotifier<List<User>> {
   PartnerSharedByNotifier(Isar db) : super([]) {
-    final query = db.users.filter().isPartnerSharedByEqualTo(true);
-    query.findAll().then((partners) => state = partners);
-    streamSub = query.watch().listen((partners) => state = partners);
+    final query = db.users.filter().isPartnerSharedByEqualTo(true).sortById();
+    query.findAll().then((partners) {
+      if (partners.isNotEmpty) {
```
fyfrey, line 47: See comment above

fyfrey (approved):

PR #14171, opened 2024-11-15
refactor(mobile): backup info box

mertalev:
## Description

The info box widget is very monolithic. As a result, the whole thing is constantly re-rendered for each progress update. This PR breaks the widget up into smaller widgets that only re-render when data that's relevant to them changes. It also removes the hidden thumbnail feature as it increases CPU usage and heat, and I don't think it's used much.

The providers are kind of weird and would probably benefit from a refactor as well, but I leave this for future work.

## How Has This Been Tested?

Tested that backup progress and info looks normal when initiating a backup.


mertalev:
The isar dev [uses](https://github.com/dart-lang/sdk/issues/40522) this a lot for short functions or functions that are only called once, and has run microbenchmarks showing improvements. There's no downside to using it when the function is only called once or if it's a short function. But it should be avoided for longer functions that are used in many places since it increases code size.

alextran1502 (approved):

PR #14199, opened 2024-11-18
fix: show tags when viewing stacked assets

### LINKED ISSUE #14179: problem with tags when stacking photos
### The bug

when stacking multiple pictures,  tags are shown only for the first picture, other pictures tags are not shown but come back when unstacked

demo

### Version of Immich Server

demo

demo

### Platform with the issue

- [ ] Server
- [ ] Web
- [ ] Mobile

```YAML
demo
```

### Your .env content

```Shell
demo
```

### Reproduction steps

1. stack multiple pictures 
2.open stack view
3.first pictures has its tags but the rest don't 
4. unstack pictures 
5. 5. all tags are back and shown normally 
...

### Relevant log output

_No response_

### Additional information

got this in the demo

michelheusschen:
When viewing a stacked asset, the tags are shown initially. However, if you switch to a different asset or go back to the first one, the tags are not visible. This happens because the `/api/stack/:id` endpoint stopped returning tags after a recent change. 

This PR fixes #14179 by including tags in the endpoint again and adds some e2e tests.

### DIFF
--- a/e2e/src/utils.ts
+++ b/e2e/src/utils.ts
@@ -11,6 +11,7 @@ import {
   PersonCreateDto,
   SharedLinkCreateDto,
   UserAdminCreateDto,
+  UserPreferencesUpdateDto,
   ValidateLibraryDto,
   checkExistingAssets,
   createAlbum,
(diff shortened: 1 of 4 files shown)


danieldietzler (approved):
Nice one!

jrasm91 (approved):

Issue #14211, opened 2024-11-18
Web search filters (not in any album, archive, favorites) are inconsistent


weathondev:
### The bug

If you dont have 'not in album', 'favorites' and 'archive' checked, it will show you pictures that are (in albums or not) (in favorites or not) (not archived) - this is inconsistent. It might be the desired behavior but then we should rename them to
'Favorites only' and 'include archive'. 
@mertalev suggested to do three-way toggles (all/matching/not matching) which is probably the best solution and we just need to think of the proper defaults.

### The OS that Immich Server is running on

Not relevant

### Version of Immich Server

v1.120.2

### Version of Immich Mobile App

-

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

1.
2.
3.
...

weathondev:
I can tackle this one after my current task but wanted to document it already as found it while on my current one.

andreiverse:
I would really like to search by 'not in archived'

alextran1502:
@andreiverse "Not in archived" is the default

jb-net441:
The behavior is different in the IOS app and Android app than the Web. The web app starting in 3.0 will show Archived photos which are not in an album. The apps will not.

meesfrensel:
With the new search UI, I think this is as the team intends, because these filters have not been changed. Therefore I'll close the issue, discussion about this might be better for the discord.

Issue #14209, opened 2024-11-18
Search view does not update when assets change (and don't match the filter anymore)


weathondev:
### The bug

Basically the same as https://github.com/immich-app/immich/issues/13460

The view does not update when e.g. i archive an item like it does when i add an item to an album.
It does get the 'archive' icon, but when i refresh the search, it's not gone - ideally it should vanish, as it doesn't match the search filter anymore.
Same for date, location, favorites, etc.

### The OS that Immich Server is running on

Not relevant

### Version of Immich Server

v1.120.2

### Version of Immich Mobile App

-

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

1. Search for something with default filters
2. Archive an asset
3. Gets an archive icon but doesn't vanish

Repeat the same with add to album while filtering 'not in album' -> it will vanish

matteonu:
I would like to work on this as a first issue. Can you assign it to me please?

Issue #13460, opened 2024-10-15
[issue] Add to album does not update filter


MickLesk:
### The bug

Hi Guys,

In a nutshell, it's about a missing update flag for active filters in the photo overview. If a filter no longer applies, it is still displayed.
The following steps are defined below (Reproduction steps)

In my case, this occurs when adding to an album, but it is also possible that it is exactly the same problem with other filters.

### The OS that Immich Server is running on

Debian 12 LXC (Docker)

### Version of Immich Server

v1.118.0

### Version of Immich Mobile App

v1.118.0

### Platform with the issue

- [X] Server
- [X] Web
- [X] Mobile

### Your docker-compose.yml content

```YAML
version: "3.8"

name: immich

services:
  immich-server:
    container_name: immich_server
    image: ghcr.io/immich-app/immich-server:${IMMICH_VERSION:-release}
    volumes:
      - ${UPLOAD_LOCATION}:/usr/src/app/upload
      - /etc/localtime:/etc/localtime:ro
    env_file:
      - stack.env
    ports:
      - 2283:2283
    depends_on:
      - redis
      - database
    restart: always

  immich-machine-learning:
    container_name: immich_machine_learning
    image: ghcr.io/immich-app/immich-machine-learning:${IMMICH_VERSION:-release}
    volumes:
      - model-cache:/cache
    env_file:
      - stack.env
    restart: always

  redis:
    container_name: immich_redis
    image: docker.io/redis:6.2-alpine@sha256:d6c2911ac51b289db208767581a5d154544f2b2fe4914ea5056443f62dc6e900
    healthcheck:
      test: redis-cli ping || exit 1
    restart: always

  database:
    container_name: immich_postgres
    image: docker.io/tensorchord/pgvecto-rs:pg14-v0.2.0@sha256:90724186f0a3517cf6914295b5ab410db9ce23190a2d9d0b9dd6463e3fa298f0
    environment:
      POSTGRES_PASSWORD: ${DB_PASSWORD}
      POSTGRES_USER: ${DB_USERNAME}
      POSTGRES_DB: ${DB_DATABASE_NAME}
      POSTGRES_INITDB_ARGS: '--data-checksums'
    volumes:
      - pgdata:/var/lib/postgresql/data
    healthcheck:
[...]

PR #14213, opened 2024-11-18
feat: Added shortcuts, shift-multi select and missing menu options to Search (Galleryviewer)

weathondev:
The Galleryviewer component (used in the Search, Individual Share page & Memories page) was lacking the ability to to
- shift multiselect
- shortcuts like ctrl A, ctrl D, delete, force delete, archive

Fixed a small issue in the search related to https://github.com/immich-app/immich/issues/13460 - It was always removing the assets added to an album, even when 'not in any album' was not on.

Also created two more issues i found while working on this, which i will tackle next:
- https://github.com/immich-app/immich/issues/14211
- https://github.com/immich-app/immich/issues/14209


alextran1502 (approved):
Nice work! Thank you so much. I wonder if we can refactor the duplicate logic/action in the gallery-view and asset-grid view in the future.

PR #14277, opened 2024-11-21
feat(server): faster metadata extraction

mertalev:
## Description

* Groups async calls where possible rather than doing them sequentially
* Moves some checks to be done synchronously without `await`
* Extracts live photo logic into a private method that metadata extraction runs
* No longer queues live photo job as part of upload pipeline
* Removes redundant `stat` calls, instead relying on exiftool
* Use async hash function to avoid blocking the event loop

Besides being significantly faster, it also makes job progress more intuitive as the number of waiting jobs goes down as one would expect without other jobs being queued in their place.

### Testing

metadata-extraction-rate

The y-axis is jobs/s. The first part is main and the second is this PR. The jump toward the end is when the queueing job finished queueing assets (the job is somewhat intensive with a large table). There is an increase on main as well if you look closely, but it's more subtle. I canceled both when their progress rate stabilized after the queueing job finished.

### DIFF
--- a/server/src/repositories/metadata.repository.ts
+++ b/server/src/repositories/metadata.repository.ts
@@ -85,6 +85,10 @@ export class MetadataRepository {
     this.logger.setContext(MetadataRepository.name);
   }
 
+  setMaxConcurrency(concurrency: number) {
+    this.exiftool.batchCluster.setMaxProcs(concurrency);
+  }
+
   async teardown() {
     await this.exiftool.end();
   }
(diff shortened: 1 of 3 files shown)


jrasm91 (changes requested):
Two parts of a live photo can be uploaded at the same time and won't be linked if their metadata extraction jobs are processed in parallel. That was why it was moved to a separate job in the first place.

jrasm91:
It might be sufficient to save the CID value to the database prior to running the link logic post of the job though.

jrasm91 (commented):
This looks pretty good but it has a ton of improvements that would have been better as separate pull requests. Can you separate out all these things? 
- exif tool concurrency control
- exif tool stat call
- live photos job change
- metadata async calls in parallel

mertalev:
Closing this PR in favor of splitting it into smaller, more focused PRs.

PR #14324, opened 2024-11-24
Update .env references to stack.env for Portainer compatibility

mordonez:
This PR updates the documentation to clarify that .env files must be renamed to stack.env for compatibility with Portainer. 

The changes include:
	•	Replacing .env references with stack.env in the instructions.


danieldietzler (changes requested):
Step 5 already describes that

mordonez:
You re right, thanks.

PR #14335, opened 2024-11-25
feat(mobile): add server cert field, similar to client cert

crisoagf:
Initial code for server certificate field.


zackpollard:
Hey, could you please provide more information on what this is supposed to achieve?

crisoagf:
So far I tested with local build + `flutter run`. Try to use without given code doesn't work, with given code works.

I'll try to add some info in some file in the repo and I'd really like to write a couple of tests. I may take a while (haven't been having a lot of free time lately), but I'll do it.

shenlong-tanwen:
@crisoagf Using the [cronet_http](https://pub.dev/packages/cronet_http) package for the HTTPClient will make flutter respect the CAs installed system-wide. That'd be the proper way to fix this. Can you please open a new PR with the cronet implementation instead? Thanks a lot for the contribution.

PR #14352, opened 2024-11-26
chore(deployment): cleanup database container args, move to using internal container ENV vars

mmomjian:
This PR makes a few formatting changes to improve the super long healthcheck line and simplify the postgres init command.

It also switches to using the in-container POSTGRES_USER and POSTGRES_DB vars, instead of assuming that the default DB_ immich vars are used.

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -129,7 +129,8 @@ services:
         pg_isready --dbname="$${POSTGRES_DB}" --username="$${POSTGRES_USER}" || exit 1;
         Chksum="$$(psql --dbname="$${POSTGRES_DB}" --username="$${POSTGRES_USER}" --tuples-only --no-align
         --command='SELECT COALESCE(SUM(checksum_failures), 0) FROM pg_stat_database')";
-        echo "checksum failure count is $$Chksum"; [ "$$Chksum" = '0' ] || exit 1
+        echo "checksum failure count is $$Chksum"
+        [ "$$Chksum" = '0' ] || exit 1
       interval: 5m
       start_interval: 30s
       start_period: 5m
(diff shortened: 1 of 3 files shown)


jrasm91 (approved):

mertalev (commented):
omg the command is so much cleaner. But I honestly prefer the healthcheck being on one line. It's ugly, yes, but it can be ugly off-screen instead of on-screen 😛

mmomjian:
I mean that’s fair. In that case, there’s no real reason we have to split up the command line either? I don’t have a strong opinion but I would think they should be consistent.

alextran1502 (approved):

akostadinov:
Running by creating systemd services with [Quadlet](https://docs.podman.io/en/stable/markdown/podman-systemd.unit.5.html). For this I already had to use internal variables. Just FYI. Thanks.

Gatsby-Lee:
I am upgrading to 1.122.3 and happened to notice that this PR switched to use "in-container POSTGRES_USER and POSTGRES_DB".

Are there any reasons to prefer to use the in-container env over  the default DB_ immich vars?
I am just curious.

mmomjian:
> I am upgrading to 1.122.3 and happened to notice that this PR switched to use "in-container POSTGRES_USER and POSTGRES_DB".
> 
> Are there any reasons to prefer to use the in-container env over the default DB_ immich vars? I am just curious.

Yes. Using the in container var is guaranteed to match with what the Postgres database is actually using, as these variables can be defined in numerous ways on the compose side.

Gatsby-Lee:
> > I am upgrading to 1.122.3 and happened to notice that this PR switched to use "in-container POSTGRES_USER and POSTGRES_DB".
> > Are there any reasons to prefer to use the in-container env over the default DB_ immich vars? I am just curious.
> 
> Yes. Using the in container var is guaranteed to match with what the Postgres database is actually using, as these variables can be defined in numerous ways on the compose side.
> 
> It also helps us prepare for future refining of the compose deployment.

well. what you're saying might be right as well.
However, in the given context and how the variables are defined, having left them as they were might be more intuitive unless immich either switch to use different DB or support more than just Postgres ( maybe it already does =) ) 

and, like @mertalev mentioned said, I personally prefer the one-line "healthcheck test" cmd since it is easier to grep and use for debugging.

anyway, since your change doesn't force things, I will go with my own preferences.

thank you

PR #14374, opened 2024-11-27
fix(web): resolve issues with user usage statistics and refactor

michelheusschen:
- apply `toLocaleString()` to the values instead of the table header
- use `$t('view_name')` for the noun form, as it differs from the verb in many languages
- show horizontal scroll bar on mobile if content doesn't fit
- change font size for video count to match the previous cells
- refactor row snippet and adjust some classes

image


danieldietzler (approved):
Awesome, thanks!

PR #14446, opened 2024-12-02
feat: original-sized previews for non-web-friendly images

eligao:
# Implementation details:
- Server: extract and keep the original-sized preview image for non-web-friendly images
  - Definition of web-friendly images:
    - From MDN: https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types
    - In the code: https://github.com/eligao/immich/blob/feat/raw-fullsize-preview/server/src/utils/mime-types.ts#L62
  - A new config for this feature: `image.fullsize`
    - `image.fullsize.enabled`: toggles original-sized previews 
      - When `image.extractEmbedded` is on, it will attempt to extract and use the JPG preview from RAW images instead of converting from RAW.
  - The result will be saved as a separate `AssetFileType.FULLSIZE` file.
 
- OpenAPI: a new `AssetMediaSize.FULLSIZE` size is available under the `viewThumbnail` API
  - It returns the `AssetFileType.FULLSIZE` file when available
  - Or redirects to the original file when it's web-friendly
  - Or otherwise redirects back to `AssetFileSize.PREVIEW`
    - Happens when an image is not web-friendly and a `FULLSIZE` preview is not yet generated.
  - HTTP 302 is used here to make full use of caching
 
- Web: photo-viewer
  - When zoomed in, the browser loads the `AssetMediaSize.FULLSIZE` preview via the `viewThumbnail` API
  - This applies to ALL images, instead of previously only loading the original file for web-friendly images.
  - The server will determine what to return or redirect, simplifying the logic on the client, and the client never knows if a fullsize preview is available anyway.
  - This file can be right-clicked and downloaded/copied for sharing.

# Some future ideas:
- Load the `FULLSIZE` image inthe mobile client. Unfortunately I'm not familiar with flutter yet.
- Download the RAW as JPEG with full EXIF preserved, for easier sharing
- Trim the RAW originals and keep the JPEG only, freeing up storage space


server/src/services/media.service.ts
```diff
@@ -204,33 +224,50 @@ export class MediaService extends BaseService {
...
+        // convert a full-sized thumbnail from original instead
+        extractedPath = StorageCore.getImagePath(asset, AssetPathType.EXTRACTED, image.preview.format);
+        // const orientation = asset.exifInfo?.orientation ? Number(asset.exifInfo.orientation) : undefined;
+        await this.mediaRepository.generateThumbnail(
+          asset.originalPath,
+          { ...image.preview, colorspace, processInvalidImages, size: Infinity },
+          extractedPath,
+        );
       }
+      fullsizePath = extractedPath;
```
mertalev, line 256: Since this is set either way, it seems you can simply the if/else above it to a single `!useExtracted` check

server/src/services/media.service.ts
```diff
@@ -204,33 +224,50 @@ export class MediaService extends BaseService {
...
-      return { previewPath, thumbnailPath, thumbhash: outputs[2] };
-    } finally {
-      if (didExtract) {
-        await this.storageRepository.unlink(extractedPath);
+      if (useExtracted) {
+        fullsizePath = extractedPath;
+      } else {
+        // did not extract or extracted preview is smaller than target size,
+        // convert a full-sized thumbnail from original instead
+        extractedPath = StorageCore.getImagePath(asset, AssetPathType.EXTRACTED, image.preview.format);
```
mertalev, line 248: It's a little weird for an extracted image to not be extracted. Maybe a different name would be more appropriate?

server/src/services/media.service.ts
```diff
@@ -204,33 +224,50 @@ export class MediaService extends BaseService {
...
+        // const orientation = asset.exifInfo?.orientation ? Number(asset.exifInfo.orientation) : undefined;
```
mertalev, line 249: Remove commented code

server/src/services/media.service.ts
```diff
@@ -204,33 +224,50 @@ export class MediaService extends BaseService {
...
+        await this.mediaRepository.generateThumbnail(
+          asset.originalPath,
+          { ...image.preview, colorspace, processInvalidImages, size: Infinity },
+          extractedPath,
+        );
```
mertalev, line 254: This should be a concurrent promise along with the others and use the decoded image buffer. The image buffer can be original size in this situation (and only in this situation).

mertalev, line 254: I'm not sure about this section in general. I think your idea is to allow a deeper zoom for RAW images even when the extracted image shouldn't be used, but this issue isn't unique to RAW images and affects many other formats. Solving this seems out of scope for the PR.

mertalev (commented):
Thank you for the PR! I mostly looked at the server code, where I think it needs some more polishing.

Re: caching, this also affects the case where the user zooms in and no extracted image exists, right? It will load the preview again.

server/src/services/media.service.ts
```diff
@@ -185,8 +195,18 @@ export class MediaService extends BaseService {
...
+      this.logger.debug(`Deleting old extracted image for asset ${asset.id}`);
+      pathsToDelete.push(extractedFile.path);
+    }
+
     if (pathsToDelete.length > 0) {
-      await Promise.all(pathsToDelete.map((path) => this.storageRepository.unlink(path)));
+      await Promise.all(
+        pathsToDelete.map(async (path) => {
+          await this.storageRepository.unlink(path);
+          await this.assetRepository.removeAssetFile(path);
```
mertalev, line 207: This should only be called for the extracted path. It's unnecessary for the others. The signature should also be changed to be based on the asset ID and file type instead of the path.

mertalev: If the file doesn't exist then there shouldn't be an entry in the DB for it imo

server/src/repositories/media.repository.ts
```diff
@@ -44,6 +44,11 @@ export class MediaRepository implements IMediaRepository {
 
   async extract(input: string, output: string): Promise<boolean> {
     try {
+      // remove existing output file if it exists
+      // as exiftool-vendord does not support overwriting via "-w!" flag
```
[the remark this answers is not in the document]
mertalev: Isn't this done with the `-overwrite_original` flag?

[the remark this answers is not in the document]
mertalev: Oh I see, it still throws an error if the output file exists without the `!`. This is okay for now - I'll make an issue/PR to address this in exiftool-vendored.

server/src/services/media.service.ts
```diff
@@ -245,29 +247,31 @@ export class MediaService extends BaseService {
       } else {
         // did not extract or extracted preview is smaller than target size,
         // convert a full-sized thumbnail from original instead
-        extractedPath = StorageCore.getImagePath(asset, AssetPathType.EXTRACTED, image.preview.format);
-        // const orientation = asset.exifInfo?.orientation ? Number(asset.exifInfo.orientation) : undefined;
-        await this.mediaRepository.generateThumbnail(
+        extractedPath = StorageCore.getImagePath(asset, AssetPathType.CONVERTED, image.preview.format);
+        // unset size to get full size
+        decodeOptions.size = undefined;
```
mertalev, line 252: Better to handle this as a ternary when `decodeOptions` is instantiated.

server/src/services/media.service.ts
mertalev: That's fine - it's a more generic name which makes it a better fit for full-res images that aren't extracted.

server/src/repositories/media.repository.ts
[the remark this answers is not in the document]
mertalev: I made a [PR](https://github.com/photostructure/exiftool-vendored.js/pull/222) for this, so we can change it if/when it's merged and in a release.

server/src/services/media.service.ts
```diff
@@ -245,29 +247,31 @@ export class MediaService extends BaseService {
...
+        // decoding from RAW requires orientation to be set
+        decodeOptions.orientation = asset.exifInfo?.orientation ? Number(asset.exifInfo.orientation) : undefined;
```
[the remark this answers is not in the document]
eligao: Not sure exactly what mean "ternary" here, but this is clean enough IMO. Making it ternary on `decodeOptions` init is too early as we need to confirm `useExtracted` first.

server/src/services/media.service.ts
eligao: This `size` cannot be handled on `decodeOptions` init, as `useExtracted` was not determined yet.
the `size` should be set when we extract from a non-RAW asset,
i.e. the `imageIsRaw || useExtracted` is false, 
or the opposite - unset when `imageIsRaw && useExtracted` is true.

server/src/services/media.service.ts
eligao: No special handling of `extractedFile` then it seems

server/src/services/media.service.ts
```diff
@@ -204,33 +225,59 @@ export class MediaService extends BaseService {
...
+        decodeOptions.size = undefined;
+        shouldConvertFromRaw = true;
       }
     }
+
+    const { info, data } = await this.mediaRepository.decodeImage(fullsizePath, decodeOptions);
+
+    const thumbnailOptions = { colorspace, processInvalidImages, raw: info };
+    const outputs = await Promise.all([
+      this.mediaRepository.generateThumbhash(data, thumbnailOptions),
```
eligao, line 272: This is indeed way slower when `data` is a full-sized buffer from RAW image, even times out on e2e `should upload and generate a thumbnail for different file types`.
Might be faster to sequentially zoom down the files, instead of delaying everything in a concurrency, which also makes the control flow ugly (note the `shouldConvertFromRaw`)

mertalev: Decoding the RAW image is *very* slow, so it's best to only do it once. This is also better in terms of minimizing disk IO (often the bottleneck during bulk import).

But it's also true that these tasks don't require such a high resolution input. Maybe it can be handled in two phases - use the full size buffer for the converted image and resize the buffer to be preview-sized as the input for the others.

mertalev: Alternatively, you can remove this code path and make a separate PR for it, potentially making it a separate setting. There's no need to block this PR for it.

eligao: Let's merge this first, and I'll make another MR to optimize the performance for RAW.

eligao:
any more comments from maintainers to get this merged, so that I can continue with previews of other images?
@danieldietzler @alextran1502

alextran1502:
@eligao no rush here

mertalev:
I think we should wait until we're done with hot fixes for the current release.

michelheusschen:
For a [DNG image](https://github.com/michelheusschen/immich-assets/blob/0fdfda8bbabd46696b235da11bda9466512d19fd/DJI_0070.dng) with a resolution of 5376x3956 and 41.3MiB in size, the following variants are generated using the default settings:
| Variant | Format | Resolution | Size |
| - | - | - | - |
| Thumbnail | webp | 5280x3956 | 887KiB |
| Preview | jpeg | 5280x3956 | 1794KiB |
| Original | jpeg | 5280x3956 | 1794KiB |

These generated variants seem larger than expected and don't match the configured settings, with the preview and original being identical. It's also strange that the resolution has slightly increased.

eligao:
> For a [DNG image](https://github.com/michelheusschen/immich-assets/blob/0fdfda8bbabd46696b235da11bda9466512d19fd/DJI_0070.dng) with a resolution of 5376x3956 and 41.3MiB in size, the following variants are generated using the default settings:
> 
> Variant	Format	Resolution	Size
> Thumbnail	webp	5280x3956	887KiB
> Preview	jpeg	5280x3956	1794KiB
> Original	jpeg	5280x3956	1794KiB
> These generated variants seem larger than expected and don't match the configured settings, with the preview and original being identical. It's also strange that the resolution has slightly increased.

nice catch, the file sizes are indeed wrong, let me fix that and add checks to tests.
as for the sizes being slightly different, it's likely due to the crop tags in exif: 
image

mertalev:
Hi again @eligao! We've been talking about the PR and think there are a few changes it'll need for it to be merge-able.

1. It should apply to all images that aren't web compatible (using the same list as [web](https://github.com/immich-app/immich/blob/bcc438eafbd6908306a41bc82afe0fd2dfe6fecc/web/src/lib/utils/asset-utils.ts#L329), ignoring that Safari can view HEIF images)
2. It should be configurable in the image settings, similar to the thumbnail and preview sections
   - There should be an "enabled" toggle that is disabled by default
   - The target format should be configurable in the case where an embedded image is not used and a new image has to be generated
3. Rename the "converted" enum to "fullsize"

server/src/repositories/media.repository.ts
```diff
@@ -54,22 +59,63 @@ export class MediaRepository implements IMediaRepository {
         return false;
       }
     }
-
     return true;
   }
 
+  async cloneExif(input: string, output: string): Promise<boolean> {
```
mertalev, line 65: Copying *all* of the metadata seems unnecessary. It adds to the size of the output, and there are surely metadata entries that are correct for the RAW, but wrong for the extracted image.

Can you narrow down which field(s) specifically need to be added? I'd like to see if we can avoid that exiftool.read call and just apply the data in exifInfo that we already have from metadata extraction.

server/src/repositories/media.repository.ts
```diff
@@ -54,22 +59,63 @@ export class MediaRepository implements IMediaRepository {
...
-        chromaSubsampling: options.quality >= 80 ? '4:4:4' : '4:2:0',
-      })
-      .toFile(output);
+    let pipeline = this.getImageDecodingPipeline(input, options).toFormat(options.format, {
+      quality: options.quality,
+      // this is default in libvips (except the threshold is 90), but we need to set it manually in sharp
+      chromaSubsampling: options.quality >= 80 ? '4:4:4' : '4:2:0',
+    });
+
+    if (options.keepExif === true) {
```
mertalev, line 114: `=== true` is unnecessary

server/src/services/media.service.ts
```diff
@@ -209,33 +230,67 @@ export class MediaService extends BaseService {
...
     }
+
+    const { info, data } = await this.mediaRepository.decodeImage(decodeInputPath, decodeOptions);
+
+    const thumbnailOptions = { colorspace, processInvalidImages, raw: info };
+    const outputs = await Promise.all([
+      this.mediaRepository.generateThumbhash(data, thumbnailOptions),
+      this.mediaRepository.generateThumbnail(data, { ...image.thumbnail, ...thumbnailOptions }, thumbnailPath),
+      this.mediaRepository.generateThumbnail(data, { ...image.preview, ...thumbnailOptions }, previewPath),
+      shouldConvertFullsize &&
```
mertalev, line 284: I'd prefer making this a list that you conditionally append to, then waiting for that list with `Promise.all`.

e2e/src/api/specs/asset.e2e-spec.ts
```diff
@@ -1222,7 +1222,7 @@ describe('/asset', () => {
       },
     ];
 
-    it(`should upload and generate a thumbnail for different file types`, async () => {
+    it(`should upload and generate a thumbnail for different file types`, { timeout: 60_000 }, async () => {
```
mertalev, line 1225: The timeout changes should be unnecessary since the setting is disabled by default, right?

open-api/immich-openapi-specs.json
```diff
@@ -11775,6 +11777,9 @@
           "extractEmbedded": {
             "type": "boolean"
           },
+          "fullsizePreview": {
```
mertalev, line 11780: I was thinking of something more like `"fullsize": {"enabled": false, "format": "jpeg"}`, where `format` is ignored if using the extracted preview.

server/src/repositories/media.repository.ts
mertalev: I also don't see the point to keepExif here. Why is this needed?

server/src/repositories/media.repository.ts
mertalev: I still don't see what the point is to including all the EXIF fields in the image. This is a generated image with a specific purpose. It doesn't need to have anything beyond what's need to display it correctly. If anything, the fact that the extracted image has metadata is a downside that prevents sharing due to metadata leakage.

eligao: This fullsize image extracted from RAW can be more than just specifically for preview. I'm thinking of adding features like "download as JPG" and "remove RAW but keep JPG to cleanup space" for RAW files later, which would be very handy for workflows, and it's important to preserve full EXIF fields.

As for EXIF leaking, this is indeed an issue when a shared link disallows downloading and hides EXIF but original JPG files can be acquired via `thumbnail?size=fullsize`, let me address this issue.

But anyway, this `cloneExif()` is already removed for now.

server/src/repositories/media.repository.ts
eligao: this was removed, mark as resolved

server/src/repositories/media.repository.ts
mertalev: I see what you're going for, but I think this should be handled when those features are actually implemented. It can be done as part of the request to download them, etc.

eligao: Good point, let me keep only colorspace and orientation EXIF in this PR.

eligao: Also I have added ` Permission.ASSET_DOWNLOAD` requirement for `thumbnail?size=fullsize` to prevent leaking EXIF via original files

open-api/immich-openapi-specs.json
mertalev: In the past, we've gotten complaints about image settings being shared between different files, and there may be other options related to the fullsize image added down the line. For example, a quality option might be added if users are concerned about file size and want 90 or 95 quality instead of 100, Safari users might complain because they don't need fullsize images for HEIF, etc.

I think the config field should be an object to make it easier to add settings in the future. The format is maybe superfluous in a vacuum, but the fact that one setting (format) is shared between the preview and the fullsize and the other two (quality, resolution) are not can lead to confusion. Having them completely separate will be less ambiguous.

eligao: it is now a separate `config.image.fullsize` config object.

server/src/services/asset-media.service.ts
```diff
@@ -208,10 +208,18 @@ export class AssetMediaService extends BaseService {
...
     let filepath = previewFile?.path;
     if (size === AssetMediaSize.THUMBNAIL && thumbnailFile) {
       filepath = thumbnailFile.path;
+    } else if (size === AssetMediaSize.FULLSIZE) {
+      await this.requireAccess({ auth, permission: Permission.ASSET_DOWNLOAD, ids: [id] });
+      // eslint-disable-next-line unicorn/prefer-ternary
+      if (mimeTypes.isWebSupportedImage(asset.originalPath)) {
+        filepath = asset.originalPath;
+      } else {
+        filepath = fullsizeFile?.path ?? previewFile?.path;
```
eligao, line 221: I wish to 302 redirect to original/preview path in the case when fullsize file is not available, so that we can make better use of caching, but I didn't find a good way to do the redirect either in this service or the controller/sendFile() that calls it. Any suggestions?

eligao:
> For a [DNG image](https://github.com/michelheusschen/immich-assets/blob/0fdfda8bbabd46696b235da11bda9466512d19fd/DJI_0070.dng) with a resolution of 5376x3956 and 41.3MiB in size, the following variants are generated using the default settings:
> 
> Variant	Format	Resolution	Size
> Thumbnail	webp	5280x3956	887KiB
> Preview	jpeg	5280x3956	1794KiB
> Original	jpeg	5280x3956	1794KiB
> These generated variants seem larger than expected and don't match the configured settings, with the preview and original being identical. It's also strange that the resolution has slightly changed.

This issue was fixed

server/src/services/asset-media.service.ts
mertalev: You can use `res.redirect` in the controller to redirect to the other endpoint. Maybe the service can return null or a boolean to indicate whether the controller should redirect?

eligao: implemented in https://github.com/immich-app/immich/pull/14446/commits/132e23b058418d5816970483c4b93739b1030d59

server/src/controllers/asset-media.controller.ts
```diff
@@ -119,10 +121,34 @@
...
+        redirPath = 'original';
+        redirSearchParams.delete('size');
+      } else if (Object.values(AssetMediaSize).includes(targetSize)) {
+        redirPath = reqPath;
+        redirSearchParams.set('size', targetSize);
+      } else {
+        throw new Error('Invalid targetSize: ' + targetSize);
+      }
+      const finalRedirPath = redirPath + '?' + redirSearchParams.toString();
+      return res.redirect(finalRedirPath);
```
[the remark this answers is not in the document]
eligao: this should be OK as `redirPath` is exactly the original request path unaltered.

mertalev:
I haven't forgotten about this PR - I will test it soon and get back to you.

mertalev:
I did a bit of testing with this. The image generation itself works well, but it doesn't get displayed correctly in the client. The initial image shown is the preview and zooming in shows the thumbnail. Also, the "Display original photos" user setting should probably make it load the full-size image directly and skip the preview.

mertalev:
Just a heads up that I will be away for the next two weeks and won't be able to test this. But feel free to continue working on it! I'm happy for it to be merged once someone can confirm that it works as expected.

mertalev:
Hey @eligao, just checking in! Were you able to take a look at the issues raised above? Is there anything I can do to help bring this PR to the finish line?

eligao:
@mertalev thanks for checking in, sorry I've been busy recently but I still remember this PR. Let me update the branch and see if I can fix the small issues you mentioned above :)

eligao:
@mertalev I've updated the branch and did some small fixes and tests:
- The initial image shown is the preview and zooming in shows the fullsize.
- The "Display original photos" user setting should not make it load the full-size image directly, otherwise it slows the browser down.

mertalev:
Nice! I'll do some testing with it tomorrow and see if we can get this merged :)

eligao:
@mertalev just checking in, any further suggestions for this PR? 👀

mertalev (approved):
The PR works as expected, tested on web with embedded preview enabled and disabled and format set to WebP and JPEG. Disabling the full-size feature and regenerating deletes both the file and the DB entry. Requesting a full-size image where there is none redirects to the preview and uses the cached preview (tested on Chrome).

eligao:
@mertalev thanks for the tweaks :)
1. Nice catch, I totally missed the case when zooming back out.
2. Sounds good as long as it is the current behavior.

eligao:
@danieldietzler any more suggestions before we can get this merged? 👀

mertalev:
We're planning to include it in 1.131.0, just needs testing from Alex

PR #14456, opened 2024-12-02
feat(server): library refresh go brrr

etnoy:
This PR significantly improves library scanning performance. Wherever suitable, we are doing jobs in batches, and many looped database interactions are replaced with SQL queries. 

**User testimonials**
"@etnoy what on earth have you done. I tried your PR and it finished the scan for 1M assets in *37 seconds* down from 728s on main. It takes 188s just to finish *queuing* on main" -- @mertalev 

**Changes made**
* When crawling a library: Don't call stat() on the files when importing, instead put a placeholder date (9999-12-31) that will be overwritten by metadata extraction. All dates after year 9000 are ignored in the timeline
* When importing new library files: Crawl 10k files per batch and then insert each batch in a single SQL query instead adding them individually
* When checking asset filenames against import paths and exclusion patterns: Do this directly in SQL which is much faster
* When a library is deleted, hide all its assets first. Otherwise, they'll linger until the asset deletion job catches up which can take days for >1M libraries

Plus several minor cleanups and performance enhancements.

The performance improvements are at least an **order of magnitude** in library scanning.

**Benchmark 1**
A library scan with 22k items where nothing has changed since the last scan used to take 1m 22s, now it's _below 10 seconds_, an improvement of 87 percent!

**Benchmark 2**
A clean library import with 19k items takes 1m40s in main and _7 seconds_ in this PR. 
NOTE: this benchmark is only the library service scan and does not include the metadata extraction. Also, some fs calls have been migrated from the library service to the metadata service, although this should only have a minor impact on overall scan performance

**Benchmark 3**
Importing a library with >5M assets.
* Time to 1M imported (without metadata extraction): 6m50s.

No need to compare to main, you know it's fast!

**Benchmark 4**
Importing a library of 527041 files took 1m58s (without metadata extraction) in this PR.
No need to compare to main, you know it's fast!

**Bonus:**
* Greatly improved log messages related to library scans. 
This scan imports all new files:
image

This is an "idle scan", where a refresh finds no changes:
image

* More e2e tests for handling when offline files go back online, leading to one major bug fixed

**Future work:**
* Crawl for XMP sidecars instead of queuing a sidecar discovery for each asset

**Final note:**
This PR allowed me to hit a milestone of 10M assets in a single Immich instance, likely a world-first. This does require max-old-space-size=8096, but that's to be expected anyway

image


server/src/services/library.service.ts
```diff
@@ -472,27 +470,35 @@ export class LibraryService extends BaseService {
...
+    const asset = await this.assetRepository.getById(id);
     if (!asset) {
       return JobStatus.SKIPPED;
     }
 
     const markOffline = async (explanation: string) => {
       if (!asset.isOffline) {
-        this.logger.debug(`${explanation}, removing: ${asset.originalPath}`);
+        this.logger.debug(`${explanation}, moving to trash: ${asset.originalPath}`);
         await this.assetRepository.updateAll([asset.id], { isOffline: true, deletedAt: new Date() });
```
mertalev, line 491: Use `update` instead of `updateAll`

server/src/services/library.service.ts
```diff
@@ -472,27 +470,35 @@ export class LibraryService extends BaseService {
...
+    return JobStatus.SUCCESS;
+  }
+
+  private async handleSyncAsset(id: string, importPaths: string[], exclusionPatterns: string[]): Promise<JobStatus> {
+    const asset = await this.assetRepository.getById(id);
     if (!asset) {
       return JobStatus.SKIPPED;
     }
 
     const markOffline = async (explanation: string) => {
```
mertalev, line 488: Just log directly without this function since the offline status will be set at the batch level at the end.

server/src/services/library.service.ts
```diff
@@ -472,27 +470,35 @@ export class LibraryService extends BaseService {
     return JobStatus.SUCCESS;
   }
 
-  @OnJob({ name: JobName.LIBRARY_SYNC_ASSET, queue: QueueName.LIBRARY })
-  async handleSyncAsset(job: JobOf<JobName.LIBRARY_SYNC_ASSET>): Promise<JobStatus> {
-    const asset = await this.assetRepository.getById(job.id);
+  @OnJob({ name: JobName.LIBRARY_SYNC_ASSETS, queue: QueueName.LIBRARY })
+  async handleSyncAssets(job: JobOf<JobName.LIBRARY_SYNC_ASSETS>): Promise<JobStatus> {
+    for (const id of job.ids) {
+      await this.handleSyncAsset(id, job.importPaths, job.exclusionPatterns);
```
mertalev, line 476: Fetch all the assets first with `getByIds`. Group the ones to be marked offline, queued for metadata extraction, etc. as you check them, then do batched async calls at the end as needed. The try/catch for `stat` should still be scoped to the asset so one error won't torpedo the batch.

mertalev (commented):
Nice start! I think there are still a lot of untapped potential improvements here.

mertalev:
The update to `fileCreatedAt`, `fileModifiedAt` and `originalFileName` is unnecessary and can be handled in metadata extraction since this will be queued anyway. This makes the batched update for `isOffline` and `deletedAt` simpler since there'll be no values that are unique to each asset.

etnoy:
Thanks for your comments @mertalev ! I'll first attempt to do the import path and exclusion pattern checks in SQL and then move to your suggestions

server/src/repositories/asset.repository.ts
```diff
@@ -709,4 +713,27 @@ export class AssetRepository implements IAssetRepository {
...
+      .createQueryBuilder()
+      .update()
+      .set({
+        isOffline: true,
+        deletedAt: new Date(),
+      })
+      .where({ isOffline: false })
+      .andWhere(
+        new Brackets((qb) => {
+          qb.where('originalPath NOT SIMILAR TO :paths', {
```
mertalev, line 730: Use `LIKE` instead of `SIMILAR TO`. 

The exclusions and import paths are also specific to a particular library, right? So you need to specify the library in the query.

Also, can you generate SQL for this and confirm with `EXPLAIN ANALYZE` that it uses an index?

server/src/repositories/asset.repository.ts
```diff
@@ -709,4 +713,27 @@ export class AssetRepository implements IAssetRepository {
...
+
+  updateOffline(importPaths: string[], exclusionPatterns: string[]): Promise<UpdateResult> {
+    const paths = importPaths.map((importPath) => `${importPath}%`).join('|');
+    const exclusions = exclusionPatterns.map((pattern) => globToSqlPattern(pattern)).join('|');
+    return this.repository
+      .createQueryBuilder()
+      .update()
+      .set({
+        isOffline: true,
+        deletedAt: new Date(),
```
mertalev, line 725: The `status` also needs to be set. This is why I don't really like the status field. The same info is stored in multiple places so it's so easy for it to go out of sync like this.

server/src/services/library.service.ts
```diff
@@ -472,52 +479,76 @@ export class LibraryService extends BaseService {
...
 
-  @OnJob({ name: JobName.LIBRARY_SYNC_ASSET, queue: QueueName.LIBRARY })
-  async handleSyncAsset(job: JobOf<JobName.LIBRARY_SYNC_ASSET>): Promise<JobStatus> {
-    const asset = await this.assetRepository.getById(job.id);
-    if (!asset) {
-      return JobStatus.SKIPPED;
+  @OnJob({ name: JobName.LIBRARY_SYNC_ASSETS, queue: QueueName.LIBRARY })
+  async handleSyncAssets(job: JobOf<JobName.LIBRARY_SYNC_ASSETS>): Promise<JobStatus> {
+    for (const id of job.ids) {
+      const asset = await this.assetRepository.getById(id);
```
mertalev, line 485: Do `getByIds` in one call

server/src/services/library.service.ts

server/src/services/library.service.ts

etnoy:
> The update to `fileCreatedAt`, `fileModifiedAt` and `originalFileName` is unnecessary and can be handled in metadata extraction since this will be queued anyway. This makes the batched update for `isOffline` and `deletedAt` simpler since there'll be no values that are unique to each asset.

Never thought of that, I've implemented your suggestion. I'm also considering changing the initial import code to ignore file mtime, this allows us to not do any file system calls except for the crawl. Metadata extraction will have to do the heavy lifting instead

mertalev:
> > The update to `fileCreatedAt`, `fileModifiedAt` and `originalFileName` is unnecessary and can be handled in metadata extraction since this will be queued anyway. This makes the batched update for `isOffline` and `deletedAt` simpler since there'll be no values that are unique to each asset.
> 
> Never thought of that, I've implemented your suggestion. I'm also considering changing the initial import code to ignore file mtime, this allows us to not do any file system calls except for the crawl. Metadata extraction will have to do the heavy lifting instead

Would that mean you queue them for metadata extraction even if they're unchanged? You can test it but I think it'd be more overhead than the stat calls

etnoy:
> > > The update to `fileCreatedAt`, `fileModifiedAt` and `originalFileName` is unnecessary and can be handled in metadata extraction since this will be queued anyway. This makes the batched update for `isOffline` and `deletedAt` simpler since there'll be no values that are unique to each asset.
> > 
> > 
> > Never thought of that, I've implemented your suggestion. I'm also considering changing the initial import code to ignore file mtime, this allows us to not do any file system calls except for the crawl. Metadata extraction will have to do the heavy lifting instead
> 
> Would that mean you queue them for metadata extraction even if they're unchanged? You can test it but I think it'd be more overhead than the stat calls.
> 
> Edit: also if you do this with the source set to `upload`, it would definitely be worse because it would queue a bunch of other things after metadata extraction.

I was referring to new imports, files that are new to immich. I hoped to improve the ingest performance by removing the stat call. After testing, there are two issues:

* assetRepository.create requires mtime, which we can only get from stat. We could work around that by setting it to new Date(), but ideally it should be undefined
* We still check for the existence of a sidecar, and this complicates things

If we can mitigate the two issues above, I can rewrite the library import feature and do that in batches as well!

mertalev:
I don't see why fileModifiedAt needs a non-null constraint in the DB. Might just be an oversight that didn't matter because it didn't affect our usage. I think you can change the asset entity and generate a migration to remove that constraint.

For sidecar files, maybe you could add`.xmp` to the glob filter and enable the option to make the files come in sorted order? That way you could make sure they're next to each other and in the same batch.

etnoy:
> I don't see why fileModifiedAt needs a non-null constraint in the DB. Might just be an oversight that didn't matter because it didn't affect our usage. I think you can change the asset entity and generate a migration to remove that constraint.
> 
> For sidecar files, maybe you could add`.xmp` to the glob filter and enable the option to make the files come in sorted order? That way you could make sure they're in the same batch.

I might just put new Date() in at the moment to keep the PR somewhat constrained.

Regarding sidecars, I have thought about that, problem right now is that we're batching the crawled files in batches of 10k. It might be hard to do get that working alright. Maybe I'll just queue a sidecar discovery for every imported asset for now

mertalev:
I think queueing sidecar discovery would introduce a race condition where it could run before, during or after metadata extraction. Since the refresh logic is already so much better, maybe leave the import for another PR so we can think about it more.

server/src/repositories/asset.repository.ts
etnoy: I'm trying to remember how the status field works. We use it to distinguish between user-deleted and offlined items, right? What shuld it be set to here?

zackpollard:
> I think queueing sidecar discovery would introduce a race condition where it could run before, during or after metadata extraction. Since the refresh logic is already so much better, maybe leave the import for another PR so we can think about it more.

Would this matter, because won't sidecar discovery re-queue metadata extraction afterwards if it does discover a sidecar file?

mertalev:
I haven't looked closely at the sidecar discovery job either, but I think it's an issue either way since it's possible for jobs dependent on the metadata to behave differently. For example, a sidecar file that changes the orientation of the image won't be respected during thumbnail generation.

server/src/repositories/asset.repository.ts
etnoy: Added the library parameter, thanks for spotting.

What is the rationale for using LIKE instead of SIMILAR TO?

server/src/repositories/asset.repository.ts
etnoy: > The `status` also needs to be set. This is why I don't really like the status field. The same info is stored in multiple places so it's so easy for it to go out of sync like this.

In main we don't set the status field here, if we want to change this we should do that in a separate PR.

server/src/repositories/asset.repository.ts
mertalev: The Postgres devs [rather dislike it](https://dba.stackexchange.com/a/10696). Queries that can use LIKE should use LIKE, queries that need regex should use the regex operator.

etnoy: We're now using LIKE instead of SIMILAR TO :)

server/src/repositories/asset.repository.ts
```diff
@@ -239,6 +245,17 @@ export class AssetRepository implements IAssetRepository {
     return paginationHelper(items as any as AssetEntity[], pagination.take);
   }
 
+  async getAllInLibrary(pagination: PaginationOptions, libraryId: string): Paginated<AssetEntity> {
+    const builder = this.db
+      .selectFrom('assets')
+      .select('id')
+      .where('libraryId', '=', asUuid(libraryId))
+      .limit(pagination.take + 1)
+      .offset(pagination.skip ?? 0);
```
mertalev, line 254: Use `.stream` and change the signature to run an async iterator. It's waaaay better than pagination because it means the query only runs once. I'm planning on changing most pagination methods to this.

etnoy: Nice, didn't know that!

server/src/repositories/asset.repository.ts
jrasm91: I'm pretty sure this is working as intended:
| State | deletedAt | status | Description |
| - | - | - | - |
| Normal | No | Active | not in trash |
| Offline | Yes | Active | "in trash" until we have a better view to manage offline assets, but not actually "trashed" |
| Trashed | Yes | Trashed | user deleted, so actually in the trash |
| Emptied from trash | Yes | Deleted | Use deleted and emptied the trash, so pending permanent delete by the server |

server/src/services/library.service.ts
```diff
@@ -190,12 +179,12 @@ export class LibraryService extends BaseService {
     }
   }
 
-  async getStatistics(id: string): Promise<LibraryStatsResponseDto> {
-    const statistics = await this.libraryRepository.getStatistics(id);
-    if (!statistics) {
-      throw new BadRequestException(`Library ${id} not found`);
+  async getStatistics(id: string): Promise<number> {
+    const count = await this.assetRepository.getLibraryAssetCount({ libraryId: id });
+    if (count == undefined) {
```
danieldietzler, line 184:
```suggestion
    if (count === undefined) {
```

server/src/services/library.service.ts
```diff
@@ -465,63 +442,145 @@ export class LibraryService extends BaseService {
...
+          break;
+        }
+        case AssetSyncResult.UPDATE: {
+          assetIdsToUpdate.push(asset.id);
+          break;
+        }
+        case AssetSyncResult.CHECK_OFFLINE: {
+          const isInImportPath = job.importPaths.find((path) => asset.originalPath.startsWith(path));
+
+          if (isInImportPath) {
```
danieldietzler, line 473: You can simplify this a bit by `if (!isInImportPath) { this.logger.debug(...); break; }`

server/src/services/library.service.ts
```diff
@@ -465,63 +442,145 @@ export class LibraryService extends BaseService {
...
+            const isExcluded = job.exclusionPatterns.some((pattern) => picomatch.isMatch(asset.originalPath, pattern));
+
+            if (isExcluded) {
+              this.logger.verbose(
+                `Offline asset ${asset.originalPath} is in an import path but still covered by exclusion pattern, keeping offline in library ${job.libraryId}`,
+              );
+            } else {
+              this.logger.debug(`Offline asset ${asset.originalPath} is now online in library ${job.libraryId}`);
+              assetIdsToOnline.push(asset.id);
+            }
```
danieldietzler, line 483: Here you don't need the else case either if you just `break` out of the case early.

server/src/services/metadata.service.ts
```diff
@@ -584,16 +598,27 @@ export class MetadataService extends BaseService {
...
+      this.logger.verbose(`Asset ${asset.id} has no time zone information`);
+    }
+
+    let fileCreatedAt = asset.fileCreatedAt;
+    let fileModifiedAt = asset.fileModifiedAt;
+
+    if (asset.isExternal) {
+      // With external assets we need to extract dates from the filesystem, this can't be done with uploades assets as that information is lost on upload
+      fileCreatedAt = stat.mtime;
+      fileModifiedAt = stat.mtime;
```
danieldietzler, line 610: I would argue that we should still prefer exif dates, no? Those will always be more accurate

server/src/repositories/asset.repository.ts
danieldietzler: @etnoy didn't you want to use a stream?

server/src/services/library.service.ts
```diff
@@ -228,26 +218,42 @@ export class LibraryService extends BaseService {
...
-        data: { id: assetId, importPaths, exclusionPatterns },
-      })),
+    const assetImports = job.assetPaths.map((assetPath) =>
+      this.processEntity(assetPath, library.ownerId, job.libraryId),
     );
+
+    const assetIds: string[] = [];
+
+    for (let i = 0; i < assetImports.length; i += 5000) {
+      // Chunk the imports to avoid the postgres limit of max parameters at once
```
danieldietzler, line 240: Is this still a problem or does kysely already handle this?

server/src/services/library.service.ts
```diff
@@ -330,6 +336,9 @@ export class LibraryService extends BaseService {
   async handleDeleteLibrary(job: JobOf<JobName.LIBRARY_DELETE>): Promise<JobStatus> {
     const libraryId = job.id;
 
+    // Hide all assets in the library before deleting them
```
danieldietzler, line 339:
```suggestion
    // Soft delete all assets in the library
```
I personally think this comment isn't even necessary

server/src/services/library.service.ts
```diff
@@ -465,62 +443,154 @@ export class LibraryService extends BaseService {
...
+
+      progressMessage.push(`${assetIdsToOffline.length} offlined`);
     }
 
-    const isExcluded = job.exclusionPatterns.some((pattern) => picomatch.isMatch(asset.originalPath, pattern));
-    if (isExcluded) {
-      await markOffline('Asset is covered by an exclusion pattern');
-      return JobStatus.SUCCESS;
+    if (assetIdsToOnline.length > 0) {
+      //TODO: When we have asset status, we need to leave deletedAt as is when status is trashed
```
danieldietzler, line 511: This TODO does not need to be addressed here?

server/src/services/library.service.ts
```diff
@@ -465,62 +443,154 @@ export class LibraryService extends BaseService {
...
+        status: AssetStatus.ACTIVE,
+        deletedAt: null,
+      });
+      await this.queuePostSyncJobs(assetIdsToOnline);
+
+      progressMessage.push(`${assetIdsToOnline.length} onlined`);
+    }
+
+    if (assetIdsToUpdate.length > 0) {
+      //TODO: When we have asset status, we need to leave deletedAt as is when status is trashed
```
danieldietzler, line 523: Same question here?

server/src/services/library.service.ts
```diff
@@ -465,62 +443,154 @@ export class LibraryService extends BaseService {
...
   }
 
-  @OnJob({ name: JobName.LIBRARY_SYNC_ASSET, queue: QueueName.LIBRARY })
-  async handleSyncAsset(job: JobOf<JobName.LIBRARY_SYNC_ASSET>): Promise<JobStatus> {
-    const asset = await this.assetRepository.getById(job.id);
-    if (!asset) {
-      return JobStatus.SKIPPED;
-    }
+  @OnJob({ name: JobName.LIBRARY_SYNC_ASSETS, queue: QueueName.LIBRARY })
+  async handleSyncAssets(job: JobOf<JobName.LIBRARY_SYNC_ASSETS>): Promise<JobStatus> {
```
danieldietzler, line 451: In this function (and kind of in general) there is a lot of logic (or at least lines of code I need to read) that are purely added complexity for the sake of logs. I get that logs are neat, but I would personally argue that some of them don't add any value at all. IMO logs should primarily indicate and help understand an error.

server/src/services/library.service.ts
```diff
@@ -546,63 +616,153 @@ export class LibraryService extends BaseService {
...
+    if (crawlCount === 0) {
+      this.logger.log(`No files found on disk for library ${library.id}`);
+    } else if (importCount > 0 && importCount === crawlCount) {
+      this.logger.log(`Finished crawling and queueing ${crawlCount} file(s) for import for library ${library.id}`);
+    } else if (importCount > 0) {
+      this.logger.log(
+        `Finished crawling ${crawlCount} file(s) of which ${importCount} file(s) are queued for import for library ${library.id}`,
+      );
+    } else {
+      this.logger.log(`All ${crawlCount} file(s) on disk are already in library ${library.id}`);
```
danieldietzler, line 669: This is a good example of the logging-induced complexity

server/src/services/library.service.ts
```diff
@@ -546,63 +616,153 @@ export class LibraryService extends BaseService {
...
+    this.logger.log(`Scanning library ${library.id} for assets missing from disk...`);
+
+    const existingAssets = usePagination(JOBS_LIBRARY_PAGINATION_SIZE, (pagination) =>
+      this.assetRepository.getAllInLibrary(pagination, job.id),
+    );
+
+    let currentAssetCount = 0;
+    for await (const assets of existingAssets) {
+      if (assets.length === 0) {
+        throw new InternalServerErrorException(`Failed to get assets for library ${job.id}`);
```
danieldietzler, line 740: I don't think I've ever seen that exception. What does it do? Why do we need it?

server/src/services/library.service.ts
```diff
@@ -546,63 +616,153 @@ export class LibraryService extends BaseService {
...
+
+      currentAssetCount += assets.length;
+
+      await this.jobRepository.queue({
+        name: JobName.LIBRARY_SYNC_ASSETS,
+        data: {
+          libraryId: library.id,
+          importPaths: library.importPaths,
+          exclusionPatterns: library.exclusionPatterns,
+          assetIds: assets.map((asset) => asset.id),
```
danieldietzler, line 751: Very minor nit: I generally find `assets.map(({ id }) => id)` cleaner.

server/src/repositories/asset.repository.ts
etnoy: I'm basically copying what's already in getAll() and as I said above using a stream will likely have to be a followup to keep the changes reasonable. Would that be OK?

server/src/repositories/asset.repository.ts
```diff
@@ -726,4 +747,64 @@ export class AssetRepository implements IAssetRepository {
...
+      )
+      .execute();
+
+    return result.map((row) => row.path as string);
+  }
+
+  async getLibraryAssetCount(options: AssetSearchOptions = {}): Promise<number | undefined> {
+    const { count } = await this.db
+      .selectFrom('assets')
+      .select(sql`COUNT(*)`.as('count'))
```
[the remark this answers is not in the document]
etnoy: You are not able to see my previous comment? 

image

server/src/services/library.service.ts
etnoy: Trust me, if you scan or rescan a library with >1M assets you really need these logs

server/src/services/library.service.ts
```diff
@@ -546,63 +616,153 @@ export class LibraryService extends BaseService {
...
+
+    const affectedAssetCount = Number(offlineResult.numUpdatedRows);
+
+    if (affectedAssetCount === assetCount) {
+      this.logger.log(
+        `All ${assetCount} asset(s) were offlined due to import paths and/or exclusion pattern(s) in ${library.id}`,
+      );
+
+      return JobStatus.SUCCESS;
+    } else if (affectedAssetCount === 0) {
```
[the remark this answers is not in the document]
etnoy: Technically, you're right, but isn't it more readable with the else? It keeps the logic grouped up in one paragraph

server/src/services/library.service.ts
etnoy: I removed it

server/src/services/library.service.ts
etnoy: Fixed

server/src/services/library.service.ts

server/src/repositories/asset.repository.ts
danieldietzler: I don't think it'll (in total) be more changes, but it's probably fine to keep it as is if you prefer that.

server/src/repositories/asset.repository.ts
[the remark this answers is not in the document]
danieldietzler: https://github.com/immich-app/immich/blob/392f9f205cd47178a72513c7b009e62b7344b88a/server/src/repositories/activity.repository.ts#L61

server/src/services/library.service.ts
danieldietzler: Can we at least reduce them or have more generic logs instead of dedicated if/else structures and variables just for logging?

server/src/services/library.service.ts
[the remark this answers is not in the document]
danieldietzler: I personally find large `if` structures that branch with `else if` more complex/hard to read compared to being able to quickly see "ah it's not this. ah it's not this. ah it's not this. ah this is it"

server/src/services/library.service.ts
etnoy: I replaced this with a single log statement

server/src/services/library.service.ts
etnoy: Done. Let me know if you think this is better

server/src/services/library.service.ts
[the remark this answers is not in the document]
etnoy: This has been cleaned up

server/src/services/library.service.ts
etnoy: Fixed

server/src/services/library.service.ts
etnoy: Fixed

server/src/services/library.service.ts
etnoy: @danieldietzler when I first migrated this PR to kysely I had to batch it here to avoid errors

alextran1502:
You call on merging this one @mertalev

zackpollard:
> You call on merging this one @mertalev

If it's ready for review I would like to take a look at it too before it is merged, I can take a look tomorrow morning.

etnoy:
> > You call on merging this one @mertalev
> 
> If it's ready for final review I would like to take a look at it too before it is merged, I can take a look tomorrow morning.

Thanks, I'd apprecate it!

server/src/repositories/asset.repository.ts
etnoy: I would need help implementing it this way since I'm not sure how it should be done. But if it is OK to follow this up later, I'd prefer that.

server/src/repositories/asset.repository.ts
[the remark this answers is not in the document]
etnoy: Done!

mertalev:
@etnoy I made a few changes, let me know if there's anything that seems off.

server/src/services/library.service.ts
mertalev: I reduced the amount of logging code a bit because I also found it excessive. Logging is good, but it's also important to have a high signal/noise ratio and avoid affecting code readability.

etnoy:
> @etnoy I made a few changes, let me know if there's anything that seems off.

I had a look and it's a big improvement, thanks for helping me clean this up

zackpollard:
Didn't get to this today sorry, on my list for tomorrow 😅

mertalev (approved):

zackpollard (approved):
LGTM, nice work. You have some failing unit tests after the latest changes, but it's good to go after those are fixed I think

etnoy:
@zackpollard good to go now that the release is done?

PR #14497, opened 2024-12-05
feat(web): chromecast support for shared content

omricarmel:
* supporting chromecast casting through chrome browser 
* only on publicly shared assets without password 
* directing the casting receiver to download asset endpoint 
* using default receiver 
somewhat resolves https://github.com/immich-app/immich/discussions/7030

image

### DIFF
--- a/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
+++ b/web/src/lib/components/asset-viewer/asset-viewer-nav-bar.svelte
@@ -15,6 +15,7 @@
   import UnstackAction from '$lib/components/asset-viewer/actions/unstack-action.svelte';
   import KeepThisDeleteOthersAction from '$lib/components/asset-viewer/actions/keep-this-delete-others.svelte';
   import CircleIconButton from '$lib/components/elements/buttons/circle-icon-button.svelte';
+  import CastSender from '$lib/components/cast/cast-sender.svelte';
   import ButtonContextMenu from '$lib/components/shared-components/context-menu/button-context-menu.svelte';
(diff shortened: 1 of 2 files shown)


bo0tzz:
How does this relate to #13966?

etnoy:
I might be biased, but I suggest we combine our efforts in #13966. That PR has come much further along, including private assets and video controls

omricarmel:
@bo0tzz  this pr is simpler and less comprehensive.
 
On the other hand i do not hold a state and offload playback controls to chrome's cast popup interface.

as @etnoy mentioned i only allow casting publicly accessible assets ( so that chromecast will be able to reach them on its own ).

omricarmel:
> I might be biased, but I suggest we combine our efforts in #13966. That PR has come much further along, including private assets and video controls

@etnoy Definitely. I didn't know about the other pr when working on this one. 

Even though this one would be easier to finish at its scope of operation and complexity.

omricarmel:
Seeing the diff between the PRs, This style implementation has merit. offloading the session playback handling to chrome and not allocating privilege to the cast device.  The PRs are complementary, this one directs the cast device to an asset upon request (hence only public assets) and the other puts the web viewer into casting mode hooking all asset views.

Eventually they should be unified but this pr can be completed before the more comprehensive one.

zackpollard:
Hey, we are going to push ahead with the existing PR, but we will pull some inspiration from this PR wrt the loading code. We have already had discussions around authentication and ability to totally disable the option to cast that this doesn't implement. Also, I understand you don't want to initiate a cast session and just want to cast a single picture, however this isn't the direction we want to pursue. Thanks for your PR and we hope to have chromecast support implemented soon!

PR #14545, opened 2024-12-07
fix(web): recent albums sort

### LINKED ISSUE #14541: recent albums doesn't work
### The bug

Recent albums dropdown show albums that are not the recent ones. One of them has the oldest photo and the two others are two years old.

obrazek
obrazek
obrazek

I've tested to change sort order, but it has no effect.

Fedora 41

### Version of Immich Server

v.1.122.1

not relevant

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

```YAML
not relevant
```

### Your .env content

```Shell
not relevant
```

### Reproduction steps

N/A

### Relevant log output

_No response_

### Additional information

_No response_

michelheusschen:
Fixes #14541 by sorting the recent albums based on the album's modify date, instead of the album's last modified asset date. I think this makes more sense, especially since `lastModifiedAssetTimestamp` can be undefined when an album has no assets.


danieldietzler (approved):
Agreed, that makes more sense. Thanks!

jrasm91 (approved):

PR #14570, opened 2024-12-08
docs: standalone psql pg_hba.conf error help

WillForan:


mmomjian:
Thanks for the PR. I really don’t know how I feel about adding such basic Postgres knowledge to our docs.

if you don’t already know how to do this, my opinion is that you should **not use a standalone postgres**.  Perhaps others disagree with me, though.

WillForan:
Thanks for looking and I can agree. The error message is relatively straight forward and easily searched (though I can only say that now with the clarity of hindsight).

The part that caught me:  I (erroneously) thought I'd already allowed all of ipv4 in my config and was convinced Immich was ignore DB_URL for a docker-compose connection (or something). It took me a while to realize the error in the log was forwarded on from the standalone server and not related to the Immich server.  ... That's probably a pretty niche confusion.

bo0tzz:
Thanks for the PR! I agree with mmomjian, I think this is basic postgresql admin knowledge that you should be able to figure out yourself if you choose to run a standalone database. I also think we shouldn't be telling people to set the auth to `trust`, because someone will blindly copy that and get their database pwned.

PR #14571, opened 2024-12-08
docs: standalone postgres: host.containers.internal

WillForan:
note about using docker's `host.containers.internal` not using 'immichdbusername' b/c don't want host.containers.internal to look like a stand in variable that should be changed


bo0tzz:
I think this is fairly standard docker networking knowledge that I'd rather not duplicate in our documentation. It's also something that will fail to work in a fair few environments and that might confuse people.

PR #14611, opened 2024-12-09
fix(server): partial fallback for hardware transcoding

mertalev:
## Description

When hardware decoding is enabled and an error occurs, there is a fallback to sw decoding + hw encoding before falling back to full software. However, this does not actually work since the dri devices aren't provided in this second call. This PR fixes this and does some refactoring around how these interfaces are passed to the command builder.

## How Has This Been Tested?

Tested that a ProRes video successfully falls back to sw decoding + hw encoding.

### DIFF
--- a/server/src/interfaces/media.interface.ts
+++ b/server/src/interfaces/media.interface.ts
@@ -130,6 +130,11 @@ export interface ProbeOptions {
   countFrames: boolean;
 }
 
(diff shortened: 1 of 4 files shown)


danieldietzler (approved):
I have no idea about the actual content but the code looks good to me lol

PR #14626, opened 2024-12-10
refactor(server): use `includeNull` in query for search suggestions

mertalev:
## Description

Extracted out from the Kysely branch, this PR refactors the search suggestion methods to use the `includeNull` option and empty string filter into the queries themselves and changes the queries to use inner joins.


danieldietzler (approved):
LGTM (minus the broken e2e tests :D)

mertalev:
I think the e2e only worked because there were assets without entries in the exif table at the time of the test. The inner join would mean null isn't one of the options in that case. Since the inner join is good to have regardless, I changed it to just add `null` to the end if `includeNull` is true. It does make the flag kind of pointless though lol

PR #14662, opened 2024-12-12
refactor(web): asset interaction

michelheusschen:
- replace store with runes
- centralize derived logic like `isAllTrashed`, `isAllUserOwned`, etc.
- add some tests

PR #14666, opened 2024-12-12
fix: reduce the number of API requests when changing route

martabal:
Reduce the number of API requests when changing route by storing the already stored value to a store. (maybe there's a better way with svelte 5?)

Example with navigating from `/photos` to `/utilities`
| Before | After |
| :---: | :---: |
| image | No request |

### DIFF
--- a/web/src/lib/components/shared-components/side-bar/recent-albums.svelte
+++ b/web/src/lib/components/shared-components/side-bar/recent-albums.svelte
@@ -4,13 +4,19 @@
   import { getAllAlbums, type AlbumResponseDto } from '@immich/sdk';
   import { handleError } from '$lib/utils/handle-error';
   import { t } from 'svelte-i18n';
+  import { userInteraction } from '$lib/stores/user.svelte';
 
   let albums: AlbumResponseDto[] = $state([]);
 
   onMount(async () => {
+    if (userInteraction.recentAlbums) {
+      albums = userInteraction.recentAlbums;
+      return;
+    }
     try {
       const allAlbums = await getAllAlbums({});
       albums = allAlbums.sort((a, b) => (a.updatedAt > b.updatedAt ? -1 : 1)).slice(0, 3);
+      userInteraction.recentAlbums = albums;
     } catch (error) {
       handleError(error, $t('failed_to_load_assets'));
     }
--- a/web/src/lib/components/forms/create-user-form.svelte
+++ b/web/src/lib/components/forms/create-user-form.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
   import FullScreenModal from '$lib/components/shared-components/full-screen-modal.svelte';
   import { featureFlags } from '$lib/stores/server-config.store';
-  import { serverInfo } from '$lib/stores/server-info.store';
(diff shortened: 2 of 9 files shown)


alextran1502:
We probably should use the rune syntax for these since store syntax will be deprecated in the future iirc

danieldietzler:
Yup, this should be some global state :)

danieldietzler:
#14662 could be a good inspiration?

michelheusschen:
Stores are indeed not deprecated. In this situation you can just replace `writable()` with `$state()` and make sure the file ends with `.svelte` or `.svelte.ts`

We do have to be careful about persisting state globally, because a user could log in with a different account. Data can also become stale after a certain period.

This PR also fixes #14509

alextran1502:
The `userInteraction` global state should be cleared when logging out, correct?

martabal:
> The userInteraction global state should be cleared when logging out, correct?

Indeed!

web/src/lib/components/shared-components/side-bar/recent-albums.svelte
```diff
@@ -4,13 +4,19 @@
   import { getAllAlbums, type AlbumResponseDto } from '@immich/sdk';
   import { handleError } from '$lib/utils/handle-error';
   import { t } from 'svelte-i18n';
+  import { userInteraction } from '$lib/stores/user.svelte';
 
   let albums: AlbumResponseDto[] = $state([]);
```
danieldietzler, line 9: You could completely drop this variable and replace it with the global `userInteraction.recentAlbums` state

danieldietzler (approved):
Code looks good to me

web/src/lib/components/shared-components/side-bar/recent-albums.svelte
martabal: Do you mean dropping `let albums: AlbumResponseDto[] = $state([]);` in favor of using only `userInteraction.recentAlbums`? If so, you have to wrap the each section in a `{#if userInteraction.recentAlbums}` because `userInteraction.recentAlbums` can be undefined. What do you prefer?

danieldietzler: Can it be undefined? You'd still set it in `onMount` if it's undefined, and `onMount` finishes before it actually mounts, right?

danieldietzler: Or is my svelte knowledge lacking here? 😅

martabal: Yes, technically it can't be indefinite, but Svelte isn't smart enough to figure that out. It only uses the type definition

image


```ts
interface UserInteractions {
  recentAlbums?: AlbumResponseDto[];
  versions?: ServerVersionHistoryResponseDto[];
  aboutInfo?: ServerAboutResponseDto;
  serverInfo?: ServerStorageResponseDto;
}
```

danieldietzler: Ah gotcha. Then I'd prefer keeping that extra variable as well

Issue #10312, opened 2024-06-14
Higher periodic CPU usage since 1.160.2


ykrasik:
### The bug

Hi,

Since upgrading to version 1.106.2 I noticed increased periodic CPU usage in both `immich-server` and `immich-machine-learning`.
There are spikes of cpu usage in `immich-machine-learning` every 6-7 minutes and in `immich-server` every 12-15 minutes:

### Before
![image](https://github.com/immich-app/immich/assets/5993890/6929c4a4-1074-4540-af15-d9c8433c9fb4)

### After
![image](https://github.com/immich-app/immich/assets/5993890/8140d249-cce4-45b7-8b45-48b6c3a7d231)

After upgrading to 1.106.2 I immediately ran the duplicates detection & thumbnail generation jobs, so in case one of those jobs got the system into an invalid state which now causes these periodic spikes - I wouldn't be able to provide the CPU graph before running those jobs.

I'm running this on an Orange Pi 5+ (Rockchip RK3588). Also a while ago I tried to get hw acceleration to work - as far as I recall I managed to make it work for transcoding but not for ML, and left it in this unfinished state. AFAICT it's working, just not utilizing the hardware. 

The logs for `immich-server` show this warning periodically:
```
[Nest] 17  - 06/14/2024, 7:55:53 AM    WARN [Api:DuplicateService~b4mce4tp] Unknown keys found: {
  "machineLearning": {
    "classification": {
      "enabled": true,
      "modelName": "google/vit-base-patch16-224"
    }
  }
}
```
My CLIP model is `ViT-H-14-quickgelu__dfn5b` and facial recognition is `buffalo_l`.

I use centralized postrgres (16.3) & redis (7.2.5) instances.

### The OS that Immich Server is running on

Ubuntu 22.04.4 LTS

### Version of Immich Server

v1.106.4

### Version of Immich Mobile App

v1.106.1.build.142

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

### Your docker-compose.yml content

```YAML
### docker-compose.yml
services:
  immich-server:
    container_name: immich-server
    image: ghcr.io/immich-app/immich-server:release
    restart: unless-stopped
    user: "$PUID:$PGID"
    extends: # uncomment this section for hardware acceleration - see https://immich.app/docs/features/hardware-transcoding
      file: hwaccel.transcoding.yml
[...]

PR #14704, opened 2024-12-15
feat(server): add Bash healthcheck script

000yesnt:
The current health check script starts up a Node JS script every time it's called. This becomes an issue in low-powered devices hosting Immich as the startup process can be CPU-intensive. The CPU usage bursts can compromise overall system stability. 
#10312 (Orange Pi 5+) and #14364 (Raspberry Pi 4) are examples of users of such devices complaining about the usage spikes, where the only solution was to disable health checking entirely.

In testing with an old desktop PC (Intel Pentium E5400, 2GB of RAM + 4GB of ZRAM), replacing the JS healthcheck with a Bash one eliminated the spikes.
image

I have tried to replicate the behavior of the original healthcheck script as closely as I could, including an early exit when the ``api`` worker is excluded or left out of IMMICH_WORKERS_INCLUDE. I also passed it through Shellcheck to prevent bugs.


jrasm91 (approved):
LGTM

danieldietzler (approved):

bo0tzz (dismissed):
Should we remove the node healthcheck code?

000yesnt:
> Should we remove the node healthcheck code?

I can delete the node script if there won't be any need for it. I should've done that eariler but I forgot to

server/bin/immich-healthcheck
```diff
@@ -1,3 +1,23 @@
 #!/usr/bin/env bash
 
-node /usr/src/app/dist/bin/healthcheck.js
+if [[ "$IMMICH_WORKERS_EXCLUDE" == *"api"* ]] ||
+   [[ "$IMMICH_WORKERS_INCLUDE" != *"api"* ]]; then
```
mmomjian, line 4: Has this been tested? AFAIK by default all workers will run, so it doesn't make sense to test for api in the include variable since most people will not have this set.

Should probably be something like `if [[ ( $IMMICH_WORKERS_INCLUDE != '' && $IMMICH_WORKERS_INCLUDE != *api* ) || $IMMICH_WORKERS_EXCLUDE == *api* ]]`

server/bin/immich-healthcheck
```diff
@@ -1,3 +1,23 @@
...
+        echo "API worker excluded, skipping";
+        exit 0;
+fi
+
+IMMICH_HOST="${IMMICH_HOST:=localhost}"
```
mmomjian, line 9: This should be `:-`, not `:=`. Same on next line.

https://tldp.org/LDP/abs/html/parameter-substitution.html

server/bin/immich-healthcheck
```diff
@@ -1,3 +1,23 @@
...
+IMMICH_PORT="${IMMICH_PORT:=2283}"
+
+result=$(curl -m 2 http://"$IMMICH_HOST":"$IMMICH_PORT"/api/server/ping)
```
mmomjian, line 12: would probably use `curl -fsS` here, because we don't need all the stderr output of curl.

server/bin/immich-healthcheck
```diff
@@ -1,3 +1,23 @@
...
+result_exit=$?
+
+if ! [ $result_exit == 0 ]; then
+        echo "Fail: exit code is $result_exit";
+        exit 1;
+fi
+
+if ! [ "$result" == "{\"res\":\"pong\"}" ]; then
```
mmomjian, line 20:
```suggestion
if [[ $result != "{\"res\":\"pong\"}" ]]; then
```

Using Bash style tests above, no reason to confuse it here

server/bin/immich-healthcheck
```diff
@@ -1,3 +1,23 @@
...
+        exit 0;
+fi
+
+IMMICH_HOST="${IMMICH_HOST:=localhost}"
+IMMICH_PORT="${IMMICH_PORT:=2283}"
+
+result=$(curl -m 2 http://"$IMMICH_HOST":"$IMMICH_PORT"/api/server/ping)
+result_exit=$?
+
+if ! [ $result_exit == 0 ]; then
```
mmomjian, line 15:
```suggestion
if [[ $result_exit != 0 ]]; then
```

mmomjian (commented):
Thanks, just a few BASHism questions!

server/bin/immich-healthcheck
000yesnt: Yes, I've tested it, but you're right. I wrote the include test to handle ``IMMICH_WORKERS_INCLUDE=microservices``, but it does make more sense to just test the excludes

server/bin/immich-healthcheck
```diff
@@ -1,3 +1,22 @@
...
+
+if [ $result_exit != 0 ]; then
+        echo "Fail: exit code is $result_exit";
+        exit 1;
+fi
+
+if [ "$result" != "{\"res\":\"pong\"}" ]; then
+        echo "Fail: didn't reply with pong";
+        exit 1;
+fi
```
mmomjian, line 22: I would explicitly `exit 0` at the end of the script. Otherwise looks good to me. Ty!

akostadinov:
Hmm, I was wondering why disk is so often queried... Is it possible to disable health check or make it not access storage so that HDDs can spin down?

paschun:
It would be great if this could also be done for the machine-learning container. It has the same issue, except launching a python healthcheck consumes 100% of one CPU core on a SBC for ~5s.

PR #14737, opened 2024-12-17
fix(web): fix missing icons 

### LINKED ISSUE #14702: Icons missing from photo toolbar when screen too narrow (web)
### The bug

Opening a photo with a web browser when the screen is too narrow will hide the leftmost icons in the toolbar. They become visible when the screen is made wider. Can't find them in the dropdown menu either.

In the app the layout is different and it doesn't have this problem. It's very close though, maybe it would be missing icons if the screen was just a tiny bit narrower than what I have?

Narrow screen:

Wide screen:

Docker

### Version of Immich Server

1.122.3

### Platform with the issue

- [ ] Server
- [X] Web
- [ ] Mobile

aviv926:
Fixes #14702

Before:

https://github.com/user-attachments/assets/e66001e2-428c-4c42-8650-e8d92d02461d

After:

https://github.com/user-attachments/assets/56946995-1788-4948-b63a-3ba72cf35106


jrasm91 (approved):
I guess this improves on the current state, but ideally the overflowed icons go into the _overflow menu_ :laughing:

ajpanton:
> I guess this improves on the current state, but ideally the overflowed icons go into the _overflow menu_ 😆

Yeah the scrolling ability isn't very visible, and I would personally never guess it's scrollable and would create the same bug report again if I didn't already.

I would propose one of three options:
1. Replace the left-most icon with something that shows that there are more icons. Not really sure what this would look like. A horizontal three dots would've been good, if it was actually the left-most icon, but it's not, as there's the "back" arrow. Maybe a fade-out on the left-most icon? A different-looking left-arrow?
2. Add a bar under the icons showing the horizontal scroll position, making it obvious that you're currently all the way to the right and there might be something more to the left.
3. Put the additional icons in the overflow menu.

jrasm91:
(3) is the only viable solution IMO.

PR #14769, opened 2024-12-18
Feat/cli delete duplicates

chefkoch:
Quick'n'Dirty first try to implement a feature to delete assets via cli not only for those that were uploaded, but also for those ones which were identified as duplicates.

see also #12181

### DIFF
--- a/cli/src/commands/asset.ts
+++ b/cli/src/commands/asset.ts
@@ -32,6 +32,7 @@ export interface UploadOptionsDto {
   dryRun?: boolean;
   skipHash?: boolean;
   delete?: boolean;
+  deleteDuplicates?: boolean;
   album?: boolean;
   albumName?: string;
   includeHidden?: boolean;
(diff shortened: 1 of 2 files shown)


mertalev:
Others can chime in, but I don't think we really need an option for this. It's more intuitive and idempotent imo if it just handles both new and duplicate files.

chefkoch:
Thanks for your reply.  I went the "conservative way" to add the option just in case, not deleting duplicates was by design.

> It's more intuitive and idempotent imo if it just handles both new and duplicate files.

So you mean by _handles both_ the **--delete** option should delete the files wheter new files have been uploaded or duplicates have been identified?
--> Fine for me as well :-)

mertalev:
> So you mean by handles both the --delete option should delete the files wheter new files have been uploaded or duplicates have been identified?

Yup!

Sese-Schneider:
@chefkoch I plan to adjust the duplicate detection on the frontend: https://github.com/immich-app/immich/pull/14947
Is this something you also might consider to add here?

PR #14775, opened 2024-12-18
fix(server): don't delete offline library files from disk when clearing trash

etnoy:
Right now, any offline assets end up in the trash but does not allow the user to permanently delete them. They only get deleted after 30 days.

This change allows users to delete those assets from trash, both individually and in bulk


etnoy:
Closing this after feedback from the team

PR #14777, opened 2024-12-18
fix(server): don't delete offline files from disk when trash empties

etnoy:
Right now, immich deletes files after 30 days that are excluded or outside of import paths, including original files.

This is not intended, if a new exclusion pattern is added or an import path is changed and one or more assets are removed due to it, we should not be deleting original files.

This bug is in my opinion quite serious and could cause data loss.

I can't find the original bug report (github? reddit?)

Edit: This might be it: https://www.reddit.com/r/immich/comments/1gu61r3/immich_tried_to_remove_my_external_libraries_file/

Also adds some e2e tests related to offline files

### DIFF
--- a/server/src/services/asset.service.ts
+++ b/server/src/services/asset.service.ts
@@ -188,7 +188,7 @@ export class AssetService extends BaseService {
           name: JobName.ASSET_DELETION,
           data: {
             id: asset.id,
-            deleteOnDisk: true,
+            deleteOnDisk: !asset.isOffline,
           },
         })),
       );
@@ -249,6 +249,7 @@ export class AssetService extends BaseService {
 
     const { thumbnailFile, previewFile } = getAssetFiles(asset.files);
     const files = [thumbnailFile?.path, previewFile?.path, asset.encodedVideoPath];
+
     if (deleteOnDisk) {
       files.push(asset.sidecarPath, asset.originalPath);
     }
--- a/e2e/src/api/specs/asset.e2e-spec.ts
+++ b/e2e/src/api/specs/asset.e2e-spec.ts
@@ -766,7 +766,7 @@ describe('/asset', () => {
       expect(body).toEqual(errorDto.badRequest('Not found or no asset.delete access'));
     });
 
-    it('should move an asset to the trash', async () => {
+    it('should move an asset to trash', async () => {
       const { id: assetId } = await utils.createAsset(admin.accessToken);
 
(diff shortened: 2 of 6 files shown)


server/src/services/asset.service.ts
```diff
@@ -249,7 +249,18 @@ export class AssetService extends BaseService {
...
+
+    let willDelete = deleteOnDisk;
+
+    if (asset.isOffline) {
+      /* We don't want to delete an offline asset because it is either...
+           ...missing from disk => don't delete the file since it doesn't exist where we expect
+           ...outside of any import path => don't delete the file since we're not responsible for it
+           ...matching an exclusion pattern => don't delete the file since we're not responsible for it */
+      willDelete = false;
+    }
```
jrasm91, line 261: This logic already exists outside of this job. Why is it duplicated here again?

jrasm91: Did you only keep the logic inside here? Imo the job should be called with delete: true | false and then that's what should happen. It seems much clearer to just pass false when the asset is offline.

zackpollard (commented):
This looks good overall, I think just the point Jason raised, the asset service isn't where that logic should live

server/src/services/asset.service.ts
etnoy: My reasoning was that it would be safer for the future to check the isOffline right before deletion. The possible outcome is data loss which is severe and I didn't want future refactors to accidently remove the check. But I see the logic in your approach and will update accordingly.

etnoy:
> This looks good overall, I think just the point Jason raised, the asset service isn't where that logic should live

The logic still needs to be in the asset service since that's where the deletion is initiated. Ideally I feel like it could be in the trash service, but it isn't for the moment.

server/src/services/asset.service.ts
jrasm91: When an asset goes online the offline flash and the trashed flag are updated at the same time and vice versa. If a file goes offline after the job is queued I don't see any reason with still trying to delete it from disk. If a file went online after the job was queued, it queued it with `delete: false` anyways.

etnoy: > When an asset goes online the offline flash and the trashed flag are updated at the same time and vice versa. If a file goes offline after the job is queued I don't see any reason with still trying to delete it from disk. If a file went online after the job was queued, it queued it with `delete: false` anyways.

Sounds good!

jrasm91 (approved):

PR #14635, opened 2024-12-11
feat(server): Merge Faces sorted by Similarity

lukasdotcom:
A new optional parameter on the `/api/people` route allows clients to specify sorting order by similarity to a person. This sorting is used on the merge page to organize photos based on similarity.

Implements: #4347

PR #14830, opened 2024-12-20
feat(web): Adds toggle to disable sorting of faces

### LINKED ISSUE #14843: Merge sort by similarity is often empty
### The bug

Trying out the new merge "sort by similarity" mentioned in v1.123.0, but it only works sometimes.

On most people, when I enter the merge page, the list of faces is just empty. I very much doubt that Immich can't find any faces close enough to display here, as it's empty even on faces where I know there are unmerged faces that should show up here. Meanwhile, on faces where it does work, the list seems to include everything, so scrolling all the way down, the faces down there have nothing in common with the merging face. So I'm pretty sure it's a bug.

Docker

### Version of Immich Server

v1.123.0

### Platform with the issue

- [X] Server
- [X] Web
- [ ] Mobile

lukasdotcom:
Adds a way to disable sorting of faces by similarity on merge screen and the detailed face view (#14635) below search bar. Keeps similarity sorting enabled as the default.

Fixes #14762
FIxes #14843

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -1196,6 +1196,7 @@
   "sort_items": "Number of items",
   "sort_modified": "Date modified",
   "sort_oldest": "Oldest photo",
+  "sort_people_by_similarity": "Sort people by similarity",
   "sort_recent": "Most recent photo",
   "sort_title": "Title",
   "source": "Source",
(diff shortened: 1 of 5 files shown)


alextran1502:
I think a better place for this is a toggle in the merge face interface itself

Lukasdotcom:
@alextran1502 I was just about to ask if that would be a better choice.

alextran1502:
Haha, like minds

welpher:
> I think a better place for this is a toggle in the merge face interface itself

Not only in the merge face page, the detail page named face should display first

Lukasdotcom:
You mean this screen right?
image
I added it to that one, but I probably should have updated the description of the pull request.

alextran1502 (approved):
Thank you, I cleaned up the logic and the UI to make the components fit in the context. I also set the sort to false by default because it causes an error that if there are no similar faces found, then no faces would be rendered

image

welpher:
> You mean this screen right? image I added it to that one, but I probably should have updated the description of the pull request.

Yes, this screen.

welpher:
> You mean this screen right? image I added it to that one, but I probably should have updated the description of the pull request.

Yes, in this screen named face should display first

PR #14866, opened 2024-12-22
feat: Mark people as favorite

arnolicious:
Added the ability to mark people as favorite on web.

The `GET /people` Endpoint then sorts favorited people to the front of the list. 
Thus the sorting would also apply in the app, however I have not (yet) added any indicators or functionality to mark people as favorite from there.


server/src/dtos/person.dto.ts
```diff
@@ -97,6 +101,9 @@ export class PersonResponseDto {
   isHidden!: boolean;
   @PropertyLifecycle({ addedAt: 'v1.107.0' })
   updatedAt?: Date;
+  @ApiProperty()
+  @PropertyLifecycle({ addedAt: 'DEV' })
```
danieldietzler, line 105: This will most likely be `v1.124.0`

server/src/repositories/person.repository.ts
```diff
@@ -95,6 +95,7 @@ export class PersonRepository implements IPersonRepository {
       .innerJoin('face.asset', 'asset')
       .andWhere('asset.isArchived = false')
       .orderBy('person.isHidden', 'ASC')
+      .addOrderBy('person.isFavorite', 'DESC')
```
danieldietzler, line 98: Is it expected to have people ordered by isFavorite by default?

web/src/routes/(user)/people/+page.svelte
```diff
@@ -223,6 +223,29 @@
...
+      const updatedPerson = await updatePerson({
+        id: detail.id,
+        personUpdateDto: { isFavorite: !detail.isFavorite },
+      });
+
+      people = people.map((person: PersonResponseDto) => {
+        if (person.id === updatedPerson.id) {
+          return updatedPerson;
+        }
+        return person;
```
danieldietzler, line 237: When can you possibly not find your person in here?

danieldietzler (commented):
The code looks very good to me, thanks! Could you add a simple server unit test for it? :)

server/src/repositories/person.repository.ts
arnolicious: I would argue yes, for me that was the only reason to add this favorite feature: to be able to display favorite people at the start of the list, and that is all it does for now.

For Users that don't use the feature, the sorting will still be the same, since every person will have it set to false.

web/src/routes/(user)/people/+page.svelte
arnolicious: Not exactly sure what you mean, this is inside of the map, its replacing the updatedPerson inside of the person list.

And I just copied the code for the other handlers in that file :D

arnolicious:
I added a simple unit test for testing the toggling, I also tried adding a test to check the sorting behaviour, tho I have not ran it locally and it was mostly guessing that this could work :D @danieldietzler

web/src/routes/(user)/people/+page.svelte
danieldietzler: What I had in mind is that `const index = findIndex((person) => person.id === detail.id)` followed by `people[index] = updatedPerson` should also work and is imo more intuitive to read. (plus it should be a touch more efficient lol)

arnolicious: Ahh right, I refactored it like that now 👍

danieldietzler (approved):
The code LGTM

arnolicious:
Alrighty, so I pulled in `main`, refactored (that one line lol) to kysely and also added the toggle Favorite functionality to the person detail page

PR #14874, opened 2024-12-23
feat(server): use the earliest date between file creation and modification timestamps when missing exif tags

### LINKED ISSUE #14857: Date fallback should be the earliest between creation and modification dates
### The bug

I have thousands of original photos without exif tag coming from many people, and many of them (maybe 80%) has a modification date *before* the creation date (the rest has identical creation and modification dates). Sadly, immich is by default (after exif tags) just using creation date, but it seems that the creation date can be far after the real original date, which is in my case the modification date.

It's a behavior of windows when copying and not moving files, it changes the creation date while keeping the modification date intact, which seems logic as the file is created at the time of the copy, while there is no file change so the modification date remains the same. 

It would be easily fixable by taking the earliest date between the creation (which could be the original date or the copied date) and the modification date.
Here is the faulty line: https://github.com/immich-app/immich/blob/6080e6e827250b069039538cb5d08ee7492b71c1/server/src/services/metadata.service.ts#L592 

As a side effect, to fix the already uploaded photos, the Metadata re-scanning job should also rescan the crearion/modification dates.

Debian

### Version of Immich Server

v1.123.0

Doesn't matter

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
Doesn't matter
```

### Your .env content

```Shell
Doesn't matter
```

### Reproduction steps

1. On windows (and probably any system), copy an old photo without exif tag
2. Then upload the copied photo 
3. You should see the copy date in photo's date instead of the original date (which is the modification date) 
...

### Relevant log output

_No response_

### Additional information

There is many apps dropping the exif part for privacy, like Facebook, Facebook messenger, WhatsApp, mms, and more. Because of those, immich do the fallback.

Chuckame:
I also applied the date extraction for motion asset creation, tell me if I need to rollback this part, maybe it was wanted to not re-use `getDates`.

Closes #14857


server/src/services/metadata.service.ts
```diff
@@ -589,9 +589,18 @@ export class MetadataService extends BaseService {
     let dateTimeOriginal = dateTime?.toDate();
     let localDateTime = dateTime?.toDateTime().setZone('UTC', { keepLocalTime: true }).toJSDate();
     if (!localDateTime || !dateTimeOriginal) {
-      this.logger.warn(`Asset ${asset.id} has no valid date, falling back to asset.fileCreatedAt`);
-      dateTimeOriginal = asset.fileCreatedAt;
-      localDateTime = asset.fileCreatedAt;
+      // When a file is copied (but not moved) before being uploaded to immich, the target file creation
```
alextran1502, line 592: We don't need a lots of comments here, maybe just

```
// use the earliest date as the fallback date
```

docs/package-lock.json
alextran1502, line 1: Can you please remove this change from the PR?

alextran1502 (approved):
LGTM, please address the comments

docs/package-lock.json
Chuckame: I though it was expected. This has been done automatically, maybe during the devcontainer init or the `npm run tests` command 🤔 

Maybe there should be a fix on that part

server/src/services/metadata.service.ts
```diff
@@ -589,9 +589,18 @@ export class MetadataService extends BaseService {
...
+      // date is set at the current timestamp, while the modification date remains untouched, so if the
+      // user copied the asset while he did not modified the file (like cropping, rotating and more), then
+      // we use the modification timestamp as it's still the original date. If the user modified the asset,
+      // then there is no other solution except a further manual fix.
+      this.logger.warn(
```
Chuckame, line 597: @alextran1502 is `warn` the good level ? I would more see it as `debug` as it's because of the user asset not exif content, but not a server error or abnormal behavior

server/src/services/metadata.service.ts
Chuckame: So I'll remove the comment, as the variable `earliestDate` is quite explicit

Chuckame: (change reverted)

server/src/services/metadata.service.ts
```diff
@@ -589,9 +589,13 @@ export class MetadataService extends BaseService {
     let dateTimeOriginal = dateTime?.toDate();
     let localDateTime = dateTime?.toDateTime().setZone('UTC', { keepLocalTime: true }).toJSDate();
     if (!localDateTime || !dateTimeOriginal) {
-      this.logger.warn(`Asset ${asset.id} has no valid date, falling back to asset.fileCreatedAt`);
-      dateTimeOriginal = asset.fileCreatedAt;
-      localDateTime = asset.fileCreatedAt;
+      this.logger.warn(
+        `No valid date found in exif tags from asset ${asset.id}, falling back to earliest timestamp between file creation and file modification`,
+      );
+      // eslint-disable-next-line unicorn/prefer-math-min-max
```
danieldietzler, line 595: Why not do `Math.min(asset.fileModifiedAt, asset.fileCreatedAt)`?

danieldietzler (approved):

server/src/services/metadata.service.ts
Chuckame: Because it returns a number, but I can do `new Date(Math.min(asset.fileModifiedAt, asset.fileCreatedAt))` instead if you prefer

docs/package-lock.json
Chuckame: (change reverted)

server/src/services/metadata.service.ts
danieldietzler: I'd prefer that, thanks!

Chuckame: Sadly, ts linter reports it 😆
image
image
You have to choose the pill: the red or the green one ? 😄

server/src/services/metadata.service.ts
Chuckame: @danieldietzler also if you have an idea

alextran1502: Yeah debug is fine

Chuckame:
```suggestion
      this.logger.debug(
```

server/src/services/metadata.service.ts
Snowknight26: `Date.valueOf()` or `Date.getTime()` also work.
```typescript
new Date(Math.min(asset.fileModifiedAt.valueOf(), asset.fileCreatedAt.valueOf()))
```

danieldietzler: I don't want to take either pill and instead just use luxon lmao. Do you think that's overkill or would it make things nicely readable?

Snowknight26: Funny enough luxon's `min`/`max` functions require a `DateTime` instead of a `Date`, so a conversion is going to have to happen somewhere anyway, which means something like:
```typescript
DateTime.min(DateTime.fromJSDate(asset.fileModifiedAt), DateTime.fromJSDate(asset.fileCreatedAt)).toJSDate()
```
😅

Chuckame: I personally prefer a vanilla approach, especially for this kind of simple need: 

> `Date.valueOf()` or `Date.getTime()` also work.
> 
> ```ts
> new Date(Math.min(asset.fileModifiedAt.valueOf(), asset.fileCreatedAt.valueOf()))
> ```

Which one you prefer? I'll commit with this one

danieldietzler: Your latest solution is great! Sorry haha

Chuckame: great!

danieldietzler (approved):

Chuckame:
Hello there, happy new year, I hope there is nothing missing to merge 🤞

alextran1502:
Happy new year and thank you!

PR #14879, opened 2024-12-23
feat(web): Add "set as featured" option for an asset

IMBeniamin:
## Description
This is a QoL feature that enables a user to set the featured photo of a person while inspecting (in full screen) said person.
This is an alternative flow to the currently existing one: `person -> three dots -> select featured photo`
It is now possible to set a featured photo by doing this:
`person -> photo -> three dots -> use this as featured photo`
The option is hidden in all views except when a person is provided to the asset-grid (i.e. the people/id page)

Note: I have followed the "set as album cover" implementation for this. The prop is propagated from the `asset-grid` all the way down to the `asset-viewer-nav-bar`. I do not like the fact that the handle function has been duplicated in both cases, but at least it's consistent.
[handleSelectFeaturePhoto]
(https://github.com/IMBeniamin/immich/blob/de39bf2b1701069b33949d5da3e0f47cdd338672/web/src/lib/components/asset-viewer/actions/set-person-featured-action.svelte#L19)
[handleSelectFeaturePhoto](https://github.com/IMBeniamin/immich/blob/de39bf2b1701069b33949d5da3e0f47cdd338672/web/src/routes/(user)/people/%5BpersonId%5D/%5B%5Bphotos%3Dphotos%5D%5D/%5B%5BassetId%3Did%5D%5D/%2Bpage.svelte#L185)

## How Has This Been Tested?

- [x] Open the asset viewer from the timeline. Option should NOT be visible
- [x] Open the asset viewer from the album asset grid view. Option should NOT be visible
- [x] Open the asset viewer from the people page. Option IS visible
- [x] Feature photo is updated after clicking the set as featured photo option

## Screenshots

image

## Checklist:

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -1142,6 +1142,7 @@
   "set": "Set",
   "set_as_album_cover": "Set as album cover",
   "set_as_profile_picture": "Set as profile picture",
+  "set_as_featured_photo": "Set as featured photo",
   "set_date_of_birth": "Set date of birth",
   "set_profile_picture": "Set profile picture",
   "set_slideshow_to_fullscreen": "Set Slideshow to fullscreen",
--- a/web/src/lib/components/asset-viewer/actions/set-person-featured-action.svelte
+++ b/web/src/lib/components/asset-viewer/actions/set-person-featured-action.svelte
@@ -1,0 +1,29 @@
+<script lang="ts">
+  import MenuOption from '$lib/components/shared-components/context-menu/menu-option.svelte';
+  import {
+    notificationController,
+    NotificationType,
+  } from '$lib/components/shared-components/notification/notification';
+  import { handleError } from '$lib/utils/handle-error';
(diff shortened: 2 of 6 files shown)


bo0tzz:
Neat! How does this behave if there are multiple people in the photo?

IMBeniamin:
> Neat! How does this behave if there are multiple people in the photo?

The action is only shown when navigating to the asset viewer from a person. Consequently the `person` which is updated is always the one from which the asset grid has been created.

Example:

`https://<immich_host>/people/c96b6bb7-553a-434b-867f-843f0ec89e6c/photos/a65bacc5-85e9-4d81-8e67-71fa4fd6b282?previousRoute=%2Fpeople`
Would update the featured photo for `c96b6bb7-553a-434b-867f-843f0ec89e6c`

jrasm91 (approved):
LGTM

PR #14887, opened 2024-12-23
chore(docs): add reference to metric ports

Stunkymonkey:
i had difficulties to setup, because the ports collided. trying to fix this.


Stunkymonkey:
can someone please fix the labels of this MR

jrasm91 (approved):

PR #14888, opened 2024-12-23
fix: trusted proxies

### LINKED ISSUE #14886: IMMICH_TRUSTED_PROXIES has no effect
### The bug

We document `IMMICH_TRUSTED_PROXIES` in the [docs](https://immich.app/docs/install/environment-variables/). As far as I can tell this was added in #11286 , with the main effect being 

```
- app.set('trust proxy', ['loopback', 'linklocal', 'uniquelocal']);
+ app.set('trust proxy', ['loopback', 'linklocal', 'uniquelocal', ...trustedProxies]);
```

However, I don't think that this *does* anything.

Testing:
- added IMMICH_TRUSTED_PROXIES as a random IP (not in my subnet)
- restart Immich
- Immich works fine through my reverse proxy, even login attempts are logged with the X-Forwarded-Host IP, not the internal container IP

For now, I will PR to remove this feature from our docs. Long term, we should enforce only accepting headers from an IP that is actually whitelisted as a trusted proxy.

Debian

### Version of Immich Server

1.123.0

1.123.0

### Platform with the issue

- [X] Server
- [ ] Web
- [ ] Mobile

```YAML
N/A
```

### Your .env content

```Shell
N/A
```

### Reproduction steps

as above

### Relevant log output

_No response_

### Additional information

_No response_

danieldietzler:
Reference: https://expressjs.com/en/guide/behind-proxies.html

Screenshot 2024-12-23 at 20 32 53

Closes #14886

### DIFF
--- a/server/src/repositories/config.repository.spec.ts
+++ b/server/src/repositories/config.repository.spec.ts
@@ -204,7 +204,7 @@ describe('getEnv', () => {
     it('should return default network options', () => {
       const { network } = getEnv();
       expect(network).toEqual({
-        trustedProxies: [],
+        trustedProxies: ['linklocal', 'uniquelocal'],
       });
     });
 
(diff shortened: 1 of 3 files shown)


alextran1502 (approved):
Thanks

jrasm91 (approved):

